Issue #2, opened 2016-03-25
Alerts Module


sebastian-leidig:
> **Components need to register `Alert`s that are displayed hovering on the top right part of the screen.**

Currently this is done with our `AlertManager` and `AlertCtrl`.

We can probably just reimplement these as a Service and Component in Angular2. Displaying may pose a challenge, as `angular-bootstrap` is not ready for Angular2.

sebastian-leidig:
Thanks @ojauch!

The only thing left to do is some visible UI element hinting that you can discard alerts by clicking on them.

ojauch:
@sebastian-leidig I think the [ng2-bootstrap](https://github.com/valor-software/ng2-bootstrap) would be a good choice for this issue, because it provides angular 2 directives for bootstrap components like alerts. Maybe we should use prefixes for our typescript classes, because I'm running into naming issues when I want to use the alert module of ng2-bootstrap.

sebastian-leidig:
ng2-bootstrap sounds good. for helgo_db we had used [angular-ui bootstrap](https://github.com/angular-ui/bootstrap#angular-2), that is working on an Angular2 version in [a different repo](https://github.com/ng-bootstrap/core). I am a little confused which one is the best choice for us right now.

About the naming issues: Maybe we can fix that somehow using namespaces? e.g. importing the external library as a different alias? http://stackoverflow.com/a/19758638/1473411
It would be nice to keep our class names as simple as possible of course.

ojauch:
I just found out that the naming isn't an issue anymore and the README of ng2-bootstrap was just outdated. But the idea of using namespaces in the future sounds nice.

sebastian-leidig:
Great, I think this is done now.

PR #10, opened 2016-05-12
Implement global alert service (#2)

ojauch:


sebastian-leidig:
Great!
Just did some tiny changes to make it slightly clearer to understand.

And for the time being I want to be more careful with automatically disposing messages. I just learned in a class that this is bad design ;-)

sebastian-leidig:
What we do need is some visible UI element hinting that you can discard alerts by clicking on them.

PR #12, opened 2016-05-24
[#2] add bootstrap directives

ojauch:


sebastian-leidig:
@ojauch Do we still need the "normal" bootstrap library (see https://github.com/NGO-DB/ndb-core/blob/master/index.html#L27 and https://github.com/NGO-DB/ndb-core/blob/master/bower.json#L16) or is everything included through ng2-bootstrap?

ojauch:
@sebastian-leidig ng2-bootstrap doesn't come with the bootstrap stylesheets and needs them as a dependency. So we can't remove them.

Issue #27, opened 2016-08-14
Upgrade to Angular 2 RC5: Introduce NgModules


sebastian-leidig:
Angular 2 [RC5 has been released](https://angularjs.blogspot.de/2016/08/angular-2-rc5-ngmodules-lazy-loading.html). Supposedly "RC5 represents our expected public API for our 2.0.0 release". The main change is the introduction of `@NgModule`.

We should upgrade and introduce `NgModule`s in our project.
More details on Angular Modules are available in the docs: https://angular.io/docs/ts/latest/guide/ngmodule.html
And there is an upgrade guide: https://angular.io/docs/ts/latest/cookbook/rc4-to-rc5.html

Issue #7, opened 2016-03-25
Extendible Navigation


sebastian-leidig:
> **Menus should be implemented in ndb-core but "real" apps (e.g. "HELGO DB" based on ndb-core) need to supply menu items for this.**

Possible solution:
- Menu items provided by a Service. The Provider for this Service is set at root level and overwritten by a concrete app using the core.
- Navigation menu is created dynamically using that Service
- ~~RouteConfig is created dynamically using that Service (RouteConfig defines the URLs and is usually done as an Annotation, [could be done programmatically](http://blog.mgechev.com/2015/12/30/angular2-router-dynamic-route-config-definition-creation/) though)~~  
- A custom app can define its own RouterConfig Component and insert `ndb-core`'s `AppComponent` as a ["Child Router"](https://angular.io/docs/ts/latest/guide/router.html#!#milestone-3-the-crisis-center)

sebastian-leidig:
Parent/Child Routers might be a good fit for this: https://angular.io/docs/ts/latest/guide/router.html#!#milestone-3-the-crisis-center

sebastian-leidig:
the router system has changed in the recent angular 2 version: https://angular.io/docs/ts/latest/guide/router.html

sebastian-leidig:
concerning RouterConfig, `NgModules` provides some interesting possibilities: https://angular.io/docs/ts/latest/guide/ngmodule.html#!#lazy-load

PR #29, opened 2016-08-19
Add navigation module that could be extended in app implementation

ojauch:


sebastian-leidig:
closes #27 

extends core towards extensible navigation ( #7  )

PR #30, opened 2016-08-19
Export navigation module

ojauch:
see #7


app/navigation/navigation-items.service.ts
```diff
@@ -11,8 +11,8 @@ export class NavigationItemsService {
         return this.menuItems;
     }
 
-    public addMenuItem(menuItem : MenuItem) {
-        this.menuItems.push(menuItem);
+    public addMenuItem(label: string, icon: string, routerLinkParameters: any[]) {
```
sebastian-leidig: Could we keep the `MenuItem` type as parameter? 
Put an additional line in "navigation.module.ts": `export * from "./menu-item";` so it gets passed on.

Code using the service then needs to `import {MenuItem} from "ndb-core"` but I feel this would be a little cleaner?

sebastian-leidig:
There is still an outdated test failing:

> has some MenuItems
> Expected 0 to be greater than 0

a real test for the new functionality of adding menu items would be great.

sebastian-leidig:
sorry for the merge conflicts ...

@ojauch Is it necessary to import the `app.routing` in `NavigationModule` (again). Couldn't that pose a problem, because that is `forRoot()`?

ojauch:
@sebastian-leidig the navigation component needs the [routerLink] directive which it gets from the routing module

sebastian-leidig:
Shouldn't we import `RouterModule` from "@angular/router"  rather than `routing` (which includes the specific routes set for the application) for this?

sebastian-leidig:
cool, after cleaning up some module imports it works for me now.

Issue #4, opened 2016-03-25
LatestChanges Module


sleidig:
> **The user should be informed about an update of the app through a popup on first login to the new version.
> In addition a user can see the current version number and click on it to get more information.**

Code for this is already separated in the old code into the folder `app/module_latest-changes`.
Reimplement this in TypeScript/Angular2?

- [x] display the current version (at bottom of menu sidebar?)
- [x] load a file (.json) in the root directory containing the list of changes for each version of our app
- [x] display the latest changes as a page or modal (overlay)
- [x] display should be shown when user clicks on version label
- [ ] display should automatically appear the once every time the user loads a new version of the app.

sleidig:
Status: by @fschoell on PR #37 
> Still missing is the functionality to automatically show the modal if a new version is available, this would require some kind of user service which returns the last version the user has been working with.

I think this shouldn't be too difficult. Adding a new attribute to the `user` entity should be enough (although we are creating dependencies between "User" and "LatestChanges" modules in that case).

PR #37, opened 2016-09-15
Latest changes

fschoell:
Implemented a latest changes module as described in issue #4 

Still missing is the functionality to automatically show the modal if a new version is available, this would require some kind of user service which returns the last version the user has been working with.


app/rxjs-operators.ts
```diff
@@ -0,0 +1,15 @@
+// import 'rxjs/Rx'; // adds ALL RxJS statics & operators to Observable
+
+// See node_module/rxjs/Rxjs.js
+// Import just the rxjs statics and operators we need for THIS app.
```
sleidig: Why do we need this file?

fschoell: The Angular2 http get function uses Observables from the RxJS library. This file is used to include a global selection of operators. `import 'rxjs/Rx` would be fine too, as it includes the complete RxJS library, but this might blow up the application size (don't know how much of an issue this actually is, but the recommendation in the Angular2 Documentation is to import only a selection of operators, see [here](https://angular.io/docs/ts/latest/guide/server-communication.html#!#rxjs)). 

I fixed the import statements in 9dbbe44 so this file is actually used.

sleidig:
information on last used version for a user should probably be added to the `UserModule` and saved to the database through the `user` entity.

@fschoell Do you want to work on this feature within this PR or should we merge the PR now? From my side the PR is tested and good to go :+1:

fschoell:
Alright I'm merging the pull request and leave the issue open.

PR #48, opened 2017-05-22
Switched build system to angular cli

fschoell:
Switched to [angular-cli](https://github.com/angular/angular-cli)

Todo:
 - [x] Get a running system
 - [ ] Make sure the tests run (depends on #49)
 - [ ] Adapt Travis CI
 - [x] Adapt the documentation


sleidig:
Nice to see some progress :-)

I had to install Angular CLI as global separately  with `npm install -g @angular/cli` so that should be mentioned in the README's Installation section.

Then, the app starts up with `ng serve` but after logging in, clicking on the other menu items produces JS errors rather than going to these modules. Is it working for you guys?

>  Error: Cannot find 'default' in 'app/user/user.module'
seems like the paths are not working properly?

sleidig:
The actual objective of this PR is completed, i.e. Angular CLI is set up and usable. Merging this to avoid confusion over the structural changes when doing any new work.

Remaining issues about test cases ( #49 ) and Travis ( #50 ) remain as separate issues

Issue #55, opened 2017-07-04
"AlertsComponent" error: 'type' isn't known property of 'alert'


sleidig:
When running the unit tests (`ng test`) I get the following error with a failed test of `AlertsComponent`:

> Can't bind to 'type' since it isn't a known property of 'alert'. ("<alert *ngFor="let alert of alerts" [ERROR ->][type]="alert.type" dismissible="true" (close)="deleteAlert(alert)">
  {{ alert.message }}
</alert>

Seems like something is not working with the [ngx-bootstrap alerts](http://valor-software.com/ngx-bootstrap/#/alerts). It might be a name confusion because we also have a local component in our project named "alert".

fschoell:
Problem was that the AlertModule from the ngx-bootstrap package wasn't imported into the AlertsComponent test class. Fixed it in #56

sleidig:
How did I spend an hour tinkering on this without finding the solution? ;-)

sleidig:
fixed with PR #56

Issue #53, opened 2017-07-04
Remove jquery dependency?


sleidig:
Why do we need `jquery`? This is currently included as a dependency in our [package.json](https://github.com/NGO-DB/ndb-core/blob/master/package.json#L27).

Was this included for bootstrap? Then it should not be needed anymore as [ngx-bootstrap](https://github.com/valor-software/ngx-bootstrap) takes care of that within Angular only.

@fschoell Do you have any idea where this might be needed?

Otherwise, let's try to remove it and see what happens.

fschoell:
Okay let's see, I'll try it

sleidig:
removed and closed with PR #56

PR #56, opened 2017-07-04
Updated ng2-bootstrap dependency, removed jquery, fixed alert test case

### LINKED ISSUE #52: Update ng2-bootstrap library
The library `ng2-bootstrap` was renamed to `ngx-bootstrap` and needs to be updated to the latest version 1.6.6. 

The failing test 
> Can't bind to 'type' since it isn't a known property of 'alert'.
is at least related to the library, although the update might not fix it.

fschoell:
This closes #52, #53, and #55


src/app/app.module.ts
```diff
@@ -25,6 +26,7 @@ import { DashboardModule } from './dashboard/dashboard.module';
     routing,
     FormsModule,
     AlertsModule,
+    AlertModule.forRoot(),
```
sleidig: I'll rename all imports of the third-party `AlertModule` with an alias to `BootstrapAlertModule`. Not sure this is the best solution but at least it keeps things clear and distinguishable.

PR #64, opened 2017-07-05
trying to fix travis config

### LINKED ISSUE #50: Fix Travis CI
currently Travis fails with the error

> ... 'apply' of undefined ...

sleidig:
source https://medium.com/from-the-couch/angular-2-with-travis-ci-922040e01937

(tries to) fix #50

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ sudo: required
 dist: trusty
 language: node_js
 node_js:
-  - '4.2'
+  - "6"
 
 addons:
 apt:
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,10 @@ before_install:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then export DISPLAY=:99.0; fi
 
 before_script:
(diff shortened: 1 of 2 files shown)


sleidig:
@fschoell Do we need those `if [[ "$TRAVIS_OS_NAME" == "linux" ]]` parts in the travis config?

sleidig:
aww, I am actually making things worse ... ideas, anyone?

fschoell:
No we don't need that anymore, as long as we keep our testing tied to linux. Could add other os later, but for now that's not necessary in my opinion. 

Fixed the cryptojs error, it's really annoying as those types are outdated and won't match the current cryptojs library. Using require should work, although we won't have typescript support now. However, Travis is now finally working (it's only the test cases that still fail).

@sleidig Do you know why travis is running two jobs on each commit within a pull request?

sleidig:
no idea about those two jobs. Maybe because I removed some config options like branches: only: master?
Anyway, merging this as it gets the job done (thanks to you!): it fixes the basic Travis setup.

Any further tuning for Travis is welcome

PR #68, opened 2017-07-06
[WIP] Reworking the EntityMapperService and Entity 

fschoell:
Don't merge, work in progress.

@sleidig Can you take a look at it and give me your opinion?


fschoell:
I might have found an elegant solution to store the references. We can use reflection to get the properties of each entity that's given to the save method of the entitymapperservice. If those entities have a property "_id" then we now that's actually a reference to another entity and we can store the reference to that id. See [here](https://github.com/NGO-DB/ndb-core/blob/832ca78dd4bd661668182d98884dc0db480716a1/src/app/entity/entity-mapper.service.ts#L74)

However, I'm not sure if this is actually suitable as it doesn't recognize Arrays or Lists of referenced entities (would be implementable though). But I don't know if there might not be more issues with that approach, we need to take a look at that.

sleidig:
all of this (mainly prefix / _id handling and loadAll) seems to be in the `master` already. Closing this here

Issue #54, opened 2017-07-04
Add functionality to the EntityMapperService and rewrite the entity class


fschoell:
EntityMapperService:

 - [x] We need the functionality to load lists of entities (like all children). The EntityMapperService should be extended to load those lists by prefix. 
 - [x] Maybe to save lists of entities, too? 
 - [x] We should find a solution to handle references.

Entity Class: 

 - [x] The entity id should be final as it's not allowed to change it later.
 - [x] We might need to rethink the getPrefix() method. Not sure if we should allow empty prefixes. In my opinion we should make this method either abstract so it has to be implemented by any child class or we use the class name as prefix instead (this way we could detatch the id handling completely from the actual entity classes, it only knows the actual id without any prefixes).
- [x] Write documentation how new sub Entity classes should be created: [Wiki - Creating new Entity types](https://github.com/NGO-DB/ndb-core/wiki/Creating-new-Entity-types)

also see PR #68, #84

fschoell:
@sleidig Your opinion on this?

sleidig:
#### Save lists of entities
I think this probably won't be needed. Stuff should be created/updated one by one anyway?

#### Handle References
Yes, a good solution for this is definitely needed. In fact the idea was that EntityMapperService should later be extended as to resolve any references of a loaded entity before returning it, so that it contains references to instances of any linked entities.
Although loading all links at once is probably slowing down things and there should be some "lazy loading" mechanism.

For the time being, maybe create a "referenced entity" class that contains the id and a function to resolve (i.e. load) the reference?

#### Entity.getPrefix()
Yes, this is actually quite ugly. I would prefer to make the prefix completely transparent - doing something based on class names might be a good solution but seemed a little to complicated and hacky when I created the first entity stubs

sleidig:
Long, but rather insightful article about document databases vs. relational databases. I'm starting to more and more think we'd be better off with SQL.
http://www.sarahmei.com/blog/2013/11/11/why-you-should-never-use-mongodb/

We chose CouchDB only for the out of the box, full sync with the local JavaScript implementation PouchDB. Despite some doubts of not using a relational db.

Maybe there is some relational DB with 2-way sync available nowadays?

Lorsbyjm:
Hi,
ich hab einfach mal gegoogelt und das gefunden, vllt hilft es:
https://xamarinhelp.com/mobile-database-bi-directional-synchronization-rest-api/

sleidig:
[...]

PR #72, opened 2017-07-07
LoadAll functionality for the EntityMapper

fschoell:
This is basically the implementation of #54 without the references so we get the functionality to load entity lists on the master branch. The references might take a while to design and implement, hence this PR. 

@sleidig I've added test cases to load multiple entities. Should be ready to merge if you don't find anything else.


src/app/config/config.service.ts
```diff
@@ -20,7 +20,7 @@ import { Injectable } from '@angular/core';
 @Injectable()
 export class ConfigService {
 
-  version = '2.0.0';
+  version = '2.0.14';
```
sleidig: This is our project version, right? If we bump it here, we should probably bump the version in `package.json` too - or in the first place only save a version in one file. Opening an issue about this, as it has nothing to do with this PR: #73

src/app/entity/entity-mapper.service.ts
```diff
@@ -48,6 +48,29 @@ export class EntityMapperService {
     );
   }
 
+  /**
+   * Loads a list of Entity from the database whose IDs contain the prefix of the given resultEntity class.
+   *
+   * @param entityType the entity's class type (this would just be <code>Entity</code> for a pure entity).
+   * @returns A Promise containing an array with the entities.
+   */
+  public loadAll<T extends Entity>(entityType: { new(id: string): T; }): Promise<T[]> {
```
sleidig: I had actually tried to do a straight-forward type parameter like this for the `load()` function but couldn't get it to work.

`load` and `loadAll` should have similar behaviour, so I suggest we modify `load` to take the same style of parameter

src/app/entity/entity-mapper.service.ts
```diff
@@ -48,6 +48,29 @@ export class EntityMapperService {
...
+    let newEntity = new entityType('');
+    return this._db.getAll(newEntity.getPrefix()).then(
```
sleidig: Can we somehow get the prefix from the class without creating an instance?  (although I did similarly ugly stuff in `load` - which I would like to get rid of)

Confusing to create a dummy instance here and again in line 64, rather than right before line 63.

src/app/entity/entity.ts
```diff
@@ -23,25 +23,62 @@
...
 
+  /**
+   * Returns the id of this entity.
+   *
+   * Note that an id is final and can't be changed after the object has been instantiated, hence there is no
+   * <code>setId()</code> method.
+   *
+   * @returns {string} the unique id of this entity
+   */
+  public getEntityId(): string {
```
sleidig: Does `Entity` in `getEntityId()` add any useful information? (rather than `getId()`)

src/app/entity/entity.ts
```diff
@@ -23,25 +23,62 @@
...
+    return this.prefix;
   }
 
-  public getId(): string {
+  /**
+   * Returns the id of this entity including a prefix to store this entity in the database.
+   *
+   * @returns {string} the entity's id including a prefix in the form <code>prefix:entityId</code>
+   */
+  public getIdWithPrefix(): string {
```
sleidig: Do we ever need to distinguish between ID with/without prefix outside of the class? (We need the prefix alone for loading ranges of course) 

I think it would be best to hide the concept of id prefixes to anyone and only expose one unique ID without confusing developers about when to use IDs with or without the prefix.

src/app/entity/entity.ts
```diff
@@ -23,25 +23,62 @@
...
+  /**
+   * Creates an entity object with the given id. This id is final and won't be changeable after this object has been
+   * created.
+   *
+   * @param id a unique id for this entity
+   */
   constructor(id: string) {
-    this.setId(id);
+    this.entityId = id;
+    this.prefix = this.constructor.name;
```
sleidig: Nice. I love this solution :-)

src/app/entity/entity-mapper.service.ts
fschoell: Unfortunately not really. I've tried static methods, but those kind of suck in typescript, because they don't exist on instances. We need the class to make a call which we won't have whenever we work with generics.

src/app/config/config.service.ts
fschoell: Ignore that, I've only bumped up the version to test my references, because whenever there is a newer version `save()` is being called on the entity mapper to store the new version to the user document. We can decide the version handling when we start releasing, as we should have different configs for releases anyways.

src/app/entity/entity-mapper.service.ts
fschoell: Yes I'll take a look into it, consistency would be nice.

src/app/entity/entity.ts
fschoell: Hmm I'll have a look into that, not quite sure how to handle this in an elegant way.

src/app/entity/entity.ts
fschoell: It has if we need getter for both ids (the entity id and the database id). Because then it makes those ids distinguishable. If we find another solution for the database id, then we can drop the Entity.

sleidig: I guess the `getIdWithPrefx()` is what you mean by "database id"? Hmm, thinking about separation of concerns, the best way might be to push that whole prefix thing out of Entity and into the database layer? Entity shouldn't worry about saving/identifying objects in the database and that is - as far as I can see - the only use of dealing with prefixes.

src/app/entity/entity-mapper.service.ts
fschoell: Okay it's done. Let me know your opinion on this and then I'll adapt the test cases.

src/app/entity/entity.ts
fschoell: Okay, it's still in the `Entity` class and we can't really move it out from there, because PouchDB needs the `_id` field with a database id (id + prefix) for saving those entities. However, I've moved it to a static function, this way we can drop at least the `getIdWithPrefix()` method. Let me know your opinion on that.

src/app/entity/entity-mapper.service.ts
```diff
@@ -73,14 +73,19 @@ export class EntityMapperService {
   }
 
   public save<T extends Entity>(entity: T): Promise<any> {
-    // TODO: how to save 'references' of this Entity to other Entities?
-    //      e.g. a 'Child' may have 'FamilyMember's who are Entity instances of their own
-    //      and should be saved separately in the database
+    entity['_id'] = EntityMapperService.getDatabaseIdByEntity(entity);
```
fschoell: @sleidig I found a solution to the `_id` field problematic. We could just dynamically add this field here and thereby remove the complete database id functionality from the entity class.

fschoell:
Ok, I think this is the way to go, I've adapted all test cases and cleaned up the code. If you agree, then it should be ready to merge. 

It's kind of a bummer that we can't make the getPrefix() method static but it just doesn't seem like typescript's generics are powerful enough to handle this. If you have another idea how we could solve this let me know.

fschoell:
Additional changes: As we moved the database id handling out of entity, I think it's more appropriate to use the term `type` instead of `prefix`.

src/app/entity/entity-mapper.service.ts
```diff
@@ -26,18 +26,27 @@ import { Entity } from './entity';
 @Injectable()
 export class EntityMapperService {
 
+  private static getDatabaseIdByEntity<T extends Entity>(entity: T): string {
+    return EntityMapperService.getDatabaseId(entity.getType(), entity.getId());
+  }
+
+  private static getDatabaseId(type: string, id: string): string {
```
sleidig: reading through this I initially got confused. Suggesting to rename this to `generateDatabaseId()`, `computeDatabaseId()` or something similar

src/app/entity/entity-mapper.service.ts
```diff
@@ -26,18 +26,27 @@ import { Entity } from './entity';
...
    *
-   * @param resultEntity An (empty) instance of an Entity class with its ID set to the one to be searched.
-   *          (This is necessary because TypeScript generic types are not available at runtime.)
+   * @param entityType a class that implements <code>Entity</code>.
+   * @param id the id of the entity to load.
    * @returns A Promise containing the resultEntity filled with its data.
    */
-  public load<T extends Entity>(resultEntity: T): Promise<T> {
-    return this._db.get(resultEntity.getIdWithPrefix()).then(
+  public load<T extends Entity>(entityType: { new(id: string): T; }, id: string): Promise<T> {
```
sleidig: :-)

src/app/entity/entity.ts
```diff
@@ -23,25 +23,47 @@
  */
 export class Entity {
 
-  private _id: string;
+  /**
+   * The unique id of this entity.
+   */
+  private readonly entityId: string;
```
sleidig: Why should this be `entityId` and not simply `id`?

src/app/entity/entity.ts
```diff
@@ -23,25 +23,47 @@
...
-    return '';
+  /**
+   * Creates an entity object with the given id. This id is final and won't be changeable after this object has been
+   * created.
+   *
+   * @param id a unique id for this entity
+   */
+  constructor(id: string) {
+    this.entityId = id;
+    this.type = this.constructor.name;
```
sleidig: What do inheriting classes like `User` have to do to ensure `type` is set correctly?

Do they have to explicitly call the super constructor? (and is `this.constructor.name` then really the name of the child class?)
Or do they have to include this line into their constructor(s) explicitly?

sleidig: Oh, you contributed the explanation already in `user.spec.ts`. Nice :+1:

src/app/entity/entity-mapper.service.ts
fschoell: Alright, renamed it to `createDatabaseId()` as we don't really computate or generate anything ;)

src/app/entity/entity.ts
fschoell: I've left this to be `entityId` on purpose, simply because it will create more distinguishable datasets in the database as otherwise we would have an `id` and an `_id` field. And it's hidden through the getter and constructor in the entity class so the name doesn't matter anyways.

PR #74, opened 2017-07-08
Changelog.json 404 not found

aankitkumar31:
fixed the issue Changelog.json 404 not found path name is given incorrect in latest-changes module and there is no file named with chnagelog.json


src/changelog.json
```diff
@@ -0,0 +1,30 @@
+[
```
sleidig: This duplicate file shouldn't be needed as it is already in src/assets/changelog.json. Or does this have a specific use here again? @aankitkumar31 ?

aankitkumar31: i had remove extra changelog

PR #75, opened 2017-07-08
change password module

aankitkumar31:
develop new change password module under user account module.


sleidig:
There are a few small formating issues (run `ng lint` to see the problems):
* attributes ("instance fields") should be declared before functions, even before the constructor
* spaces between `) {` and after the comma in lists of multiple parameters

Having a consistent format of all code in the projects makes it a lot easier to read, so our repository is set up to only allow merging after such issues are fixed.

aankitkumar31:
fixed the formating issues

src/app/service/change-password.service.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  public setNewPassword(password) {
+    const cryptKeySize = 256 / 32;
+    const cryptIterations = 128;
+    const cryptSalt = CryptoJS.lib.WordArray.random(128 / 8).toString();
+    const hash = CryptoJS.PBKDF2(password, cryptSalt, {
+      keySize: cryptKeySize,
+      iterations: cryptIterations
+    }).toString();
+    console.log('a');
```
sleidig: Please remove any debug functions from the code before a PR. (best to exclude such lines when making a commit)

src/app/service/change-password.service.ts
```diff
@@ -0,0 +1,31 @@
...
+import 'rxjs/add/operator/map';
+declare const require: any;
+const CryptoJS = require('crypto-js');
+@Injectable()
+export class ChangePasswordService {
+
+  constructor() { }
+
+  public name: string;
+  public lastUsedVersion: string;
```
sleidig: What is this `lastUsedVersion` used for?

src/app/service/change-password.service.ts
```diff
@@ -0,0 +1,31 @@
...
+import {Http,Headers} from '@angular/http';
+import 'rxjs/add/operator/map';
+declare const require: any;
+const CryptoJS = require('crypto-js');
+@Injectable()
+export class ChangePasswordService {
+
+  constructor() { }
+
+  public name: string;
```
sleidig: What is the `name` used for?

src/app/user/user-account/user-account.component.ts
```diff
@@ -16,7 +16,7 @@
  */
 
 import { Component, OnInit } from '@angular/core';
-
+import { ChangePasswordService } from '../../service/change-password.service';
```
sleidig: Ẁe are sorting our scripts by functional modules (i.e. `Navigation`, `User`, ...) rather than into `service`s and `component`s. Thereby, everything that is related and interacting with each other (e.g. everything having to do with the user) is in the same folder, easy to find, and independent of other code.

src/app/user/user.routing.ts
```diff
@@ -18,6 +18,7 @@
 import { Routes, RouterModule } from '@angular/router';
 import { UserAccountComponent } from './user-account/user-account.component';
 import { ModuleWithProviders } from '@angular/core';
+import { User } from './user';
```
sleidig: Why is the import of `User` needed here?

src/app/user/user-account/user-account.component.ts
```diff
@@ -16,22 +16,28 @@
...
 
   ngOnInit() {
+    this.user = this.sessionService.getCurrentUser();
   }
 
   changePassword( pwd , rpwd ) {
     if ( pwd === rpwd ) {
-      this.ChangePasswordService.setNewPassword( pwd );
+      // TODO: update the password for this remote database user first and deny the password change if that fails
+      this.user.setNewPassword( pwd );
```
sleidig: Currently the local login (through the `User` object) and the remote login with the database server are tightly linked. We need to ensure that the password is the same at local and remote.

Should we do remote password changes in the `User` class itself? Or take care of it in the UserAccountComponent or a separate service? @fschoell, what do you think?

fschoell: Hmm that's a good question. Especially when we are going to provide offline functionality in future this get's pretty tricky. I think we should postpone this until we figure out how we might seperate local and remote session handling, like proposed in #78 and while we're on that, we should rethink this whole login system, too.

sleidig:
all of this is already in 'dev' branch. Further changes should probably be done during developing the new session system.

PR #84, opened 2017-07-14
[WIP] Testing relational-pouch for saving references

fschoell:
see #54 

@sleidig So I've found this PouchDB plugin [relational-pouch](https://github.com/pouchdb-community/relational-pouch) which I think is exactly what we need as we have pretty structured data with lots of relationships. 

Started a new branch on top of `entity_mapper` so we can start testing this. Maybe you could take a look into it, too. It looks fine but I cannot get it working, because of [this error](https://pouchdb.com/errors.html#could_not_be_cloned). There seems to be a general problem with non-JSONfiable objects, so it's kind of weird that `put()` actually worked.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ cache:
     - node_modules
 
 script:
-  - ng lint
+  - ng lint --type-check
   - karma start karma.conf.js --single-run # Instead of ng test
 #  - ng e2e # Not implemented currently
   - ng build
--- a/package-lock.json
+++ b/package-lock.json
@@ -2,16 +2,81 @@
   "name": "ndb-core",
   "version": "0.0.0",
   "lockfileVersion": 1,
+  "requires": true,
   "dependencies": {
     "@angular/cli": {
       "version": "https://registry.npmjs.org/@angular/cli/-/cli-1.1.3.tgz",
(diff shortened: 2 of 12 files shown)


sleidig:
Looks like `relational-ṕouch` is packaging exactly what we started implementing into a nice plugin. On first sight, this seems a good solution for us.

If jsonifying classes is not acceptable for pouch we should maybe include a `getJSON()` function in `Entity` base class to filter the appropriate data for storing?

fschoell:
Hmm that isn't the optimal solution either. We get relations but there are some problems: 

1. We can only use integers as id, which means we can't use the username as id for example. This requires some rewrite of the session handling code. 
2. We need to extract the ids manually from nested objects. The library for example can load a child and the school it is going to, but it won't be a nested json object. See here: [https://github.com/pouchdb-community/relational-pouch#managing-relationships]()

Lorsbyjm:
@fschoell @sleidig just to bring it up again, ist this still a possible solution?

right now there are two branches, entity_mapper and entity_ampper_test. The second one is a few commits before the first one. Could the first one be deleted?

fschoell:
Let's keep them open until we decided how to implement the reference issue. They both offer partly implementations that can be used for testing and discovering (dis-) advantages of the certain approach.

sleidig:
Closing this due to the huge difference to the current code base.

Issue #86, opened 2017-07-25
Overview of pinned dependency versions and workarounds


fschoell:
Let's collect issues with dependencies upgrades here, so we have an overview which dependencies are pinned to a certain version and why and which workarounds are currently in place. 

 - ~~`pouchdb` and its typings are pinned to 6.1.1~~

fschoell:
PouchDB 6.3.4 creates the following error: 
```
WARNING in ./src/app/database/pouch-database-manager.service.ts
42:0-14 "export 'plugin' (imported as 'PouchDB') was not found in 'pouchdb'
```

Might be a problem with pouchdb's types `@types/pouchdb` which are outdated (those are for the 6.1.1 version of pouchdb). We should wait with pouchdb updates until newer typings are released.

sleidig:
Maybe better document this in a more visible/durable place - either in the README or the wiki?

fschoell:
I think this is rather an "issue", as those are hopefully not pinned to a certain version forever. Just until the bugs are resolved by the 3rd party developers.

sleidig:
closing this, as pouchdb is running on its latest version again

PR #88, opened 2017-07-26
Changed the dependency handling

fschoell:
I had multiple issues with dependencies again, so I decided to fix our dependency handling. This is the PR that includes:

 - Added a `npm-shrinkwrap.json` which saves the state of the `node_modules` folder and will be used by `npm install` instead of the `package.json`. This insures that we all are working with the same version of **all** installed packages (even of their dependencies). After an dependency upgrade, just run `npm shrinkwrap` to update the shrinkwrap file. In future this might be replaced by `package-lock.json` which is generated by the latest npm version automatically, however this still has some bugs and shrinkwrap is backward compatible to older npm versions, too. 
 - Changed the version handling to only update to minor versions (using `~` instead of `^`)
 - Updated dependencies
 - Pinned the zone.js version due to a bug, see #86


fschoell:
@sleidig those changes are only dependency updates and refactoring, so as long as you don't get any issues with those dependencies it's fine to merge (despite of the amount of changes).

PR #93, opened 2017-12-08
[WIP] Angular material

fschoell:


sleidig:
@TheSlimvReal, is this adding more than the switch to material in den `dev` branch? Otherwise we can maybe close this PR/branch and consider `dev` as the foundation for everything based on angular material?

sleidig:
According to the branch graph https://github.com/NGO-DB/ndb-core/network this code fully went into SchoolsList, ChildAndSchoolMerge and then ChildSchoolConnection branches. I would suggest we remove all those branches and only keep the last one in line, otherwise merging will become more and more confusing ...

PR #116, opened 2018-05-31
Offline Functionality

sleidig:
### Visible/Frontend Changes
- [x] app can now be used offline (supported through @angular/pwa)
- [x] icons changed to [font-awesome](https://fontawesome.com/v4.7.0/icons/)
- [x] modal dialog (instead of a simple alert message) shown during initial sync that prevents login

- [x] fixed Entity types' prefix/id handling as this got broken by `ng build --prod`


sleidig:
This is ready to be reviewed and merged.
The offline functionality can be tested here also: https://demo.aam-digital.com/

codingfabi:
I checked the Demo and the Code and it looks good and works approprietly.
One more Idea though:
How about adding a Button to the Top-Bar so you can synchronize the Database on click during a session?
Couldnt find such a feature.

sleidig:
That's a good point. There is an icon in the status bar being shown whenever PouchDB is currently syncing in the background. Maybe this can become an active button.
Did you observe the sync behaviour? When is the local data getting uploaded again after the client comes online? Only after a complete site refresh?

PR #130, opened 2018-06-11
Adaptions to HELGO needs

sleidig:
To get started with a pilot use of the system at our partner, fields and features have to be slightly adapted to their current use cases.


sleidig:
Sorry for this mess of a PR!
This includes a number of hotfixes and smaller features added that were somewhat urgent during introduction of the software at our pilot project.

I suggest this to be merged into `master` to have a common code base again (with or without detailed review).

From now onwards I will strictly keep one branch/PR per bugfix or feature to keep code review feasible. To avoid delay for deployment, I will keep a side branch `helgo_live` that have the latest merges without reviews before `master` receives them.

Issue #44, opened 2016-10-07
HELGO "Children" entity, list and details


sleidig:
Reimplement the basic child entity from the existing app. Keep this to the basics and handle any challenges in separate issues after this is done.
- [x] Child Entity (including unit tests)
- [x] Children List Component
- [x] Child Details Component
- [x] ~~~Routing (with lazy loading?)~~~

Things to be tackled after this issue is done:
Image/file uploads, linking between child and school, family & guardians

sleidig:
closed through #131, adapted to current data model in #130

PR #131, opened 2018-06-11
editing & creating children

TheSlimvReal:
Creating/Editing children works with the formular. IDs are made with the uniqID package.

### DIFF
--- a/src/app/children/child-details/child-details.component.html
+++ b/src/app/children/child-details/child-details.component.html
@@ -17,7 +17,8 @@
 
 <div class="row">
   <div class="col-lg-10">
-    <h1 class="page-header section-child"><i class="fa fa-child fa-fw"></i> {{ child.name }} [{{ child.pn }}]</h1>
+    <h1 *ngIf="!creating" class="page-header section-child"><i class="fa fa-child fa-fw"></i> {{ child.name }} [{{ child.pn }}]</h1>
+    <h1 *ngIf="creating" class="page-header section-child"><i class="fa fa-child fa-fw"></i>Adding new Child</h1>
   </div>
 </div>
 
@@ -203,17 +204,7 @@
         <mat-select [disabled]="!editable" formControlName="gender">
           <mat-option *ngFor="let g of genders()" [value]="gender[g]">{{gender[g].toLowerCase()}}</mat-option>
         </mat-select>
-        <!--<input matInput formControlName="gender" placeholder="Gender" title="gender" type="text">-->
       </mat-form-field>
-
-      <!--<mat-form-field>-->
-      <!--<mat-select [(value)]="selected">-->
-      <!--<mat-option>None</mat-option>-->
-      <!--<mat-option value="option1">Option 1</mat-option>-->
-      <!--<mat-option value="option2">Option 2</mat-option>-->
-      <!--<mat-option value="option3">Option 3</mat-option>-->
-      <!--</mat-select>-->
-      <!--</mat-form-field>-->
 
       <mat-form-field>
         <input matInput formControlName="project" placeholder="Project Number" title="project" type="number">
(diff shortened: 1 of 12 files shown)


src/app/children/child-details/child-details.component.html
```diff
@@ -17,13 +17,9 @@
 
 <div class="row">
```
sleidig: Do those divs and their classes still have meaning? Let's simplify and remove anything that's not strictly necessary (including the commented out sections of the template).

src/app/children/child-details/child-details.component.ts
```diff
@@ -15,21 +15,35 @@
...
   gender = Gender;
+
+  genders() : Array<string> {   //Transforms enum into an array for select component
+    let genders = Object.keys(this.gender);
+    return genders.slice(0, genders.length / 2);
+  }
+
   genderSelector;
   selectedGender;
   FamiliyTableSettings = {
```
sleidig: I think family members and socialworkers are currently not supported currently? Then before merging into master to close #44 we should remove these things completely for now.

src/app/children/child-details/child-details.component.ts
```diff
@@ -15,21 +15,35 @@
...
   templateUrl: './child-details.component.html',
   styleUrls: ['./child-details.component.css']
 })
-export class ChildDetailsComponent implements OnInit {
-  child = new Child('test');
-  editable = false;   // editable schould be enabled by default and be activated by button
+export class ChildDetailsComponent {
+
+
+  child: Child = new Child("testID");
```
sleidig: creating the new Child here is irrelevant, as it will be created in the constructor properly, isn't it?

src/app/children/child-details/child-details.component.ts
```diff
@@ -89,26 +103,67 @@ export class ChildDetailsComponent implements OnInit {
...
-    const params = this.route.snapshot.params;
-    const childId = params['id'];
-
-    this.entityMapperService.load<Child>(Child, childId)
-      .then(child => this.child = child);
-    // this.editable = false;
-    this.genderSelector = Object.keys(this.gender).filter(k => !isNaN(Number(k)));
-    this.selectedGender = '0';
+  save() {
+    this.child.name = this.form.get("name").value;
```
sleidig: wouldn't it be possible to avoid all these assignments using Angular's two-way data binding? https://angular.io/guide/template-syntax#binding-syntax-an-overview

src/app/children/child-details/child-details.component.ts
TheSlimvReal: It will throw an null pointer error when you leave child undefinded so this was the easiest workaround.

src/app/children/child-details/child-details.component.ts
TheSlimvReal: when you use 2 way bindings, you always override the loaded child, so there is not easy way to cancel the changes you made. This way you have a layer in between and we also can define which of the fields are optional and which are necessary for creation/editing

src/app/children/child-details/child-details.component.ts
sleidig: ok, fair point - I think I have done the same in other places ...
Shall we make it a convention to use `new Child('')` in such cases?

src/app/children/child-details/child-details.component.ts
sleidig: ah, that's interesting. This together with `initializeForm()` looks like a lot of boilerplate code but maybe it's a good solution.
I tried to solve the same issue at other places: https://github.com/NGO-DB/ndb-core/blob/864fbf6a45d09dfec889651ae4ed2e7e78c92302/src/app/ui-helper/entity-subrecord/entity-subrecord.component.ts#L59-L69 and https://github.com/NGO-DB/ndb-core/blob/864fbf6a45d09dfec889651ae4ed2e7e78c92302/src/app/children/notes/note-details/note-details.component.ts#L47  Not sure what's the best way overall ...

Issue #42, opened 2016-10-07
Reimplement "School" entity, list and details view


sleidig:
Reimplement the School module of the existing prototype for the HELGO project.
- [ ] School Entity
- [ ] School List Component
- [ ] School Details Component

codingfabi:
what exactly should the school details feature look like? just a list of childs visiting this school? or more details of the school itself?

sleidig:
`School` entities should have some general details about the school:
- Medium [select] (i.e. the primary language in which classes are taught)
- Board [select] (i.e. which government organisation is accrediting the school)
- Private School [checkbox]
- Address [text]
- Website
- School Timing [text]
- Working Days [text]
(this is what is currently in use)

Also, a list of the students studying in that school queried from the database would be good

TheSlimvReal:
I figured out a design for the connection between schools and children:

- every school has an array of ID's from children visiting this school
- every child has the ID of the school the child is visiting
- when creating/editing a child, a school can be selected from the already created schools in the database
- Routing is possible between:
 - SchoolList to SchoolDetails
 - SchoolDetails to ChildDetails and vice versa

sleidig:
Sounds good. For the displaying and navigation between Child and School you can use the `ChildBlockComponent` and `SchoolBlockComponent`.

But if you are saving the link on school as well as child side, isn't that redundant and in danger of becoming inconsistent when an edit is not properly implemented (or simply disturbed by connectivity problems)?

In the long run I would like a history of schools a child has attended (i.e. 2015: class 9 at School X; 2016: class 10 at School X; 2017: class 11 at School Y). But for now let's keep it simple for the first implementation.

TheSlimvReal:
I was also thinking about implementing it on just one side. But ist'nt it a kind of slow when you always have to look through every child in every school object to find the school a child is visting, when you just save the child ID's in the school object.

sleidig:
Yeah, might be slow, especially with many child records in the db. But saving redundant information is a big anti-pattern. To improve performance we could set up an index in pouchdb if necessary (Although the search, which is doing that, seems really slow as well on the pilot site).
relational-pouch (see #108) would probably do this internally.

Maybe we should go for the desired endresult right away:

- new Entity type "SchoolEnrollment" with child, school, dateFrom, dateUntil, grade
- Child is considered in the School, whose dateFrom-dateUntil interval falls into the current date
- an index (or relational-pouch setup) give efficient access to all enrollments of a child or a school

[...]

PR #156, opened 2018-07-26
[WIP] school detail and school list feature

SuppenGeist:
see #42 
Implement the school details view


codingfabi:
I think it is almost completed... I hope we can finish it tomorrow, so stay tuned.

Lorsbyjm:
Please also look at sentry. There was an error raised in your files

TheSlimvReal:
> 
> 
> What's the status on this? Is this anywhere near completion?
> The social workers have some new schools to add now in December/January. If this still takes a while, maybe I will implement a simplified version for immediate use.

The SchoolDetail screen is finished, but now I also wanted to bring the new ChildSchoolRelations into play. This takes me a little bit more time but I think it would be nice to finally have this implemented.

src/app/entity/EntityRelation.ts
```diff
@@ -0,0 +1,7 @@
+import {Entity} from './entity';
+
+export abstract class EntityRelation extends Entity {
```
sleidig: In which way is `EntityRelation` different from `Entity`? Does it really require a separate class for this?

src/app/schools/schoolsShared/schools.services.ts
```diff
@@ -0,0 +1,15 @@
+import {Injectable} from '@angular/core';
+import {School} from './school';
+import {EntityMapperService} from '../../entity/entity-mapper.service';
+import {Observable} from '../../../../node_modules/rxjs/Rx';
+
+@Injectable()
+export class SchoolsServices {
```
sleidig: Naming convention is to use "service" in singular (i.e. file name "schools.service.ts", class name "SchoolsService")

sleidig:
I do understand why this is taking long - quite a complex challenge with the relationship modelling between `Child` and `School`.

Have you looked into solutions that keep the relationship loading out of the entity classes (i.e. move/remove `Child.getSchools()` and `School.getStudents()`)? Maybe having that code in the ChildrenService and SchoolService - or in the EntityMapper?
To move those Entity classes around, save them to the database - and first and foremost to keep them simple - I like the idea that Entity subtype classes only contain business logic. The idea was initially to let an EntityMapper class take care of all the loading, saving and also the resolution of relationships.

sleidig:
I am taking out the completed features of the school details (everything up to a278a7d8a4c52fe2a5cc4007d1cbd86ca00af7ce ) into a separate PR: #204. So that we can make this available to users within the next days already.

Issue #187, opened 2018-11-29
update npm / third-party libraries


Lorsbyjm:
create new branch
run npm audit and check whats wrong
for more information talk to Lukas :D

---
Remaining alerts:
- pouchdb-authentication uses outdated url-parse (fixed, npm release pending: https://github.com/pouchdb-community/pouchdb-authentication/issues/249)
- @angular-devkit/build-angular uses outdated webpack-dev-server (to be fixed soon: https://github.com/angular/angular-cli/issues/13342)

sleidig:
`npm audit` shows security vulnerabilities of our dependencies/libraries. Seems to be the same as what github is analysing: https://github.com/NGO-DB/ndb-core/network/alerts ?

I think someone should look into a major update of all libraries, including to [Angular 7](https://blog.angular.io/version-7-of-angular-cli-prompts-virtual-scroll-drag-and-drop-and-more-c594e22e7b8c).
Any volunteers?

RichardNutt:
see https://update.angular.io/

**Before Updating**

- [ ] Switch from HttpModule and the Http service to HttpClientModule and the HttpClient service. HttpClient simplifies the default ergonomics (You don't need to map to json anymore) and now supports typed return values and interceptors. Read more on angular.io
- [ ] Remove deprecated RxJS 6 features using rxjs-tslint auto update rules
- [ ] Once you and all of your dependencies have updated to RxJS 6, remove rxjs-compat.
- [ ] If you use the Angular Service worker, migrate any versionedFiles to the files array. The behavior is the same.

**During the Update**

- [ ] Angular now uses TypeScript 3.1, read more about any potential breaking changes
- [ ] Update to v7 of the core framework and CLI by running ng update @angular/cli @angular/core in your terminal
- [ ] Update Angular Material to v7 by running ng update @angular/material in your terminal. You should test your application for sizing and layout changes.

**After the Update**

- [ ] Stop using matRippleSpeedFactor and baseSpeedFactor for ripples, using Animation config instead.

sleidig:
I think we'll have to pin the typescript dependency to `<3.2` for the time being.

PR #209, opened 2019-01-07
update Angular & dependencies

sleidig:
see issue: #187


sleidig:
This works for me know. @liwde @RichardNutt can you check and confirm?

codingfabi:
grafik

thats part of my stacktrace right there, so i would not recommend merging. but maybe @liwde or @RichardNutt could run a check of the branch aswell

liwde:
typescript

This is the error-message I'm getting when running `npm i` -- and it already was when we first worked on this. It says, no correct version of typescript is installed, even though it is (manually).

liwde:
I tried it on my second Laptop as well and it worked. Just as a heads up for everyone: If `npm install` does not work as expected, try `npm ci` for fetching all dependencies. This will

a) Clear node_modules before doing anything
b) Just use the `package-lock.json` --> no changed files in git, much faster

I will (squash) merge this now.

PR #212, opened 2019-01-07
Hotfixes 2.5.1

sleidig:
Several small but important bug fixes.

see commit messages


sleidig:
merged hotfixes without review to avoid long delays

PR #217, opened 2019-01-17
Use PouchDB Browser

liwde:
PouchDB is able to run both in NodeJS as well as in a Browser. In a browser, IndexedDB will be used, for NodeJS, however, Leveldown is the backend, which needs a binary. Per default, PouchDB will install support for both scenarios, with the binary failing to install on some versions of node due to prebuilds not being available (i.e. 9, 11).

There are officially supported versions of PouchDB ("presets") that don't come with this dependency. We should use these stripped versions to have less dependencies in our project, allowing a faster and more stable npm install.

See https://pouchdb.com/custom.html for more information.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1532,14 +1532,6 @@
       "integrity": "sha1-kbR5JYinc4wl813W9jdSovh3YTU=",
       "dev": true
     },
-    "abstract-leveldown": {
-      "version": "4.0.3",
-      "resolved": "https://registry.npmjs.org/abstract-leveldown/-/abstract-leveldown-4.0.3.tgz",
-      "integrity": "sha512-qsIHFQy0u17JqSY+3ZUT+ykqxYY17yOfvAsLkFkw8kSQqi05d1jyj0bCuSX6sjYlXuY9cKpgUt5EudQdP4aXyA==",
-      "requires": {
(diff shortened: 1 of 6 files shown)


liwde:
Be sure to disable `useTemporaryDatabase` in the `config.default.json` to not use the MockDB when testing. The error at the Notes tab is also present in the current master, I guess there are some issues with the foreign keys in the dev database...


I just checked, there are. `Note:1536258354699` references a Child `new` that is not in the db. I'll open an issue for that, the error message could be a bit more descriptive...

Lorsbyjm:
I looked at it and its starting and running, but I directly get some errors, see screenshot.

pouchdb-browser

liwde:
@Lorsbyjm, did you perform any actions to get this issue? Is this on a fresh local DB (in an Incognito Tab)?

I could not reproduce this issue on both my computers on both fresh and already-present local databases. Maybe this is related to the database-indices you added recently? These indices would be created by map-reduce-functions and might have issues with unexpected old values in the DB (i.e. text where only numbers are expected). Maybe they were added to your local PouchDB during development/testing, but are maybe not present in this branch?

Lorsbyjm:
Ok guys, should be good. While starting in icognito mode it has the same error like the master.

Issue #214, opened 2019-01-08
Tracking height and weight of children


sleidig:
Special request of the social workers using:
App should get a new section to enter height and weight of children at a specific date, as this will be tracked regularly from this year onwards.

- [x] Create new Entity subtype
- [x] Add table of heigh-weight measures to the ChildDetailsComponent ("Health" section)
- [ ] Design a page for batch entry of measures for a group of children, similar to "add a month's attendance". (Maybe this should be generalized to use a generic component for both attendance and heigh-weight entry?)
- [ ] Design a optional graphical layout of weight and height development over time. 
- [ ] Find a comfortable way to switch between foot/inches and m/cm

codingfabi:
Dumb question, but what system should we use for the height? foot and inches? or cm?

codingfabi:
started implementing in branch HealthCheckFeature

sleidig:
> Dumb question, but what system should we use for the height? foot and inches? or cm?

Not a dumb question at all. I think they would mostly use feet and inches in India. Let's start with that. But if you have any idea on how to make configurable, that's great

codingfabi:
Idea: With the table, implement an option to show the measurements in a graph over time, so you can track the childs growth visually.
we should search a good library to do that.

TheSlimvReal:
For displaying we could use a pipe that formats the weight/height into the correct format. One library I found for [metric pipes](https://github.com/vimalavinisha/angular2-metric-conversion-pipes) looks like it offers most of the functionality but I didn't test how well it works.

sleidig:
A pipe to just adapt the displaying sounds like a great way to go. That will give us flexibility for more localization in the future, also. I would suggest we write metric units into the entities/database and modify the view as needed (for now, switching between feet/inches and meters should be enough).

TheSlimvReal:
We should then define one type how we store it in the DB that will never change and only use the conversions in the displaying and reading of data. A problem might be how the user defines the unit (e.g. how do we know that the user enters a length in meters and not in feet). So we should propably create some components that let you select a metric unit and enter the value in a input field. Internaly we can use these two informations to determine the correct conversion.

sleidig:
The basic feature is functional and merged with PR #224. The other points mentioned above remain. Maybe we should create separate issues for those?

codingfabi:
I'll close this issue to keep the issues organized. The main features are implemented with pr#224. the open points will be put into a seperat issue with low priority.

PR #224, opened 2019-02-07
Health check feature

codingfabi:
see issue: #214 

### Visible/Frontend Changes
- [x] Implemented a sub component for child details showing the HealthChecks of the Child

- [x] Implemented a new subclass of entity called HealthCheck
- [x] Implemented methods for saving and loading HealthChecks into the child details component

### Pushing this PR as [WIP] because there are few things i need help with: 
- [x] Theres a bug not showing the health checks in the table when you load the page. you need to sort the content of the table first. i made the laoding of the health checks asynchron, but it didnt help.
- [ ] we need to work with units for the height and weight measurements here before pushing to the master. there should be a way to switch between foot/cm and kg/pounds. also we should somehow make the users use one unit when storing a new healthCheck.  (-> #241)


sleidig:
Do you have a special reason to display the form for adding a health check directly and disabling the integrated Add feature of the EntitySubrecordComponent?

I would tend to keep the interface as consistent as possible, doing it the same way everywhere.

src/app/children/child-details/child-details.component.ts
```diff
@@ -170,4 +195,36 @@ export class ChildDetailsComponent implements OnInit {
...
+           if(doc.child===this.child.getId()){
+                tempArray.push(doc);
+            }
+            else{
+              console.log(doc);
+            }
+       })
+       ).catch(error => console.log("ERROR FROM LOADHEALTHCHECKS"));
+       console.log(tempArray);
+       this.HealthCheckRecords=tempArray;
```
sleidig: Try putting the `this.HealthCheckRecords=tempArray;` inside the `then(...)` block above. (Can't test it because I am getting other errors)

Async can be a mess: whatever is inside the `then(...)` is only called later, so when your code `this.HealthCheckRecords=tempArray;` that is outside the `then(...)` is called immediately, the tempArray is still empty. And when the tempArray is finally filled, you are not assigning it to this.HealthCheckRecords again.

sleidig:
The socialworkers are using cm and kg as units. Let's not over-engineer it with a feature to switch units. But the unit should at least be visible in the UI (and maybe in the database?).

codingfabi:
> 
> 
> Do you have a special reason to display the form for adding a health check directly and disabling the integrated Add feature of the EntitySubrecordComponent?
> 
> I would tend to keep the interface as consistent as possible, doing it the same way everywhere.

I couldn't figure out how to use the add button in my case, so I made it possible to hide it. But I think i figured it out now so I'll try to implement a solution ASAP

codingfabi:
Okay so i finally managed to fix the bugs and implement the component properly by using the subentity-record. So i would really appreciate if someone would take a second look and point out the flaws... ;) thanks :)

src/app/children/child-details/child-details.component.html
```diff
@@ -297,7 +297,7 @@ <h3>Coaching Attendance</h3>
         </mat-form-field>
       </div>
 
-      <div>
+      <!-- <div>
```
sleidig: Don't forget to delete the commented code before the final merge

src/app/children/child-details/child-details.component.ts
```diff
@@ -28,7 +28,7 @@ import uniqid from 'uniqid';
 import {AlertService} from '../../alerts/alert.service';
 import {School} from '../../schools/school';
 import {ChildrenService} from '../children.service';
-import {HealthCheck} from '../HealthCheck';
+import {HealthCheck} from '../health-checkup/HealthCheck';
```
sleidig: for consistency the file name should better be health-check.ts.
Angular CLI and Webstorm manage to relate the camel case and hyphonated spellings automatically

src/app/children/children.service.ts
```diff
@@ -189,6 +190,15 @@ export class ChildrenService {
     );
   }
 
+  getHealthChecksOfChild(childId: string): Observable<HealthCheck[]>{
```
sleidig: this won't scale well with a larger database as it is loading every health record instead of using an index. But as we'll hopefully have a generic solution for this soon, I think this is alright here.

src/app/children/health-checkup/HealthCheck.ts
```diff
@@ -15,8 +15,8 @@
  *     along with ndb-core.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import { Entity } from '../entity/entity';
-import {Child} from '../children/child';
+import { Entity } from '../../entity/entity';
+import {Child} from '../child';
 
 export class HealthCheck extends Entity{
```
sleidig: better use `/**` style code comments to describe the attributes below. With that we can possibly generate some documentation automatically in the future (e.g. https://typedoc.org/guides/doccomments/).

Usually the IDE even automatically creates a comment template for you when you type `/**` in the line above a statement and press enter.

src/app/children/health-checkup/health-checkup.component.html
```diff
@@ -0,0 +1 @@
+<app-entity-subrecord [records]="records" [columns]="columns" [newRecordFactory]="generateNewRecordFactory().bind(this)"></app-entity-subrecord>
```
sleidig: what does `.bind(this)` do?

src/index.html
```diff
@@ -1,9 +1,10 @@
+
 <!doctype html>
 <!--
   ~     This file is part of ndb-core.
   ~
   ~     ndb-core is free software: you can redistribute it and/or modify
-  ~     it under the terms of the GNU General Public License as published by
+  ~     it Pnder the terms of the GNU General Public License as published by
```
sleidig: accidental change?

sleidig:
Generally, can you say something about the process developing this feature: What was difficult or particularly confusing? Would be great if we can further improve the framework accordingly

src/app/children/child-details/child-details.component.html
codingfabi: I cleaned up all the unnecessary comments now

src/app/children/child-details/child-details.component.ts
codingfabi: Changed the Filename and fixed the imports

src/app/children/health-checkup/health-checkup.component.html
codingfabi: It was heritage code from some thing i tried. I fixed it.

src/app/children/health-checkup/HealthCheck.ts
codingfabi: I get what you're saying and implemented most of the comments using /**. Anyway, I dont see the connection from the commenting style to the code snipped you used...

codingfabi:
In my opinion, the feature is mainly working now and can be merged into the master. we should then look for a way to desplay the units of height and weight for further development.

src/app/ui-helper/entity-subrecord/entity-subrecord.component.ts
```diff
@@ -140,4 +142,12 @@ export class EntitySubrecordComponent implements OnInit, OnChanges {
     col.selectValues = col.allSelectValues.filter(v => v.value.includes(input) || v.label.includes(input));
   }
 
+  changeVisibilityOfAddButton() {
+    if (this.showButton === true) {
```
TheSlimvReal: Can be replaced with this.showButton = !this.showButton

src/assets/config.default.json
```diff
@@ -6,6 +6,6 @@
     "remote_url": "https://demo.aam-digital.com/db/",
     "timeout": 60000,
     "outdated_threshold_days": 0,
-    "useTemporaryDatabase": true
+    "useTemporaryDatabase": false
```
TheSlimvReal: I think these changes should not be pushed

src/app/children/health-checkup/health-check.ts
```diff
@@ -0,0 +1,29 @@
...
+import { Entity } from '../../entity/entity';
+import {Child} from '../child';
+
+export class HealthCheck extends Entity {
+
+    static ENTITY_TYPE = 'HealthCheck';
+
+    child: String;      // child_id of the child the health check is made for
+    date: Date;
+    height: number;     // stores the height of a child in cm
```
sleidig: Should we name the attribute explicitly with the unit, e.g. `heightCm` to avoid any misunderstandings? (would be ugly but clear ...)

Can you use JavaDoc style code comments here as well? (`/**` instead of `//`)

src/app/children/health-checkup/health-checkup.component.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { HealthCheckupComponent } from './health-checkup.component';
+
+describe('HealthCheckupComponent', () => {
+  let component: HealthCheckupComponent;
+  let fixture: ComponentFixture<HealthCheckupComponent>;
+
+  beforeEach(async(() => {
+    TestBed.configureTestingModule({
+      declarations: [ HealthCheckupComponent ]
```
sleidig: I think you are missing some imports here, that's why the test fails.

src/app/children/health-checkup/health-checkup.component.ts
```diff
@@ -0,0 +1,71 @@
...
+export class HealthCheckupComponent implements OnInit {
+  records = new Array<HealthCheck>();
+  /**
+   * Column Description for the SubentityRecordComponent
+   * The Date-Column needs to be transformed to apply the MathFormCheck in the SubentityRecordComponent
+   */
+  columns: Array<ColumnDescription> = [
+    new ColumnDescription('date', 'Date', 'date', null,
+    (v: Date) => this.datePipe.transform(v, 'yyyy-MM-dd')),
+    new ColumnDescription('height', 'Height', 'number'),
```
sleidig: at least the labels here need to state the unit (i.e. "in cm" and "in kg")

src/app/children/health-checkup/health-check.ts
codingfabi: I don't think that the heightCm thing is a good idea, because it suggests that there is a height variable with a different unit... i would keep it like that for the moment and maybe refactor when we take care of the units properly...

src/app/children/health-checkup/health-checkup.component.ts
codingfabi: done

src/app/children/health-checkup/health-checkup.component.spec.ts
codingfabi: It took some time but I think I managed to fix this

src/app/ui-helper/entity-subrecord/entity-subrecord.component.ts
codingfabi: Done

Issue #218, opened 2019-01-17
No Error Handling / check for missing attributes in Notes


liwde:
When opening the Notes with `"useTemporaryDatabase": false` in the `config.default.json`, a pretty non-descriptive error-message is shown in a red Message-Toast: "missing(undefined)"

In the console, a few more messages appear: The DB could not find an entry to be loaded (`{"status":404,"name":"not_found","message":"missing","reason":"missing"}`) and a later error where `hasPhoto()` is called on `undefined`. It seems like the Child related to a Note is not found resulting in the error.

In fact, `Note:1536258354699` references a child `Child:new` that does not exist. So to fix this now, we could just delete the Note. However, we need to also think about fixing the root cause:

# 1. No Error Handling

`child-block.component.ts` has no error handling:
```ts
ngOnInit() {
    if (this.entityId !== undefined) {
      this.entityMapper.load(Child, this.entityId).then(child => {
        this.entity = child;
      });
    }
  }
```

I'm not sure if we could let an error here bubble up to the containing `notes-manager.component.html` to somehow remove the child from the Note, or whether a simple in-place message "Child deleted" would be better.

However, this relates to...

# 2. Foreign Key Handling

I don't know what happened to `Child:new`, but it must have been deleted somehow. Today, this is possible from our UI, which directly uses the EntityMapperService to delete the Entity. However, this does not delete foreign keys (e.g. in Notes) or even Entities (e.g. the ChildSchoolRelationship). Are there already ideas on how to handle this? Otherwise, we should think about a concept on how to handle cases like this.

sleidig:
Somewhat related: The current live system was broken due to a record with `note.date = null`, causing the NotesManager to remain blank without showing any notes. Console logged the error

> Uncaught (in promise): TypeError: Cannot read property 'getTime' of null

We do need better error checking. Maybe this should best go into the Entity sub-classes like `Note` itself (should we define getters?) and fill it with a reasonable default value or some kind of "Null Object". Then we don't have to repeat ourselves with error checking conditions all over the place.

sleidig:
I would suggest to limit this issue to the error handling / attribute checking.

@liwde, do you want to create a separate issue for the Foreign Key Handling / cascading delete / ... question? That sounds like a big but important topic.

liwde:
@sleidig, I completely agree, these are two different issues. I opened #220 for the long-term solution, so we can focus on handling the immediate effect and error-handling for Notes here.

[...]

Issue #220, opened 2019-01-23
Cascading delete (remove related entities / references when deleting an entity)


liwde:
# Issue Summary
When deleting an entity that is referenced in other entities (e.g. its id is stored in Note entities, etc.), we are currently neither deleting these related records nor removing the reference. This leads to dead references and subsequently to errors (see #874). We therefore need to implement a more sophisticated delete functionality.

Requirements:
- ? id of deleted entity is removed from any other record that remains in the database (e.g. Notes referencing a deleted child are edited and that id is removed from the Note)
- ? cascading delete: related entities can be deleted automatically together with the primary entity that is deleted (e.g. all ChildSchoolRelation entities for a child are deleted when that child is deleted)
  - how to configure this or ask the user whether to do a cascading delete?

Possible approaches:
- ?

-----
## original comment
As discussed in #218, currently we do not look at foreign key relationships when deleting entities at all. When someone deletes e.g. a child, the notes relating to that child keep the reference to a (now deleted) Entity causing some errors. Aside from the error-handling, we should also focus on this root-issue.

I can imagine that a feasible workaround would be to extend `entity.ts` to have some (abstract) lifecycle-method `onBeforeDelete(em: EntityMapperService)`, where each Entity-Subclass (like Child) can define steps to remove foreign references through the EntityMapperService passed as an argument.

This should work relatively independent of the Database layer just within our Entity system. It should be especially easy for Entities like ChildSchoolRelation that should be deleted with the Child.

What are your thoughts on this?

Lorsbyjm:
I am not really in to the EntityMapperService, but we should definetly not have "dead" relations. The sougesstion sounds good and not to complicated

sleidig:
This is somewhat related to the overall topic of how we manage related entities in the database (#108). But relational-pouch doesn't seem to offer this at the moment either https://github.com/pouchdb-community/relational-pouch/issues/82

sleidig:
If we implement a cascading delete of related records, we should consider restricting the rights to delete to admins only. Anyway, deletion is usually something users wouldn't want, as they loose all the context. Archiving entities (i.e. flagging them to be hidden) would often be a better option, I guess.

sleidig:
For ChildSchoolRelations (and maybe some other types of entities) there is really no point to keep them if a child is indeed deleted. So at least for some entities this cascading delete is needed (#547), while for other entities deleting in general may not be wanted (#371).

sleidig:
[...]

PR #225, opened 2019-02-07
Fix dangling relations at Notes

liwde:
see issue: #218 

### Visible/Frontend Changes
- [x] No more cryptic error messages ("missing(undefined)") from pouchdb
- [x] child-block.component: Will not be displayed if child was deleted from the database
- [x] child-select.component: Will lose selection if preselected child was deleted from the database

- [x] Introduce an option to make Alerts silent, so they won't be displayed at the UI

### DIFF
--- a/src/app/alerts/alert.service.ts
+++ b/src/app/alerts/alert.service.ts
@@ -33,7 +33,9 @@ export class AlertService {
 
   addAlert(alert: Alert) {
     this.alerts.push(alert);
-    this.openSnackBar(alert);
+    if (!alert.silent) {
+      this.openSnackBar(alert);
+    }
     this.logToConsole(alert);
   }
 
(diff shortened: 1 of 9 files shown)


src/app/alerts/alert.service.ts
```diff
@@ -33,7 +33,9 @@ export class AlertService {
 
   addAlert(alert: Alert) {
     this.alerts.push(alert);
-    this.openSnackBar(alert);
+    if (!alert.silent) {
+      this.openSnackBar(alert);
```
sleidig: I am submitting a commit, refactoring the silent and the switch-case to decide the display of alerts into one system that's more explicit than the original `openSnackBar()` function

src/app/database/pouch-database.ts
```diff
@@ -107,7 +107,7 @@ export class PouchDatabase extends Database {
 
 
   private notifyError(err) {
-    this.alertService.addWarning(err.message + '(' + err.id + ')');
+    this.alertService.addWarning(err.message + ' (' + err.status + ')', true);
```
sleidig: If we silence the error reporting here, we should make sure that higher level modules consistently do report what fails. But I guess that's the whole point of this PR anyway.

src/app/children/child-block/child-block.component.ts
```diff
@@ -22,8 +22,8 @@ export class ChildBlockComponent implements OnInit {
     if (this.entityId !== undefined) {
       this.entityMapper.load(Child, this.entityId).then(child => {
         this.entity = child;
-      }).catch(err => {
-        console.log(err);
+      }).catch(() => {
+        // No special error handling here, as the database will report the technical error and the UI catches the entity being undefined
```
sleidig: As the database errors are silenced for the user, maybe we should display a (user-friendly) alert here?

liwde: I thought about this, but just handling it silently then seemed like the better option. I hope, you can understand my reasoning (I found it very hard to put it in words):

Imagine this: Someone added a Note about some children. One of these children was then deleted. You could now (in theory) display a message whenever this child's child-block is deleted, i.e. at the Notes Overview. However, there is nothing the end user can do about it except editing the Note and then saving it (as the child-select component won't preselect children that do not exist).

Or you can hide the child-block altogether (which is what is happening now anyways, as we have nothing we could show (neither name nor picture, we have just the (technical) ID)), and don't display a message. When the Note is edited, the problem will solve itself as described above.

Once the child is deleted, there is nothing the user can do other than accepting this -- so why bother him/her? It is, however, important, to inform the user about the consequences of deleting a child in the first place -- but that's a different topic.

liwde:
I really like your changes, the alert-display is an elegant way of handling this.

src/app/children/child-block/child-block.component.ts
sleidig: Yes, I get your point. That's true.
I was thinking along something like: The user reads a note talking about a child and wonders why that child is not referenced. Without a notification the behaviour of the UI is not transparent. But it's probably even more important that - as you point out - the user can't do anything about it anyway.

At the core of this is #220 anyway: this error here should be a fringe case hardly ever occurring.

Go ahead and merge!

Issue #211, opened 2019-01-07
Latest Changes popup is always shown (should only show once for each version)


sleidig:
The `LatestChanges` module is supposed to automatically show a popup of the changelog whenever the user logs in for the first time after a version update.
However, the popup currently shows after every login (without version changes) for most users.

sleidig:
This is of course happening with the in-memory, development MockDatabase. The bug is for a proper CouchDB setup however. To test change your config.json to `"useTemporaryDatabase": true`.

Unfortunately I can't replicate it anymore, although it was the case for myself over the last weeks ...

Issue #161, opened 2018-08-02
font-awesome font is not available when offline


sleidig:
The icons are just displayed as boxes when the computer doesn't have a internet connection. While other resources are cached and usable offline, the webfont is somehow missing.

sleidig:
Can't reproduce anymore ...

sleidig:
the issue does still occur

Issue #148, opened 2018-07-13
Update (ServiceWorker) notification or reload


sleidig:
When a new version is loaded to the server, users still see the old app as it is cached using `@angular/pwa`. It seems the user actually has to do a had reload (Ctrl + F5) to get the update.

Users should receive a notification when there is a new version on the server with a link to load/refresh to get the update.

It seems like Service Workers offer some functionality to achieve this: https://angular.io/guide/service-worker-communications#swupdate-service

sleidig:
Any progress here or shall I take this up?

Lorsbyjm:
You can take it, markus has not ben there for a while

Lorsbyjm:
Any further comments on that or should we take away High Priority?

PR #227, opened 2019-02-14
Improved Updates

sleidig:
see issues: #148, #211, #161

### Visible/Frontend Changes
- [x] Notification message is displayed to the user when a new version of the app is available on the server and can be used after a reload
- [x] FIX: Icons (font-awesome) are properly cached and displayed when offline
- [x] FIX: Changelog popup is only displayed once after a version update

- [x] Remove `User.lastUsedVersion` from User entity. Now using a browser cookie to remember the last used version.


sleidig:
To test the offline (service worker) related features, refer to https://angular.io/guide/service-worker-getting-started#service-worker-in-action-a-tour

Issue #108, opened 2018-05-23
Consider using relational-pouch


sleidig:
The PouchDB plugin [relational-pouch](https://github.com/pouchdb-community/relational-pouch) could be interesting for our implementation of `PouchDatabase` to simplify and optimize the management of different Entity types in the database. (also see #54).

---

Pro's using _relational-pouch_
- some indices may be added automatically and for other indices the syntax is easier (https://github.com/NGO-DB/ndb-core/blob/master/src/app/children/children.service.ts)
- as schemas are objects, this might help towards generating schemas from user-written configuration instead of code? (#222)

Con's using _relational-pouch_
- the [schema configuration](https://github.com/pouchdb-community/relational-pouch#dbsetschemaschema) is quite verbose and somewhat confusing to me
- relational-pouch would be yet another wrapper between PouchDB and out EntityMapperService, which already does similar things (we are actually using the same _id trick as they do: https://github.com/pouchdb-community/relational-pouch#how-does-it-work)
- looks like we still would have to create some indices on our own: https://github.com/pouchdb-community/relational-pouch#dbrelfindhasmanytype-belongstokey-belongstoid
- existing code (and data!) needs to be migrated

---

Anyway, we need a system to automatically create IDs and related objects (making all the duplicated code in [ChildrenService](https://github.com/NGO-DB/ndb-core/blob/master/src/app/children/children.service.ts) obsolete. Either through relational-pouch or implemented on our own in EntityMapperService.

sleidig:
Things like https://github.com/NGO-DB/ndb-core/pull/127/commits/2a1cc141ef455cb18b4386a4aaece8d14ca93ff4 are rather boiler-plate code that relational-pouch seems to cover out of the box. Maybe it's worth to using it.

Lorsbyjm:
@TheSlimvReal @liwde Take a look at this. It seems to solve some of our problems for many to many relationships.

liwde:
I think we talked about this with @fschoell a while ago, who was looking into this.
Coming back to it now, it seems like one main problem would be that we'd need to migrate all databases to the format the plugin uses:

```ts
db.rel.makeDocID({ "type": "author", "id": 123 });
// author_1_0000000000000123
db.rel.makeDocID({ "type": "author", "id": "onetwothree" });
// author_2_onetwothree
```

Another question is what to do with the MockDB, as adding map-reduce queries on the fly is possible (and might even be possible in a generic way), but reimplementing this plugin is not feasible.

[...]

PR #233, opened 2019-03-08
Basic EntitySchema system

sleidig:
see issue: #260, #108

This is an attempt to simplify creation of new Entity subclasses by implementing a EntitySchema system that allows EntityMapper to automatically handle relationships between Entities.

I am still actively working on the architecture for it, this PR is a snapshot to welcome any discussion.

Have a look into the Unit Tests [entity.spec.ts](https://github.com/NGO-DB/ndb-core/blob/relational-pouch/src/app/entity/entity.spec.ts) and [entity-schema.spec.ts](https://github.com/NGO-DB/ndb-core/blob/relational-pouch/src/app/entity/entity-schema.spec.ts) to get examples of how it could be used.

### Current Status:
- [x] basic interfaces for `EntitySchema` defined and used
- [x] implement schema in Entity subclasses (User, Child, School, ...)
- [x] basic documentation (-> [Wiki: Creating new Entity types](https://github.com/NGO-DB/ndb-core/wiki/Creating-new-Entity-types), [Wiki: EntitySchema System](https://github.com/NGO-DB/ndb-core/wiki/EntitySchema-System))
- ~~fix MockDatabase to still work with the new system~~ (problem actually was the missing schema for `Child`

further additions will be done in later PRs:
- [ ] design the schema to allow Entity types to be used as data types and automatically transform them between objects and reference IDs on save/load.
- [ ] design automatic creation of indices based on the schema annotation and provide automatic querying methods for these (replacing the duplicated codes in `ChildrenService`, etc.)
- [ ] design the schema to allow arrays of other datatypes
- [ ] replace the explicit schema object with annotations for the attributes of the Entity classes.
- [ ] unit tests for save - load comparison of every Entity type to check the schema is correct (?)


sleidig:
The logically flow looks something like this:

entityschema_load

1. Entities are requested through the EntityMapper as before (`entityMapper.load(...)`)
2. Data from the database is "loaded" into an Entity instance to combine the raw data with the class (`resultEntity.load(data)`)
3. `Entity.load()` uses the schema defined statically for its Entity class to transform and filter the raw data while loading it into the attributes. (`Entity.schema.transformDatabaseToEntityFormat(...)`).

The process of saving an Entity to the database works similarly through `entity.rawData()` doing filtering and transformation of data with `EntitySchema.trasnformEntityToDatabaseFormat()`. This can for example replace a full object with it's reference id.
entityschema_save

sleidig:
One thing leads to another ...
I am now also changing the Entity system further:

1. Each Entity type (e.g. `Child`, `School`, ...) needs to be registered initially at `EntityMapperService.registerEntityType(Entity)`. This can be done in the related Module's constructor.
2. EntityMapperService keeps a registry of Entity types. Therefore we don't need to pass the constructor of the type into the load method anymore (type can be inferred from the id prefix or passed in explicitly); i.e. `entityMapper.load<Child>(childId)` instead of the current `entityMapper.load<Child>(Child, childId)`

The explicit registering of types is necessary because somewhere down the line we need the Database from Angular's dependency injection on startup to be able to create indices.

TheSlimvReal:
> 
> 
> One thing leads to another ...
> I am now also changing the Entity system further:
> 
>     1. Each Entity type (e.g. `Child`, `School`, ...) needs to be registered initially at `EntityMapperService.registerEntityType(Entity)`. This can be done in the related Module's constructor.
> 
>     2. EntityMapperService keeps a registry of Entity types. Therefore we don't need to pass the constructor of the type into the load method anymore (type can be inferred from the id prefix or passed in explicitly); i.e. `entityMapper.load<Child>(childId)` instead of the current `entityMapper.load<Child>(Child, childId)`
> 
> 
> The explicit registering of types is necessary because somewhere down the line we need the Database from Angular's dependency injection on startup to be able to create indices.

When registering the entities from their related module (e.g. for child in the children-module), might there be a problem when using the child class in a different module (e.g school-module) without visiting the children-list screen first? Because this would mean that the code in the child-module has not been executed.

TheSlimvReal:
You mentioned that if an object hast an attribute of type `Entity`, it will be saved to the database only with the `ID` of the `Entity`. How do you want to handle this attribute when loading the object from the database. Will the `EntityMapper` also load all the entities that you have an id for or will there be some kind of lazy loading of the `Entities` in the attributes?

sleidig:
> When registering the entities from their related module (e.g. for child in the children-module), might there be a problem when using the child class in a different module (e.g school-module) without visiting the children-list screen first? Because this would mean that the code in the child-module has not been executed.

It certainly shouldn't rely on users visiting sites in a specific order. This is a challenge I am still working on. I am trying to do entity registration when the module is loaded (which - except for Modules with special lazy loading by Angular - should happen on startup, as far as I understand). But I couldn't fully get it to work yet (especially in unit tests).

sleidig:
> You mentioned that if an object hast an attribute of type `Entity`, it will be saved to the database only with the `ID` of the `Entity`. How do you want to handle this attribute when loading the object from the database. Will the `EntityMapper` also load all the entities that you have an id for or will there be some kind of lazy loading of the `Entities` in the attributes?

For starters I would leave it to the consuming code to load the related entity (that's what we currently have also).

To improve on this, lazy loading would be best. Maybe during loading of the primary entity the EntityMapper could set some kind of load function for the related entities that only gets trigger once these are accessed. Maybe we should research a little how other ORM libraries or Angular for the lazy loading of modules do lazy loading.

TheSlimvReal:
I checked your latest commits and the wiki. It looks really good. But I think the unit tests should be passing and maybe all entities should be transformed to the new entity schema before merging it to the master.

TheSlimvReal:
The decorators were now merged into this branch but are only used in one of the classes and still shows the bug where the attribute is declared read only. I think we should decide for one way to go (with or without decorators).

sleidig:
The decorators are great to simplify the interface. My new proposal is a use case like this:

```
class TestClass extends Entity {
  @DatabaseField('string')
  fieldUndefined: string;

  @DatabaseField('string')
  fieldWithDefault = 'default';

  @DatabaseField('date', { generateIndex: true } )
  fieldDate: Date;
}
```

Where @DatabaseField is
`function DatabaseField(dataType: string, options: SchemaFieldOptions = {})`

src/app/entity/schema/entity-schema.service.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+
+    expect(entity.otherDate.getFullYear()).toBe(2018);
+    expect(entity.otherDate.getMonth()).toBe(0);
+    expect(entity.otherDate.getDate()).toBe(1);
+
+    const rawData = entitySchemaService.transformEntityToDatabaseFormat(entity);
+    expect(rawData.otherDate).toBe(data.otherDate);
+  });
+
+  it('schema:month converts to correctly between string and Date objects', function () {
```
TheSlimvReal: Maybe change the description of this test

src/app/children/attendance/attendance-month.ts
```diff
@@ -36,15 +37,22 @@ export class AttendanceMonth extends Entity {
...
-  student: string; // id of Child entity
-  remarks = '';
-  institution: string;
+  @DatabaseField() student: string; // id of Child entity
+  @DatabaseField() remarks: string = '';
+  @DatabaseField() institution: string;
+
 
   private p_month: Date;
+  @DatabaseField({dataType: 'month'})
```
TheSlimvReal: Why do you set the getter as the `DatabaseField`?

src/app/entity/schema/entity-schema.service.spec.ts
sleidig: Do you have a suggestion? Is the `schema:month` confusing - or which part do you stumble over?

src/app/children/attendance/attendance-month.ts
sleidig: If the decorator is on the private p_month, the field name saved to the database would be p_month rather than month, too. Having the name "month" seems much cleaner

src/app/entity/schema/entity-schema.service.spec.ts
TheSlimvReal: No, I just don't understand the sentence grammatically. I think you only need to remove the 'to'.

src/app/children/attendance/attendance-month.ts
TheSlimvReal: That makes sense but does this work when loading the object from the database to? Will the database field month be mapped to the p_month?

src/app/entity/schema/entity-schema.service.spec.ts
sleidig: oh, yes. You are absolutely right - don't know how I missed that after your first comment

src/app/children/attendance/attendance-month.ts
sleidig: I added another unit test in 1fc1923 that proves it is working. As the getter and setter run on the same property name, it's not an issue - and I wanted the loading to use the setter. I agree however, that it makes the code somewhat more confusing

PR #234, opened 2019-03-09
fix error in demo data (#232)

### LINKED ISSUE #232: Bug in demo-data.ts (only occurs in March)
In March, demo-data.ts will create attendance-information for February. It assumes that numbering of days starts at 1 (as is the default in JavaScript):

```ts
    current1.dailyRegister[1].status = AttendanceStatus.PRESENT;
    current1.dailyRegister[2].status = AttendanceStatus.PRESENT;
//...
    current1.dailyRegister[27].status = AttendanceStatus.PRESENT;
    current1.dailyRegister[28].status = AttendanceStatus.LATE;
```

However, AttendanceMonth.updateDailyRegister() creates the `dailyRegister` with indices starting at 0 (because array.push() is used). Therefor, `dailyRegister[28]` is undefined and its `.status` cannot be set.

@sleidig, is this an issue with the demo data and all indices should be reduced by 1, or should the dailyRegister behave differently?

sleidig:
fixes #232

### DIFF
--- a/src/app/database/demo-data.ts
+++ b/src/app/database/demo-data.ts
@@ -122,6 +122,7 @@ export class DemoData {
     current1.student = '1';
     current1.month = new Date();
     current1.month = new Date(current1.month.getFullYear(), current1.month.getMonth(), current1.month.getDate() - 10);
+    current1.dailyRegister[0].status = AttendanceStatus.LATE;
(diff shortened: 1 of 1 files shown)


sleidig:
Already merged through #237

Issue #216, opened 2019-01-12
Use ChildSchoolRelations instead of SchoolID


TheSlimvReal:
Currently, we have a SchoolId tag at each child to indicate, which school it is visiting. We also have the ChildSchoolRelations, which store information, on which child visited which school for which period of time. Now we need to replace all usages of the SchoolId tag with the usage of the child school relations, so we only store the information of the attended school at one place.
When this is done, we need to port the old data to the new data model.

TheSlimvReal:
There is another problem I ran into. Currently it's fairly easy to access the school's name from a child object, because it's just saved there. But if we now have the ChildSchoolRelation which will get us to the school object, where du we save that school object? For the moment I created extra classes that hold a child, a school and the relation object. This is needed, because in the child-list-component, we need the child object, the class (relation) and school name (school). Is there maybe a cleaner way to go?

TheSlimvReal:
I pushed the current status on the branch ["child_school_relation_impl"](https://github.com/NGO-DB/ndb-core/tree/child_school_relation_impl) but I dont want to go much further till this questions are answered.

sleidig:
We could add to `EntityMapper` and when loading a Child entity, immediately also load the current School entity and save that reference to the School object in an attribute like `Child.currentSchool: School` (same goes for the current class/grade, which will often be needed.
Then before saving the Child entity back to the database, we probably have to delete those references to avoid bloating the database.

Would be awesome if we could just save all related objects as references, like `Child.previousSchoolEnrollments: ChildSchoolRelation[]` and do lazy loading - i.e. only get the objects from the database once they are required. But that's maybe getting us in over our heads?

Once again, related to #108 (relational-pouch)

sleidig:
> There is another problem I ran into. Currently it's fairly easy to access the school's name from a child object, because it's just saved there.

Actually we only have the schoolId, not the name itself in the Child currently. But that is mostly to keep writing to the database simple, for convenience maybe we should save object references (see my comment above)

> But if we now have the ChildSchoolRelation which will get us to the school object, where du we save that school object? For the moment I created extra classes that hold a child, a school and the relation object. This is needed, because in the child-list-component, we need the child object, the class (relation) and school name (school). Is there maybe a cleaner way to go?

Isn't ChildSchoolRelation already that class, holding the combined information? Maybe here also the EntityMapper should add the object references during runtime. Or you can use the IDs in ChildSchoolRelation similar to how now the Child.schoolId is used?

[...]

Issue #207, opened 2019-01-05
Use EntitySubrecordComponent for childSchoolRelations


sleidig:
Currently previous schools of a child (`childSchoolRelation` entities) are displayed through a custom component `ViewSchoolsComponentComponent`. As these are simply entities related to a Child, similar to Attendance or Note entities, we can use the generic `EntitySubrecordComponent` to display/edit them. This will also provide a consistent UI across different sections.

* [x] use `EntitySubrecordComponent` for child's schools table
* [ ] table of schools of the child should display class --> #398 
* [ ] make the current school readonly (or bring up the edit popup of the table below?) and make it clear that users should probably add a new entry to the school history rather than change the existing entry --> #317

sleidig:
Also, the UX can be improved further: 

* [ ] table of schools of the child should display class
* [ ] make the current school readonly (or bring up the edit popup of the table below?) and make it clear that users should probably add a new entry to the school history rather than change the existing entry

sleidig:
remaining related points have their own issue tickets

PR #239, opened 2019-03-15
Child school relation impl

TheSlimvReal:
see issue: #216 

This branch introduces the ChildSchoolRelations.
This relations descripte a child visiting a school for a certain time period and class.

This relation should be used instead of giving each child a schoolId which this child is visting at the moment.

I removed the schoolId and schoolClass tags from the Child class and introduced the new ChildWithRelation class. This class gets a child and relation object and performs exactly like the old child object (with schoolId and schoolClass). This could maybe be refactored sometimes.

I also generated database indices and queries to fetch this relations effectivly.

### Visible/Frontend Changes
- [x] 
- [ ] 

- [x] 
- [ ]


src/app/children/childWithRelation.ts
```diff
@@ -0,0 +1,33 @@
+import {Child} from './child';
+import {ChildSchoolRelation} from './childSchoolRelation';
+
+export class ChildWithRelation extends Child {
```
sleidig: Why do we need a separate class ChildWithRelation? I think it would be much clearer if we just modify the existing Child class (and leave the schoolRelation undefined where necessary). Or am I missing something?

src/app/children/children.service.ts
```diff
@@ -80,7 +99,72 @@ export class ChildrenService {
...
+          map: `(doc) => {
+            if (!doc._id.startsWith("${ChildSchoolRelation.ENTITY_TYPE}")) return;
+            if (doc.end) return;
+            emit(doc.schoolId);
+            }`
+        },
+        by_date: {
+          map: `(doc) => {
+            if (!doc._id.startsWith("${ChildSchoolRelation.ENTITY_TYPE}")) return;
+            emit(doc.childId + (new Date(doc.start)).getTime());
```
sleidig: maybe for readability during debugging add a separator character: `emit(doc.childId + '_' + (new Date(doc.start)).getTime());`

src/app/children/children.service.ts
```diff
@@ -198,4 +274,23 @@ export class ChildrenService {
...
+    return this.queryLatestRelation(childId)
+      .then(relation => {
+        if (relation) {
+         return this.entityMapper.load<School>(School, relation.schoolId)
+        }
+        return null;
+      });
+  }
+
+  async getViewableSchools(childId: string): Promise<SchoolWithRelation[]> {
```
sleidig: In which sense are the Schools returned by `getViewableSchools()` "viewable"? Maybe we could name this in a more obvious way? The code looks like it loads all schools of a child?

src/app/children/view-schools-component/edit-school-dialog/edit-school-dialog.component.html
```diff
@@ -10,7 +10,7 @@ <h2 mat-dialog-title>{{creating ? 'Add' : 'Edit'}} school for {{child.name}}</h2
 </mat-form-field>
 
 <mat-form-field>
-  <input matInput [(ngModel)]='childSchoolRelation.schoolClass' placeholder='Class' title='class' type='text'>
+  <input matInput [(ngModel)]='childSchoolRelation.class' placeholder='Class' title='class' type='text'>
```
sleidig: `class` is or may be a protected keyword of TypeScript, that's why I renamed the attribute to `schoolClass` initially.

src/app/children/view-schools-component/view-schools.component.ts
```diff
@@ -0,0 +1,89 @@
...
+import {ChildrenService} from '../children.service';
+import {SchoolWithRelation} from '../../schools/schoolWithRelation';
+
+@Component({
+  selector: 'app-view-schools-component',
+  templateUrl: './view-schools.component.html',
+  styleUrls: ['./view-schools.component.scss']
+})
+
+export class ViewSchoolsComponent implements OnInit, OnChanges {
```
sleidig: Could this use our `EntitySubrecordComponent`? That might avoid writing some of this code that's similar in other components.

src/app/entity/PassableEntityConstructor.ts
```diff
@@ -0,0 +1,3 @@
+import {Entity} from './entity';
+
+export type PassableEntityConstructor<T extends Entity> = new(id: string) => T;
```
sleidig: Good idea to define a proper type for this rather than always having to search for complicated type definition!

I would simply name it `EntityConstructor` though - we don't have a non-passable one anyway.

As this is just a one liner and closely related, maybe we add it to the entity.ts file rather than in a separate one?

src/app/schools/school.ts
```diff
@@ -1,24 +1,23 @@
...
 
 export class School extends Entity {
-  static ENTITY_TYPE = 'School';
+  public static ENTITY_TYPE = 'School';
 
-  name = '';
-  medium = '';
-  privateSchool: boolean;
-  academicBoard = '';
+  public name = '';
```
sleidig: attributes are "public" by default. Any reason to add it here explicitly for some of the attributes?

src/app/schools/schoolWithRelation.ts
```diff
@@ -0,0 +1,32 @@
+import {ChildSchoolRelation} from '../children/childSchoolRelation';
+import {School} from './school';
+
+export class SchoolWithRelation {
```
sleidig: Same discussion point as for `ChildWithRelation`: It would be easier to understand if there is just a `School` class that integrates this directly in my opinion.

src/app/schools/schools-list/schools-list.component.html
```diff
@@ -35,7 +35,7 @@ <h1>Schools List</h1>
 </mat-expansion-panel>
 
 <div *ngIf="schoolDataSource">
-  <table mat-table [dataSource]="schoolDataSource" matSort class="mat-elevation-z8 table-list" width='100%'>
+  <mat-table mat-table [dataSource]="schoolDataSource" matSort class="mat-elevation-z8 table-list" width='100%'>
```
sleidig: Angular Material documentation recommends to use `<table mat-table>` (https://material.angular.io/components/table). I think that has some advantages over `<mat-table>` regarding how browsers handle it. Any particular reason to change it here?

src/app/ui/search/search.component.ts
```diff
@@ -1,8 +1,8 @@
 import { Component, OnInit } from '@angular/core';
 import {Database} from '../../database/database';
 import {Child} from '../../children/child';
-import {School} from '../../schools/school';
 import {Entity} from '../../entity/entity';
+import {School} from '../../schools/school';
```
sleidig: if you notice it, try to avoid committing such small, unrelated changes. The less changes, the easier is the review and merging :-)

src/app/children/childWithRelation.ts
TheSlimvReal: When we want to move this to the original child class, we need some functionality in the entity mapper that prevents the additional information (like class and school) to be saved to the db. Maybe we could say the nested objects are not stored at all or we only store attributes without a leading "_". If you want we can have a call about that topic this week.

src/app/children/children.service.ts
TheSlimvReal: Maybe you could call it just getSchoolWithRelation. It's viewable in that sense, that you can display the name of the related school instead of just the schoolId which is saved in the relation.

src/app/children/view-schools-component/edit-school-dialog/edit-school-dialog.component.html
TheSlimvReal: Oh sorry I dind't know that. I was not sure, what the current version of it was.

src/app/children/childWithRelation.ts
sleidig: That's true. The introduction of an EntitySchema (see #233) would solve this by making the decision what is saved to the database more explicit.

src/app/children/view-schools-component/view-schools.component.ts
TheSlimvReal: I will talk to @codingfabi and @christophscheuing what they think about it.

src/app/children/view-schools-component/edit-school-dialog/edit-school-dialog.component.html
sleidig: Funny enough, I never saw any issues with using `class` but it's probably better to avoid it.

src/app/schools/school.ts
TheSlimvReal: Your right, I am more used to do it that way but I will change it back to keep it in line with the project-style.

sleidig: Maybe also discuss it in the next meeting. If most people feel it's clearer if we explicitly add "public", we can also do that across all classes.

src/app/children/childWithRelation.ts
TheSlimvReal: So what do you recommend? Should we leave it like this until you are done with #233 or should we implement something temporary?

src/app/children/view-schools-component/view-schools.component.ts
TheSlimvReal: Turns out @christophscheuing is already working on it (#207). But I think it would be good to merge these changes soon because I also edited this component.

sleidig: ah, nice. Maybe you guys can discuss directly which change should go in first and in which way.

src/app/children/childWithRelation.ts
sleidig: Might be that #233 leads to a couple of interface changes that touch many files, which then creates complicated merges ... So I would prefer getting this done independent of it first.

What if we just add class and school to the `Child` class anyway - and accept that for a brief period that information will be written to the database unintentionally until the Schema system is in master? And as soon as the option is available, we can exclude those fields from the database schema.
At first glance, I don't see that causing major problems. What do you think?

TheSlimvReal: When this is anyway just a temporary solution, can't we stick to the ChildWithRelation and SchoolWithRelation for now? I think it shows very clearly where changes will happen when #233 is ready and it would be good @christophscheuing to continue with his task.

Issue #78, opened 2017-07-11
Rework Login Process: Introduce a remote session service


fschoell:
We should decouple the session services to avoid bugs such like #70 and #31 and to generally make the login process more transparent. 

I think the best way to solve this is to introduce a remote and a local session service with seperate states. Might even be possible to abstract the session service and then only overwrite particular methods.

### Requirements/Considerations:
- [ ] Simple and robust initial login (when local DB has to be synced before login)
- [ ] Smooth UX during initial sync (minimize needed sync size, clear UI feedback, ...)
- [ ] Simple handling of online/offline login differences with robust fallbacks
- [ ] After offline login, redoing remote login later on in background?
- [ ] Keep all authentication (local/remote) encapsulated at one place only (i.e. simplify `User` and `Database` modules to remove/hide all related)

sleidig:
Making this the thread to discuss a redesign of the login process overall, there are a few more points to consider:

* Currently we need to ensure that the password is only changed synchronuously in the local database and for the remote server (also see #75 )
* Keeping login information in the same database as other data produces huge delay for the first login on a new machine. At least we need to use an independent database for user data.

fschoell:
Some thoughts:

> Keeping login information in the same database as other data produces huge delay for the first login on a new machine. At least we need to use an independent database for user data that we sync independently.

Is this delay actually a problem? A partly synced database is probably not usable anyways. Maybe we should perform a full sync every time a new user logs in before he can actually do anything. 

Seperating the databases could still be wise though.

Maybe we should think about another remote login process. I've came up with some possibilities to decouple the session handling so we won't have the double login, but it's probably too much of an overhead:

* Key authentication would eliminate the need to synchronize to the remote server as the password change would just concern the private key. But this is kind of a hassle to set up and a lot of overhead. 
* Some token based approach would be possible either, so every new user get's a token which he enters at the first login, and then we use refresh tokens to keep the sessions alive. However, this lowers security (we will lose control over the remote sessions) and as soon as we might lose a token (or worse, the server loses session information like the auth key), the sessions get invalid and the user needs to enter a new token (which, in contrast to a password, he won't remember or have with him all the time). 
[...]

Issue #175, opened 2018-10-12
Set up CouchDB server to provide _bulk_get


sleidig:
PouchDB is giving a note in the console 

> POST https://couchdb.SERVER.com/DB_NAME/_bulk_get?revs=true&latest=true 400 (Bad Request)
> The above 400 is totally normal. PouchDB is just detecting if the remote supports the _bulk_get API.

`_bulk_get` is supposed to speed up replication of the database significantly, which might help solve the long initial loading times for us after first login (https://pouchdb.com/2015/10/06/pouchdb-5.0.0-five-years-of-pouchdb.html)

I couldn't find any documentation on how to make the CouchDB server support bulk_get, however ...

codingfabi:
This looks quiet familiar, but only quiet. Normaly when im working with pouchDB, i get errors of this kind:

_The above 400 is totally normal. PouchDB is just detecting if the remote supports the _bulk_get API._

Which PouchDB and CouchDB Versions are we using?
According to this discussion: https://github.com/pouchdb/pouchdb/issues/3424 it should work as far as i know.
its hard to help with this issue without any real insights of the server, do you have any more infos?

sleidig:
It looks like nowadays the `_bulk` api requests are actually accepted from our CouchDB server, so this ticket can probably be closed.

PR #246, opened 2019-04-04
New Session

liwde:
see issue: #78

Blogpost on the Software Design: https://blog.liwde.de/posts/2018/der-ewige-kreis/ (in German)

- [x] Completely rewrite the session service
- [x] Remove DatabaseManagerService and merge it's tasks into the session
- [x] Three parts for the new session: `local-session.ts` for local db and auth, `remote-session.ts` for remote db-connection and `session.service.ts` putting both together as Angular Service
- [x] Reworked login flow: See https://blog.liwde.de/posts/2018/der-ewige-kreis/#login-ablauf or the respective comments in the code
- [x] New Mock-Session to handle non-synced mock-db during development
- [x] Retry to connect when user was offline and goes online
- [x] Live-Sync with offline/online state change recognition
- [x] Unit Tests (some are already there, but we need more)
- [x] Documentation


liwde:
@sleidig @TheSlimvReal,

it would be great, if you could start looking at this. Feel free to ask, if you've got any questions!
An overview of the architecture is laid out here: https://blog.liwde.de/posts/2018/der-ewige-kreis/#redesign, with the final setup here: https://blog.liwde.de/posts/2018/der-ewige-kreis/#refactoring-und-fazit

I'm currently working on offline/online detection.

Some help regarding the tests would be much appreciated. To test the whole thing, integration tests would be nice, but they keep failing on premise (and sometimes toggle locally). If you've got ideas how to improve this, I'd be really happy to hear them :)

liwde:
Online/Offline-Checking now works out of the box, the UI can just grab the session and use `sessionService.getConnectionState().getStateChangeStream().subscribe()` or ``sessionService.getConnectionState().getState() === ConnectionState.connected` to get the values

src/app/session/local-session.ts
```diff
@@ -0,0 +1,130 @@
...
+      // all other cases must throw an error
+      throw error;
+    }
+  }
+
+  /**
+   * Wait for the first sync of the database, returns a Promise.
+   * Resolves directly, if the database is not initial, otherwise waits for the first change of the SyncState to completed (or failed)
+   */
+  public async waitForFirstSync() {
```
sleidig: Can we sketch an activity diagram to make the states entirely clear? Even a snap of a drawing would help a lot as a starting point.

src/app/session/login-state.enum.ts
```diff
@@ -1,31 +1,23 @@
...
+ *     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *     GNU General Public License for more details.
+ *
+ *     You should have received a copy of the GNU General Public License
+ *     along with ndb-core.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+// Login at the local database
+export enum LoginState {
+  loginFailed, // wrong credentials
```
sleidig: JsDoc?

src/app/session/remote-session.spec.ts
```diff
@@ -0,0 +1,16 @@
...
+ *     (at your option) any later version.
+ *
+ *     ndb-core is distributed in the hope that it will be useful,
+ *     but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *     GNU General Public License for more details.
+ *
+ *     You should have received a copy of the GNU General Public License
+ *     along with ndb-core.  If not, see <http://www.gnu.org/licenses/>.
+ */
```
sleidig: empty file?

src/app/session/remote-session.ts
```diff
@@ -0,0 +1,116 @@
...
+ *     but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *     MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *     GNU General Public License for more details.
+ *
+ *     You should have received a copy of the GNU General Public License
+ *     along with ndb-core.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+
+/**
```
sleidig: move docs directly above class

liwde:
@sleidig Thank you for the review! I'll have a look at the code next Thursday, but I've already created some sequence diagrams detailing out what happens in which scenarios and how the three different states changed. I checked in the documentation and will add the generated SVGs as well next week -- for now there is a preview in my blog post's appendix.

src/app/session/local-session.ts
liwde: I added sequence diagrams and hope they help

src/app/children/child-details/child-details.component.spec.ts
```diff
@@ -93,15 +91,14 @@ describe('ChildDetailsComponent', () => {
         AlertService,
         DatePipe,
         PercentPipe,
-        { provide: SessionService, useValue: mockedSession},
-        DatabaseManagerService,
+        databaseServiceProvider,
```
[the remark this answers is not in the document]
liwde: @TheSlimvReal mentioned the same thing. But wouldn't that leave us with a cyclic dependeny, where SessionModule imports DatabaseModule for the Database, but DatabaseModule imports SessionModule for the Session to be used in the DatabaseServiceProvider?

src/app/session/local-session.spec.ts
```diff
@@ -1,22 +1,16 @@
...
- *     You should have received a copy of the GNU General Public License
- *     along with ndb-core.  If not, see <http://www.gnu.org/licenses/>.
- */
-
-export enum SessionStatus {
-  loggedIn,
-  loggedOut,
-  loginFailed
-}
+/*
```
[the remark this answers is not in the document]
liwde: Yes, there are currently no tests for the local and the remote session. I'm not sure what to test here, but I'll think about it

src/app/session/synced-session.service.spec.ts
```diff
@@ -0,0 +1,278 @@
...
+            expect(sessionService.getLoginState().getState()).toEqual(LoginState.loggedIn);
+            expect(sessionService.getSyncState().getState()).toEqual(SyncState.completed);
+            expect(sessionService.getConnectionState().getState()).toEqual(ConnectionState.connected);
+
+            expect(sessionService.isLoggedIn()).toEqual(true);
+            expect(sessionService.getCurrentUser()).toBeDefined();
+        });
+
+        it('has the correct state after Logout', () => {
+            sessionService.logout();
```
TheSlimvReal: I would sugest to log the user in before calling logout. This way the test would also pass without even calling logout thus it does not properly test the function.

src/app/session/synced-session.service.spec.ts
```diff
@@ -0,0 +1,278 @@
...
+            expect(sessionService.getLoginState().getState()).toEqual(LoginState.loggedOut);
+            expect(sessionService.getConnectionState().getState()).toEqual(ConnectionState.disconnected);
+
+            expect(sessionService.isLoggedIn()).toEqual(false);
+            expect(sessionService.getCurrentUser()).not.toBeDefined();
+        });
+    });
+
+    // These tests mock the login-methods of local and remote session.
+    // We cannot test wether the StateHandlers are in correct state, as these are set in the sub-classes themselves.
```
TheSlimvReal: *whether

liwde:
@sleidig Any more input from your side?

TheSlimvReal:
I collected a list of ToDo's from the comments of @sleidig:

- [ ] Check if the DatabaseModule can be removed or refactored it with the session
- [ ] Change enum-names to be upper case
- [ ] Explain why the `pulledChanges` in the `syncStatus` is commented-out
- [ ] Reset the config.json to use the local DB on default
- [ ] Logout when being offline should be working
- [ ] Login in for the initial sync when offline should create an konstistent state

liwde:
> Explain why the pulledChanges in the syncStatus is commented-out

In the initial design I didn't include that sync status. Live Sync is working (tested across different browsers) and a sync completion message is shown. Is that enough or do we need the additional message? (We could add it under the `case SyncState.STARTED`.)

liwde:
Hi everyone,
I'm back and addressed the open issues. Please have a look again, but the issues from the last reviews were resolved :)

src/app/session/mock-session.service.ts
```diff
@@ -0,0 +1,73 @@
...
+        return this.loginState;
+    }
+    public getSyncState(): StateHandler<SyncState> {
+        return this.syncState;
+    }
+    public getDatabase(): Database {
+        return this.database;
+    }
+    public login(username, password): Promise<LoginState> {
+        if (username === 'demo' && password === 'pass') {
```
TheSlimvReal: It would be good if the login method uses the users of the mock BD. When working with the admin panel or creating new users it would be helpful if these things can also be tested in the temporary database.

liwde:
At least in Firefox, indexeddb (which pouchdb heavily uses) is disabled – when I tried it, that message was actually shown at login. I didn't try Chrome, though.
You can reset your data in the devtools, it's mostly just clearing the indexeddb. Chrome even has a simple option in the devtools to delete everything of a webapp at once.

src/app/session/mock-session.service.ts
liwde: I'll have a look at this. As far as I remember, this piece was adapted from the old session/Database-Manager, but what you say makes sense.

TheSlimvReal:
Okay when I start the app in chrome and disable the internet before login, the error message is correct. However, after turning on the internet again, the app starts trying to load the db immediately (without pressing login button) and the popup stays for ever.

liwde:
> However, after turning on the internet again, the app starts trying to load the db immediately (without pressing login button) and the popup stays for ever.

This is (at least in part) as designed: When you are offline, the login will automatically retry in the background. It does not make a difference, whether you are online or not. However, the popup should not stay forever. I'll have a look.

sleidig:
from a UX perspective I think it would be best to start the retry in background only after a successful (offline) login. In that case going online really just triggers background sync and no visible action so it is not intrusive.

liwde:
@sleidig I agree and added this other behavior. I also added batching to the non-live sync, which seems to speed up the initial sync by factor 4 (4min->1min)

src/app/session/mock-session.service.ts
sleidig: you could even consider some inheritance from LocalSession (`class MockSession extends LocalSession`) and override the diverting parts

liwde: The `MockSession` must (at least indirectly) extend the abstract `Session` for Dependency Injection. I'd rather not make `LocalSession` extend `Session`, as we'd need to implement the additional APIs then that don't make sense at the `LocalSession`...

liwde: @TheSlimvReal The demo user is now pulled from the database instead of being hardcoded. However, it is still defined in the MockSession – should we move it to the Demo Data generation?

liwde: @sleidig This is my first time working with demo data generation, and I really like it! It feels much better to have the demo user in there (see https://github.com/Aam-Digital/ndb-core/pull/246/commits/9ede8b53a42de38a790641a9a55fd6ff2524454f)

sleidig:
Another bug when offline:
1. try to log in with wrong credentials (correctly, the error message is displayed - we should change it's wording a little bit though as it doesn't have to be a first login. Maybe the same "Username or password incorrect!" that is shown when online and a second line "If you are trying to log in for the first time on this device, please make sure you have a working internet connection.")
2. log in with correct credentials. You are logged in but immediately end up in a login-logout-loop sending you forth and back between dashboard and login form

sleidig:
Everything else looks good!

liwde:
> we should change it's wording a little bit though as it doesn't have to be a first login

There was some flawed logic on my side, and fixing it required a bit more work. I introduced a separate SyncState `ABORTED`, as this better describes what happens to the initial sync when we are offline. This way, the UI can decide to show the correct error message.

sleidig:
> @sleidig I agree and added this other behavior. I also added batching to the non-live sync, which seems to speed up the initial sync by factor 4 (4min->1min)

is this batching actually the same as this issue: https://github.com/Aam-Digital/ndb-core/issues/175 ?

liwde:
> is this batching actually the same as this issue: #175 ?

Looks like it. At least I found that adding the batch option increased response sizes and sped up the sync. PouchDB default seems to be 25. And `_bulk_get` is used in some requests.

src/app/session/mock-session.service.ts
sleidig: :-)

Issue #263, opened 2019-05-14
EntitySchema should use code annotations rather than an explicit object


sleidig:
_also see the related Epic #260_

The EntitySchema system currently uses an explicit object to define the schema that has to be separatedly written by the developer. This is duplicating a lot of information, repeating attribute names and data types that are already written in the TypeScript code of the class itself and creates room for mistakes e.g. when renaming a class attribute and forgetting to rename the schema field. Therefore it would be better to use code annotations, similar to the Angular framework's system, instead of the explicit assignment of the schema (`static schema = Entity.schema.extend({...});`)

sleidig:
implemented with #233 thanks to @TheSlimvReal

PR #266, opened 2019-05-17
[WIP] Schema decorators

TheSlimvReal:
This targets issue #263.
The PR targets to remove the schemaline object and replace it with annotations that are more readable.
To look at an example go to the aser.ts file.
The concrete but very provisionally implementation can be found in the file db-decorators.ts but should not be seen as final.

### DIFF
--- a/src/app/children/aser/aser.ts
+++ b/src/app/children/aser/aser.ts
@@ -17,21 +17,10 @@
 
 import { Entity } from '../../entity/entity';
 import {WarningLevel} from '../attendance/warning-level';
-import { dbProperty } from '../../entity/db-decorators';
+import {dbClass, dbProperty} from '../../entity/db-decorators';
 
-
+@dbClass('Aser')
 export class Aser extends Entity {
-  static ENTITY_TYPE = 'Aser';
-  static schema = Entity.schema.extend({
-    'child': 'string',
-    'date': 'Date',
-    'hindi': 'string=',
-    'bengali': 'string=',
-    'english': 'string=',
-    'math': 'string=',
-    'remarks': 'string',
-  });
-  
 
   static ReadingLevels = [
     'Nothing',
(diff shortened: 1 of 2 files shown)


src/app/entity/db-decorators.ts
```diff
@@ -0,0 +1,23 @@
+import {Entity} from './entity';
+
+export function dbClass(entityType) {
+  return (constructor) => {
+    constructor.ENTITY_TYPE = entityType;
```
sleidig: Nice!

src/app/children/aser/aser.ts
```diff
@@ -47,13 +37,19 @@ export class Aser extends Entity {
     'Division'
   ];
 
-
+  @dbProperty('string')
   child: string; // id of Child entity
+  @dbProperty('Date')
```
sleidig: Is it possible to use the TypeScript typings of the annotated class attributes implicitly (i.e. instead of writing 'Date' or 'string' explicitly in the annotation, infer that from `date: Date`)?

sleidig:
Cool, this looks really good!

We should settle on an intuitive naming of the annotations, I would start with a capital letter, like Angular annotations do.

Maybe `@Entity` and `@EntityField` are slightly more descriptive? (also, we could orient ourselves at other established ORM libraries like Hibernate [->](http://www.techferry.com/articles/hibernate-jpa-annotations.html#Entity)

src/app/children/aser/aser.ts
TheSlimvReal: You can retrieve the type by using the `reflact-metadata` API, but to get the more sophisticated information like to determine if it is an option parameter is not supported yet (one aproach can be found [here](https://stackoverflow.com/a/34701754/10713841)). But I think that is way to "hacky". So should we just retrieve the type information automatically or just leave it as it is with the string regex?

Issue #254, opened 2019-05-02
Create a fully functioning admin dashboard


codingfabi:
To make the software completly usable, it is essential that some of the users are granted admin privileges and are able to do perform CRUD on users. Therefor, we should consider creating a functioning admin UI. For the UI, we should consider the following: 

- [x] A list that shows all users (use EntitySubrecordComponent) -> #63 
- [x] Within the List, an option to change the password of the user or even delete the user -> ? (related: #17 (password change))
- [x] A option to grant the user admin privileges
- [x] A button to create a new user using a form

Furthermore, we need to work out the following: 

- [ ] how to store the role of the user? AFAIK, it is stored as a boolen `isUser==true/falls`, is this the best way? (related: #128 closed origin discussion)
- [x] the admin pannel should only be accesable if the user is a admin
- [x] Can we use the existing admin interface and extend it? Or should we create a new one and use the exisiting as "super-admin-interface"

Due to the relevance  of this feature for the basic standalone use of the database by the social workers, this issue is getting set as High Priority.
Discussions about the details of implementation are welcome.

codingfabi:
For development use the branch _adminUI_

sleidig:
I added a few links to the list pointing to existing issues on these.

I would strongly suggest that we work for every point (also) with a separate issue and PR, so that we'll be able to merge things more regularly (even if the feature might not be visible to users yet). That would help us avoid stagnation and merge conflicts, imo.

sleidig:
> * the admin pannel should only be accesable if the user is a admin

For this we have `AdminGuard` (assuming that the way we manage the admin status are safe, which is a general question); for usage see e.g. https://github.com/NGO-DB/ndb-core/blob/249f96fcff29f90ebca1a845043cfac94f3b38b4/src/app/navigation/navigation/navigation.component.ts#L42

Additionally, the final enforcement of access rights happens at the database on the server. (As everything else is happening locally on the user's machine it could always be circumvented or hacked more easily). Changing users is not possible for our normal CouchDB user accounts for example.


> * Can we use the existing admin interface and extend it? Or should we create a new one and use the exisiting as "super-admin-interface"

I think the user management should be a separate page, which would probably deserve its own top-level entry in the menu.

codingfabi:
Okay so here is what @christophscheuing and I have done so far: 

- we created a list with all users only accessable for the users that are granted the admin role inside the application
[...]

Issue #199, opened 2018-12-13
Generalize (base) UI components


sleidig:
To avoid code duplication and duplicate work, we should look into common UI components and generalize them.
We can look into different approaches: 

- develop an abstract base class to be extended for different use cases
- develop a generic component that may accept some config variable to adapt it to slightly different use cases (e.g. like `EntitySubrecordComponent`, `FilterSelection<T>`)

This will also help us to make more of the UI user-configurable through the UI itself, rather than specifically hard-coding each screen.

Related to this: #6 

## Components to be generalized:
- [x] Children list filters (filter children by center, status, ...); to be reused in attendance manager and other places
- [ ] Entity Select Form Element (we have `ChildSelectComponent` but this should be generalized into an "EntitySelectComponent" to reuse it for users, schools, etc.)
- [ ] Entity Block Component (i.e. generalization of `ChildBlockComponent`, `SchoolBlockComponent`)
- [x] List Component (i.e. generalization of `ChildListComponent`, `SchoolListComponent`)
- [x] Details Component (i.e. generalization of `ChildDetailsComponent`, `SchoolDetailsComponent`)

vlle1:
the selection menu (mat-form field / mat-button-toggle-group) of html in notes-manager.component, attendance-manager.component is redunant. (add this to Components to be generalized)

sleidig:
most important components have been refactored and generalized in #550 .
Closing this - new issues can be opened for specific individual components.

PR #271, opened 2019-05-19
[WIP]AdminUI

codingfabi:
!!!Please note that this is highly WIP, I opened a pull requests because I could need some help and another pair of eyes to look at it!!
see issue: #254 

### Visible/Frontend Changes
- [x] List of all curent Users with CRUD functionality, only accessable for admins

- [x] enabling storing of users
- [x] enabling deleting of users
- [x] enabling user loading from database
- [ ] enabling updating users including password


codingfabi:
Let me explain the wokraround (commit 7e8d06c) because it may seem like the dummest thing in existence but it actually is necessary:

`generateNewRecordFactory(){
    return () => {
      const newUser= new User('');
      return newUser;
    };
  }`
when creating new entitys with the entity-subrecord component, we need to initilize a new entity before we edit the rest of the information. in case of the user, the id is dependant on the username and therfore dependant on the admins input for the username. 

so i tried to catch the creation of a new user in the entity subrecord, but it is not possible to get to the password the admin has entered. therefore, i had to do this piece of whatever you want to call it just to get the whole thing running:
`if(record.getType()==='User'){
      let user: User;
      user=new User((record as User).name);
      user.setNewPassword('pass');
      user.admin=(record as User).admin;
      user.name=(record as User).name;
      record=user;
    }`

so to get it running, it was enough but it is basically unusable. 
See Issue #254 for further discussions

sleidig:
As the user management with passwords is a somewhat special case, it might be easiest to implement it's own template without EntitySubrecordComponent. In fact even ChildrenList and SchoolList don't use a common component yet - although that's on the roadmap (#199).

Otherwise, we could think about extending `ColumnDescription` with another attribute, a "saveFunction" that defines a special way how to write the field? Or could we make the `setPassword()` function of User into a simple setter (that still converts the plaintext password into the encrypted password object)?

src/app/ui-helper/entity-subrecord/entity-subrecord.component.ts
```diff
@@ -49,8 +50,22 @@ export class EntitySubrecordComponent implements OnInit, OnChanges {
...
+  /**
+   * 
+   * @param record uses the input from the form field to create a record
+   * curently has ugly workaround for entity type user because the id is dependent from the users input
+   * this workaround is highly WIP because its buggy aswell for creating a user twice somehow. 
+   */
   save(record: Entity) {
+    if(record.getType()==='User'){
+      let user: User;
+      user=new User((record as User).name);
```
sleidig: Don't you lose the "_rev" field of an existing user object when creating a new user like this? That would make PouchDB reject changing it.

TheSlimvReal:
As far as I got the state of the PR what needs to be done is to correlctly implement the `signupUser` function for the Pouch and probably also the moch database. Added to that I think there is a more elegant solution to the `save` method in the entity subrecord component. As @sleidig mentiond, it might makes sense to use a custom component here.

TheSlimvReal:
Closing in favor of #389

Issue #213, opened 2019-01-08
child photos not showing anymore


sleidig:

codingfabi:
Can you provide a screenshot, error message or anythin? cant really recreate this problem with the demo database.

sleidig:
To recreate:

1. create a folder `src/assets/child-photos` and an image `XYZ.jpg` inside it
2. change the projectNumber of a child to `XYZ`

Expected outcome: the child photo src is set to `XYZ.jpg` and the picture is displayed

Actual outcome: the child photo src is still the default "assets/child.png"

RichardNutt:
Note: If you provide photos for existing children (eg. `/assets/child-photos/1.jpg`) it will display them at first. If you change the projectNumber to something else (`1` to `42`) and then revert the change (back to `1`) it won't display the old image, even though it was once loaded.

sleidig:
@RichardNutt: What's the status on this, displaying individual photos?

I tried again recently and still can't get images to be displayed at all. As this is a nice and very visible feature, the users have asked about a fix for this a couple of times already ;-)

PR #305, opened 2019-07-01
Photo fix

RichardNutt:
see issue: #213

### Visible/Frontend Changes
- added a "has photo" checkbox when editing child; check it when a file is deposited onn the server

- removed onPhotoError() method, since it just set "hasPhoto" to false, regardless of the photos existence

### DIFF
--- a/src/app/children/child-details/child-details.component.ts
+++ b/src/app/children/child-details/child-details.component.ts
@@ -110,7 +110,7 @@ export class ChildDetailsComponent implements OnInit {
       dropoutType:    [{value: this.child.dropoutType,    disabled: !this.editing}],
       dropoutRemarks: [{value: this.child.dropoutRemarks, disabled: !this.editing}],
 
-      hasPhoto: [this.child.hasPhoto]
+      photoFile: [this.child.photoFile]
     });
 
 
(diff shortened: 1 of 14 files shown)


src/app/children/child-details/child-details.component.ts
```diff
@@ -109,6 +109,8 @@ export class ChildDetailsComponent implements OnInit {
       dropoutDate:    [{value: this.child.dropoutDate,    disabled: !this.editing}],
       dropoutType:    [{value: this.child.dropoutType,    disabled: !this.editing}],
       dropoutRemarks: [{value: this.child.dropoutRemarks, disabled: !this.editing}],
+
+      hasPhoto: [this.child.hasPhoto]
```
sleidig: should we maybe save the filename instead? That'd give us more flexibility

RichardNutt: Done.

sleidig:
@RichardNutt If my additions look good (and clear) to you, go ahead and press the merge button.

src/app/children/child.ts
```diff
@@ -100,9 +108,9 @@ export class Child extends Entity {
   }
 
   public getPhoto() {
-    if (!this.photoFile) {
+    if (!this.photoFile || this.photoFile === '') {
```
RichardNutt: empty string already evaluate to false

Issue #223, opened 2019-02-07
Implement an "add photo"-functionality


RichardNutt:
Currently there is no way of adding new photos through the app...
There might be a problem with the error-handling when displaying images, see Issue #213.

sleidig:
Allowing users to add photos is definitely needed.

Connected to the larger issue of how we can deal with any kinds of files: #89

liwde:
Nextcloud config: https://github.com/nextcloud/server/issues/3131#issuecomment-489179454

sleidig:
done through Nextcloud integration with PR #306 
Big thank you to @RichardNutt !

PR #306, opened 2019-07-01
Blob service

RichardNutt:
see issue: #223

### Visible/Frontend Changes
- [x] Photo-Functionality on the details page
- [x] Photo-functionality on the block/list page

- [x] Nextcloud authentification currently not working / Auth. header missing / CORS

### ToDo
- [x] Fix tests
- [x] Write tests
- [x] Fix buffer size (bufferArrayToBase64 method breaks for large images)
- [x] Check whether pictures exists on the nextcloud server
(- [ ] Check if the login to the nextcloud server succeeds (see [here](https://github.com/perry-mitchell/webdav-client/issues/163)))
- [x] Store pictures for future requests (only load them once)
- [x] Utilize folder structure on the nextcloud-side

Documentation:
- [x] needs some documentation what exactly the webdav url in the config.json should be. (for Nextcloud it seems to be "nextcloud-domain.com/remote.php/webdav")
- [x] needs documentation that the Nextcloud-Server has to be tweaked to allow the remote connections

Uploading:
- [ ] no error message displayed if upload fails (e.g. because credentials are not set)
- [x] file extension (e.g. ".png" gets lost during upload), therefore Nextcloud website doesn't recognize the uploaded image when browsing there
(Note: We would have to save the image file type somewhere to retrieve images this way.)
- [x] credentials don't get saved persistently (or decrypted correctly). After reload I am getting 401 errors from the webdav routes until I enter the credentials again on the user account form

User-Account:
- [x] no feedback after clicking "Update" to save the service credentials. We should display an alert to confirm and probably reset the form fields
- [x] it is not completely intuitive what the "Password" field in the "Image Service" section means. Maybe put something more elaborate along the lines of "enter your user-account password to encrypt and save the new cloud image service credentials"
- we could use tabs or ExpansionPanel like in the ChildDetailsComponent to separate the User Account and Image Service Account forms more clearly
- Chromium warns in the console that "Password field is not contained in a form: (More info: https://goo.gl/9p2vKq)"


src/app/app-config/app-config.model.ts
```diff
@@ -26,4 +26,10 @@ export interface IAppConfig {
     outdated_threshold_days: number;
     useTemporaryDatabase: boolean;
   };
+
+  webdav: {
+    remote_url: string;
+    username: string;
```
sleidig: This poses a security risk - even without logging in the webdav credentials can be seen, so files readable & writable by anyone. Not sure how we can resolve this - it must be done dynamically only after user authorisation somehow ...

RichardNutt: Agreed. Were you able to get the nextcloud running?

src/app/children/child-details/child-details.component.html
```diff
@@ -55,8 +56,14 @@ <h1 class="page-header section-child">
...
-          <img [src]="child.getPhoto()" class="child-pic" alt="child's photo">
+          <img *ngIf="!creatingNew && !editing" [src]="image" class="child-pic" alt="child's photo">
+
+          <input style="display: none" type="file" (change)=uploadChildPhoto($event) #fileUpload>
+          <button *ngIf="creatingNew || editing" class="child-pic-upload" (click)="fileUpload.click()">
+            <mat-icon class="upload-icon" fontIcon="fa-plus-circle"></mat-icon>
+          </button>
           <br>
           <mat-form-field *ngIf="creatingNew || editing" class='photo-filename'>
             <input matInput formControlName="photoFile" placeholder="Photo File Name" title="photoFile" type="text" >
```
sleidig: once the upload is working and tested, we can remove the input field for the file name again

src/app/children/child-details/child-details.component.ts
```diff
@@ -133,16 +138,19 @@ export class ChildDetailsComponent implements OnInit {
           this.initForm();
           this.entityMapperService.load<School>(School, this.child.schoolId)
             .then(school => this.currentSchool = school);
+          this.blobService.getImage(this.child.photoFile)
```
sleidig: maybe we could extract the image loading into a component or directive, so that this doesn't have to be repeated in child-details, child-block and other places.

src/app/webdav/blob-service.service.ts
```diff
@@ -0,0 +1,84 @@
+import { Injectable } from '@angular/core';
+import { AppConfig } from '../app-config/app-config';
+import webdav from 'webdav';
+import { DomSanitizer, SafeUrl } from '@angular/platform-browser';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class BlobServiceService {
```
sleidig: `BlobServiceService` --> `BlobService`?

src/app/webdav/blob-service.service.ts
```diff
@@ -0,0 +1,84 @@
+import { Injectable } from '@angular/core';
+import { AppConfig } from '../app-config/app-config';
+import webdav from 'webdav';
+import { DomSanitizer, SafeUrl } from '@angular/platform-browser';
+
+@Injectable({
+  providedIn: 'root'
```
sleidig: the `providedIn: 'root'` annotation overrides the `providers: []` statement in a module where you provide the service and always provides it in the root module. In that case we wouldn't need the BlobModule at all and can just use `providedIn: 'root'`. Otherwise we can remove the `providedIn` here and use the module to explicitly import it somewhere

src/app/app-config/app-config.model.ts
sleidig: We could put this into the couchdb as a config record instead. Then at least an initial successful login is required to access it.

Also, I've considered creating a second database besides the one containing all the entities that would hold all configuration and would be readonly for normal users, requiring special permissions to change anything.

RichardNutt:
Changed the nextlcloud `remote.php` file to correctly set CORS headers
[remote.zip](https://github.com/Aam-Digital/ndb-core/files/3556856/remote.zip)

src/app/app-config/app-config.model.ts
RichardNutt: Now an encrypted version of the password gets saved, which can be accessed by decrypting with the user password.

sleidig:
> Changed the nextlcloud `remote.php` file to correctly set CORS headers
> [remote.zip](https://github.com/Aam-Digital/ndb-core/files/3556856/remote.zip)

@vlle1 where do you think we should add documentation for administrators, who would need to adapt their Nextcloud server to make it work together with our app?

TheSlimvReal:
I have read through the current code and made a few notes/tasks:
- did you come up with a (secure) solution on how to treat the nextcloud password?
- The logic of saving and loading pictures should not be placed in each component. Maybe move it to the `ChildrenService` or even create a new service that could treat any kind of photos for later use
- Maybe create a little graphic that describes how the authentication with the nextcloud works/should work and also how the procedure of storing and loading works
- what exactly should be happening when the user is offline? Will it be just like before?
- Do we need to create an option which restrics the loading of picture if someone does not have the proper bandwidth?

RichardNutt:
> * did you come up with a (secure) solution on how to treat the nextcloud password?
Currently the password is encrypted with each users password and then stored. This is something I would change in the future, but i'd like to finish this feature first.

> * The logic of saving and loading pictures should not be placed in each component. Maybe move it to the `ChildrenService` or even create a new service that could treat any kind of photos for later use
What do you mean by this? How would a new service be different than the blob-service? (Yes, image should be loaded once and then stored somewhere, as noted by @sleidig.)

> * Maybe create a little graphic that describes how the authentication with the nextcloud works/should work and also how the procedure of storing and loading works
Added to ToDo (Documentation).

> * what exactly should be happening when the user is offline? Will it be just like before?
Yes.

> * Do we need to create an option which restrics the loading of picture if someone does not have the proper bandwidth?
I would rather convert the image to lower resolutions that can be loaded if the user has low downstream speed. Again this is something I would look into after this feature works.

Also I don't know if there is a better method to supply node modules than editing the webpack-configs. Feels a bit hacky :)

TheSlimvReal:
> What do you mean by this? How would a new service be different than the blob-service? (Yes, image should be loaded once and then stored somewhere, as noted by @sleidig.)

I am talking about this code section in the `ngOnInit` method which is also marked as a duplicate code by GitHub:
```
this.blobService.getImage(this.entity.getId().replace('child:', ''))
        .then( arrayBuffer => this.image = this.blobService.bufferArrayToBase64(arrayBuffer));
```

Otherwiese I would say we can probably merge this branch with its current state after a little testing and we create some issues for the other points mentioned. Good work.

src/app/webdav/blob-service.service.ts
```diff
@@ -0,0 +1,115 @@
...
+   */
+  public async getDir(path: string): Promise<string> {
+      const contents = await this.client.getDirectoryContents(path);
+    return JSON.stringify(contents, undefined, 4);
+  }
+
+  public async doesFileExist(name: string): Promise<boolean> {
+    if (!this.fileList && !this.currentlyGettingList) {
+      this.currentlyGettingList = new Promise((resolve, reject) => {
+        this.getDir('').then(list => {
```
TheSlimvReal: you can also use await here to make the code a little more readable

src/app/children/notes/notes.component.ts
```diff
@@ -49,8 +49,9 @@ export class NotesComponent implements OnInit {
 
 
   generateNewRecordFactory() {
-    // define values locally because "this" is a different scope after passing a function as input to another component
-    const user = this.sessionService.getCurrentUser().name;
+    // define values locally because "this" is a different scope after passing a function as input to another component#
+    // TODO: currentUser undefined in tests
```
sleidig: if it's just required for tests, we should try to mock the currentUser value.
But I think your change makes the code more stable in general and is a good addition :+1:

src/app/user/user.ts
```diff
@@ -40,11 +46,18 @@ export class User extends Entity {
...
     this.password = {'hash': hash, 'salt': cryptSalt, 'iterations': cryptIterations, 'keysize': cryptKeySize};
+
+    // update encrypted nextcloud password
+    this.blobPasswordEnc = CryptoJS.AES.encrypt(this.blobPasswordDec, password).toString();
   }
 
   public checkPassword(givenPassword: string): boolean {
     // compares given password to the stored one of this user
     // therefore hashes the given password string and compares it with the sored hash
+    if  (this.hashPassword(givenPassword) === this.password.hash) {
```
sleidig: I think writing a getter for `blobPasswordDec` would be better than doing the decryption here. The method name `checkPassword` gives the impression that there will be no side effects

src/app/webdav/blob-service.service.ts
```diff
@@ -0,0 +1,131 @@
+import { Injectable } from '@angular/core';
+import { AppConfig } from '../app-config/app-config';
+import webdav from 'webdav';
+import { DomSanitizer, SafeUrl } from '@angular/platform-browser';
+import { SessionService } from 'app/session/session.service';
+
+@Injectable()
+export class BlobService {
```
sleidig: `BlobService` isn't making it obvious what this does. Now the service seems tightly related to webdav and saving/loading images, maybe it could be `WebdavFileService` or `CloudFileService` or something similar?

src/app/webdav/blob-service.service.ts
```diff
@@ -0,0 +1,131 @@
...
+  public connect(password: string = this.sessionService.getCurrentUser().blobPasswordDec) {
+    // clear the promise that retrieves the root dir
+    this.currentlyGettingList = null;
+    this.fileList = null;
+
+    if (this.sessionService.getCurrentUser() != null) {
+      this.client = webdav.createClient(
+        AppConfig.settings.webdav.remote_url,
+        {
+          username: 'nextclouduser',
```
sleidig: we might consider letting users link personal nexcloud accounts (with their own username). At least for HELGO people already have their own Nextcloud account as well.

src/app/webdav/blob-service.service.ts
```diff
@@ -0,0 +1,131 @@
...
+  /**
+   * Reinitialize the client for the nextcloud server
+   * @param password login password
+   */
+  public connect(password: string = this.sessionService.getCurrentUser().blobPasswordDec) {
+    // clear the promise that retrieves the root dir
+    this.currentlyGettingList = null;
+    this.fileList = null;
+
+    if (this.sessionService.getCurrentUser() != null) {
```
sleidig: if currentUser is null, wouldn't that crash the system if it's trying to set the default value `connect(password: string = this.sessionService.getCurrentUser().blobPasswordDec)`? Maybe better to move that assignment into the method here inside the if block.

src/app/webdav/blob-service.service.ts
```diff
@@ -0,0 +1,131 @@
...
+import { SessionService } from 'app/session/session.service';
+
+@Injectable()
+export class BlobService {
+
+  // TODO Check connection/login success?
+  private client: any;
+  // private defaultImage: SafeUrl;
+  private fileList: string;
+  private currentlyGettingList: Promise<boolean>;
```
sleidig: This might be a good use case for a `BehaviorSubject<boolean>` (https://www.learnrxjs.io/subjects/behaviorsubject.html)

A BehaviorSubject is basically an Observable with a default value. It's a little confusing to wrap your head around in the beginning but it can be quite useful.

EDIT: made a commit with suggestion to simplify this.

patch-webpack.js
```diff
@@ -0,0 +1,13 @@
+const fs = require('fs');
+const f = 'node_modules/@angular-devkit/build-angular/src/angular-cli-files/models/webpack-configs/browser.js';
+
+fs.readFile(f, 'utf8', function (err,data) {
```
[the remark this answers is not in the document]
RichardNutt: Otherwise `ng build --prod` fails, see [here](https://travis-ci.org/Aam-Digital/ndb-core/builds/584295792?utm_source=github_status&utm_medium=notification). Let me know if you know a cleaner way to circumvent this.

src/app/children/child-block/child-block.component.spec.ts
```diff
@@ -10,10 +10,20 @@ import {MockDatabase} from '../../database/mock-database';
...
+import { MatSnackBar } from '@angular/material/snack-bar';
+import { Overlay } from '@angular/cdk/overlay';
+import { userInfo } from 'os';
+import { User } from 'app/user/user';
+import { MockBlobService } from 'app/webdav/mock-blob-service';
 
 describe('ChildBlockComponent', () => {
   let component: ChildBlockComponent;
   let fixture: ComponentFixture<ChildBlockComponent>;
+//  let sessionService = { getCurrentUser: () => new User('TestUser')};
```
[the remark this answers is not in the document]
RichardNutt: removed that part. Regarding the password: There is currently only one nextcloud account for all pictures/users. The password gets encrypted with each users login password and then gets saved to the DB.

src/app/children/child-details/child-details.component.ts
```diff
@@ -191,4 +193,14 @@ export class ChildDetailsComponent implements OnInit {
...
     this.location.back();
   }
+
+  /**
+   * hands over the selected file to the blobService together with the childID
+   * @param event
+   */
+  uploadChildPhoto(event) {
+    console.log(event);
+    this.blobService.setImage(event.target.files[0], this.child.getId().replace('child:', ''));
```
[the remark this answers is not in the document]
RichardNutt: Will fix that tomorrow. Thanks :)

src/app/children/child-details/child-details.component.ts
```diff
@@ -52,6 +54,7 @@ export class ChildDetailsComponent implements OnInit {
   eyeStatusValues = ['Good', 'Has Glasses', 'Needs Glasses', 'Needs Checkup'];
   vaccinationStatusValues = ['Good', 'Vaccination Due', 'Needs Checking', 'No Card/Information'];
 
+  image: SafeUrl;
```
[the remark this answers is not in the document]
RichardNutt: No, removed it. In a previous version photos weren't stored with each child and thus was saved here.

src/app/children/notes/notes.component.ts
RichardNutt: Agreed, but I probably won't fix this on this branch, since its tangential to the blob service. (I will open a new issue for this.)

src/app/user/user.ts
RichardNutt: removed that part; made decryptBlobPassword() public and used it in session service.

src/app/webdav/blob-service.service.ts
RichardNutt: I will rename it.

src/app/webdav/blob-service.service.ts
RichardNutt: The proposed fix currently breaks the service. I will look into it though.

src/app/children/notes/notes.component.ts
sleidig: as I said, your change is good in general - so I think we can also just remove the TODO here

src/app/webdav/blob-service.service.ts
RichardNutt: `  public async doesFileExist(name: string): Promise<boolean> {
    const fileList = await this.getFileList();`
does not work for me. Reverted the change for now.

src/app/children/notes/notes.component.ts
RichardNutt: Ok :) Will do that.

src/app/app-config/app-config.model.ts
RichardNutt: Currently the cloud-service password has to be entered manually on the user-account page. It then gets encrypted with the login password as key and afterwards the encrypted password gets saved in the database.

src/app/children/attendance/add-day-attendance/add-day-attendance.component.spec.ts
```diff
@@ -28,8 +30,9 @@ describe('AddDayAttendanceComponent', () => {
         MatIconModule, MatProgressBarModule, NoopAnimationsModule,
         EntityModule],
       providers: [
-        {provide: ChildrenService, useClass: ChildrenService},
+        ChildrenService,
```
sleidig: we should mock the ChildrenService to make it better testable and avoid having to provide all those dependencies of ChildrenService in these tests. I started refactoring for one test: https://github.com/Aam-Digital/ndb-core/pull/306/commits/60d838ef1211c45f69d8686d179303f2b8c51415

src/app/webdav/cloud-file-service.service.spec.ts
```diff
@@ -1,23 +1,72 @@
...
+    });
+
+    cloudFileService = TestBed.get(CloudFileService);
+    cloudFileService['client'] = clientSpy;
+    sessionService = TestBed.get(SessionService);
+  });
+
+  it('.connect() should call sessionService.getCurrentUser()', () => {
+    cloudFileService.connect('user', 'pass');
+    expect(sessionService.getCurrentUser).toHaveBeenCalled();
```
sleidig: is the call to `getCurrentUser` really the relevant thing to test? Isn't the call to webdav or the creation of the client object what really matters from an outside perspective?

src/app/webdav/cloud-file-service.service.spec.ts
```diff
@@ -1,23 +1,72 @@
...
+    spyOn(cloudFileService, 'getDir').and.returnValue(new Promise((resolve, reject) => {resolve('"basename": "filename"'); }));
+    expect(await cloudFileService.doesFileExist('filename')).toBe(true);
+    expect(await cloudFileService.doesFileExist('nonexistant')).toBe(false);
+  });
+
+  it('should get images', async() => {
+    spyOn(cloudFileService, 'doesFileExist').and.returnValue(new Promise((resolve, reject) => {resolve(true); }));
+    await cloudFileService.getImage('filepath');
+    expect(clientSpy.getFileContents).toHaveBeenCalledWith('filepath');
+  });
```
sleidig: you could add one more test for the case when there is no image and the default image should be returned

sleidig:
A few observations from functional user testing:

User-Account:
- I like the solution with the account settings as part of the user page :+1: 
- [ ] no feedback after clicking "Update" to save the service credentials. We should display an alert to confirm and probably reset the form fields
- [ ] it is not completely intuitive what the "Password" field in the "Image Service" section means. Maybe put something more elaborate along the lines of "enter your user-account password to encrypt and save the new cloud image service credentials"
- we could use tabs or ExpansionPanel like in the ChildDetailsComponent to separate the User Account and Image Service Account forms more clearly
- Chromium warns in the console that "Password field is not contained in a form: (More info: https://goo.gl/9p2vKq)"

RichardNutt:
I added the bullet points to the TODO-List. Thank you :)

sleidig:
sorry for the merge conflicts. This is because we moved all the folders around a little bit (#394). Maybe the easiest way to fix it is to take the folders you changed and copy them to their new location of the master branch (replacing the files that are there and don't include your changes yet). If you get stuck, let me know.

Issue #268, opened 2019-05-19
create documentation


vlle1:
# general todos
- [x] remove redundant documentation, keep number of places small.
- [x] make clear differentiations between documentation for developers (->github pages) and documentation for users (->github wiki)

# developer-oriented documentation
- [x] write instructions how to use compodoc in wiki #276 
- [ ] **document more classes using standard JSDoc** (see 'documentation coverage' in generated compodoc), this is automatically exported to the compodoc. #277 
- [x] automatically use documentation via github pages @TheSlimvReal 
- [x] transfer as much as possible from /doc/technical to markdown (/doc/compodoc_sources) #279 
- [ ] add graphic for routing to the compodoc (see https://compodoc.app/guides/routing.html) #278
- [x] change branch from where compodoc is generated in .travis.yml back to 'master'.
# user-oriented documentation
- [x] translate and add specification sheet
- [x] update functional requirements (in specification sheet) #280 
- [x] create 'how to' file from functional  requirements #281 
- [x] copy wiki files into '/doc/wiki' folder (wiki can later be automated)
- [x] **automate wiki**:  automatically add markdown files from 'doc/wiki' to github wiki page @TheSlimvReal

sleidig:
So where are the docs on GitHub Pages actually? What's the URL?

It probably makes sense to close and delete the existing wiki completely (at least for developer documentation) in favour of the new docs.

vlle1:
* you can find the new documentation in this github repository under http://aam-digital.github.io/ndb-core/ or just navigate in this repository to > Code > envrionment.
* we intend to use the github wiki not for developing the project but only for how to use the 'product'.

sleidig:
I love the compodoc docs! :+1: 

I'd suggest to rename the compodoc section "developer wiki" to avoid confusions between that and the GitHub Wiki. Maybe the "developer wiki" section should actually become three sections (following https://www.divio.com/blog/documentation/): "Tutorials", "How Tos" and "Explanations" (maybe there are better titles for them).
Would you delete the GitHub Wiki pages if they are in compodocs, so that we don't keep parallel content?

Is it possible to move the generate sections (Modules, Classes, Injectables, Guards, Interfaces, Miscellaneous, Documentation Coverage) under a top-level menu entry called "Reference" (or a similar name)?

vlle1:
hey, removing duplicate is part of this issue already, but I also moved the other ideas of you to another issue (see #293 )

sleidig:
[...]

PR #320, opened 2019-07-11
[WIP] fixed links between articles in developer guide and links to used images

vlle1:
minor changes, onlyaffecting developer guide (compodoc).


.travis.yml
```diff
@@ -64,4 +64,4 @@ deploy:
     local_dir: doc/compodoc
     github_token: $GH_REPO_TOKEN
     on:
-      branch: master
+      branch: compodoc
```
sleidig: won't that prevent updates of the docs from the master branch, e.g. when in the future a PR is adding new code as well as the related documentation to the master?

Should this maybe become 
```
branch:
  - master
  - compodoc
```
?

vlle1: yeah we tried to do something like this (@TheSlimvReal  is the expert concerning travis)...
unfortunately that didn't work (syntactically a string is required) and writing 'branches: [...] ' fixed that issue syntactically (no error is thrown), but instead, the command is never executed..
So my plan was to keep 'compodoc' for now so that I can look at the outcome without merging into master all the time. I see no big issue there (functionality/...) and added changing it back later to my todo in the issue #268 . If you don't like that solution we could also leave out that condition and override the compodoc on each branch (leads to far more inconsistency).

sleidig: ah, I see. Seems like you can set multiple branches in `on:` but a regex could be a workaround:

```
on:
  all_branches: true
  condition: ${TRAVIS_PULL_REQUEST_BRANCH:-$TRAVIS_BRANCH} =~ ^(master|compodoc)$
```
https://stackoverflow.com/a/48959539/1473411

vlle1: Ok, I just used your code snippet, it seems to do the right thing. 
I didn't know that you could use regex there. Thanks.

I'd still be interested if this works as well:
```if: branch IN (compodoc, master)``` (https://docs.travis-ci.com/user/conditions-v1)

**Comment by sleidig** [comment]
Shall we merge these changes so far and open a new PR if more are necessary?

File: doc/compodoc_sources/guidelines/documentation.md [hunk 2]
```diff
@@ -1,9 +1,9 @@
...
+# Documentation for Developers - how to contribute
 * with every push to the master, travis generates this Documentation of the Project via Compodoc automatically. **All you have to do is to Comment your code with basic JSDoc!** Further information: https://compodoc.app/guides/comments.html
 * New code should always be documented while writing!
-* If you want to **add or edit separate additional documentation** like this page, work on branch compodoc and edit the markdown files under doc/compodoc_sources. Make sure to kee the summary.json up-to-date.
+* If you want to **add or edit separate additional documentation** like this page, work on branch compodoc and edit the markdown files under doc/compodoc_sources. If you add a new markdown file you also have to add it to summary.json.
 
-###additional information
+### additional information
 * the script "compodoc" in the package.json is used by travis to generate the documentation!
 * Install compodoc manually with npm i -save -dev @compodoc/compodoc
```
TheSlimvReal: for console input please use markdowns `>` highliting:
- Install compodoc manually with
> npm i -save - dev \@compodoc/compodoc

This makes it more clear where the command begins/ends and gives the user the ability to just 'scan' the document for the command he is looking for.

doc/compodoc_sources/guidelines/documentation.md
```diff
@@ -1,9 +1,9 @@
...
 * run compodoc manually (you should instead use the documentation provided online!) with npm run compodoc (this calls the script compodoc in the package.json).
```
TheSlimvReal: maybe change _(you should instead use the documentation provided online!)_ to _(the latest documentation for the master can always be found online)_ to make more clear what we are trying to say with this comment.

Issue #325, opened 2019-07-18
Saving status information


Lorsbyjm:
Saving a child can take some time and you are not notified when it is done. You only see that the childs' name is displayed in the headline and that there is an edit button in the top right corner.

codingfabi:
I implemented a notification in the Branch ChildSavingStatus

Issue #324, opened 2019-07-18
Validation of child attributes when saving


Lorsbyjm:
Right now there is no validation and every value can be empty. It is nicer if there are some required attributes (at least in the frontend) like name, date of birth ...

codingfabi:
I implemented some validations in the branch ChildSavingStatusAndFormValidation and opened a PR for it. Please add some comments on the PR if we need more validation with the form.

codingfabi:
Should we implement a saving notification for other entitys aswell? we could implement the alert service into the basic entity save method in the EntityManagerService...

Thoughts @sleidig @liwde @TheSlimvReal ?

liwde:
Hm. In principle, I'd say yes, validation might be useful in a central place. Maybe there is already something in the EntitySchemaService, that allows annotating restrictions on certain fields of an entity?

However, you bring this already into the form that doesn't necessarily "know" about the entity. And this is absolutely the right thing, as then Angular can handle most of the validation without needing to go to the entity. So: thinking about it in a central place is good, but we'd still need to implement it everywhere else in the forms as well -- at least that is what I'm thinking...

sleidig:
I am a big fan of a generalized solution. But we shouldn't mix the data/logic oriented services with the UI providing alerts to users. Maybe slowly we can work towards a generic component as the base for all forms (ChildDetails, SchoolDetails, ...) that includes something like this?

Also, from a UX perspective form field validation with an error message at each field would be awesome in the long run (see [live example](https://angular.io/generated/live-examples/form-validation/stackblitz.html) from the [Angular Docs](https://angular.io/guide/form-validation))

PR #327, opened 2019-07-25
Child saving status and form validation

codingfabi:
see issues: #324 and #325 

### Visible/Frontend Changes
- [x] Added a alert for succesfull saving of a child and a failed form validation

- [x] Added a validation for the child editing form to make some of the information required


src/app/children/child-details/child-details.component.ts
```diff
@@ -75,12 +75,12 @@ export class ChildDetailsComponent implements OnInit {
     this.form = this.fb.group({
       name:           [{value: this.child.name,           disabled: !this.editing}, Validators.required],
       // gender:         [{value: this.child.gender}], // reactive forms seem broken for mat-select, using ngModel instead
-      projectNumber:  [{value: this.child.projectNumber,  disabled: !this.editing}],
-      dateOfBirth:    [{value: this.child.dateOfBirth,    disabled: !this.editing}],
+      projectNumber:  [{value: this.child.projectNumber,  disabled: !this.editing}, Validators.required],
```
sleidig: The users may want to add candidates to the app, who are not enrolled in the project yet and don't have an official projectNumber. They could assign some other temporary numbers of course but it would be better to not enforce artificial constraints.

I think apart from name only the center makes sense to be required. Other information may not be available to the social workers yet when they first enter a child

src/app/children/child-details/child-details.component.ts
```diff
@@ -154,10 +155,14 @@ export class ChildDetailsComponent implements OnInit {
           this.router.navigate(['/child', this.child.getId()]);
           this.creatingNew = false;
         }
+        this.alertService.addInfo('Saving Succesfull');
         this.switchEdit();
       })
       .catch((err) => this.alertService.addDanger('Could not save Child "' + this.child.name + '": ' + err));
+  } else {
+    this.alertService.addDanger('Form ist not valid!!');
```
sleidig: At the moment, it is unclear (or even impossible?) for the user to see what he/she should do to make the form valid.

We have to mark the required form fields visibly as "required" and it would be much easier for users if we highlight the individual invalid field(s) with a specific error message, as in the [live example](https://angular.io/generated/live-examples/form-validation/stackblitz.html) from the [Angular Docs](https://angular.io/guide/form-validation)), for example.

codingfabi: Good point, I'll look further into that :)

src/app/children/child-details/child-details.component.ts
codingfabi: i removed the _required_ tags from all attributes except for name and center

codingfabi:
Allright, so I tried to implement the requestet changes and mark the required fields, but I didn't manage to find a proper solution the styling just doesnt work. 
Here is what i tried: 

- I wanted to implement the styling use in the [live demo](https://stackblitz.com/angular/lmmqnkkarmbg) of angular that sleidig recommendet, but it didnt work out. 
- I tried to implement the warning messages from PrimeNG, with wich if worked before and actually like. But the styling failed again.
- So my workaround was to just mark the invalid fields as red and put a hint into the message thrown by the allert service.

I would love to merge this PR ASAP, since the feature is kind of working and i'll be on vacation for three weeks. I will have a look at a proper implementation when I return, thats why i left the code for the PrimeNG Messages etc in the branch. We can discuss if it makes sense to import another node package, I would put this feature into production because i think it is really helpfull, even without perfect styling.

src/app/children/child-details/child-details.component.ts
```diff
@@ -154,10 +158,22 @@ export class ChildDetailsComponent implements OnInit {
           this.router.navigate(['/child', this.child.getId()]);
           this.creatingNew = false;
         }
+        this.alertService.addInfo('Saving Succesfull');
         this.switchEdit();
       })
       .catch((err) => this.alertService.addDanger('Could not save Child "' + this.child.name + '": ' + err));
+  } else {
+    let errorstring = 'Form is not valid: \n';
+    if (this.form.get('name').invalid) {
```
TheSlimvReal: I think this if-clauses can be changed to a more general behaviour

src/app/children/child-details/child-details.component.html
```diff
@@ -69,14 +69,24 @@ <h1 class="page-header section-child">
           <mat-form-field style='width: 300px;'>
             <input matInput formControlName="name" placeholder="Name" title="name" type="text">
           </mat-form-field>
-
+<!--           <div *ngIf="this.form.get('name').touched && this.form.get('name').invalid" class="alert alert-danger">
```
TheSlimvReal: Will this be re-implemented?

src/app/children/child-details/child-details.component.ts
```diff
@@ -80,7 +83,7 @@ export class ChildDetailsComponent implements OnInit {
       motherTongue:   [{value: this.child.motherTongue,   disabled: !this.editing}],
       religion:       [{value: this.child.religion,       disabled: !this.editing}],
 
-      center:         [{value: this.child.center,         disabled: !this.editing}],
+      center:         [{value: this.child.center,         disabled: !this.editing}, Validators.required],
```
TheSlimvReal: why exaclty is the center required?

codingfabi: You've got to ask @sleidig on this one...

sleidig: as we use the `center` for basic filters and grouping it somewhat makes sense to require it. we could also keep it optional as well, however.

codingfabi:
I finally brought this to a mergable state, so if either @sleidig or @TheSlimvReal could give a final review, I would be really thankfull

karma.conf.js
```diff
@@ -31,7 +31,10 @@ module.exports = function (config) {
       require('@angular-devkit/build-angular/plugins/karma')
     ],
     client:{
-      clearContext: false // leave Jasmine Spec Runner output visible in browser
+      clearContext: false, // leave Jasmine Spec Runner output visible in browser
+      jasmine: {
+        random: false
+      }
```
sleidig: Do you really need to deactivate random order of unit tests?

It can certainly be a pain in the ass during development but unit tests should be independent of each other so randomization shouldn't break anything and might help to discover unintended side effects.

src/app/child-dev-project/children/child-details/child-details.component.ts
```diff
@@ -41,11 +41,14 @@ export class ChildDetailsComponent implements OnInit {
...
   schools: School[] = [];
 
+  validateForm = false;
   form: FormGroup;
   healthCheckForm: FormGroup;
   creatingNew = false;
   editing = false;
   gender = Gender;
+  nameinvalid: boolean;
+  centerinvalid: boolean;
```
sleidig: are these actually used?

src/app/child-dev-project/children/child-details/child-details.component.html
```diff
@@ -69,14 +69,25 @@ <h1 class="page-header section-child">
...
           </mat-form-field>
+          </div>
 
+          <div>
           <mat-form-field>
             <input matInput formControlName="center" placeholder="Center" type="text">
           </mat-form-field>
+          <div *ngIf="this.form.get('center').invalid && validateForm" class="validationWarning">
+            Center is required
+          </div>
```
sleidig: when looking into the styling to move the error message a little closer to the field, I found the documentation for mat-error: https://material.angular.io/components/form-field/overview#error-messages
This is probably what we would want to use instead of our own div. Or do you have a specific reason to do it this way?

mat-error would also automatically use an optimized styling:
image

codingfabi: i implemented mat-error and it worked fine

src/app/child-dev-project/children/child-details/child-details.component.ts
codingfabi: no, i cleaned it up

karma.conf.js
codingfabi: to be honest, i cant even remember why I did this in the first place. I changed it back

codingfabi:
Okay so I implemented your requested changes and everything is working out fine, except for the codeclimate checks. @sleidig @TheSlimvReal please review this again and feel free to merge if the codeclimate checks can be ignored.

PR #333, opened 2019-08-08
feat(demo): add services to auto-generate demo data

sleidig:
Generate more realistic dummy data for demo purposes and implement are modular, flexibly extendable system for other modules to integrate their DemoDataGenerators.

* [x] Implement a module DemoData generator system
* [x] Documentation how to implement new demo data generators

* [x] Demo data for Children
* [x] Demo data for Schools
* [x] Demo data for ChildSchoolRelations
* [x] Demo data for Notes
* [x] Demo data for AttendanceMonths
* [x] Demo data for Aser
* [x] Demo data for EducationalMaterials
* [x] Demo data for HeightWeight


sleidig:
I still need to implement generators for a few more entity types but appreciate a review of the general system and how it can be simplified or made more clear.

Should we add a README to the DemoData Module folder? (In addition to docs for compodoc)

liwde:
I like this system! The faker library seems really powerful and the right thing to use here.

It took me a few seconds to understand how the ChildSchoolRelations can build on top of existing schools and children, but using Angular's injection system is a great way of doing so.

I'm not sure about monkey-patching faker to support birthdays, but that's a really minor thing.

sleidig:
Thanks for the feedback!

Totally agree about the monkey patching, I'll work on wrapping the whole thing in a proper Angular Service. Also in general, maybe we can manage to simplify the system further

Issue #281, opened 2019-05-23
create 'how to' file from functional requirements


Lorsbyjm:
Write an How to for each functional requirement off the application

Lorsbyjm:
see  https://www.divio.com/blog/documentation/

PR #342, opened 2019-08-22
How to docu

Lorsbyjm:
see issue: #281 

### Visible/Frontend Changes
- [x] added new Help section to navigation
- [x] Help section displays the how to


Lorsbyjm:
Had to upload the package-lock.json here because otherwise the changes made in package.json are not used. Is this ok? @liwde @sleidig

src/app/app.module.ts
```diff
@@ -45,10 +45,13 @@ import {environment} from '../environments/environment';
...
 import {AdminModule} from './admin/admin.module';
 import {CookieService} from 'ngx-cookie-service';
+import { HelpComponent } from './help/help.component';
+import { MarkdownModule} from 'ngx-markdown';
 
 @NgModule({
   declarations: [
-    AppComponent
+    AppComponent,
+    HelpComponent
```
sleidig: I would prefer to make this a separate module. That way it will be easier to include, replace or remove the feature in the future, if necessary. `AppModule` really is just tying things together without providing own functionality. In the future we might have separate AppModule implementations for different user organisations also.

src/app/help/help.component.html
```diff
@@ -0,0 +1,114 @@
+
+<div markdown ngPreserveWhitespaces>
+  # How to ...?
```
sleidig: How can we avoid the duplication between this an the How_To.md file?

sleidig:
> Had to upload the package-lock.json here because otherwise the changes made in package.json are not used. Is this ok? @liwde @sleidig

adding the latest package-lock.json to the commit is correct. package.json and package-lock.json should be committed together.

The package-json.lock is fixing the exact versions and dependencies while package.json is defining a broader range of allowed versions. We should keep both up to date in the repo.

src/app/help/help.component.html
Lorsbyjm: <markdown [src]="'path/to/file.md'" (load)="onLoad($event)" (error)="onError($event)"></markdown>
 
You can bind the file, but that was not working for me. Always have problem with the relative path.

src/app/app.module.ts
Lorsbyjm: Ok, I created a new module and moved the component in it.

src/app/help/help.component.html
Lorsbyjm: just used the wrong path :/. Now it works.

angular.json
```diff
@@ -26,7 +26,7 @@
               "src/ndb-theme.scss",
               "src/styles.css"
             ],
-            "scripts": []
+            "scripts": ["node_modules/marked/lib/marked.js"]
```
sleidig: why do we need this? (i.e. what does it actually do?)

src/app/help/help.component.html
sleidig: awesome!

we still have two copies of the .md file in the repo, however. Can we let compodoc somehow grab it from the assets folder maybe?

doc/wiki/How_To.md
```diff
@@ -0,0 +1,111 @@
+# How to?
+## This document should help you as a user to understand all functionalities of the application.
+
+##Introduction
+After logging in you will be navigated to the dashboard screen.
+Here you see the navigation at the left border and some fancy dashboards in the content area.
+They give you some information about children and can be defined by you, but later more about that (Verlinkung section Dashboard).
```
sleidig: `(Verlinkung section Dashboard)`

doc/wiki/How_To.md
```diff
@@ -0,0 +1,111 @@
...
+1. On the Children List Screen click on the button with the plus (and the text "Add New") at the upper right side.
+  This will open an empty child view
+
+2. Fill in the needed attributes for the new child.
+After this click "Save" in the top right corner.
+You can see that saving was successful when the child's name is the headline.
+
+### ... edit a child?
+1. On the Children List Screen search for the child you want to edit and click on the corresponding entry in the list.
+It is the best to click on the name or the project number (PN), cause clicking on the school will navigate you to the school's detail page.
```
sleidig: maybe we should consider changing this confusing behaviour? ;-)

sleidig:
This looks good and is working well!

angular.json
Lorsbyjm: As the library is using marked parser you will need to add node_modules/marked/lib/marked.js to your application.

If you are using Angular CLI you can follow the angular.json example below...

"scripts": [
+ "node_modules/marked/lib/marked.js"
]

doc/wiki/How_To.md
Lorsbyjm: Sure, create a new issue ;)

src/app/help/help.component.html
Lorsbyjm: I think so. I just forgot the one in the doc folder

doc/wiki/How_To.md
```diff
@@ -0,0 +1,150 @@
...
+
+
+## Children
+The *Children* section gives you all information about the children you interact with.
+In the list you can search and filter the records and compare different kinds of information.
+The child details page shows all information across the different sections that is related to the selected child.
+
+To enter the Children section click on the button in the navigation. You will be navigated to the Children List Screen.
+
+###... create a child?
```
TheSlimvReal: missing whitespace

src/assets/HowTo.md
```diff
@@ -0,0 +1,111 @@
+# How to?
```
TheSlimvReal: Delete this or the other `HowTo.md` and use one file for both app and documentation

doc/wiki/How_To.md
Lorsbyjm: There is no dashboard section right now. Should create a new issue for that.

PR #349, opened 2019-10-11
DIsable closing the menu on ESC

MoritzGiessmann:
Hi!

We spoke at #DigitaleNachtKA, and I didn't find any issue matching the "escape closes navigation" problem. Besides having 500MB of dependencies this might be the first repo I ever checked out and it worked out of the box after following the README.md. Thumbs up for that!

### Visible/Frontend Changes
- [x] Sidebar navigation no longer closes on pressing escape. It still works

### DIFF
--- a/src/app/ui/ui/ui.component.html
+++ b/src/app/ui/ui/ui.component.html
@@ -46,7 +46,7 @@
 <mat-sidenav-container class="app-content mat-typography">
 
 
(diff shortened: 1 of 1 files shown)


sleidig:
Hi Moritz, thanks a lot for chipping in with the bugfix! If you like, you are more than welcome to contribute from time to time - we appreciate every hand helping us to make life easier for social workers :-)

Thanks for the feedback also. Yeah, these npm dependencies with all their sub-dependencies somehow tend to get huge. Glad that the setup worked as planned, though.

Issue #126, opened 2018-06-06
Export of data from EntityListComponents / DataTables


sleidig:
The user should be able to simply export (download) the data from any data-table or list component in order to continue working with the data outside of our app. For this a simple .csv export should be accessible from the UI.

see #125 (admin/backup.service.ts) for possible implementation. Probably inject the `PapaParseService` directly in the component and use it to parse to CSV from the data.
Ideally we should avoid duplicating code between all the different entity types' views and design something generic.

Lorsbyjm:
First try to realise the csv dowload in the UI, even with douplicate code. 
Then we can assaign this to the Milestone technical Improvment and write the generic code.
But of course you can do both in one :)

sleidig:
The csv export is already implemented in the AdminModule.
Ideally, we could develop a component that extends `MatDataTable`, adding a download button to export whatever data is displayed in a table anywhere in the app.

mspoeri:
I will play around a little with this feature but don´t want to block anyone whose willing to implement this faster as me so i´m not going to assign it to me.

sleidig:
The export is basically already implemented. Can be adapted or build upon the `BackupService` and then added to the relevant UIs.

https://github.com/NGO-DB/ndb-core/blob/67aad641c397fe130f8d58dec1975a90fc3ffc8d/src/app/admin/admin/admin.component.ts#L42-L45

https://github.com/NGO-DB/ndb-core/blob/67aad641c397fe130f8d58dec1975a90fc3ffc8d/src/app/admin/backup.service.ts#L28-L44

jana47:
You can now download some data as csv in the children list component. It's not all of the data and the code is not pretty good. But the exported file looks much better and is more user-friendly than the file exported from the admin view.
Should only the admin be able to download the data? And which components need a csv export?

sleidig:
Great that this is getting started!

Do you want to create a pull request and upload your work so far? I'll be happy to give you feedback if I can think of any improvements.

I think generally everybody should be able to download data - although we might change our minds if users start to then mostly work in excel instead of our app ;-)

Children data is most important. Schools maybe too. But in the end it would be nice to also integrate it into `EntitySubrecordComponent` so that download is available for any table.

jana47:
I created a pull request, so you can review it. I will have a look at the EntitySubrecordComponent.

PR #352, opened 2019-10-21
Csv export feature

TheSlimvReal:
see issue: #126 

The backup service was extend to create backups (.csv and .json files) from any kind of data.
This way it can be used to create a CSV file from a components local data.
If a export button should be available in a component, only the new `ExportDataComponent` needs to be included, which handels the creation of the downloadable data.

### Visible/Frontend Changes
- [x] Created `Export CSV` feature for children list


TheSlimvReal:
I also change the `tslint.json` file because of a warning message. The removed feature is now checked by the TS-Compiler

src/app/admin/export-data/export-data.component.ts
```diff
@@ -0,0 +1,47 @@
...
+    const blobData = this.getFormattedBlobData();
+    const link = this.createDownloadLink(blobData);
+    link.click();
+  }
+
+  private createDownloadLink(blobData): HTMLAnchorElement {
+    const link = document.createElement('a');
+    link.setAttribute('style', 'display:none;');
+    document.body.appendChild(link);
+    link.href = window.URL.createObjectURL(blobData);
```
liwde: We should make sure to properly destroy the object url, as it's lifecycle is bound to `document`. This might be a memory leak, if the user exports lots of things.

https://developer.mozilla.org/en-US/docs/Web/API/URL/revokeObjectURL
https://angular.io/api/core/OnDestroy

src/app/admin/services/backup.service.ts
```diff
@@ -15,35 +15,36 @@ export class BackupService {
...
-          {data: results, fields: resultFields},
-          {quotes: true, header: true});
-      });
+      .then(results => this.createCsv(results));
   }
 
-
+  createCsv(data): string {
+    const resultFields = ['_id', '_rev'];
+    data.forEach(d => {
```
liwde: I find it a bit hard to understand what this does, but am also unsure how to simplify. A really minor thing is to replace `indexOf` with `includes`. Also having mixed `.forEach()` and `for ... in` feels a bit strange.

All in all, the backup service could benefit from using async functions instead of using Promises directly, but that's just my personal opinion.

TheSlimvReal: I don't think the async await syntax would help much here. There is barely more then one asynchronous call per funtion.

src/app/admin/export-data/export-data.component.ts
TheSlimvReal: I implemented something, but I am not quite sure how to test if it actually works.

vlle1:
I promise i will review this next time (maybe) :D

PR #357, opened 2019-10-29
Add a proxy server to mitigate future CORS issues

liwde:
This will automatically proxy calls to from the dev-server to the demo backend, thus mitigating CORS issues.
Keep in mind that any non-default config.jsons must be updated as well to use the proxy.
Details on the proxy configuration can be found in the webpack docs: https://webpack.js.org/configuration/dev-server/#devserverproxy

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ Adapt the settings, especially regarding the CouchDB server that should be used 
 
 ### Development server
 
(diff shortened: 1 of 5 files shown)


liwde:
@RichardNutt, this is something you could use for your Nextcloud development as well,  isn't it?

Issue #252, opened 2019-04-25
Can't scroll to bottom


jana47:
For example on the dashboard or the attendance register I can't scroll to the bottom. Only on full screen.

sleidig:
Confirmed. Actually, this is an issue for me no matter if in fullscreen or not.

I think this has something to do with the HTML elements' size and positioning. We had similar problems before - unfortunately I can't find anything about it anymore.

sleidig:
Original bug (seems to be a problem in Firefox, Chromium displays fine):
![image](https://user-images.githubusercontent.com/1682541/68740582-af04c400-05eb-11ea-8aec-5914a67f55a4.png)

sleidig:
After PR:
![image](https://user-images.githubusercontent.com/1682541/68740774-276b8500-05ec-11ea-8dc9-c0d55a86b148.png)

PR #363, opened 2019-11-12
Change html to not overlap bottom screen

NatrixAeria:
see issue: #252 

By wrapping the contents of the `<mat-sidenav-content>` element into a div, we now have a scrollable `<mat-sidenav-content>` whose contents can be padded correctly:
```diff
- <mat-sidenav-content class="sidenav-content">
+ <mat-sidenav-content>
+   <div class="sidenav-content">
      <router-outlet *ngIf="isLoggedIn()"></router-outlet>
      <app-login *ngIf="!isLoggedIn()"></app-login>
+   </div>
  </mat-sidenav-content>
```

I hope this changes are not too hacky.

### DIFF
--- a/src/app/ui/ui/ui.component.css
+++ b/src/app/ui/ui/ui.component.css
@@ -35,7 +35,7 @@ app-ui {
   left: 0;
   right: 0;
   bottom: 0;
-  padding-bottom: 30px;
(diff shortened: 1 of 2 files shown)


sleidig:
Do you have any idea why removing the styles and adding the div actually helped?

Issue #326, opened 2019-07-25
Remove empty language filter in School List filter panel


Lorsbyjm:
The filter for languages has one empty element, this could be deleted.
~~Furthermore the langue should always be englisch (Deutsch -> German)~~

![image](https://user-images.githubusercontent.com/9371498/61896809-033d4880-af16-11e9-8ce0-f9cffb042671.png)

sleidig:
I think the filter use dynamically generated from the data. So it's probably just a matter of changing the child details of existing entities?

PR #375, opened 2019-11-26
Do not create empty media filters in school list

NatrixAeria:
see issue: #326 

`ngOnInit` does not create media that are empty for the filter selection any more.


sleidig:
Cool :+1: 
Can you implement the same fix for the "Center" filter of the ChildrenListComponent also?

Issue #328, opened 2019-07-25
Saving new note has no visible reaction


Lorsbyjm:
If you create a new note an click save the note is not displayed in the list. If you leave the section and come back the note is shown. Should trigger a reaload after saving a new issue ?!

sleidig:
Confirmed. This is a bug and has to be fixed. If we can, avoiding a complete reload would be nice as the production systems can have a lot of Notes, leading to noticeable loading times.

PR #376, opened 2019-11-26
Show recently added note after creation

Ma27:
When adding a new Note, it appears in the list of notes immediately
after saving.

see issue: #328

### Visible/Frontend Changes
- [x] Note appears in the full list after it was created

- [ ] *n/a*

### DIFF
--- a/src/app/children/notes/notes-manager/notes-manager.component.ts
+++ b/src/app/children/notes/notes-manager/notes-manager.component.ts
@@ -135,11 +135,18 @@ export class NotesManagerComponent implements OnInit, AfterViewInit {
     newNote.date = new Date();
     newNote.author = this.sessionService.getCurrentUser().name;
 
-    this.showDetails(newNote);
+    this.showDetails(newNote, true);
   }
 
 
-  showDetails(entity: Note) {
-    this.dialog.open(NoteDetailsComponent, {width: '80%', data: {entity: entity}});
+  showDetails(entity: Note, newNote: boolean = false) {
+    const nextNote = this.dialog.open(NoteDetailsComponent, {width: '80%', data: {entity: entity}});
+
+    if (newNote) {
+      nextNote.afterClosed().subscribe(val => {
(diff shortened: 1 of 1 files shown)


src/app/children/notes/notes-manager/notes-manager.component.ts
```diff
@@ -140,6 +140,10 @@ export class NotesManagerComponent implements OnInit, AfterViewInit {
 
 
   showDetails(entity: Note) {
-    this.dialog.open(NoteDetailsComponent, {width: '80%', data: {entity: entity}});
+    const nextNote = this.dialog.open(NoteDetailsComponent, {width: '80%', data: {entity: entity}});
+
+    nextNote.afterClosed().subscribe(val => {
+      this.notesDataSource.data = [val].concat(this.entityList);
```
sleidig: also, because the class property `this.entityList` is used as the starting point whenever the user changes filters, we should add the new item to this so it doesn't disappear again suddenly.

This seems to work for me:
```
nextNote.afterClosed().subscribe(val => {
      this.entityList.push(val);
      this.applyFilterSelections();
});
```

Ma27: fixed.

src/app/children/notes/notes-manager/notes-manager.component.ts
```diff
@@ -140,6 +140,10 @@ export class NotesManagerComponent implements OnInit, AfterViewInit {
 
 
   showDetails(entity: Note) {
-    this.dialog.open(NoteDetailsComponent, {width: '80%', data: {entity: entity}});
+    const nextNote = this.dialog.open(NoteDetailsComponent, {width: '80%', data: {entity: entity}});
+
+    nextNote.afterClosed().subscribe(val => {
```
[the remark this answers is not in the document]
Ma27: Ouch :man_facepalming: good catch!

Fixed, this part is now only called if a new note is added :)

Issue #368, opened 2019-11-25
Increase size of textbox of Note


sleidig:
Writing reports is one of the most used features. Users work in the popup Note Details screen, writing often 1-2 paragraphs. The textbox however is currently by default only 2 lines high and doesn't display the full report.

Set a larger default size of the main text box.
Maybe this can be relative to the total screen height to make best use of the available space while still keeping the full form (including the other fields) on the screen without scrolling?

vlle1:
I instead of changed number of rows dynamically used this tag: cdkTextareaAutosize
It basically does the same thing but cleaner (now the text box is exactly as big as the input and there never is another scrollbar, instead the scrollbar of the whole form is used) .
Also I restructured the other points to be in a better logical order and more compact (using fxFlex).

PR #377, opened 2019-11-26
change layout of note-details

vlle1:
see issue: #368


sleidig:
shifted around the layout a little more in order to highlight the most important form fields within the sea of similar looking elements.

PR #378, opened 2019-11-26
Also allow to perform login() with enter if user-field is currently selected

### LINKED ISSUE #361: Login with click "Enter"
Pressing the enter button in the LoginComponent should also trigger the login process.
Currently the button has no effect.

Ma27:
Fixes #361 

### Visible/Frontend Changes
- [x] Pressing enter on `username` field causes login.

- *n/a*

### DIFF
--- a/src/app/session/login/login.component.html
+++ b/src/app/session/login/login.component.html
@@ -28,7 +28,7 @@
 
     <p>
     <mat-form-field class="login-field">
-      <input matInput placeholder="Username" type="text" [(ngModel)]="username" required autofocus>
+      <input matInput placeholder="Username" type="text" [(ngModel)]="username" required autofocus (keyup.enter)="login()">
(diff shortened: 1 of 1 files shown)


sleidig:
Could you also add a <form> tag wrapping the login form as part of this PR? Chromium is display the following warning in the console: https://www.chromium.org/developers/design-documents/create-amazing-password-forms

sleidig:
wouldn't it actually be better to jump focus to the password field when pressing enter in the username field rather than login in?

Ma27:
> wouldn't it actually be better to jump focus to the password field when pressing enter in the username field rather than login in?

I don't think so: when I press enter in a form, I'd expect the form to be submitted (when you correct a typo in the username field and the password is already entered, the form should be submittable from the username-field IMHO). To navigate between fields, `Tab` can be used.

PR #384, opened 2019-12-03
refactor(logging): update to latest sentry sdk and use LoggerService 

sleidig:
improved (remote) logging


src/app/logging/logging.service.ts
```diff
@@ -1,13 +1,17 @@
...
-import {LogLevel  as RavenLogLevel, RavenOptions} from 'raven-js';
 import {LogLevel} from './log-level';
+import * as Sentry from '@sentry/browser';
 
 
-Raven
-  .config('https://bd6aba79ca514d35bb06a4b4e0c2a21e@sentry.io/1242399')
-  .install();
+Sentry.init({
+  dsn: 'https://bd6aba79ca514d35bb06a4b4e0c2a21e@sentry.io/1242399',
```
liwde: Should these values be moved to appconfig?

liwde: This makes it somehow very hard to catch errors during startup, as the appconfig is injected somewhere in the process. Storing the messages in some kind of queue until we initialized the client would be a bigger change... Maybe we should set up a different issue tracking this. What do you think?

sleidig: Yeah, this would actually belong into the config.json ... but when I tried now I was indeed failing to get the config ready loaded while still having things cleanly separated.
I opened a separate issue for this now (#385) so that the cleaned up logging can be release already.

Issue #381, opened 2019-11-29
Add paginatior for big tables


Schottkyc137:
At the moment, big tables used in, for example, the notes-manager or the children-list display every single element in a big scroll view at once. 
In my opinion, a better user experience can be achieved using a paginator that displays the entries throughout different pages. The first test of this has also shown better performance since not every entry has to be loaded and rendered at once. 

Personally, I do not know how those entries are being used in a temporal context. However, I can imagine that the working set might only include a few months of entries. If this is the case, and the database allows it, it might be a good idea in a second step to only load a few months worths of data and then load more if needed.

sleidig:
I am also in favour of paginated lists where the number of elements is really large. To help performance, this should not just be a UI change but actual load only part of the entities from the database (however, these need to be in the required order).

sleidig:
These long lists without pagination have become a big problem for performance (and thereby for usability). Assigning this high priority.

Issue #221, opened 2019-01-31
Refactor getNotesOfChild to EntityMapper


codingfabi:
I wanted to implement a function to get all Health Checks for a certain loaded child.
I wanted to do it familiar to the way we load notes of a child into the ChildDetailComponent. 
But i dont know how to implement a index in pouchDB that i can query. 

We should refactor and rethink the following:

```
 getNotesOfChild(childId: string): Observable<Note[]> {
    const promise = this.db.query('notes_index/by_child', {key: childId, include_docs: true})
      .then(loadedEntities => {
        return loadedEntities.rows.map(loadedRecord => {
          const entity = new Note('');
          entity.load(loadedRecord.doc);
          return entity;
        });
      });

    return Observable.fromPromise(promise);
  }

```
In my opinion, this belongs into the Entity Mapper and should be made generic.

liwde:
https://pouchdb.com/guides/queries.html

These indices affect two separate use cases:
1. Statistics
2. Traversing relations between entities

I'm fine with statistics not being part of the EntityMapper, but traversing relations is a general problem: Health Checks, Notes, ChildSchoolRelationships, ...

Shouln't the Entity to be queried by provide the map-function?

```ts
export class Note extends Entity {
  static ENTITY_TYPE = 'Note';
  static dbViews : any = [/* All Design Docs relevant for the Entity */]
}
```

The EntityMapper may then be queried like this:

```ts
this._entityMapper.getRelational<Child, Note>(childId);
```

At the point, this method is called, the EntityMapper should make sure that for both Child and Note the dbViews have been added to PouchDB and use the right index to execute the Query and construct the Entities.

(There needs to be put some more thought into having metadata for the dbViews so that the EntityMapper can pick the correct one, but this seems like a solution to me)

liwde:
@TheSlimvReal How did you implement this for the ChildSchoolRelations?

sleidig:
The idea behind our architecture of `Entity` and `EntityMapper` is the [Data Mapper pattern](https://martinfowler.com/eaaCatalog/dataMapper.html), i.e. to keep the Entity class/classes free of any database related concerns and simply focus on the "business logic" in them.

Therefore, I would try to put that stuff into `EntityMapper` (originally, the idea was also to implement subclasses of EntityMapper for different Entity type, e.g. a `ChildMapper`. But so far there was no real use for it).

[...]

Issue #145, opened 2018-07-07
Adapted Note UI for Meetings and Excursions


sleidig:
The `Note` entity is also used to document a child's participation (or absence) in meetings. At the moment users create two Notes after a meeting: one linked to all the children who were present and one linked to all the children who were absent, each with an appropriate title. To make it easier to document and understand participation the UI should be extended:

- [x] Note categories "Guardians Meeting", "Children Meeting", "Excursion/Trip" can be selected and cause an automatic change of the NoteDetails UI
- [x] Note entity uses an array with 3 fields (`childId: string`, `present: boolean`, `remarks: string`) instead of simple `children: string[]` array. By default remarks is empty, present is true.
- [x] Meeting Notes have a table of linked children in addition to the selection list. When a child is selected from the list it is added to the table where the user can select whether the child was present/absent and if necessary write a short text remark related to that child.
- [x] In the Notes table of a Child, Notes of meeting categories are displayed in a different color than other notes and differently colored for a child who was absent or present.

codingfabi:
Was this done with PR #154 ?
If so, can you close the issue?

sleidig:
No, this is a different feature related to Notes and still open.

PR #386, opened 2019-12-11
Note additions

Schottkyc137:
see issue: #221, #381, #145 

### Visible/Frontend Changes

- [x] Meeting Notes now have a table of children that were present/absent
- [x] Users can choose whether a child was present or absent and set remarks 
- [x] This table is suitable for small screens as well as big screens
- [x] The absence/presence of a single child will be reflected by suitable colors in the child's `notes-list` component
- [x] The `Note-Manager`-component uses a paginator

- [x] `Note` now is now it's own module and is no longer a sub-component of the `ChildrenModule`
- [x] A `NotesService` has been implemented to fetch resources from the `EntityMapper` and distribute updates throughout every listening component
- [x] Minor changes have been made to the `EntitySubrecord` and `ChildSelect`-components to enable these new features
- [x] Demo-Data now also incorporates a generator for meeting-notes where certain children are marked as absent/present with suitable remarks


sleidig:
sorry for the merge conflicts with the master branch, this will be mostly due to the moving of files into a new folder structure (#394).

Is this otherwise complete and ready for review and merge?

Schottkyc137:
Concerning features and other changes, this is done.
Currently, I'm only working on improving the test coverage & fixing other codeclimate errors

src/app/child-dev-project/children/children.service.ts
```diff
@@ -4,7 +4,7 @@ import { Child } from './model/child';
 import { EntityMapperService } from '../../core/entity/entity-mapper.service';
 import { AttendanceMonth } from '../attendance/model/attendance-month';
 import { Database } from '../../core/database/database';
-import { Note } from '../notes/model/note';
+import { NoteModel } from '../notes/note.model';
```
sleidig: Any particular reason to rename the `Note` class to `NoteModel`? I think we should keep this consistent for all Entity classes and personally favour the short `Note` but we can put this up for discussion

src/app/child-dev-project/children/notes-list/notes-list.component.ts
```diff
@@ -30,36 +33,35 @@ export class NotesComponent implements OnInit {
...
 
   ngOnInit() {
     this.route.paramMap.subscribe(params => {
       this.childId = params.get('id');
+    });
 
-      this.childrenService.getNotesOfChild(this.childId)
-        .subscribe(results => this.records = results.sort((a, b) => {
-          return (b.date ? b.date.valueOf() : 0) - (a.date ? a.date.valueOf() : 0); } ));
+    this.entityMapperService.loadType<NoteModel>(NoteModel).then(notes => {
```
sleidig: Loading every note and doing in-memory filtering may become a performance issue. `ChildrenService.getNotesOfChild()` uses a pouchdb database index and should (in theory) be the better choice.

Did you have any particular issues that pushed you to change this here?

src/app/child-dev-project/notes/attendance.model.ts
```diff
@@ -0,0 +1,21 @@
+
+export class AttendanceModel {
```
sleidig: This sounds very similar to our `AttendanceMonth` class and the (completely separate) system for keeping students' school attendance. Maybe `MeetingNoteAttendance` or something like that can make its purpose more clear?

src/app/child-dev-project/notes/note-detail/note-detail.component.html
```diff
@@ -0,0 +1,140 @@
...
+                               fxFlex>
+      </app-child-presence-list>
+      <app-child-presence-list [note]="entity"
+                               [recordForm]="recordForm"
+                               [present]="false"
+                               [label]="'Absent'"
+                               class="padding-bottom"
+                               fxFlex>
+      </app-child-presence-list>
+    </div>
```
sleidig: Let's extract this into a separate component as well so the responsive version can be easily re-used (and the template here will be shorter and more manageable). Possibly such a new component could also include the `app-child-select` component above and be somewhat of an alternative to the simple `app-child-select`

src/app/core/ui-helper/AbstractDetailsComponent.ts
```diff
@@ -0,0 +1,45 @@
+import { Inject, Input, OnInit, ViewChild } from '@angular/core';
+import { MAT_DIALOG_DATA, MatDialogRef } from '@angular/material/dialog';
+import { ConfirmationDialogService } from './confirmation-dialog/confirmation-dialog.service';
+import { EntityMapperService } from '../entity/entity-mapper.service';
+import { Entity } from '../entity/entity';
+
+export abstract class AbstractDetailsComponent<T extends Entity> implements OnInit {
```
sleidig: I like the idea to factor out that abstract class :+1:

src/app/child-dev-project/children/children.service.ts
Schottkyc137: I have done this to distinguish the new `NoteModel` from the old `Note` during refactoring from a component to a module. I kept the change to make it clear that this was a model. Since I agree that this is no longer needed, I will revert the changes

src/app/child-dev-project/children/notes-list/notes-list.component.ts
Schottkyc137: I had an issue with this but could resolve this in another way. Until now I thought that these two ways of getting the data were equal in performance. If that's not the case, I will revert these changes as well

Schottkyc137: Looking at the code right now, the issue was not resolved. `ChildrenService` filters if the children-array of the Note contains the child-id (which is never true since the Array is of type `AttendanceModel`). Is it possible to change the database quarry to check if any of it's `AttendanceModel`s contains the id?

sleidig: For the PouchDb/CouchDB based Database queries should be more efficient, the database then only loads the required documents and returns them (and hopefully the developers of pouchdb have figured out a way to do that reasonably efficient).

The query index is also created in the ChildrenService:
https://github.com/Aam-Digital/ndb-core/blob/134d808a86fde5983c81e93f2ac1e6d0a6bba98b/src/app/child-dev-project/children/children.service.ts#L229-L243

Beware, unfortunately the function has to be a string containing the javascript function (to prevent the build optimization from messing this up). You can also refer to the PouchDB guide about how those query functions work: https://pouchdb.com/guides/queries.html

Basically you'll just have to change that function to `emit()` the childId from its new nested place.

sleidig:
closing this as PR #431 is built upon this branch

sleidig:
closing this as PR #431 is built upon this branch

Issue #17, opened 2016-06-30
UserAccount password change


sleidig:
The user needs a simple "user account" page where she can change her password (and maybe in the future edit other account details).
The same component could also be used for admins to view or edit another user's account or create a new account.

**Attention: Users are not just normal Entity instances**
We currently have a slightly complex concept where a user exists as an Entity in the normal database and _additionally_ as a CouchDB user object (int the _users db). Passwords need to be changed in both objects in parallel!

- see https://github.com/Aam-Digital/ndb-core/issues/272#issuecomment-494941923 for more details
- also refer to the [add-user script](https://github.com/Aam-Digital/ndb-setup/blob/master/add-user.js) which currently does this job outside of the UI
- we should also consider changing the architecture around user accounts overall (see #272)

-------
**After revising the technical requirements, the steps for implementation are as follows**
(copied from @TheSlimvReal below):

- [ ] Create `UserAccountService`
- [ ] Create `ChangePassword` function
- [ ] `GET` request to `/db/_user/org.couchdb.user:<user>` with username and password in authorization header to get user data 
- [ ] add password field to user object with the unencrypted password
- [ ] `PUT` request to `/db/_user/org.couchdb.user:<user>` with new object and username and old password authorization header
- [ ] Update user object in local database

aankitkumar31:
develop UserAccount View

Lorsbyjm:
check branch change pw

sleidig:
basic user account view is done. However, backend work to properly change to password (which is linked between local database and remote server) is pending

codingfabi:
a user view is already existing. only functionality to change password is missing.

codingfabi:
@sleidig could u create a test user for me so i can test the password change feature? i dont want to mess up the only user that we have...

sleidig:
Another user is available now:
Username: Tester
Password: pass

liwde:
Work on the improved session continues and the interface supports an "offline" state that can be read and listened to. This way we can make sure the password may only be changed when the user is online.

However, currently the offline state is not really set inside the improved session. I'm currently focusing on feature parity, but after this we can look into how to set the online/offline state.

codingfabi:
What is the status on this? @liwde can you give a quick update if this is possible to do now with the new session implemented?

TheSlimvReal:
- [ ] Create `UserAccountService`
- [ ] Create `ChangePassword` function
[...]

PR #389, opened 2019-12-18
Admin ui v2

TheSlimvReal:
see issue: #254

This PR reimplements the adminUI similar to #271.
In the previous PR we tried to implement the user list with the EntitySubrecordComponent which lead to some design problems that were to difficult to fix.

Therefore we changed to a own component which displays the registered users and allows to edit them. 

### Visible/Frontend Changes
- [x] A list overview of the registered users
- [x] A dialog to add and edit users


src/app/core/admin/user-details/user-details.component.ts
```diff
@@ -0,0 +1,70 @@
...
+import { MatSnackBar } from '@angular/material/snack-bar';
+
+@Component({
+  selector: 'app-user-details',
+  templateUrl: './user-details.component.html',
+  styleUrls: ['./user-details.component.scss'],
+})
+export class UserDetailsComponent implements OnInit {
+  creating: boolean = true;
+  private readonly defaultPasswordValue: string = 'password';
```
sleidig: I wouldn't set defaults (only placeholders as hints for the user). Let's better leave the form invalid until the user enters something properly.

src/app/core/admin/user-details/user-details.component.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+  }
+
+  saveUser() {
+    const {username, password, admin} = this.userForm.getRawValue();
+    if (this.creating) {
+      this.user = new User(username);
+    } else if (username !== this.user.name) {
+      // Can't be reached because changing username is not allowed
+      this.entityMapper.remove<User>(this.user);
```
sleidig: careful. this may have side effects if we every reference the user entity id (because now it will be different). To be honest, I am not sure why we disallow username changes but I think it's probably because the username directly makes part of the _id and we wanted to keep the id immutable.

Maybe we should disallow username changes in the form also?

src/app/core/admin/user-details/user-details.component.ts
```diff
@@ -0,0 +1,70 @@
...
+      this.user = new User(username);
+    }
+    this.user.name = username;
+    if (password !== this.defaultPasswordValue) {
+      this.user.setNewPassword(password);
+    }
+    this.user.admin = admin;
+    this.entityMapper.save<User>(this.user)
```
sleidig: Saving the `User` entity is not sufficient in our current architecture - we have an additional user document in CouchDB's "_users" database that is used for authorization of access to the server database. New users and passwords need to be synchronously done there as well (this is what made #17 so complicated).

sleidig:
Maybe better to break this down into steps? Let's make this completely read-only and merge it. Then as soon as possible work out the issues of saving of users across the two databases (or re-design the approach to be simpler).

PR #394, opened 2020-01-10
Refactor code folder structure

sleidig:
Attempt to simplify the code structure with its huge number of folders at the same levels:

1. The application code is split within the `src/app/` directory into modules providing general features and abstract components (_core_) and concrete feature modules for users' use cases (_child-dev-project_).
2. moved other entities with their components out of the _children_ directory. These should become separate modules in the long run (but we need to remove circular dependencies here first)

Reviewers: Please have a look whether the suggested folder structure seems clear and reasonable (no need to check all diffs, unfortunately this PR touches and moves pretty much every file)

### Visible/Frontend Changes
_none_ 

_no functional changes_


sleidig:
> I really like the new structure. I think it's good that you moved some front end components like the `EntitySubrecord` to the core folder because they are very generic and reusable. However the circular dependencies are definitly something we should fix. In my opinion nothing of the `core` folder should have a dependency to the`child-dev-project`. This way we can later create a reusable Angular library with the features of the core code.

Absolutely agree. We should work towards this and list the remaining cross-dependencies. There are also tools to support this kind of dependency constraints, that we could consider: https://nx.dev/angular/guides/monorepo-tags

Issue #373, opened 2019-11-25
Note doesn't display long list of linked children properly


sleidig:
**Describe the bug**
If a note is linked to many children (e.g. to document a group activity) our component cuts of the list of children rather than wrapping it to multiple lines.

PR #399, opened 2020-01-28
Note children display

KatSB:
see issue: #373 

### Visible/Frontend Changes
- [x] In Create Note: long list of children is wrapped in multiple lines
- [x] In Create Note: Label of input field Children stays where it is, instead of moving around


codingfabi:
Perfect, feel free to merge ;)

Issue #402, opened 2020-02-01
Implement EntitySchema Datatype "recursive" to consider annotations within referenced object


sleidig:
For more complex Entity classes it would be helpful to recursively use schema annotations within referenced objects to control which properties of such an object will be written to the database.

For example:
```
class ComplexEntity extends Entity {
  @DatabaseField() title: string;
  @DatabaseField('recursive-schema') details: Description;
}

class Description {
  @DatabaseField() id: number;
  otherStuff: string; // won't be serialized to the database as it doesn't have the annotation
}
```
should result in a json object like this: `{ "title": "foo", "details": { "id": 123 } }`

---
* [ ] decide on a name for the "datatype" (e.g. "recursive-schema")
* [ ] write unit tests for all variations
* [ ] implement functionality

TheSlimvReal:
Is this now done with #403 and can be closed?

sleidig:
Yes, this is the thing

Issue #390, opened 2020-01-05
AttendanceMonths can have hidden logical conflicts


sleidig:
**Describe the bug**
If two users trigger the creation of a new AttendanceMonth at a similar time (e.g. by doing the first "add a day's attendance' of the month or by visiting the child details view at the beginning of a new month), this can result in two different entities with different ids for the logically same record. This is especially likely to happen if some devices are offline at that time and the newly created records only get synced later.

The usual conflict system of CouchDB doesn't trigger here, because on database level these are two separate objects (the ids are currently randomly generated).

sleidig:
A possible solution could be to set deterministic ids for AttendanceMonth records (e.g. a combination of the child id and the month for which this is set). That way the conflict is at least ending up as a conflict on database level as well and can be handled with more general conflict resolution tools.

PR #403, opened 2020-02-01
Bug fixes and improvements to AttendanceDay

sleidig:
Fix problems with the attendance records. Especially an issue that the roll call (add day's attendance) feature wrongly jumped to a certain date instead of the selected date.

see issues: #390, #402 

### Visible/Frontend Changes
- [x] Fix problem that "add day's attendance' displayed previously entered status from a different than the selected day
- [x] Fix problems when using attendance records in different timezones (causing records to be shifted by one day)
- [x] Make hidden duplicates (two records with different IDs referring to the same AttendanceMonth) less likely by better ID generation.

- [x] Add EntitySchema datatype "array" to configure (de)serialization of arrays through the `@DatabaseField()` annotation
- [x] Add EntitySchema datatype "schema-embed" (_**suggestions for a better name welcome!**_) to configure (de)serialization of complex objects through the `@DatabaseField()` annotation, allowing annotations in the class type of that object recursively.
- [x] Add EntitySchema datatype "date-only" to configure (de)serialization of date objects to a simple string without time (YYYY-MM-dd) through the `@DatabaseField()` annotation
- [x] Refactor the components for the "add day's attendance" feature, splitting and simplifying the code

### DIFF
--- a/src/app/core/entity/schema/entity-schema.service.spec.ts
+++ b/src/app/core/entity/schema/entity-schema.service.spec.ts
@@ -108,9 +108,68 @@ describe('EntitySchemaService', () => {
     entitySchemaService.loadDataIntoEntity(entity, data);
 
     const expectedDate = new Date(2018, 1); // month indices start at 0!
-    expect(entity.month.toDateString()).toEqual(expectedDate.toDateString());
+    expect(entity.month).toEqual(expectedDate);
 
     const rawData = entitySchemaService.transformEntityToDatabaseFormat(entity);
     expect(rawData.month).toEqual('2018-2');
+  });
+
+
+
+  it('schema:array converts contained dates to month for saving', function () {
+    class TestEntity extends Entity {
+      @DatabaseField({ arrayDataType: 'month' }) dateArr: Date[];
+    }
+    const id = 'test1';
+    const entity = new TestEntity(id);
+    entity.dateArr = [ new Date('2020-01-01'), new Date('2020-02-02') ];
+
+    const rawData = entitySchemaService.transformEntityToDatabaseFormat(entity);
+
+    expect(rawData.dateArr).toEqual(['2020-1', '2020-2']);
+  });
+
+  it('schema:array converts contained month strings to dates when loading', function () {
+    class TestEntity extends Entity {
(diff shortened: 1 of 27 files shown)


src/app/core/entity/schema/entity-schema.service.spec.ts
```diff
@@ -108,9 +108,88 @@ describe('EntitySchemaService', () => {
...
+
+    entitySchemaService.loadDataIntoEntity(entity, data);
+    expect(entity.details.month).toEqual(new Date(2020, 0));
+
+
+    entity.details.otherStuff = 'foo';
+    const rawData = entitySchemaService.transformEntityToDatabaseFormat(entity);
+    expect(rawData.details.month).toEqual('2020-1');
+    expect(rawData.details.otherStuff).toBeUndefined();
+  });
```
TheSlimvReal: It would be good to have at least on test for the array datatype annotation

src/app/core/entity/schema/entity-schema.service.spec.ts
```diff
@@ -108,9 +108,88 @@ describe('EntitySchemaService', () => {
...
+    const expectedDate = new Date('2018-01-15');
+    expect(entity.day).toEqual(expectedDate);
+
+    const rawData = entitySchemaService.transformEntityToDatabaseFormat(entity);
+    expect(rawData.day).toBe('2018-01-15');
+  });
+
+
+
+  it('schema:array converts contained dates to month for saving', function () {
```
sleidig: This is the test for the array datatype

src/app/core/entity/schema/entity-schema.service.spec.ts
sleidig: Absolutely. There actually is a unit test for tart in the entity-schema.service.spec

Should we split out these tests into individual files for each datatype transformer? The setup would be a little repetitive but maybe easier to navigate

TheSlimvReal: ah sorry I must have overlooked it. Then I have nothing to complain about. I guess it might be viable to split this tests up. I guess when we decide to add more data types we can also introduce this

Issue #398, opened 2020-01-25
"Previous Schools": display class and grade percentage


sleidig:
The "Previous Schools" table in the "Education" section of the Child Details should manage the class and results for each line as well.

- "class" (string) -> which class/grade the student was attending at that school and time
- "results" (float) -> total marksheet average results (in India usually displayed as a percentage, e.g. he scored 78% in class 10 exams)

Both should be displayed and edited in the table similar to the columns that are already part of the table.

Schottkyc137:
Is the result-field going to be a new field or can that be computed out of some other data?
Also, the previous schools are not editable, is that something that should be implemented as well?

sleidig:
The results would be a simple input field for the time being. In the mid to long run it would be exciting to have a more detailed feature for grades down to the individual subject (something like a digital copy of the marksheet with all its details) but for that we first should draft possible UI layouts and plan it out. (feel free to think about that as well and propose something, however!)

Issue #317, opened 2019-07-09
Reactivate use of ChildSchoolRelations to determine current school of a Child


sleidig:
> PR #316 has reverted the improvements made in #216 due to some problems. This code should be brought back and fixed or we should find another way of using the ChildSchoolRelation records to determine the current school of a child.

**Objective:**
Currently we are managing and storing the school of a child in two places (as the schoolId and schoolClass properties of `Child` and as `ChildSchoolRelation`s through the "Previous Schools" feature). Ideally only ChildSchoolRelations are necessary and from the date range the current school and class (if any) can be inferred and displayed.

- the form fields for school and class in the Child Details should be readonly. Users need to understand (maybe tooltip?) that changes are to be made through the "Previous Schools" table.
- an easy (and performance optimized) method is necessary to get the current school and class of a child
- a migration function is needed to help switch that existing data to the new system. Migration should automatically generate a ChildSchoolRelation record for a Child's current schoolId property before removing that property.

PR #407, opened 2020-02-04
[WIP] Previous schools

Schottkyc137:
see issue: #398 

### Visible/Frontend Changes
- [x] The "Previous Schools tab" in the "Education" section of a child displays grade and the class
- [x] The Edit-Form can be used to edit those new values (grade and class)

- [x] The "Edit School Dialog component" now uses the Abstract Details component


codingfabi:
Is this still WIP? Looks done to me...

Schottkyc137:
I still have some things to fix since this commit uses an outdated version of master

sleidig:
@Schottkyc137 Ideally make PRs against the master for each change that can somehow stand on its own. That makes it much easier to review and quickly merge those parts of you changes that are done instead of collecting a large, long opened PR

christophscheuing:
I'm confused about how we are supposed to handle the attended school class with respect to the time spent at a certain school: If the period of time spent at a given school extends over several classes, should this school classes be mentioned consequently (with separating commas?) in the school class field? Or should there be a different entry for every school class / year, meaning that the period from the start to the end date cannot be longer than one school year?

The way we handle school classes is of importance for issue #317 on how to determine the currently attended school class.

sleidig:
I had one line/entry per school year in mind. That way it should be much easier to manage information about a particular class like the results for that year, etc.  And even if a student should repeat the same class, that becomes more obvious if there are two lines, one for each year, for this.

I didn't think about the constraint, that the form should only allow one year periods, but I agree with you. (although maybe there could also be some kind of training programmes that last 1.5 years? That would make the case to just leave it up to users how they want to use it)

Issue #396, opened 2020-01-15
Use prettier with pre commit hooks


codingfabi:
@TheSlimvReal and me discussed using a tool to guarantee lint-confirm code. 
I got some experience working with [prettier](https://prettier.io/) and would volunteer to do a setup for our project. But I wanted to hear your thoughts about adding pre commit hooks and forcing contributers into a certain code format?

codingfabi:
For those who are further interessted:

Prettier can be used with or without pre-commit Hooks. 
If would only be usefull to implement it with a pre-commit tool like [husky](https://github.com/typicode/husky). 
Then, prettier automatically fixes your tsLint warnings and formats your code. For example, adding missing semicolons or replacing double with single quotes. It can also arrange html tags and removes trailing whitesspaces etc. 

I think it could save us a lot of time because it could prevent us from fixing broken CI - Tests that only fail because of linting errors.

sleidig:
This sounds cool. Additionally a pre-commit hook could warn if there are linting errors that cannot be fixed automatically (if that even exists).

PR #408, opened 2020-02-04
Use prettier

codingfabi:
see issue :#396

tl;dr:
Prettier in combination with pre-commit hooks and husky formats the code automatically with predefined optioins before your commits. this gurantees a generalized code style and prevents the lint-checker from failing. 

- [x] install the prettier package
- [x] create a prettier config file to manage formatting options
- [x] install husky and configure a pre-commit hook
- [x] refactor the whole projects using those two packages


codingfabi:
Feel free to discuss the defined formatting options and give me your opinion on the whole thing

TheSlimvReal:
How do you get this hook executed? It did not work when I commited something in WebStorm.

sleidig:
at least to some extent tslint offers the same thing, doesn't it? The prettier docs claim there still is a point to use both together (https://prettier.io/docs/en/integrating-with-linters.html and there are integration plugins) but I am not fully sure. Do you have experience with both, @codingfabi ?

Definitely a good idea to include this in pre-commit hooks!

codingfabi:
> 
> 
> How do you get this hook executed? It did not work when I commited something in WebStorm.

In VSCode, it worked on my machine automatically with the commit. Did you npm install? 
It could also be a problem with access authorization, because prettier needs a CL run as administrator. 
maybe that is your problem here, try to run _prettier --write '/**/*.js'_  in your command line in webstorm and give me the output

codingfabi:
> 
> 
> at least to some extent tslint offers the same thing, doesn't it? The prettier docs claim there still is a point to use both together (https://prettier.io/docs/en/integrating-with-linters.html and there are integration plugins) but I am not fully sure (`tslint --fix` can also auto-format your code). Do you have experience with both, @codingfabi ?
> 
> Definitely a good idea to include this in pre-commit hooks!

Yes I have a project at work where we use both. We could also consider to add jsut tslint --fix as a pre-commit hook. But I like the easy configuration from prettier without touching the tslint highlighting. But that is a personal preference. What do you think?

sleidig:
I am slightly afraid of maintaining similar things in two different places. But if you have positive experience with that setup let's just do that. 
Should we use one of the integration packages to sync rules from tslint to the new prettier config then? ( https://prettier.io/docs/en/integrating-with-linters.html) I am happy with the current tslint config but also open to adaptions.

src/app/child-dev-project/aser/model/aser.ts
```diff
@@ -22,21 +22,8 @@ import { DatabaseEntity } from '../../../core/entity/database-entity.decorator';
...
-    'Read Paragraph',
-  ];
-  static MathLevels = [
-    'Nothing',
-    'Numbers 1-9',
-    'Numbers 10-99',
-    'Subtraction',
-    'Division',
-  ];
+  static ReadingLevels = ['Nothing', 'Read Letters', 'Read Words', 'Read Sentence', 'Read Paragraph'];
```
sleidig: is there a way to force consistent linebreaks (all or nothing) but leave the choice between breaks or not to the user? (eslint seems to have that option)

src/app/child-dev-project/attendance/child-attendance/child-attendance.component.spec.ts
```diff
@@ -156,33 +155,30 @@ const ATTENDANCE_ENTITIES: AttendanceMonth[] = (() => {
   return data;
 })();
 
-
```
sleidig: I do like the opportunity to add 2-3 empty lines separating different logical blocks

src/app/child-dev-project/aser/model/aser.ts
sleidig: apparently not possible with prettier ...

Issue #374, opened 2019-11-25
make BMI feature more resilient and flexible


sleidig:
The Heigh & Weight tracking is used regularly already and from the user observation some possible improvements came up:
* users are currently manually calculating their measurements (which they do in feet and inch) to centimeters before entering them. We could offer an option to enter heigh in inches directly (and convert it to centimeters before saving it to the database so that we always have the same unit in storage)
* weight is tracked more often than height. can we calculate the BMI anyway, using the last recorded height? Or find some other nice way to adapt our UX to this real life workflow?

codingfabi:
Would the input in feet and inches be a string or a number? 
Can you get some information wether the users would make an input like 6'2" oer 6,2?

I would implement a toggle button to switch input. I think transforming the input will be quite stressfull while using the entity subrecord component but I think there's some workaround to figure it out. Any thoughts?

codingfabi:
i made a WIP PR for a basic workaround inside the EntitySubrecordComponent. I would love to hear your feedback because I think this solution has got several smells

Schottkyc137:
Would it be a good idea for this to allow the user to have some sort of "settings" where he can set what kind of unit he wants to use for weight & health-tracking?

PR #417, opened 2020-02-19
[WIP] basic switch for height input

codingfabi:
see issue: #374 

### Visible/Frontend Changes
- [x] switch for input unit

- [x] automized recalculation of height from feet/inches to cm 
- [ ] take last height input as default for new record

### DIFF
--- a/src/app/core/ui-helper/entity-subrecord/entity-subrecord.component.ts
+++ b/src/app/core/ui-helper/entity-subrecord/entity-subrecord.component.ts
@@ -1,3 +1,4 @@
+import { HealthCheck } from './../../../child-dev-project/health-checkup/model/health-check';
 import { Component, Input, OnChanges, OnInit, OnDestroy, SimpleChanges, ViewChild } from '@angular/core';
 import { MatDialog } from '@angular/material/dialog';
 import { MatSnackBar } from '@angular/material/snack-bar';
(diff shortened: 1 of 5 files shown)


src/app/core/ui-helper/entity-subrecord/entity-subrecord.component.ts
```diff
@@ -231,4 +234,11 @@ export class EntitySubrecordComponent implements OnInit, OnChanges, OnDestroy {
   isReadonlyInputType(inputType: ColumnDescriptionInputType): boolean {
     return inputType === ColumnDescriptionInputType.FUNCTION || inputType === ColumnDescriptionInputType.READONLY;
   }
+
+  transformInput(height: number, entity: Entity, colName: string): number {
+    if (this.transformFlag && entity.getType() === 'HealthCheck' && colName === 'height') {
+      height = height * 30.48;
+    }
+    return height;
+  }
```
sleidig: Adding the explicit transformation function for HealthCheck to the generic EntitySubrecordComponent is not ideal. How about instead extending the `ColumnDescription` with an optional inputTransformer property similar to the `formatter`?
 https://github.com/Aam-Digital/ndb-core/blob/9c34a807453fbae4934eb3ac7a497087fdc96078/src/app/core/ui-helper/entity-subrecord/column-description.ts#L43

That could also allow us to remove the `@Input() transformFlag` here which would better apply to a single column anyway.

Issue #366, opened 2019-11-25
Search bar is behaving weird


sleidig:
**Describe the bug**
The search box (top right) is not working smoothly:
* [x] searching for something into the search box and then deleting it again doesn't reset the result. Previous results remain visible in the dropdown even after clicking somewhere else and clicking back into the empty searchbox.
* [ ] It takes a very long time to load search results (at least on the first search query). A loading animation would be helpful user feedback to signal the status to the user.

**Additional context**
**_Search depends on pouchdb indices so this can only be debugged when setting `useTemporaryDatabase` to false in the config.json_**

Lorsbyjm:
fixed the first point in the pull request by removing a failure in one of the called functions. Furthermore I check the input before sending the query to the database to contain at least a char or a number. Otherwise even an space would be searched and that takes long.
Here, we could also use harder restrictions as a minimum of three letters before going to the database ??

Did not work on the second point yet and the searching is still to long. Animation needed

Lorsbyjm:
@sleidig  For the second problem, the long loading. Instead of using a loading icon  we could add an information block to autolaod alias the search result and display there a text like: "please wait". Furthermore you could display other texts like "please be more specific, there are to many items matching your search" . What do you think? Good idear or too much overhead?

sleidig:
Showing a text to give explanation is a smart idea, especially for the cases when we are not triggering search yet ("Please enter more specific search term, there are too many items matching your search.") and when there are no results ("Sorry, no results are matching your search.").

I didn't completely understand your other suggestion with the alias but "please wait" might stop people while they are still typing, which is not really the intention, is it?

Do you know if the delay is mostly due to preparing result items or the initial search request?

tomwwinter:
The search is now only carried out from three characters upwards.

However, a search like "M Mu" will also trigger a search. Here we should perhaps check whether the first word has at least three characters?

Besides, now:
* The current status of the search is displayed
* The system waits a short time before an entry triggers a search
* An already running search is cancelled (technically only ignored) and replaced with the new search query (switchMap)

sleidig:
> However, a search like "M Mu" will also trigger a search. Here we should perhaps check whether the first word has at least three characters?

I don't see this as a big problem. The goal is to limit the number of results displayed (and loaded - for performance). At least the first goal is still reached.

PR #424, opened 2020-03-24
optimized searchbar: only words, removed bug in parse rows

Lorsbyjm:
see issue: #366 

### Visible/Frontend Changes
- [x] searchbar dissapers when searchfield is empty
- [x] searching does not take that long

- please check the regex expression, it should allow only strings with at least one characters or number
- remeber: 
Search depends on pouchdb indices so this can only be debugged when setting useTemporaryDatabase to false in the config.json

### DIFF
--- a/src/app/child-dev-project/children/child-block/child-block.component.html
+++ b/src/app/child-dev-project/children/child-block/child-block.component.html
@@ -1,12 +1,12 @@
 <span *ngIf="entity" (mouseenter)="showTooltip()" (mouseleave)="hideTooltip()">
-  <img [src]="entity?.photo" class="child-pic">
+  <img [src]="entity?.photo?.value" class="child-pic">
   {{entity?.name}} <span style="font-size: x-small">({{entity?.projectNumber}})</span>
 </span>
 
 <div style="position:absolute;" *ngIf="tooltip">
   <div class="mat-elevation-z5 child-tooltip" (mouseenter)="showTooltip()" (mouseleave)="hideTooltip()" fxLayout='row'>
     <div fxFlex='30'>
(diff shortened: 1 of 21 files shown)


sleidig:
I noticed here that the previous merge in the master of the ChildPhotoService had broken the photos in the search result. After some forth and back I ended up with a larger change (moving the loading of the photos to a new EntitySchemaDatatype).
@Lorsbyjm (or @RichardNutt ?): please do a review of my commits here as well - and point out any parts that need to be corrected, cleaned up or better documented.

Lorsbyjm:
@sleidig what do you want to achive. Now no pictures are shown. 
Which properties do I need to set?

sleidig:
Oh, you are absolutely right. there was another bug that converted an undefined photoFile name wrongly and broke images if you have not set up the Nextcloud integration. This should be fixed now. I also shared a test account for nextcloud in our slack in case you want to test with that as well.

Lorsbyjm:
I do not have enough experience to review the code, but it looks like it is working.
But in the database are still children with an old entity schema, so there are no pictures shown, these children should be deletet. (To test search for Jonas)

Furthermore the loading takes even more time if for every search reslut the image is loaded from the nextcloud. We might think about limiting the search result to an special threshold and not show it if there are more. But to ask the user for more input.

sleidig:
limiting it (maybe also just start after input of at least 3 characters) might be a good idea. The photo load should work asynchronously however - first return the path to the default image (or one that has been uploaded by the server admin) and only later replacing that if a photo is returned from the cloud.
The "Jonas" doc had an invalid photoFile path saved - probably due to the early bug.

PR #427, opened 2020-03-30
WIP: Search information panel

Lorsbyjm:
see issue: #366 
### Visible/Frontend Changes
- [x] information displayed when no search results found
- [x] added a search-information component and class
 
To test search fo k, then there should be a information "no results found"
This is an idea how to do it, the code might be ugly and needs to be updated. So give me your input.

### DIFF
--- a/src/app/core/ui/search/search-information/model/search-information.ts
+++ b/src/app/core/ui/search/search-information/model/search-information.ts
@@ -1,0 +1,9 @@
+import { Entity } from '../../../../entity/entity';
+import { DatabaseEntity } from '../../../../entity/database-entity.decorator';
+
+
+@DatabaseEntity('SearchInformation')
+export class SearchInformation extends Entity {
+
+    text: string;
+}
--- a/src/app/core/ui/search/search.component.html
+++ b/src/app/core/ui/search/search.component.html
@@ -8,6 +8,7 @@
     <mat-option *ngFor="let res of results" [value]="''">
       <app-child-block *ngIf="res.getType()==='Child'" [entity]="res" [ngClass]="{'result-inactive': !res.isActive()}"></app-child-block>
       <app-school-block *ngIf="res.getType()==='School'" [entity]="res"></app-school-block>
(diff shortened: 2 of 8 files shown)


src/app/core/ui/search/search-information/model/search-information.ts
```diff
@@ -0,0 +1,9 @@
+import { Entity } from '../../../../entity/entity';
+import { DatabaseEntity } from '../../../../entity/database-entity.decorator';
+
+
+@DatabaseEntity('SearchInformation')
+export class SearchInformation extends Entity {
```
sleidig: extending from the `Entity` class and using the annotations is important if you want to save instances to the database. If you don't need that, you can use a "normal" class or simple variables without this overhead.

src/app/core/ui/search/search.component.html
```diff
@@ -8,6 +8,7 @@
     <mat-option *ngFor="let res of results" [value]="''">
       <app-child-block *ngIf="res.getType()==='Child'" [entity]="res" [ngClass]="{'result-inactive': !res.isActive()}"></app-child-block>
       <app-school-block *ngIf="res.getType()==='School'" [entity]="res"></app-school-block>
+      <app-search-information *ngIf="res.getType()==='SearchInformation'" [entity]="res"></app-search-information>
```
sleidig: I used the `app-child-block` and `app-school-block` components to build on the reusable presentation of these Entities. for displaying the search information text, that is only relevant within the SearchComponent you could even just directly display it here, maybe using a html `span`. A separate component can also be nice (not sure whether it's an overkill here).

It might be easier to keep that search-information separate from the results array and it's *ngFor, then you don't have to make sure it has the same format like entities with their getType(), etc.
e.g. `<mat-option *ngIf="searchInformationText"><span class="search-information">{{searchInformationText}}</span></mat-option>`

sleidig:
Closing this. I think the search-bar loading UX issues can be picked up in a fresh branch more easily, as this PR has been idle and conflicting changes to the code base are coming up in the master branch.

PR #431, opened 2020-04-14
Note Additions + Child school relations update

christophscheuing:
see issue: #317; see also issue: #407 

-----
This branch includes PR #386 (Note Additions).
Collecting the open issues from that branch here:

- [x] child-select and the presence list elements should be directly next to each other (maybe move the child-select down, above the presence list). That will make it easier for new users to understand the connection.
- [x] How to switch children between present/absent is not obvious for non-technical users. We could try whether a toggle [Present|Absent] instead of the arrow makes it more transparent.
- [x] Meeting notes are very highlighted in the Notes section of the Child Details view currently, because most other Notes just have white background. I'm not sure how to improve that easily however and don't think this is a blocker. But maybe you have any ideas?


src/app/child-dev-project/attendance/attendance-details/attendance-details.component.spec.ts
```diff
@@ -40,7 +40,7 @@ describe('AttendanceDetailsComponent', () => {
         UiHelperModule, EntityModule],
       providers: [
         {provide: Database, useClass: MockDatabase},
-        {provide: MatDialogRef, useValue: {beforeClose: () => { return { subscribe: () => {}}; }}},
+        {provide: MatDialogRef, useValue: {beforeClosed: () => { return { subscribe: () => {}}; }}},
```
sleidig, line 43: why was this changed? according to the Angular API Reference it should be `beforeClose` (https://material.angular.io/components/dialog/api#MatDialogRef)

christophscheuing: Ooops, sorry, this must have been a typing error.

PR #432, opened 2020-04-21
fixed not showing search button on xs device

vlle1:
simply deleted an fxHide.xs in ui.component


vlle1:
The main problem is that the synchronisation and search is not really performant and you need to wait long.
Nevertheless, the feature is working in theory.

Issue #409, opened 2020-02-07
Upgrade to Angular 9


sleidig:
Angular 9 was released! :rocket: 
https://blog.angular.io/version-9-of-angular-now-available-project-ivy-has-arrived-23c97b63cfa3

* [x] do the upgrade process with the help of https://update.angular.io/
* [x] also do a (careful) upgrade of other dependencies (e.g. with `ncu`)
* [x] summarize the most important changes of Angular 8 -> 9 and Typescript 3.5 -> 3.7 for all of us here in this issue

PR #443, opened 2020-05-23
Angular 9 upgrade

sleidig:
see #409 
Update to Angular 9 and update all other dependencies.

This article gives a good overview of the changes & improvements: https://blog.angular.io/version-9-of-angular-now-available-project-ivy-has-arrived-23c97b63cfa3


vlle1:
I got it running and from my perspective we can merge this PR. 

Note, that there are some vulnerabilities in the used packages (see npm audit). For example papaparse could be upgraded to a never version without the dos vulnerability, but then some code in the adminmodule must be changed.

sleidig:
Good point. I created a ticket for upgrading papaparse: #444

PR #456, opened 2020-06-27
fix(search): fix weird behaving in search bar

tomwwinter:
see issue: #366 

### Visible/Frontend Changes
- [x] show current state of the search is displayed ("Searching", "No Results", ..)
- [x] wait a short time before search input triggers a search (user typing)
- [x] minimum of three characters in search

- [x] An already running search is cancelled (technically only ignored) and replaced with the new search query (switchMap)

Closing #366

### DIFF
--- a/src/app/core/ui/search/search.component.html
+++ b/src/app/core/ui/search/search.component.html
@@ -9,7 +9,7 @@
       title="Search"
       [matAutocomplete]="autoResults"
       [(ngModel)]="searchText"
-      (ngModelChange)="searchInputChanged($event)"
+      (ngModelChange)="searchStringChanged.next($event)"
       #searchInput
     />
   </mat-form-field>
@@ -79,7 +79,7 @@
       title="Search"
       [matAutocomplete]="autoResults"
       [(ngModel)]="searchText"
-      (ngModelChange)="searchInputChanged($event)"
+      (ngModelChange)="searchStringChanged.next($event)"
     />
     {{ searchInput.focus() }}
     <button mat-icon-button (click)="toggleSearchToolbar()">
--- a/src/app/core/ui/search/search.component.html
+++ b/src/app/core/ui/search/search.component.html
@@ -1,26 +1,89 @@
-<div fxHide.xs="true" fxHide="false">
-  <mat-form-field fxFill class="search-field">
-    <span matPrefix><mat-icon class="search-icon" fontIcon="fa-search"></mat-icon>&nbsp;</span>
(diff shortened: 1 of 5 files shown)


src/app/core/ui/search/search.component.ts
```diff
@@ -28,9 +37,91 @@ export class SearchComponent implements OnInit {
...
+        switchMap((value) => {
+          return value;
+        })
+      )
+      .subscribe(
+        (result: { queryResults: any; searchTerms: string[] }) => {
+          this.handleSearchQueryResult(result);
+        },
+        (error) => {
+          console.error("SearchQuery failed", error);
```
sleidig, line 58: Let's use "LoggingService" for this to get it properly registered. Possibly also an error to the user through AlertService?

src/app/core/ui/search/search.component.ts
```diff
@@ -28,9 +37,91 @@ export class SearchComponent implements OnInit {
...
+      this.showSearchToolbar = false;
+    }
+  }
+
+  toggleSearchToolbar() {
+    this.showSearchToolbar = !this.showSearchToolbar;
+  }
+
+  searchInputChanged(searchString) {
+    this.searchStringChanged.next(searchString);
```
sleidig, line 104: can we call this directly in the template (and reduce the number of methods in the class here)?

src/app/core/ui/search/search.component.html
```diff
@@ -1,26 +1,89 @@
...
+    #autoResults="matAutocomplete"
+    (optionSelected)="clickOption($event.option)"
+  >
+    <mat-option
+      *ngIf="
+        searchInput.value.length > 0 &&
+        searchInput.value.length < MIN_CHARACTERS_FOR_SEARCH
+      "
+    >
+      <p>Insert at least {{ MIN_CHARACTERS_FOR_SEARCH }} characters</p>
```
sleidig, line 27: I would give the "operational" mat-options a different styling than the actual search result items.

maybe something like
```
result-hint {
    background-color: lightgrey;
    font-style: italic;
}
```

src/app/core/ui/search/search.component.ts
tomwwinter: loggingService and alertService added

src/app/core/ui/search/search.component.ts
tomwwinter: done.

src/app/core/ui/search/search.component.html
tomwwinter: good point, updated the styles - I would stay with the hints

Issue #462, opened 2020-06-30
Improve initial index building its user feedback


sleidig:
The app is extremely slow in some instances (for a general overview of performances issues see #257). Especially after initial load or update, it feels like the app is broken because certain data takes several minutes to load. After all data indices are built, the app starts to become fast and comfortable to use.

We need to optimize the indexing of the database and/or provide clear feedback to users so they understand that the app is still during its initial setup process.

sleidig:
Observations from extensive local performance testing.

Test case:
- ChildrenService.createAttendanceAnalysisIndex and ChildrenService.createNotesIndex indices
- production data (~4200 AttendanceMonth documents


It makes a huge difference whether the query is run for the first time (!!!):
- first query: 17000 - 19000 ms (avg_attendance_index/three_months, avg_attendance_index/last_month)
- second query: 30 - 40 ms

Not a big difference between building indices in parallel (Promise.all) or in sequence (individual awaits):
- in squence: 39-41 seconds (avg_attendance_index/three_months + notes_index/by_child)
- in parallel: 37-38 seconds (avg_attendance_index/three_months + notes_index/by_child)

CPU speed has ??? impact:
- ... couldn't get the tests to run with 4x CPU slowdown

Other observations:
- for every index view PouchDB creates a separate IndexedDB

PR #463, opened 2020-06-30
Performance Improvements for Indexing

sleidig:
see issue: #462

### Visible/Frontend Changes
- [x] extended database sync indicator to provide details on other background processes like indexing

 UI: indexing-indicators

- [x] Performance tests to benchmark indexing (results detailed in #462)
- [x] Creating and querying indices is now done through the new `DatabaseIndexingService` instead of directly on `Database`
- [x] Trigger initial indexing immediately after login (together with creation of the index) to prepare database for faster queries by user


src/app/utils/performance-tests.spec.ts
```diff
@@ -0,0 +1,123 @@
...
+    jasmine.DEFAULT_TIMEOUT_INTERVAL = 150000;
+  }));
+
+  it("sync initial and indexing", async () => {
+    // delete previously synced database; uncomment this to start with a clean state and test an initial sync.
+    // await deleteAllIndexedDB(db => true);
+
+    const session = TestBed.inject<SessionService>(SessionService);
+    const syncTimer = new Timer(true);
+    await session.login("demo", "pass");
```
TheSlimvReal, line 63: is it okay to have credentials as plaintext here?

TheSlimvReal (approved):
Looks very good, just a few small notes.

src/app/utils/performance-tests.spec.ts
sleidig: good point. even though the credentials are trivial this shouldn't be in the code base

sleidig: a really clean solution would probably be an .env file to set environment variables and put that file on .gitignore. That way we could even easily configure these in the CI as environment variables.

But for now, I am taking the shortcut on just putting variables with placeholders to finally get this PR out of the door.

Issue #476, opened 2020-07-28
Invalid fields not highlighted after failing to save Child Details


sleidig:
**Describe the bug**
When required fields are missing in the ChildDetails form when trying to save, the user sees a generic message "Form invalid, required fields missing" but it remains unclear which of the fields are required.

Form fields that are required are nicely highlighted with an error message after the user clicks into the form field and leaves it again without entering a value. This needs to be displayed for all invalid fields after failing to save as well.

**To Reproduce**
1. Go to 'Children' section
2. Click on 'Add New' to open an empty Child Details form to create a new child
3. Press save without entering anything --> only the error toast message appears
4. Click into a required field (e.g. "Center") and click somewhere else again, leaving the focus --> the field error is displayed

PR #477, opened 2020-07-28
Input error fix

RichardNutt:
see issue: #476

### Visible/Frontend Changes
- error message more specific
- Required fields get marked upon pressing save

### DIFF
--- a/src/app/child-dev-project/children/child-details/child-details.component.html
+++ b/src/app/child-dev-project/children/child-details/child-details.component.html
@@ -76,7 +76,7 @@
 
         <div fxFlex>
           <mat-form-field style='width: 300px;'>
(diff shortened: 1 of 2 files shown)


src/app/child-dev-project/children/child-details/child-details.component.html
```diff
@@ -76,7 +76,7 @@ <h1 class="page-header section-child">
 
         <div fxFlex>
           <mat-form-field style='width: 300px;'>
-            <input matInput formControlName="name" placeholder="Name" title="name" type="text">
+            <input matInput formControlName="name" placeholder="Name" required="true" title="name" type="text">
```
sleidig: I'm reverting this as it doesn't seem necessary for the bugfix and to keep definition of required fields in one place only (in the initForm() of child-details.component.ts)

Issue #461, opened 2020-06-29
Search not working in Demo / temporary database


sleidig:
In the MockDatabase (in-memory demo database) the global search through the searchbar isn't working. The query does not have a mock implementation in the `MockDatabase` class.

--> add an implementation and test the search

PR #484, opened 2020-08-18
Search in temporary database

christophscheuing:
see issue: #461

Global searchbar (for names of children and schools) implemented in mock database without searchindex so far.


sleidig (approved):
Awesome, finally a working search in demo! :smile: 

I adapted it a bit to use the individual searchIndices each Entity type generates, so that it really gives the same results as the real database version.
(see for details of those indices, basically an entity can generate an array containing any kind of search keywords: https://github.com/Aam-Digital/ndb-core/blob/fa93a7b5ae370eb147a775ba991c7aecb97102ca/src/app/core/entity/entity.ts#L165-L183)

As a side note:
Make sure you pull the latest version of master branch before creating new branches, there was a package.json file here that was a bit outdated and clashed when trying to merge into master.

Issue #479, opened 2020-08-03
Exclude inactive/dropout children from child-select box


sleidig:
In forms our child-select component allows users to add Children to an activity using autocomplete. Suggestions from the autocomplete dropdown should only display "active" children, to make the list easier to navigate. We already filter inactive students from the default children list, etc.

Maybe we should add an item at the bottom of the suggestion list "show inactive children" which will then also include the inactive kids, to provide some fallback possibility to link inactive children which would otherwise become impossible?

PR #488, opened 2020-09-01
Hide inactive children in selection

RichardNutt:
see issue: #479

### Visible/Frontend Changes
- [x] Display only active children by default
- [x] Add button to show all


src/app/child-dev-project/children/child-select/child-select.component.ts
```diff
@@ -72,10 +75,18 @@ export class ChildSelectComponent implements OnChanges {
 
     this.suggestions = this.allChildren.filter((child) => {
       const key = "" + child.name + " " + child.projectNumber;
-      return key.toLowerCase().includes(this.searchText);
+      const isActive = child.isActive();
+      const matchesKey = key.toLowerCase().includes(this.searchText);
+      return this.showOnlyActiveChildren ? matchesKey && isActive : matchesKey;
```
sleidig, line 80:
```suggestion
      if (this.showOnlyActiveChildren && !child.isActive()) {
        return false;
      }
      return key.toLowerCase().includes(this.searchText);
```

Having the two conditions separate makes it a bit easier to read in my opinion.

sleidig (approved):
Looks good :+1: 

Great to have the option to extend the selection to inactive children as well.

Issue #500, opened 2020-09-22
Evaluate whether all components in child detail view only need the child


TheSlimvReal:
The submenu in the `child-details` view should be configurable by the final user. Optimally the user only needs to provide the name of the subfolder and the component which is displayed in this subfolder. The best case would be, that each of the components that can be selected, only needs the `Child` object as input parameter. This would allow the components to be very reusable and the configuration of the `child-details` view to be very simple.

Steps to do:

- [x] Check which components already only need the `Child` object as input
- [x] Create strategies for the other components, so they are only dependent on the `Child` object
- [x] Create an issue for each component that needs to be changed

_relates to #222_

sleidig:
done in PR #508

PR #508, opened 2020-10-08
Child Details Subcomponents

TheSlimvReal:
see issue: #500

Besides the basic information, all the other components in the child details section only get the child object as an input. This is a step towards the goal, that the components which will later be displayed on the child details screen should be customizable.

I also created two new components for the `health` and the `dropout` information. These components are able to alter the child object of the surrounding child details class. This means clicking the save button in either of these components, saves the whole child entity and not just the `health` and `dropout` fields. 

### Visible/Frontend Changes
- Additional save buttons for the `health` and `dropout` seection
- The form in `health` and `dropout` is always editable but only saved once the save button is pressed

- All used components in the child details view only depend on the child object

### DIFF
--- a/src/app/child-dev-project/children/child-details/child-details.component.html
+++ b/src/app/child-dev-project/children/child-details/child-details.component.html
@@ -387,20 +387,10 @@
         Education
       </mat-expansion-panel-header>
 
-      <div
-        *ngIf="child.schoolId"
-        matTooltip="To edit the current school or class, please add to the school history below"
-      >
-        <em>
-          Currently attending class {{ child.schoolClass }} at
-          <app-school-block [entityId]="child.schoolId"></app-school-block>
-        </em>
-      </div>
-
       <div>
         <h3>School History</h3>
         <app-previous-schools
-          [childId]="child.getId()"
+          [child]="child"
           (changedRecordInEntitySubrecord)="changedRecordInEntitySubrecord()"
         >
         </app-previous-schools>
--- a/src/app/child-dev-project/children/child-details/child-details.component.html
+++ b/src/app/child-dev-project/children/child-details/child-details.component.html
@@ -409,25 +409,7 @@
         </mat-icon>
         Attendance
       </mat-expansion-panel-header>
-
-      <div fxLayout="row wrap" fxLayout.lt-lg="column" fxLayoutGap="2em">
-        <div fxFlex="45">
-          <h3>School Attendance</h3>
-          <app-child-attendance
-            [childId]="child.getId()"
-            institution="school"
-            [showDailyAttendanceOfLatest]="true"
-          ></app-child-attendance>
-        </div>
-        <div fxFlex="45">
-          <h3>Coaching Attendance</h3>
-          <app-child-attendance
-            [childId]="child.getId()"
-            institution="coaching"
-            [showDailyAttendanceOfLatest]="true"
-          ></app-child-attendance>
-        </div>
-      </div>
+      <app-grouped-child-attendance [child]="child"></app-grouped-child-attendance>
     </mat-expansion-panel>
 
 
--- a/src/app/child-dev-project/children/child-details/child-details.component.html
+++ b/src/app/child-dev-project/children/child-details/child-details.component.html
@@ -386,17 +386,11 @@
         </mat-icon>
         Education
       </mat-expansion-panel-header>
-
-      <div>
-        <h3>School History</h3>
-        <app-previous-schools [child]="child"></app-previous-schools>
-      </div>
-      <br />
-
-      <div>
-        <h3>ASER Results</h3>
-        <app-aser [child]="child"></app-aser>
-      </div>
+      <h3>School History</h3>
+      <app-previous-schools [child]="child"></app-previous-schools>
+      <br/>
+      <h3>ASER Results</h3>
+      <app-aser [child]="child"></app-aser>
     </mat-expansion-panel>
 
     <mat-expansion-panel expanded="false" title="Attendance"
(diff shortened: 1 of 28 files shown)


src/app/child-dev-project/children/child-details/child-details.component.html
```diff
@@ -470,131 +431,10 @@ <h3>Coaching Attendance</h3>
...
-        </mat-form-field>
-      </div>
-
-      <div>
-        <h3>Height & Weight Tracking</h3>
-        <app-health-checkup></app-health-checkup>
-      </div>
+      <app-health-form [child]="child"></app-health-form>
+      <br/>
+      <h3>Height & Weight Tracking</h3>
```
sleidig, line 436: should we move that part into the `app-health-form` in order to get a single component for the whole section/panel?

src/app/child-dev-project/children/child-details/dropout-form/dropout-form.component.ts
```diff
@@ -0,0 +1,20 @@
...
+  styleUrls: ["./dropout-form.component.scss"],
+})
+export class DropoutFormComponent {
+  @Input() child: Child;
+
+  constructor(private entityMapper: EntityMapperService) {}
+
+  save() {
+    // Saving will save all the fields of the child, not just the dropout information
+    // Undoing changes that were not saved yet can be done by re-entering the view
```
sleidig, line 17: what does "re-entering the view" mean?

src/app/child-dev-project/children/children.module.ts
```diff
@@ -146,6 +149,10 @@ import { AttendanceAnalysisComponent } from "../attendance/attendance-analysis/a
     AttendanceAnalysisComponent,
     HealthCheckupComponent,
     PreviousSchoolsComponent,
+    GroupedChildAttendanceComponent,
+    HealthFormComponent,
+    DropoutFormComponent,
+    DropoutFormComponent,
```
sleidig, line 155: duplicate line?

sleidig:
dropout component has lost some styling:
image

src/app/child-dev-project/children/child-details/child-details.component.ts
```diff
@@ -58,13 +57,7 @@ export class ChildDetailsComponent implements OnInit {
...
     "",
   ];
-  eyeStatusValues = ["Good", "Has Glasses", "Needs Glasses", "Needs Checkup"];
-  vaccinationStatusValues = [
-    "Good",
-    "Vaccination Due",
-    "Needs Checking",
-    "No Card/Information",
-  ];
+
```
sleidig, line 60: how will we hide fields from the "basic information" section based on config? Do you already have an approach for this?

A temporary solution could be to make that section into a separate component as well and create copies for more than one variant.

sleidig: either way it could be a good idea to move the basic information section into a sub component, that way the responsibility of ChildDetailsComponent is clearer and it is less code in one file.

sleidig (commented):
looks good overall, once you check the last minor glitches :+1: 

from a UX perspective I am not totally happy with the separate save buttons in the sections. It would be nicer to tie changes back to the "global" save button at the top. Couldn't we simply do another input for the global editMode on/off into the components and let users only save through the top button?

src/app/child-dev-project/children/child-details/dropout-form/dropout-form.component.ts
TheSlimvReal: moving back to child list and clicking on the child again, anything that reloads the initial child from the database

src/app/child-dev-project/children/child-details/child-details.component.ts
TheSlimvReal: This would probably be the most stable approach. However I would at least try to make an attempt where we define the displayed information in the config file and see how the styling will become. We also should discuss soon, how we define the attributes on the `child` and `school` entity.

TheSlimvReal:
> 
> 
> looks good overall, once you check the last minor glitches 👍
> 
> from a UX perspective I am not totally happy with the separate save buttons in the sections. It would be nicer to tie changes back to the "global" save button at the top. Couldn't we simply do another input for the global editMode on/off into the components and let users only save through the top button?

But are you even sure that users will understand this connection to the `edit` button on the top of the page? When you scroll down to the health or the dropout section you cannot see these buttons anymore. 
It will also make the the configuration step much more difficult because we would need to define another concept how other information besides the `child` object are passed to the subcomponent.

sleidig:
> But are you even sure that users will understand this connection to the `edit` button on the top of the page? When you scroll down to the health or the dropout section you cannot see these buttons anymore.

That's true! far from ideal, even though it is the status quo. If we stick with that we should fix the title section with the name and edit/save buttons to always remain visible and scroll only the parts below.

Saving already saves all sections implicitly, so the global save button actually better reflects what is actually happening. But I do see the point of the clear and simple interface, so I am also fine with individual save buttons.

In any case, I would like to have a consistent editing approach, so if we keep individual buttons they should also have a read-only mode that first needs to be activated to edit the fields.

src/app/child-dev-project/children/child-details/child-details.component.html
TheSlimvReal: I think its alright to have multiple things displayed in one subfolder. We have the same in the education section.

src/app/child-dev-project/children/child-details/child-details.component.ts
sleidig: ok, I think that will happen in a separate, follow-up PR

sleidig (approved):
Looks and feels fairly intuitive now with the buttons. Ready to merge in my opinion
(when merging please sqash the commits of this PR into one clearly named commit following semantic release format: https://github.com/angular/angular.js/blob/master/DEVELOPERS.md#-git-commit-guidelines)

PR #509, opened 2020-10-13
Config lazy loading

christophscheuing:
Trying to get lazy loaded modules into the general config (with dynamic import() in the best case)


sleidig:
No easy solution at the moment. Postpone this, as it is a minor detail

PR #518, opened 2020-10-25
Basic info

TheSlimvReal:
see issue: #510 

- Moved Basic-Info to separate component
- Created abstract class which handles most of the logic for all form subcomponents
- Changed health and dropout form to use reactive forms instead of two-way-bindings

### Visible/Frontend Changes
- Each form subcomponent has its own `change`, `edit` and `cancel` button
- `Delete` button moved to end of page

### Architectural/Backend Changes
- [x] `Basic-Info`, `Health` and `Dropout` are now separate but quite similar subcomponents

Issue #493, opened 2020-09-08
Configuration to enable/hide sections of ChildDetails


sleidig:
_also see Summary #222 Configuration instead of Code_

To adapt the app to use cases of different organisations, as a first step we at least need to be able to hide certain sections through configuration without changing the code base. The main sections (Health, Educational Material, ...) should all be configurable to be displayed/hidden.

~Exact implementation to be discussed. Config documents in database?~
Use `ConfigService` and the json file configuration introduced in PR #504 

UI to edit this config is part of a later issue and not required right now.

sleidig:
Do we simply hide/show sections (with `*ngIf`) or do we load and insert components dynamically, like the implementation for dashboard widgets in #504 ?

PR #522, opened 2020-10-28
Child details config

TheSlimvReal:
see issue: #493 

After refactoring the `ChildDetailsComponent` (#518) the accordion can now be filled with the configuration provided in the config file. An approach similar to the `DashboardComponent` was chosen to reuse most of the code.

- The accordion in the ChildDetailsComponent is now created with the config file


TheSlimvReal:
replaced by #526

Issue #523, opened 2020-10-30
Style Basic-Info form dynamically


TheSlimvReal:
Currently the form in the basic info (and also in the health and dropout form) component are hard-coded in the HTML file.
In the future we will allow customers to modify the attributes of an entity and therefore the forms should automatically be generated from a configuration. 
This configuration should list **which attribute** is being displayed and **what input** type should be used (e.g. text, datepicker, select, ...), together with all the required information for this input type.

TheSlimvReal:
merged with #526

Issue #510, opened 2020-10-17
Basic Info in separate component


TheSlimvReal:
Move the basic info field of the child details section in to a separate component. This is should make the child details component only to be a wrapper for all the component that will be displayed inside of it.
To still allow users to edit and save changes, these buttons should be moved to the new basic info component as well. For a reference see the `dropout form component` and the `health form component`. In the end only the `delete` button should be left on child details component. Maybe put it after the `accordion` so it is not clicked on mistake. 

- [x] Create new subcomponent for the basic info with form and current layout
- [x] Remove `edit`, `cancel` and `save` button from child details
- [x] Move `delete` button to bottom of page

PR #526, opened 2020-11-02
Configurable Child Details

TheSlimvReal:
see issue: #493, #523 and #510

This PR makes the ChildDetails component configurable. Configurable are the panels which are displayed in the accordion and which attributes are displayed in the BasicInfo, Health and Dropout formular.

### Visible/Frontend Changes
- Each Form panel (BasicInfo, Health, Dropout) has its own `edit`, `save` and `delete` button
- The Form panels have a more generic styling
- The `delete` button moved to the bottom of the page

- The panels in the accordion of the child details section are created dynamically from the configuration
- The form panels (BasicInfo, Health, Dropout) all use the same component with a different configuration specified in the configuration file


src/app/child-dev-project/children/child-details/child-details.component.spec.ts
```diff
@@ -1,91 +1,103 @@
...
+  let routeObserver: Subscriber<any>;
   const mockedRoute = {
-    paramMap: of({ get: () => "new" }),
-    data: of({ icon: "child" }),
+    paramMap: new Observable((observer) => {
+      routeObserver = observer;
+      observer.next({ get: () => "new" });
+    }),
+    data: of({
+      panels: [
```
sleidig, line 28: if we don't have specific unit tests requiring all the panels, we can reduce this here in the spec to a minimal config, maybe just with one Form and one special component panel

src/app/child-dev-project/children/child-details/form/form.component.ts
```diff
@@ -0,0 +1,144 @@
...
+    if (changes.hasOwnProperty("child")) {
+      this.initForm();
+    }
+  }
+
+  onInitFromDynamicConfig(config: any) {
+    this.child = config.child;
+    this.config = config.config;
+    this.initForm();
+    if (!config.child.name) {
```
sleidig, line 48: that seems slightly hacky ... who knows how field names and required fields change in the future. Maybe we should add an explicit flag `isCreatingNew` to the config object instead?

src/app/child-dev-project/children/child-details/form/form.component.ts
```diff
@@ -0,0 +1,144 @@
...
+    this.enablePhotoUpload = this.childPhotoService.canSetImage();
+  }
+
+  async save(): Promise<Child> {
+    this.checkFormValidity();
+    this.assignFormValuesToChild(this.child, this.form);
+    try {
+      await this.entityMapperService.save<Child>(this.child);
+      if (this.creatingNew) {
+        this.router.navigate(["/child", this.child.getId()]);
```
sleidig, line 66: does angular prevent routing to the current route? Maybe we could omit the condition and just generally navigate after save?

src/app/child-dev-project/children/child-details/form/form.component.ts
```diff
@@ -0,0 +1,144 @@
...
+      }
+      this.alertService.addInfo("Saving Successful");
+      this.switchEdit();
+      return this.child;
+    } catch (err) {
+      this.alertService.addDanger(
+        'Could not save Child "' + this.child.name + '": ' + err
```
sleidig, line 73: can we get the form panel title as context for the log message?

src/app/child-dev-project/children/child-details/form/form.component.ts
```diff
@@ -0,0 +1,144 @@
...
+    this.form = this.fb.group(this.buildFormConfig());
+  }
+
+  private checkFormValidity() {
+    // errors regarding invalid fields wont be displayed unless marked as touched
+    this.form.markAllAsTouched();
+    this.validateForm = true;
+    if (!this.form.valid) {
+      const invalidFields = this.getInvalidFields();
+      this.alertService.addDanger(
```
sleidig, line 125: do we still need these alerts if we display the errors inline beneath the invalid form field?

sleidig (commented):
The code looks pretty solid and clean. And you even passed all the test coverage and other checks! :smiley: 

Thanks also for the good PR description.

Functionality also seems to work well.

A few tiny design issues I noticed:
- [x] some gaps appeared or are missing (see screenshot)
- [ ] somehow the notes panel is not colored anymore
- [ ] Edit / Save button would look a bit more prominent and clean if we can add it on the right side of the panel heading. Not sure whether this is doable with reasonable effort in the template though.

image

src/app/core/config/config-fix.json
```diff
@@ -108,13 +108,223 @@
...
+                    "components": [
+                        {
+                            "title": "",
+                            "component": "Form",
+                            "config": {
+                                "cols": [
+                                    [{
+                                        "input": "select",
+                                        "id": "health_vaccinationStatus",
+                                        "placeholder": "Peter Status",
```
sleidig, line 248:
```suggestion
                                        "placeholder": "Vaccination Status",
```
:smile:

src/app/child-dev-project/children/child-details/form/form.component.ts
TheSlimvReal: Currently it is not passed, because the title is set by the surrounding component. This way we can ensure a general styling for the titles. 
The alternative approach would be to let each component take care of displaying an title itself.

I understand that you want to show the user where this alert is coming from, but because it is shown after the user clicks the save button of a specific form-component, I think the user understands that the error is connected with this form.

src/app/child-dev-project/children/child-details/form/form.component.ts
TheSlimvReal: After trying it out, I did not see any reasons why not to do it

src/app/child-dev-project/children/child-details/form/form.component.ts
TheSlimvReal: On a mobile device, the field which is invalid might not be visible on the users screen, when he presses the save button. In this case the user does not get any direct feedback why the saving did not work and he has to search on his own for the problem. In this scenario the alert-snackbox gives the user valuable feedback.

TheSlimvReal:
Because the color bug is also on the master, I think it got introduced with #521 and I missed it at my review of this branch

TheSlimvReal:
About  the `edit`, `save` and `cancel` button. I think the `save` button at the end of the form makes sense, because users will fill out the form top to bottom and especially on mobile devices they would have to scroll back to the top to save. 
For the `edit` and `cancel`  button I can see why it should be at the top, I just dont really see how we could put it in the panel header with the current structure. Additionally, only the form components would have this edit button in the header and would look significantly different to other headers.
What we could do is put it at the top of the form but I dont know if that is better than the current solution.

src/app/child-dev-project/children/child-details/child-details.component.ts
```diff
@@ -37,161 +32,63 @@ import { UntilDestroy, untilDestroyed } from "@ngneat/until-destroy";
...
-    }
-  }
-
-  private assignFormValuesToChild(child: Child, form: FormGroup) {
-    Object.keys(form.controls).forEach((key) => {
-      const value = form.get(key).value;
-      if (value !== null) {
-        child[key] = value;
-      }
+  private addChildToConfig() {
```
sleidig, line 74: This looks a bit cumbersome. Any chance to avoid this step?

src/app/child-dev-project/children/child-details/form/form.component.html
```diff
@@ -0,0 +1,153 @@
+<form [formGroup]="form">
+    <div
+            fxLayout="row wrap"
+            fxLayout.xs="row wrap"
+            fxLayout.md="row wrap"
+            fxLayout.sm="row wrap"
+    >
+        <div *ngFor="let col of config.cols" fxFlex>
+            <div *ngFor="let row of col">
+                <div *ngIf="row.input === 'photo'" class="child-pic-container">
```
sleidig, line 10: all the details inside this should actually be its own component to separate this.

We can refactor this later, however

sleidig (approved):

Issue #511, opened 2020-10-17
Generate children list column group selector from config


sleidig:
The ChildrenListComponent allows users to select different views, i.e. different groups of columns to be displayed in the table. The names and assigned columns for this selection should be generated from config rather than fixed in code.

![image](https://user-images.githubusercontent.com/1682541/96333680-a7effc80-106b-11eb-8737-e14370cae298.png)

sleidig:
completed with #550

PR #527, opened 2020-11-03
WIP: Make list column groups configurable

christophscheuing:
see issue: #511 

Getting column groups for children list from config
Added "defaultColumnGroup" and "columnGroupSmallDisplay" in config, but made them optional in children-list.component.ts

One test still failing because accessing config.columnGroups[0].name as fallback in children-list.component.ts (line 90). Could someone please help me how to mock this in the spec.ts in order not to access an empty config.columnGroups?


TheSlimvReal:
See #550

PR #528, opened 2020-11-03
Sentry/Color fixes

### LINKED ISSUE #515: Cannot read property 'ok' of undefined [Sentry]
The sentry.io logs regularly show errors

>  Cannot read property 'ok' of undefined

https://sentry.io/organizations/ewb/issues/1866604280/

It's not clear where this is coming from. Needs to be debugged and fixed

### LINKED ISSUE #516: e.$implicit.getColor is not a function [Sentry]
The sentry.io logs regularly show errors

>  e.$implicit.getColor is not a function

https://sentry.io/organizations/ewb/issues/1930664778

It's not clear where this is coming from. Needs to be debugged and fixed

RichardNutt:
see issue: #515 #516 #529

This should fix the sentry errors. Unfortunately I was not able to reproduce the errors myself.

Also fixes the issue that colors were missing in the child notes tab. (I included this here since otherwise this would probably lead to merge conflicts due to the other changes.)

- [x] added undefined/null check to fix #515 
- [x] moved the getColor function to notes-of-child to fix #516
- [x] fixed demo data in light of #529 
- [x] fixed getColorForId()/getColor() in note.ts


src/app/core/entity/entity-mapper.service.ts
```diff
@@ -94,7 +94,7 @@ export class EntityMapperService {
       entity
     );
     const result = await this._db.put(rawData, forceUpdate);
-    if (result.ok) {
+    if (result?.ok) {
```
TheSlimvReal, line 97: result can be `null`?

TheSlimvReal (approved):
Looks very clean. Good work

src/app/core/entity/entity-mapper.service.ts
RichardNutt: Apparently. I think this causes #515. It's the only time we access a variable called 'ok'.

src/app/child-dev-project/notes/demo-data/notes_group-stories.ts
```diff
@@ -1,37 +1,40 @@
 import { WarningLevel } from "../../warning-level";
 
+/**
+ * Contains demo group notes. This data is used to create new Note instances with the provided values which then get saved to the MockDB.
+ */
 export const noteGroupStories = [
   {
-    category: "GUARDIAN_MEETING",
+    category: { name: "Guardians' Meeting", color: "#E1F5FE", isMeeting: true },
```
sleidig, line 8: because these objects are still transformed by the schema, I think just `category: { name: "Guardians' Meeting" }` should be sufficient and we can leave the color definition to the config file. Let's try that

src/app/child-dev-project/notes/model/note.spec.ts
```diff
@@ -115,4 +115,12 @@ describe("Note", () => {
     n6.togglePresence("1");
     expect(n6.attendances[0].present).toBe(false);
   });
+
+  it("should return colors", function () {
```
sleidig, line 119: :+1: great to have a test here

src/app/child-dev-project/notes/demo-data/notes_group-stories.ts
RichardNutt: Just to clarify: Do you want me to rewrite the to-database-transformation so that this would work? As of now we check if the category object matches the (stringified) one in the config file (including color, isMeating properties) in the InteractionSchemaDatatype-class.

Issue #533, opened 2020-11-18
Configurable Children List


TheSlimvReal:
The columns of the table in the children list component should be created dynamically from the configuration file.

sleidig:
completed with #550

Issue #512, opened 2020-10-17
Get filter options for list components from config


TheSlimvReal:
The available filters for the `child-list` and the `school-list` components should be loaded from the configuration file. The config should define how the filter is called and for which values it filters. In the list component this config should be used to create the correct filters and display them as it is currently done. The view of the component should be the same as before.
A problem might be filters which use a function. This might be a question that can be discussed in the whole team. 

- [ ] Figure out structure for filter configuration in the `child-list` / `school-list` config object
- [ ] Create config for all the currently available filters
- [ ] Write a function which reads the filter config from the config file and creates the filters
- [ ] Remove hard coded filters from the list components

sleidig:
completed with #550

PR #534, opened 2020-11-19
Configurable child list

TheSlimvReal:
see issue: #533

### Visible/Frontend Changes
- -- 

- Table columns are built from conifg file
- Removed  `ChildrenService` usage from `ChildrenList`
- Added some components to fetch the `ChildSchoolRelation` themselves


src/app/child-dev-project/children/children-list/children-list.component.html
```diff
@@ -144,178 +144,37 @@ <h1>{{ listName }}</h1>
...
+        <div *ngIf="col.type === 'date'">
+          {{ child[col.id] | date: "longDate" }}
+        </div>
+
+        <app-child-block
+                *ngIf="col.type === 'child-block'"
+                [entity]="child"
+        ></app-child-block>
+
+        <app-list-relation
```
sleidig, line 164: Actually this is a school block - or a child-school-relation. Name should be more explicit

src/app/child-dev-project/children/children-list/children-list.component.spec.ts
```diff
@@ -88,13 +143,16 @@ describe("ChildrenListComponent", () => {
...
+        { provide: ActivatedRoute, useValue: routeMock },
       ],
     }).compileComponents();
   }));
 
   beforeEach(() => {
     fixture = TestBed.createComponent(ChildrenListComponent);
     component = fixture.componentInstance;
+    const router = fixture.debugElement.injector.get(Router);
+    fixture.ngZone.run(() => router.initialNavigation());
```
sleidig, line 155: is this really best practice to go into the `ngZone` manually?

src/app/child-dev-project/children/children-list/children-list.component.ts
```diff
@@ -228,24 +216,15 @@ export class ChildrenListComponent implements OnInit, AfterViewInit {
...
-        this.centerFS.initOptions(centers, "center");
-
-        this.applyFilterSelections(replaceUrl);
-      });
-    this.childrenService
-      .getAttendances()
-      .pipe(untilDestroyed(this))
-      .subscribe((results) => this.prepareAttendanceData(results));
+  private async loadData(replaceUrl: boolean = false) {
+    this.childrenList = await this.entityMapperService.loadType<Child>(Child);
```
sleidig, line 220: The point of using `ChildrenService` is to let that load the school relation or other related data. Why should this be changed?

src/app/child-dev-project/children/children-list/children-list.component.ts
```diff
@@ -78,8 +70,8 @@ export class ChildrenListComponent implements OnInit, AfterViewInit {
         "age",
         "schoolClass",
         "schoolId",
-        "attendance-school",
-        "attendance-coaching",
+        "school",
```
sleidig, line 73: these columns are still about attendance, aren't they? the new naming doesn't make that clear.

src/app/child-dev-project/children/children-list/children-list.component.ts
```diff
@@ -257,49 +236,25 @@ export class ChildrenListComponent implements OnInit, AfterViewInit {
...
-    while (!found && i < this.columnGroups.length) {
-      if (this.columnGroups[i].name === columnGroupName) {
-        found = true;
-        group = this.columnGroups[i];
-      }
-      i++;
+    // When components, that are used in the list (app-list-attendance), also listen to the mediaObserver, a new
+    // mediaChange is created once this used component is displayed (through column groups change). This may
+    // re-trigger the settings for small screens. Therefore, we only allow a change ever 0.5 seconds to prevent this.
+    if (this.ready) {
```
sleidig, line 249: we really need to find a cleaner solution to deal with the resizing ...

src/app/child-dev-project/children/children-list/list-relation/list-relation.component.html
```diff
@@ -0,0 +1,3 @@
+
+<app-school-block *ngIf="id === 'schoolId'" [entityId]="relation?.schoolId"></app-school-block>
```
sleidig, line 2: maybe just inline this in the list-component, to reduce the number of components a bit?

src/app/child-dev-project/children/children-list/list-relation/list-relation.component.ts
```diff
@@ -0,0 +1,25 @@
...
+})
+export class ListRelationComponent implements OnChanges {
+  @Input() id: string;
+  @Input() child: Child;
+  relation: ChildSchoolRelation;
+
+  constructor(private childrenService: ChildrenService) {}
+
+  ngOnChanges(changes: SimpleChanges): void {
+    if (changes.hasOwnProperty("child")) {
```
sleidig, line 19: we gotta think and test performance of this approach. I would prefer to do loading once (for the complete child entity) and then have sub-components and columns work only as "display components" that get `@Input`s without using services to load data.

Did you have a particular reason to change and split this up?

sleidig (changes requested):
Sorting by a column is broken now. Otherwise looks good

src/app/child-dev-project/children/children-list/children-list.component.ts
TheSlimvReal: This makes it independent of the entity that is being displayed. I got rid of the `ChildrenService` and only use the `EntityMapper`. This means data is only loaded once it is actually required.

src/app/child-dev-project/children/children-list/children-list.component.ts
TheSlimvReal: This is done to align it with the `id`s that were used in the configuration for the columns (and is in this case the value which is passed to the attendance-component). Once #512 is done, a filter can be generated just by providing the id of that column.

src/app/child-dev-project/children/children-list/children-list.component.ts
TheSlimvReal: Yes, I don't like this either. The resizing configurations are scattered across the whole application. I would like to define it only once per view (Dashboard, School, ...).

src/app/child-dev-project/children/children-list/children-list.component.ts
TheSlimvReal: I think it is time to agree on an architecture which will always be present, e.g. there is always a child entity which stands in a relation to schools/team and we use this to optimize the application to this scenario or we want even this behavior to be configurable and every entity can stand in a relation to any other.

src/app/child-dev-project/children/children-list/list-relation/list-relation.component.ts
TheSlimvReal: This was done to loose the coupling between `ChildrenListComponent` and `ChildrenService`. This would be a necessary step if we want to use the same component for the children and school (list-)overview.

TheSlimvReal:
See #550

Issue #513, opened 2020-10-17
Create config structure for entities


TheSlimvReal:
(Some) fields of the `child`, `school` entity should be configurable. This means that it is not predefined which attributes the class has but some of the classes get extended with new attributes that are defined in the config file. For example one organization wants a `center` attribute on the child class while another wants the `address`. For this we need to figure out a concept how this can be done in our config file. An approach is to do it incremental. We start with attributes that are very basic (just text or a date) and only they can be defined with a configuration. Later we can create concepts for more difficult attributes.

PR #536, opened 2020-11-24
Configurable Entity Classes

TheSlimvReal:
see issue: #513

### Visible/Frontend Changes
--

- Attributes can now be added to an entity by defining them and their schema in the config file
- Some attributes of the child class that are never referenced directly in the code were moved to the config


src/app/child-dev-project/children/children.service.ts
```diff
@@ -23,12 +24,14 @@ export class ChildrenService {
...
     childPhotoService: ChildPhotoService,
-    private logger: LoggingService
+    private logger: LoggingService,
+    private entityConfigService: EntityConfigService
   ) {
     this.entitySchemaService.registerSchemaDatatype(
       new LoadChildPhotoEntitySchemaDatatype(childPhotoService)
     );
     this.createDatabaseIndices();
+    this.entityConfigService.addConfigAttributes<Child>(Child);
```
TheSlimvReal, line 34: A point for discussion is, where this line should be inserted. There are a few options:

- In the constructor of the `EntityMapperService`. This would ensure that all entities are registered when they are needed. But it will also register entities that might not be needed at all and it creates dependencies from the  `EntityMapperService` towards all entities and the `EntityConfigService`.
- In the load and save functions of the `EntityMapperService` this would ensure that an entity is only registered once it is actually needed. A list of registered entities would ensure that each entity is only registered once. This appraoch would also introduce the dependency towards the `EntityConfigService` and I didn't yet find a way to get the class (e.g. `Child`) when I only have an object of this class (e.g. `new Child("...")`). This is the case in the `load` function of the `EntityMapperService`. However, the class is needed in order to set up the `schema` of this class.
- In the `app.module.ts` constructor. This would also mean that all classes are registered no matter if they are used or not.

sleidig (approved):

src/app/child-dev-project/children/children.service.ts
TheSlimvReal: But modules can load this class without using the module (through the `EntitiyMapperServic`). If this happens before the `ChildrenModule` was used, it would result in the `Child` class not being configured.

src/app/child-dev-project/children/model/child.ts
```diff
@@ -34,43 +34,23 @@ export class Child extends Entity {
...
   @DatabaseField() center: string = "";
   @DatabaseField() admissionDate: Date;
   @DatabaseField() status: string = "";
 
-  /** current school (as determined through the ChildSchoolRelation docs) set during loading through ChildrenService */
-  schoolId: string = "";
-  /** current class (as determined through the ChildSchoolRelation docs) set during loading through ChildrenService */
-  schoolClass: string = "";
-
-  @DatabaseField() address: string = "";
```
sleidig, line 51: We need to make really sure that we are not _ever_ missing an entity field in the config, otherwise people would just delete properties from objects in the database whenever they save them!
Can we think of any safeguards? Or should we consider switching the approach to one where fields not mentioned in the schema are simply saved without transformation as well?

TheSlimvReal: In that case, I would be worried about saving too much with the entity, if we just save everything. But what about the revisions? Aren't older revisions still on the database?
And on the other hand, its still just a property on the entity that would go lost in the rare case that a broken config is actually merged into the master. I don't see how a change like this would be easily approved for a merge.

src/app/child-dev-project/children/children.service.ts
TheSlimvReal: I now even moved it to `app.module` to ensure that no entity can be loaded or saved before it has been registered. 
To make it more "lazy", it could be moved to the `EntityMapper` but this would introduce new dependencies.

TheSlimvReal:
Live at https://config-documentation-test-depl.herokuapp.com/

src/app/child-dev-project/children/model/child.ts
sleidig: Any data loss is absolutely unacceptable. at least we need to be extremely aware when any changes are done. but now that a lot of properties were moved from class to config I don't have a definite overview that we didn't loose any even right now.

I am thinking whether we can at least write a unit tests that flags changes ...

src/app/core/config/config-fix.json
```diff
@@ -503,5 +503,30 @@
                 }
             ]
         }
+    },
+
+    "entity:Child": {
+        "attributes": [
+            {"name": "gender", "schema": { "dataType": "string" } },
```
sleidig, line 510: gender, mothertongue and religion are still defined in the class, also. Should we define every field in the config - or remove every field that is already defined in the class?

sleidig (approved):

src/app/core/config/config-fix.json
TheSlimvReal: I removed the duplicates from the config. We should soon clean up all helgo dependent attributes from the entities and tests, so we are not stuck with them on future deployments

Issue #405, opened 2020-02-02
Move CI from Travis to GitHub Workflow


TheSlimvReal:
GitHub recently published their own continuous integration tool. This allows to create better looking reports which might help to distinguish whether problems happened in the `lint`, `test` or `built` step.
It might also make results more reproducible (Travis sometimes has different results on the same code).

TheSlimvReal:
Tasks:

- [ ] Create Pipeline for install, test (lint+unit+[upload to codecov](https://github.com/codecov/codecov-action)), build and deploy to gh-pages
- [ ] Use caching to speed up builds
- [ ] Add compodoc and semantic release workflow for merges to the master branch
- [ ] Add publish workflow for tagged commits

PR #537, opened 2020-11-24
added new workflow for the whole build pipeline and also added caching

TheSlimvReal:
see issue: #405

- Travis is no longer used
- The pipeline is now implemented in GitHub workflows
- Caches are implemented to allow for faster builds
- `npm ci` is used instead of `npm install` to ensure reproducible builds


TheSlimvReal:
Replaced by #545

PR #551, opened 2020-12-08
Storybook

sleidig:
add storybook for easier UI development and as a component reference library

see https://storybook.js.org/docs/angular/get-started/introduction


src/storybook-examples/Button.stories.ts
```diff
@@ -0,0 +1,39 @@
+// also exported from '@storybook/angular' if you can deal with breaking changes in 6.1
+import { Story, Meta } from '@storybook/angular/types-6-0';
+import Button from './button.component';
+
+export default {
```
TheSlimvReal, line 5: Should we really keep the examples?

TheSlimvReal (approved):
It would now be good to deploy this for the master. Maybe on gh-pages

PR #556, opened 2020-12-11
Split Docker workflow

TheSlimvReal:
This workflow should run faster and independently run the tests and deploy the image


sleidig:
This looks pretty good and with very reasonable CI durations :+1: 

Can we get this into master soon? Currently we are running blind without automated tests.

What we need additionally is the two travis jobs:
- semantic release on every commit on master
- publish to docker hub on every tag

TheSlimvReal:
Not necessary due to speedup in #545

PR #562, opened 2020-12-15
Build and deploy application with Docker and Heroku

TheSlimvReal:
see issue: #405

The idea is to create a Dockerfile which allows to install and run the application without any requirements to the host (besides docker). The image installs all necessary dependencies and runs the application when executed. For each PR the Docker image is built and deployed to heroku.

Because we want to create a minimal image which only consist of the necessary files to run the application, everything that is not necessary is disposed. In the first step the image is only built for the first stage and the coverage report is uploaded from there. In the second step the image is fully built and uploaded.

### Visible/Frontend Changes
-- 

- Tests and build runs in GitHub
- Creating a docker image for each PR
- Uploading and deploying the image on Heroku
- Creating semantic release for each push on the master
- Uploading the cache for each push on the master
- Uploading the image for each push on the master


.github/workflows/pull-request-closed.yml
```diff
@@ -0,0 +1,15 @@
...
+on:
+  pull_request:
+    types: [closed]
+
+jobs:
+  remove-heroku-app:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Create app
```
sleidig, line 11:
```suggestion
      - name: Remove app
```

.github/workflows/pull-request-update.yml
```diff
@@ -0,0 +1,56 @@
...
+        uses: paambaati/codeclimate-action@v2.7.5
+        env:
+          CC_TEST_REPORTER_ID: ${{ secrets.CODE_CLIMATE_ID }}
+        with:
+          coverageLocations: ${{github.workspace}}/img/app/coverage/lcov.info:lcov
+  deploy-prod-image:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - name: Set up Docker Buildx
```
sleidig, line 35: there is some duplication here with the setup for building on master. Is there any feasible way to let them use the same config?

sleidig (approved):
check the docker username, otherwise ready for merge

.github/workflows/pull-request-update.yml
TheSlimvReal: I dont think so and you can not refer to a job in a different Workflow

PR #573, opened 2020-12-19
Schema datatype map

sleidig:
- [x] add new datatype for EntitySchema system that transforms JavaScript `Map` to a simple object


TheSlimvReal (approved):
Nice construction of the datatype. I like how the nesting is handled

PR #575, opened 2020-12-19
refactor(*): generate a default uuid in Entity base class

sleidig:
as we have already started using auto-generated uuids in several places, make this available by default.

- [x] make id paramter in the `Entity` base class constructor optional. If no id is given, generate a uuid
- [x] remove uniqid package as it has the same purpose as uuid package


TheSlimvReal (approved):
Clean solution, I like it

PR #586, opened 2020-12-29
Entity Subrecord adaptions

sleidig:
- [x] inline editing (in table view) of `EntitySubrecordComponent` can be disabled
- [x] column values can be calculated through a more flexible function, defined in the column definition

> Please merge without sqashing.

### DIFF
--- a/src/app/child-dev-project/children/child-block/child-block.component.html
+++ b/src/app/child-dev-project/children/child-block/child-block.component.html
@@ -2,7 +2,7 @@
       (mouseenter)="showTooltip()" (mouseleave)="hideTooltip()" (click)="showDetailsPage()"
       [ngClass]="{ 'inactive': !entity.isActive }">
   <img [src]="entity?.photo?.value" class="child-pic">
-  {{entity?.name}} <span style="font-size: x-small">({{entity?.projectNumber}})</span>
+  {{entity?.name}} <span style="font-size: x-small" *ngIf="entity?.projectNumber">({{entity?.projectNumber}})</span>
 </span>
 
 <div style="position:absolute;" *ngIf="tooltipVisible">
--- a/src/app/core/entity-components/entity-subrecord/entity-subrecord.component.spec.ts
+++ b/src/app/core/entity-components/entity-subrecord/entity-subrecord.component.spec.ts
@@ -7,14 +7,13 @@ import { MockDatabase } from "../../database/mock-database";
 import { ConfirmationDialogService } from "../../confirmation-dialog/confirmation-dialog.service";
 import { Database } from "../../database/database";
 import { Entity } from "../../entity/entity";
-import { ColumnDescription } from "./column-description";
 import { ColumnDescriptionInputType } from "./column-description-input-type.enum";
 import { By } from "@angular/platform-browser";
 import { SimpleChange } from "@angular/core";
 import { NoopAnimationsModule } from "@angular/platform-browser/animations";
 import { MatNativeDateModule } from "@angular/material/core";
 
-fdescribe("EntitySubrecordComponent", () => {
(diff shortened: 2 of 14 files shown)


src/app/core/entity-components/entity-subrecord/entity-subrecord.component.ts
```diff
@@ -60,6 +60,11 @@ export class EntitySubrecordComponent implements OnInit, OnChanges {
    */
   @Input() detailsComponent: ComponentType<ShowsEntity>;
 
+  /**
+   * Whether the records can be edited directly in the table.
+   */
+  @Input() inlineEditable: boolean = true;
```
TheSlimvReal, line 66: I think `editable` or `allowEditing` would be more clear.

TheSlimvReal (changes requested):
A good extension of the subrecord component but it took me much to long to understand what it is actually meant for. I think the description could be much clearer and I also see an overlap with the `formatter` function.

src/app/core/entity-components/entity-subrecord/entity-subrecord.component.ts
sleidig: "inline" because there is also the functionality to open a popup and edit an entry there, which is independent of the editing inline as a table row directly

src/app/child-dev-project/children/child-block/child-block.component.html
```diff
@@ -2,7 +2,7 @@
       (mouseenter)="showTooltip()" (mouseleave)="hideTooltip()" (click)="showDetailsPage()"
       [ngClass]="{ 'inactive': !entity.isActive }">
   <img [src]="entity?.photo?.value" class="child-pic">
-  {{entity?.name}} <span style="font-size: x-small">({{entity?.projectNumber}})</span>
+  {{entity?.name}} <span style="font-size: x-small" *ngIf="entity?.projectNumber">({{entity?.projectNumber}})</span>
```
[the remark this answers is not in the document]
sleidig: yes, that may become necessary, we could also then merge the different block components (child block, school block) and move the differences into config only. I'd leave it as is for now until we come across the need, however.

src/app/core/entity-components/entity-subrecord/entity-subrecord.component.ts
sleidig: would `disableEditingInline` be more readable? that would also be more similar to the broadly used `disabled` property

TheSlimvReal: Yes I think thats a good idea

TheSlimvReal: Yes I think thats a good idea

src/app/core/entity-components/entity-subrecord/entity-subrecord.component.spec.ts
```diff
@@ -33,4 +47,91 @@ describe("EntitySubrecordComponent", () => {
...
+      columns: new SimpleChange(undefined, component.columns, true),
+    });
+    fixture.detectChanges();
+
+    const tdColumns = fixture.debugElement.queryAll(By.css("td"));
+    expect(tdColumns[0].nativeElement.innerText).toBe("0.9");
+    expect(tdColumns[1].nativeElement.innerText).toBe("90%");
+  });
+
+  it("formats MONTH automatically", () => {
```
TheSlimvReal, line 116: Maybe you could also add an test for the DAY, since it is also hard coded into the component.

TheSlimvReal (approved):
Nice! I really like how you changed the ColumnDescription to an interface. I think it becomes much more readable and just having one value function seems to work nicely.

PR #589, opened 2021-01-04
Configurable Enums

sleidig:
generalized the existing implementation to configure the available interaction-types (Note.category) in the config. Now any kind of select value collections can be created and used through schema datatypes and structural directives

- [x] added schema datatype `"configurable-enum"` (`@DatabaseField({ dataType: "configurable-enum", innerDataType: INTERACTION_TYPE_CONFIG_ID }) category: InteractionType;`) to convert between id and full enum value from the config for any config entry defining a "configurable enum" collection
- [x] added directive `*appConfigurableEnum` (`<div *appConfigurableEnum="let item of 'interaction-type'"></div>`) to iterate through all options of a "configurable enum" defined in the config


src/app/core/configurable-enum/configurable-enum-directive/configurable-enum.directive.ts
```diff
@@ -0,0 +1,51 @@
+import { Directive, Input, TemplateRef, ViewContainerRef } from "@angular/core";
+import {
+  CONFIGURABLE_ENUM_CONFIG_PREFIX,
+  ConfigurableEnum,
+  ConfigurableEnumValue,
+} from "../configurable-enum.interface";
+import { ConfigService } from "../../config/config.service";
+
+/**
+ * Enumerate over all {@link ConfigurableEnum} value for the given enum config id.
```
TheSlimvReal, line 10:
```suggestion
 * Enumerate over all {@link ConfigurableEnum} values for the given enum config id.
```

src/app/core/configurable-enum/configurable-enum.interface.ts
```diff
@@ -0,0 +1,35 @@
+/**
+ * Interface specifying overall object representing an enum with all its options
+ * as stored in the config database
+ */
+export type ConfigurableEnum<
+  T extends ConfigurableEnumValue = ConfigurableEnumValue
```
TheSlimvReal, line 6: what does this line mean?

src/app/core/configurable-enum/configurable-enum.module.ts
```diff
@@ -0,0 +1,63 @@
...
+  imports: [CommonModule],
+  exports: [ConfigurableEnumDirective],
+})
+export class ConfigurableEnumModule {
+  constructor(
+    private configService: ConfigService,
+    private entitySchemaService: EntitySchemaService
+  ) {
+    this.entitySchemaService.registerSchemaDatatype(
+      new ConfigurableEnumDatatype(configService)
```
TheSlimvReal, line 60: I think there is a problem. If a component does not use the `ConfigurableEnumDirective`, the constructor of this module is not executed. This means if the component loads an entity with an `ConfigurableEnumDatatype` attribute, the schema service will not be able to parse that object. I think the registration have to happen at another place.

TheSlimvReal (commented):
Very nice work. I really like how flexible this new datatype is and how well it is integrated into the directive. I think that will be used a lot in the future.

src/app/child-dev-project/notes/model/note.spec.ts
```diff
@@ -118,9 +145,21 @@ describe("Note", () => {
...
   it("should return colors", function () {
     const note = new Note("1");
-    note.category = { name: "test", color: "#FFFFFF" };
+    note.category = { id: "", label: "test", color: "#FFFFFF" };
     expect(note.getColor()).toBe("#FFFFFF");
     note.warningLevel = WarningLevel.URGENT;
     expect(note.getColor()).toBe(WarningLevelColor(WarningLevel.URGENT));
   });
+
+  it("transforms interactionType from config", function () {
```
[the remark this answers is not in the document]
sleidig: I had added this as an explicit test case before changing the implementation in the sense of test driven development because the general schema test was ignoring the special transformation before

src/app/child-dev-project/notes/notes-manager/notes-manager.component.ts
```diff
@@ -67,7 +67,7 @@ export class NotesManagerComponent implements OnInit {
     private sessionService: SessionService,
     private media: MediaObserver,
     private entityMapperService: EntityMapperService,
-    private configLoader: NoteConfigLoaderService,
+    private configService: ConfigService,
```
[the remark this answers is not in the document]
sleidig: thanks, that's true. I guess that is left over from replacing configLoader ...

src/app/core/configurable-enum/configurable-enum.interface.ts
sleidig: `ConfigurableEnum` (as the type how an enum collection is stored in the config) is an array of objects implementing `ConfigurableEnumValue` interface by default. But the objects in the config can also have additional properties (like a color, etc.), then e.g. `ConfigurableEnum<InteractionType>` (where `InteractionType` extends the basic `ConfigurableEnumValue`) as a type already also includes type hints for the additional peroperties that InteractionType has.

src/app/core/configurable-enum/configurable-enum.module.ts
sleidig: It seems module constructors can be used for setup: https://www.bennadel.com/blog/3180-ngmodule-constructors-provide-a-module-level-run-block-in-angular-2-1-1.htm

I also checked and confirmed now that the constructor is only called once even if the module is imported multiple times.

TheSlimvReal (approved):
Looks good!

PR #593, opened 2021-01-10
Fix Webdav proxy

TheSlimvReal:
The proxy for Webdav is specified in the nginx config file `default.conf` and dependes on a environment variable.
This variable is now injected into the config file to fix the proxy.

- Webdav url injection

### DIFF
--- a/deploy/nginx/default.conf
+++ b/deploy/nginx/default.conf
@@ -32,7 +32,7 @@ server {
 
     location /webdav {
         rewrite /webdav/(.*) /$1 break;
-        proxy_pass ${WEBDAV_URL};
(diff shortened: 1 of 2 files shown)


deploy/nginx/default.conf
```diff
@@ -32,7 +32,7 @@ server {
 
     location /webdav {
         rewrite /webdav/(.*) /$1 break;
-        proxy_pass ${WEBDAV_URL};
+        proxy_pass $WEBDAV_URL;
```
sleidig, line 35: why removing the brackets? This used to work and is required if https://serverfault.com/a/755541/511220 is correct

sleidig (approved):

Issue #558, opened 2020-12-13
Save config in the database (or deploy config)


sleidig:
The individual configuration for an instance of a single customer should be easy to switch and update - and not clog up our main core repository.

It could either be saved in a (separate, readonly) database or somehow managed as part of the deploy process.

TheSlimvReal:
Moving the config to the DB sounds like the right way to later allow people to customize their UI from the app.
However for the beginning I like being able to edit the config file as a whole.
Therefore I suggest to create a section in the admin area where the config file can be uploaded and downloaded.
At the beginning a default config will be selected and can be replaced by uploading a new one

TheSlimvReal:
Steps:

- [ ] Extend config service to save a config file into the db
- [ ] Extend config service to create a file from a config loaded from the db
- [ ] Create UI component in admin section do upload and download config
- [ ] Set fallback to a default config file

PR #594, opened 2021-01-10
Save config in DB

TheSlimvReal:
see issue: #558 

### Visible/Frontend Changes
- New section in the admin panel to upload and download the config file 

- Initially, the config is loaded from the db with a fallback to a local config
- Added hook for registering functions that should run, when the config changes
- Move loading of routes, navigation and entity configs
- New Config Entity
- Increased test coverage for the admin component
- Removed a library for saving files and replaced it with plain javascript


src/app/core/admin/admin/admin.component.ts
```diff
@@ -71,6 +75,19 @@ export class AdminComponent implements OnInit {
...
 
+  downloadConfigClick() {
+    const jsonString = this.configService.exportConfig();
+    this.startDownload(jsonString, "text/json", "config.json");
+  }
+
+  uploadConfigFile(file: Blob) {
+    this.readFile(file)
+      .then((res) =>
+        this.configService.saveConfig(this.entityMapper, JSON.parse(res))
```
sleidig, line 86: use dependency injection instead of passing around the dependency here?

src/app/core/config/config.module.ts
```diff
@@ -1,10 +1,11 @@
 import { NgModule } from "@angular/core";
 import { CommonModule } from "@angular/common";
 import { ConfigService } from "./config.service";
+import { EntityModule } from "../entity/entity.module";
 
 @NgModule({
   declarations: [],
-  imports: [CommonModule],
+  imports: [CommonModule, EntityModule],
```
sleidig, line 8: not necessary if dendency added to configmodule

src/app/core/config/config.service.ts
```diff
@@ -1,35 +1,63 @@
...
+import { BehaviorSubject } from "rxjs";
 
 @Injectable({
   providedIn: "root",
 })
 export class ConfigService {
-  constructor() {}
+  private static CONFIG_KEY = "CONFIG_ENTITY";
+  private config: Config = new Config(ConfigService.CONFIG_KEY);
+  public configNotifier: BehaviorSubject<Config>;
```
sleidig, line 14: clarify name, maybe `configUpdated` or `currentConfig`?

sleidig (approved):
approving this as a first step towards the goal. We need to rework the startup process as soon as possible, though (#595). This is introducing tighter coupling and dependencies at the moment.

PR #580, opened 2020-12-21
[WIP] Event Attendance (EPIC)

sleidig:
Collecting implementation in multiple steps:

- [x] PR #555 
- [x] PR #560 
- [x] PR #569 
- [x] PR #585
- [x] attendancestatus as config & dropdown for all attendance-status in note details attendance list (PR #616)
- [x] demo data generation for event attendance (PR #617)
- [x] AttendanceBlocks in list (PR #621)
- [x] Dashboard widgets for Attendance
- [x] migration of previous attendancemonth records to new events (PR #651)
- [x] DB Indices (PR #648)
- [x] disable attendance analysis and attendance monthly forms until they are re-implemented
- (later) implement a system for monthly attendance (entering monthly totals manually instead of day by day details)

TODO Refactorings:
- [x] remove `AttendanceDay`
- [x] remove `MeetingNoteAttendance`
- [x] create and assign a sensible interactionType category to Notes that are Events relating to a RecurringActivity
- [x] rename activity-setup component to roll-call-setup
- [x] rename attendance-days component to attendance-calendar
- [x] remove AttendanceStatus enum

TODO UX:
- [x] display a different details popup when opening a month's event attendance popup from an individual child (currently the popup shows averages instead of data for the focused child)
- [ ] ~select-field (dropdown) for user -> RecurringActivity.assignedTo~ (--> #681)
- [ ] only isMeeting interaction-types allowed in activitiy
- [ ] notes dashboards should only count non-event notes (or ones where kid is present)
- [ ] when I am in Child Details, I can not select a different month in the calendar for the attendance.
- [ ] AttendanceService events_index filters by EventNote but we do want to allow users to create one-off events in notes section also ...

PR #596, opened 2021-01-11
PPI Adaptions

sleidig:
### Depends on
- [ ] merge Event Attendance Feature (#580) into master
- [x] load config (or at least its id) dynamically (#594)
- [ ] disable actions (create new player/team/activity only allowed for managers) based on user's isAdmin flag and make configurable which actions require the permission in config
- [x] configurable attendance status options

### TODOs
- [ ] load columns of sections using entitysubrecord-component from config
- [x] add select fields to note & recurringAcitivty: programme, isMixedEvent, 
- [x] adapt available attendance status options: present, absent, excused


sleidig:
can be handled completely through config in database and the epic/event-attendance branch

PR #602, opened 2021-01-15
Change password

TheSlimvReal:
see issue: #17 

What is not yet working is the password for the nextcloud. The workflow the adaption of this password still has to be specified.

### Visible/Frontend Changes
- Error messages for wrong password and failed process

- The password change functionality is implemented


sleidig:
Seems to work well at its core :+1: 

A few things still to do:

- [ ] disable the password change form if the user is in temporary/demo mode
- [ ] error message needs to be displayed in form if old password is not correct
- [ ] error message needs to be displayed in form if new passwords don't match
- maybe the validations can best be done with Angular Custom Validators (so we can easily add more validations also): https://blog.angular-university.io/angular-custom-validators/
- [ ] it might be worthwhile to switch the form to use ReactiveForms (especially for the validators) and `matInput`s
- [ ] I am still seeing a 401 in the console after successful password change (`POST http://localhost:4200/db/app/_revs_diff 401 (Unauthorized)`)

and points for discussion:
- should we move all the logic from the User.setPassword method also into the UserAccountService?

sleidig (approved):

TheSlimvReal:
aam_error_message

Some problems with the error message. This [issue](https://github.com/angular/components/issues/4580) seems to be a common problem. I will try to find a fix.

PR #604, opened 2021-01-15
Add componentConfig to detailsComponent for EntitySubrecordComponent popup

sleidig:
- [x] allow detailsComponent of SubrecordEntityComponent to receive a custom config object to adapt to different contexts


TheSlimvReal (approved):
Changes look very solid, I like the test case

PR #616, opened 2021-01-25
Configurable attendance status [Event Attendance Part +1]

sleidig:
- replace hard-coded attendance status options with a configurable enum loaded from config
- update children attendance list in NoteDetails with proper selectable version that is offering all options

the code overall is becoming bloated with too many components and files and complicated demo setups. I will try to simplify this once all the basic functionality is implemented for the epic.

### DIFF
--- a/src/app/core/config/config-fix.json
+++ b/src/app/core/config/config-fix.json
@@ -58,6 +58,8 @@
       }
     ]
   },
+
+
   "enum:interaction-type": [
     {
       "id": "",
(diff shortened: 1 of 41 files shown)


src/app/core/config/config-fix.json
```diff
@@ -161,6 +163,53 @@
...
+      "label": "Absent",
+      "style": "attendance-A",
+      "countAs": "ABSENT"
+    },
+    {
+      "id": "LATE",
+      "shortlabel": "L",
+      "label": "Late",
+      "style": "attendance-L",
+      "countAs": "PRESENT"
```
TheSlimvReal, line 187: I am not so sure if `countAs` is at the right place here. 
I can see the use case here but I am worried that in future requirements, multiple `countAs`  fields would be required for different calculations. However I think this change can be done latter without much additional work.

TheSlimvReal (approved):

PR #617, opened 2021-01-25
Demo event attendance [Event Attendance Part +2]

sleidig:
generate demo events including individual attendance for every RecurringActivity

### DIFF
--- a/src/app/core/config/config-fix.ts
+++ b/src/app/core/config/config-fix.ts
@@ -1,4 +1,8 @@
-{
+import { defaultAttendanceStatusTypes } from "./default-config/default-attendance-status-types";
+import { defaultInteractionTypes } from "./default-config/default-interaction-types";
+
+// prettier-ignore
+export const defaultConfig = {
   "navigationMenu": {
     "items": [
       {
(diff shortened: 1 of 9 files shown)


src/app/core/config/config-fix.ts
```diff
@@ -1,4 +1,8 @@
-{
+import { defaultAttendanceStatusTypes } from "./default-config/default-attendance-status-types";
+import { defaultInteractionTypes } from "./default-config/default-interaction-types";
+
+// prettier-ignore
+export const defaultConfig = {
```
TheSlimvReal, line 5: we should be careful to never put any logic in this file

TheSlimvReal (approved):
The demo data looks good, but when I am in Child Details, I can not select a different month in the calendar for the attendance. Is this intended?

sleidig:
> when I am in Child Details, I can not select a different month in the calendar for the attendance. Is this intended?

This is not supported yet because the calendar is only showing events of the first aggregation period of the table (i.e. the most recent month). We might be able to extend this feature though. Will add this to the list for the epic

src/app/core/config/config-fix.ts
sleidig: that's an important remark. After you raised this I looked at typescript typings for it, but it seems to be far from trivial to define a JSON type ...

Issue #615, opened 2021-01-24
Config is not applied after initial sync


sleidig:
**Describe the bug**
After clearing all storage and loading the app, I see the default config (dashboard + forms, ...) instead of the one uploaded into the database.

**To Reproduce**
1. Test on a site with a server database and upload a custom config to it before following these steps:
2. Clear storage, removing existing IndexedDb to imitate a very first login for a user
2. Login and wait for initial sync
3. You see the default dashboard, not the one from the config in the db; even logging in and out does not help.
4. Reload the site. You now see the expected dashboard configured in the db

PR #620, opened 2021-01-26
Fix Reloading Config

TheSlimvReal:
see issue: #615 

### Visible/Frontend Changes
- The config is applied at the initial sync 
- The current route is not lost when logging in/out

- The config is reloaded, once the sync status changes to `COMPLETED`
- After the config is loaded, a routing to the same location is started, to apply config that affects the first page that is already loaded

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -27,6 +27,7 @@ import { EntityConfigService } from "./core/entity/entity-config.service";
 import { Child } from "./child-dev-project/children/model/child";
 import { SessionService } from "./core/session/session-service/session.service";
 import { SyncState } from "./core/session/session-states/sync-state.enum";
+import { ActivatedRoute, Router } from "@angular/router";
 
 /**
  * Component as the main entry point for the app.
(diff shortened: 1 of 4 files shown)


src/app/app.component.ts
```diff
@@ -42,13 +45,22 @@ export class AppComponent implements OnInit {
...
     // The EntityMapperService needs the SessionServiceProvider which needs the AppConfig to be set up.
     // If the EntityMapperService is requested to early (through DI), the AppConfig is not ready yet.
     // TODO fix this with https://github.com/Aam-Digital/ndb-core/issues/595
     configService.loadConfig(entityMapper);
+    // Reload config once the database is synced
+    sessionService
+      .getSyncState()
+      .waitForChangeTo(SyncState.COMPLETED)
+      .then(() => configService.loadConfig(entityMapper))
+      .then(() => router.navigate([], { relativeTo: this.activatedRoute }));
```
sleidig, line 63: if this is necessary to apply changes done by our `RouterService, the navigate should probably happen there in the `reloadRouting`?

sleidig (approved):

PR #624, opened 2021-01-29
refactor: implement a EntityTestingModule for easier integration tests

sleidig:


TheSlimvReal (approved):
Sounds like a good idea, I guess you will use it in the Attendance epic

PR #629, opened 2021-02-02
Simplify Index/Query code refactorings

sleidig:
see commits

> please merge without squashing!


src/app/core/entity/entity.ts
```diff
@@ -112,6 +112,28 @@ export class Entity {
...
+    let indices = [];
+
+    // default indices generated from "name" property
+    if (this.hasOwnProperty("name")) {
+      indices = this["name"].split(" ");
+    }
+
+    return indices;
+  }
+  set searchIndices(value) {
```
TheSlimvReal, line 132: Is there a reason why this function is implemented?

src/app/core/ui/search/search.component.ts
```diff
@@ -97,7 +97,7 @@ export class SearchComponent implements OnInit {
 
     const searchTerms = searchString.split(" ");
 
-    const queryResults = await this.dbIndexingService.queryIndex(
+    const queryResults = await this.dbIndexingService.queryIndexRaw(
```
TheSlimvReal, line 100:
```suggestion
    const queryResults = await this.dbIndexingService.queryIndexDocs<Entity>(
```
In a later step the results are transformed into entities, I did not find a reason why the raw results are needed

src/app/core/entity/database-indexing/database-indexing.service.ts
```diff
@@ -73,10 +79,39 @@ export class DatabaseIndexingService {
...
+    indexName: string,
+    options: QueryOptions = {
+      reduce: true,
+      group: true,
+    }
+  ): Promise<any> {
+    return await this.queryIndexRaw(indexName, options);
+  }
+
+  async queryIndexRaw(indexName: string, options: QueryOptions): Promise<any> {
```
TheSlimvReal, line 114: With the changes in `search.compontent.ts`, this function can be made private
```suggestion
  private async queryIndexRaw(indexName: string, options: QueryOptions): Promise<any> {
```

TheSlimvReal (approved):
I like how the proposed changes removed a lot of duplicate code and more values could be  typed. I propose to make `queryIndexRaw` completely hidden (private) and only usable through `queryIndexStats` and `queryIndexDocs`.

Another thought I had: The `createIndex` function additionally expects a type and a constructor function as parameters. This way a developer has to specify the return type of the index while creating the index instead of when the index is used. To make it completely type-safe, the `createIndex` function would return a function that can be used to query the index later, this would replace the `queryIndexDocs` and `queryIndexStats` functions. 
In the future this could be used to automatically create indices for entities by assigning the query function to a property of the entity once the index is created. 
E.g. The `Child` object has a function `getAserResults: () => Aser[]` with the annotation `@Index(Aser, 'childId')`. This annotation automatically triggers the creation of the index and the returned query function will override the `getAserResults` function of the child class.

src/app/core/entity/entity.ts
sleidig: SchemaService is writing all fields' values and I received errors when I didn't define a setter. Alternatively we could introduce a special case in the schemaService for searchIndices again and not load them - but I preferred to keep it general there and add this peculiar empty setter here.

src/app/core/database/query-options.ts
```diff
@@ -0,0 +1,14 @@
+/**
+ * Basic query options supported by {@link Database}
+ */
+export interface QueryOptions {
```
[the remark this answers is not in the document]
sleidig: Should we list all options or only those we "consciously" support?

I guess at almost everything is based on PouchDB the best way is to allow all PouchDB options.

src/app/core/ui/search/search.component.ts
sleidig: I like the idea to hide `queryIndexRaw` (in fact I had been thinking the same) but the search component is currently transforming the results to either Child or School entities as they can be both. We could parse into any type based on the prefix but that would require a central register of all Entity types (probably something using Angular DI with `multi: true` would be great).

As the priority is finally getting the event attendance epic merged, I would argue to postpone all this a bit.

src/app/core/entity/entity.ts
TheSlimvReal: Ok makes sense

src/app/core/ui/search/search.component.ts
TheSlimvReal: I can agree to postponing it. Making the search component more flexible is not really required yet I gues

PR #634, opened 2021-02-05
feat(core): allow hiding/filtering of entity subrecord columns through config

sleidig:


TheSlimvReal (approved):
Good idea, just a bit sad to see that so many similar changes have to be done

sleidig:
Agreed. In the long run we may want to implement a configurable EntitySubrecordPanelComponent similar to the FormComponent to replace all this similar code

Issue #559, opened 2020-12-14
Allow EntityLists without ColumnGroup selection and only fixed columns


sleidig:
Currently, the generic EntityListComponent requires to set a `columnGroup` config in the config-fix.json and always displays a toggle button group to select different column groups. For more simple use cases I want to display the list with just a fixed set of columns and not show any selection to the user.

Suggestion:

make the `columnGroup` object in the config completely optional. If it doesn't exist, just display all defined columns and hide the column group selection UI.

TheSlimvReal:
I would not consider this very high priority. The column selection is not visible if only one column group is defined and the creation of a column group does not cost much time.

PR #635, opened 2021-02-05
ColumnGroups and Filters are optional in config

TheSlimvReal:
see issue: #559

- Column groups are optional in config and default to all columns
- Filters are optional in config and default to no filters


sleidig (approved):
Great! This will make configuration more comfortable!

Issue #579, opened 2020-12-21
Use configurableEnum for config based select fields


TheSlimvReal:
The core allows configurable enum components (see #236, #589, #599). This shoud be used for other `"input": "select"` fields in the config-fix.json:

- [x] Create an enum for the `DocumentStatus` in the config
- [x] Replace the `select` fields which use the `DocumentStatus` options by the new `configSelect` component (has_BplCard, has_rationCard, ...)

<del> - [ ] do the same for `health_vaccinationStatus`
<del> - [ ] do the same for `health_eyeHealthStatus`

PR #638, opened 2021-02-05
Use config enum

TheSlimvReal:
see issue: #579 
I only implemented the first two tasks and crossed of the second two. The latter two did not make sense, because both the selections are only used once.

### Visible/Frontend Changes
--

- Created enum for document status used by child entity

### DIFF
--- a/src/app/core/entity-components/entity-details/EntityDetailsConfig.ts
+++ b/src/app/core/entity-components/entity-details/EntityDetailsConfig.ts
@@ -22,3 +22,37 @@ export class PanelConfig {
   creatingNew?: boolean;
   config?: any;
 }
+
+export class FormConfig {
+  /**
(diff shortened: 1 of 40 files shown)


src/app/core/entity-components/entity-details/EntityDetailsConfig.ts
```diff
@@ -22,3 +22,37 @@ export class PanelConfig {
   creatingNew?: boolean;
   config?: any;
 }
+
+export class FormConfig {
```
sleidig, line 26: :smiley:

sleidig: Where is this used? Can you add a jsDoc for the whole class explaining that?

sleidig (approved):

Issue #546, opened 2020-11-30
Pagination for EntitySubrecordComponent


sleidig:
The embedded tables (e.g. the Notes and Attendance sections in the Child details) don't have pagination yet and become very long. They should also generally use pagination and be limited to maybe 10 entries.

PR #644, opened 2021-02-09
Pagination for entity subrecord component

christophscheuing:
see issue: #546

### Visible/Frontend Changes
Pagination now generally available in the EntitySubrecordComponent.

But the pagination settings are not being stored. This means that every time calling the EntitySubecordComponent, the pagination is set with the standard values defined in entity-subrecord.component.ts

Saving the paginations settings to the user settings would need some reflection as the EntitySubecordComponent is called from different contexts.


src/app/core/entity-components/entity-subrecord/entity-subrecord.component.ts
```diff
@@ -155,6 +164,25 @@ export class EntitySubrecordComponent<T extends Entity>
     }
   }
 
+  ngAfterViewInit() {
+    this.recordsDataSource.sort = this.sort;
```
sleidig, line 168: I had not even realized sort is broken. Thanks for adding and fixing that! :smiley:

src/app/core/entity-components/entity-subrecord/entity-subrecord.component.ts
```diff
@@ -106,6 +111,10 @@ export class EntitySubrecordComponent<T extends Entity>
   private screenWidth = "";
 
   @ViewChild(MatSort) sort: MatSort;
+  @ViewChild(MatPaginator) paginator: MatPaginator;
+
+  paginatorPageSize = 10;
+  paginatorPageIndex = 0;
```
sleidig, line 117: as we don't save the index, it is not necessary to track and update it explicitly here in the component

sleidig (approved):
Great! This looks good.

On testing I realized the UI looks a bit weird if the multiple side-by-side tables of attendance have different lengths, so I added a tweak to sync all subrecord components pageSize

Issue #592, opened 2021-01-08
Add PWA Install button to our interface


sleidig:
*As a user with limited computer experience I want to easily install Aam Digital "as an app" in order to allow me quickly find and open it everytime and make it feel like "a real app".*

**Objective**
As a Progressive Web App (PWA) our application can be install and look almost like a native app rather than like a simple website. However, multiple users have been unable to find the option in the browser to "install" the PWA or add it to their home screen. We should make this much easier to find. Also, every browser displays the option a bit differently, making it hard to guide users.

Ideally a user should be able to install the PWA with a single click.

**Proposed Solutions & Alternatives**
Add a button to the top or left menu / toolbar that always shows the option to install the pwa from within our own interface.

**Additional context**
https://web.dev/customize-install/

sleidig:
some work and discussion had previously happened in this PR: #645 . May be useful to look over when picking this up again.

christophscheuing:
Hint on how to test PWA functionality / service workers on localhost: https://stackoverflow.com/a/59785846

PR #645, opened 2021-02-09
Added Add to Homescreen Button

HiddenTr0ll:
see issue: #592

### Visible/Frontend Changes
- [x] Added an "Add to Homescreen" Button

### DIFF
--- a/src/app/core/navigation/navigation/navigation.component.ts
+++ b/src/app/core/navigation/navigation/navigation.component.ts
@@ -15,7 +15,7 @@
  *     along with ndb-core.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import { Component, OnInit, HostListener } from "@angular/core";
+import { Component, HostListener } from "@angular/core";
 import { MenuItem } from "../menu-item";
 import { AdminGuard } from "../../admin/admin.guard";
 import { NavigationMenuConfig } from "../navigation-menu-config.interface";
@@ -31,7 +31,7 @@ import { ConfigService } from "../../config/config.service";
   templateUrl: "./navigation.component.html",
   styleUrls: ["./navigation.component.scss"],
 })
-export class NavigationComponent implements OnInit {
+export class NavigationComponent {
   /** name of config array in the config json file */
   private readonly CONFIG_ID = "navigationMenu";
   /** all menu items to be displayed */
@@ -86,8 +86,6 @@ export class NavigationComponent implements OnInit {
     e.preventDefault();
     this.deferredPrompt = e;
   }
-
-  ngOnInit() {}
 
   /**
    * Show the installation prompt when button is clicked
--- a/src/app/core/navigation/navigation/navigation.component.ts
+++ b/src/app/core/navigation/navigation/navigation.component.ts
@@ -22,7 +22,8 @@ import { NavigationMenuConfig } from "../navigation-menu-config.interface";
 import { RouterService } from "../../view/dynamic-routing/router.service";
 import { ViewConfig } from "../../view/dynamic-routing/view-config.interface";
 import { ConfigService } from "../../config/config.service";
-
+import { Platform } from "@angular/cdk/platform";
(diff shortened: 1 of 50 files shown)


HiddenTr0ll:
BeforeInstallPromptEvent is only compatible with Chome (Desktop), Edge, Chrome (Mobile) and Samsung Internet (Mobile) (see https://developer.mozilla.org/en-US/docs/Web/API/BeforeInstallPromptEvent#browser_compatibility) 

TODO
[] Add an alternative, visual guide for users using not supported browser

sleidig (commented):
Cool! :smiley: 

The code looks really nice and clean, but maybe you can move it to its own angular component so we can move it around more easily if necessary (and encapsulate code). Don't hesitate to ask if you need some pointers regarding Angular for that. (The cli  `ng generate component` should get you quite far).

However, somehow I don't see the install button on the test deployment :thinking:   Maybe because I have/had the app installed before. Can someone else test and confirm the button shows up?

Do you know what happens if the user dismissed the "install app" prompt on the first visit - will we still get the event on the user's later visits?

src/app/core/navigation/navigation/navigation.component.ts
```diff
@@ -72,4 +76,30 @@ export class NavigationComponent {
...
+  public addToHomeScreen() {
+    this.deferredPrompt.prompt();
+    this.deferredPrompt.userChoice.then((choiceResult) => {
+      if (choiceResult.outcome === "accepted") {
+        console.log("User accepted the A2HS prompt");
+        this.showInstallButton = false;
+      } else {
+        console.log("User dismissed the A2HS prompt");
+      }
+      this.deferredPrompt = null;
```
TheSlimvReal, line 102: If the user denied the prompt the first time (`else` case), can the deferred prompt be called again to show the prompt a second time?

HiddenTr0ll: Yes, the button still functions after the installation has been declined.

HiddenTr0ll:
I am also encountering some issues on the test deployment, but I am noticing a pattern. On the live website [https://demo.aam-digital.com/](url) I see the warning `Site cannot be installed: Page does not work offline. The page will not be regarded as installable after Chrome 93, stable release August 2021.`, which warns about an upcomming change to the offline detection, comming in August 2021. When that warning is shown in the console, no `beforeinstallprompt` is fired and therefore no button is shown. On my local testsetup I get this warning about 50% of the time when I refresh the site. But on the occasions where the warning is missing, everything works. Coupled with this warning is a failed fetch from `ngsw-worker.js `(http://puu.sh/Hhlhw/f72c8e702f.png) - this only fails when the warning is there or vice versa. Do you have any ideas?

sleidig:
> I am also encountering some issues on the test deployment, but I am noticing a pattern. On the live website [https://demo.aam-digital.com/](url) I see the warning `Site cannot be installed: Page does not work offline. The page will not be regarded as installable after Chrome 93, stable release August 2021.`, which warns about an upcomming change to the offline detection, comming in August 2021. When that warning is shown in the console, no `beforeinstallprompt` is fired and therefore no button is shown. On my local testsetup I get this warning about 50% of the time when I refresh the site. But on the occasions where the warning is missing, everything works. Coupled with this warning is a failed fetch from `ngsw-worker.js `(http://puu.sh/Hhlhw/f72c8e702f.png) - this only fails when the warning is there or vice versa. Do you have any ideas?

We should definitely follow up on this. Unfortunately I cannot reproduce that warning or error when I try from my machine. Do you have any idea under which exact circumstances this happens? This is on demo.aam-digital.com with app version (visible bottom left after login) 2.27.1? In Chrome Browser (which version? which OS?)?

Maybe updating the angular service-worker package helps. I was planning to do an Angular upgrade the next days.

sleidig:
closing this as it has been dormant for a very long time. I am pointing the issue back to this PR in case a new developer wants to refer back to the work and discussion here.

PR #649, opened 2021-02-10
Permission Systems

TheSlimvReal:
Added a system for giving permissions for entity interactions (`CREATE`, `UPDATE`, `DELETE`) based on the config.

### Visible/Frontend Changes
- [x] Buttons for creating deleting or editing entities are disabled if user is not permitted
- [x] Tooltip is shown when button is disabled

- [x] Config for entities was extended with permissions
- [x] New service for getting permission for the current user based on an entity and interaction
- [x] New service for creating tooltips next to a given element
- [x] Entity list should also disable `new` button when no elements are present
- [x] `child/new` / `school/new` should route back if user is not allowed to create entities
- [x] Add directive to all buttons (_Only creating notes can be restricted_)
- [x] Add documentation


src/app/core/entity-components/entity-details/form/form.stories.ts
```diff
@@ -0,0 +1,97 @@
...
+import { ChildPhotoService } from "../../../../child-dev-project/children/child-photo-service/child-photo.service";
+import { SessionService } from "../../../session/session-service/session.service";
+import { EntityDetailsModule } from "../entity-details.module";
+import { Child } from "../../../../child-dev-project/children/model/child";
+import { RouterTestingModule } from "@angular/router/testing";
+import { User } from "../../../user/user";
+import { BrowserAnimationsModule } from "@angular/platform-browser/animations";
+import { EntityPermissionsService } from "../../../permissions/entity-permissions.service";
+
+export default {
```
TheSlimvReal, line 15: There is a problem when I run this in storybook: sometimes I get an error that the `EntityDetailsComponent` is part of two modules. Any idea why this happens?

src/app/core/entity-components/entity-details/entity-details.component.ts
```diff
@@ -37,6 +38,8 @@ export class EntityDetailsComponent {
   entity: Entity;
   creatingNew = false;
 
+  operationType = OperationType;
```
sleidig, line 41: always mapping these to a local variable in the components is a pain in the ass ... I am a bit undecided if we shouldn't just use the operation types as plain strings in the templates? (also considering that these might well end up being configurable later)

src/app/core/entity-components/entity-list/entity-list.component.ts
```diff
@@ -140,6 +144,9 @@ export class EntityListComponent<T extends Entity>
       this.displayColumnGroup(this.defaultColumnGroup);
     }
     if (changes.hasOwnProperty("entityList")) {
+      if (this.entityList.length > 0) {
+        this.entityConstructor = this.entityList[0].getConstructor();
```
sleidig, line 148: Doesn't that mean that anybody can create the first entity and only then permissions are enforced?

src/app/core/entity/entity-config.service.ts
```diff
@@ -28,9 +25,15 @@ export class EntityConfigService {
       );
     }
   }
+
+  public getEntityConfig(entityType: typeof Entity): EntityConfig {
```
sleidig, line 29: I think the type you want here would be `EntityConstructor` (defined in entity.ts)?

src/app/core/permissions/entity-operation.directive.ts
```diff
@@ -0,0 +1,43 @@
...
+    if (this.arguments?.operation && this.arguments?.entity) {
+      const permitted = this.entityPermissionService.userIsPermitted(
+        this.arguments.entity,
+        this.arguments.operation
+      );
+      setTimeout(() => {
+        // The timeout is required because the mat-button directive otherwise sets disabled back to false in the
+        // Form component. Other components somehow do not have this problem.
+        if (!permitted) {
+          this.renderer.setAttribute(this.el.nativeElement, "disabled", "true");
```
sleidig, line 38: do we really have to go through the native element? Is there no way to set the disabled input on the component here programmatically?

src/app/core/permissions/entity-permissions.service.spec.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  it("should be created", () => {
+    expect(service).toBeTruthy();
+  });
+
+  it("should give permission if nothing is defined", () => {
+    mockConfigService.getEntityConfig.and.returnValue(null);
+    let permitted = service.userIsPermitted(Entity, OperationType.CREATE);
+    expect(permitted).toBeTrue();
+    mockConfigService.getEntityConfig.and.returnValue({
```
sleidig, line 38: that's actually two tests here ;-)

src/app/core/entity-components/entity-details/form/form.stories.ts
```diff
@@ -0,0 +1,97 @@
...
+        { provide: EntityMapperService, useValue: {} },
+        { provide: AlertService, useValue: {} },
+        { provide: ChildPhotoService, useValue: { canSetImage: () => true } },
+        {
+          provide: SessionService,
+          useValue: { getCurrentUser: () => new User() },
+        },
+        {
+          provide: EntityPermissionsService,
+          useValue: { userIsPermitted: () => false },
```
sleidig, line 35: can we built in a tooltip to tell users that the button is deactivated because they don't have required permissions?

src/app/core/entity/entity-config.service.ts
TheSlimvReal: The `EntityConstructor` type is used when new entities of a type have to be created. `typeof Entity` defines the class of an object and allows to access the static variables and is provided through the `Entity` [class](https://github.com/Aam-Digital/ndb-core/blob/master/src/app/core/entity/entity.ts#L151). In this case only the static attribute `ENTITY_TYPE` is required and therefore `typeof Entity` provides the highest type-safety.

src/app/core/permissions/entity-operation.directive.ts
TheSlimvReal: The use of the `Renderer2` by Angular should prevent this problem. Another approach would be to get rid of the directive and instead just set the `[disabled]` directive depending on the `userIsPermitted` value.

src/app/core/permissions/entity-permissions.service.spec.ts
TheSlimvReal: I put both of these tests into one function because they both fall into the category `nothing is defined`. In the first case nothing is defined for the entity, in the second case, something is defined for the entity, but nothing is defined for this operation. 
Of course they can be split up but I did not want to make the test description to detailed.

src/app/core/entity-components/entity-list/entity-list.component.ts
TheSlimvReal: Yes. The `EntityListComponent` does currently not know which elements are used inside of it. Also, calling `child/new` in the URL still works (security through obscurity).

src/app/core/entity-components/entity-details/entity-details.component.ts
TheSlimvReal: For the moment it gives us type-safetyness and it is also easily traceable where the enums are used. It is easier to change this to just writing strings than coming back from writing strings to this approach.

src/app/core/entity-components/entity-details/form/form.stories.ts
TheSlimvReal: Done

src/app/core/entity-components/entity-list/entity-list.component.ts
TheSlimvReal: This is now solved but requires an additional parameter for the entity list component

src/app/core/entity-components/entity-details/form/form.stories.ts
sleidig: Do you get this after the storybook is reloaded live when changing the code? I have the same problem (I think some bug in storybook itself) but reloading with F5 works

TheSlimvReal: Yes thats exactly the behavior, so then I guess it is alright.

TheSlimvReal:
There is currently a problem that the `mouseenter` event is not thrown on browser other than Firefox. Porbably because the element is disabled.

doc/compodoc_sources/concepts/configuration.md
```diff
@@ -362,6 +364,28 @@ Example:
...
+    "update": ["admin"],
+    "delete": ["admin"]
+  }
+}
+```
+Buttons can be marked as part of an interaction using the `appEntityInteraction` directive:
+```
+<button
+  (click)="switchEdit()"
+  [appEntityOperation]="{entity: entity?.getConstructor(), operation: operationType.UPDATE}"
```
sleidig, line 385: The name `appEntityOperation` doesn't indicate in any way that this relates to disabling something or that it has to do with permissions. Maybe it could be called `appRestrictToEntityOperationPermission` or another way more explicit way?

src/app/core/entity-components/entity-details/entity-details.component.spec.ts
```diff
@@ -151,4 +160,12 @@ describe("EntityDetailsComponent", () => {
     expect(entityMapper.save).toHaveBeenCalledWith(testChild, true);
     expect(router.navigate).toHaveBeenCalled();
   }));
+
+  it("should call router when user is not permitted to create entities", () => {
+    mockEntityPermissionsService.userIsPermitted.and.returnValue(false);
```
sleidig, line 165: I don't get the assumption behind this test - if not allowed (to create? or what permission) then you ensure there is navigation to the new form rather than no navigation or navigation to the list/dashboard?

src/app/core/tooltip/tooltip.service.ts
```diff
@@ -0,0 +1,42 @@
...
+@Injectable({
+  providedIn: "root",
+})
+export class TooltipService {
+  constructor(
+    private overlay: Overlay,
+    private overlayPositionBuilder: OverlayPositionBuilder
+  ) {}
+
+  public createTooltip(elementRef: ElementRef): OverlayRef {
```
sleidig, line 19: I think we shouldn't be building our own tooltip module as there is MatTooltip available, which we are already using in other places. Having different implementations will inevitable lead to confusion and inconsistent styling.

Do you think it is possible to use MatTooltipModule to cover your logic from this? Or maybe we need to implement a permissionProtectedButton component instead of a directive, so we can apply styling and matTooltip on that?

src/app/core/entity-components/entity-details/entity-details.component.spec.ts
TheSlimvReal: As the name of the test says, it is expected that the router is being called if the user opens for example `/child/new` but is not permitted to create new `Child` entities. 

This should prevent users from manually navigating to the `/new` page when the `add` button is disabled on the list view.
For now the test only expects the router to be called, because I did not want to make any further assumptions on which route the router will route to.

doc/compodoc_sources/concepts/configuration.md
TheSlimvReal: I chose this name because I did not want to make it dependent on what the directive actually does. This directive should only _mark_ the buttons which belong to entity operations. 
At a later point this should allow us, for example, to log every time a `CREATE`, `DELETE` or `UPDATE` operation is performed on a entity.

TheSlimvReal:
- [ ] make name of directive more explicit

src/app/core/permissions/permissions.module.ts
```diff
@@ -0,0 +1,13 @@
+import { NgModule } from "@angular/core";
+import { CommonModule } from "@angular/common";
+import { DisableEntityOperationDirective } from "./disable-entity-operation.directive";
+import { DisabledWrapperComponent } from "./disabled-wrapper/disabled-wrapper.component";
+import { MatTooltipModule } from "@angular/material/tooltip";
+
+@NgModule({
+  declarations: [DisableEntityOperationDirective, DisabledWrapperComponent],
+  imports: [CommonModule, MatTooltipModule],
+  exports: [DisableEntityOperationDirective, DisabledWrapperComponent],
```
sleidig, line 10: should `DisabledWrapperComponent` rather stay internal and not be exported?

src/app/core/config/config-fix.json
```diff
@@ -1034,6 +1034,10 @@
   },
 
   "entity:Child": {
+    "permissions": {
+      "create": ["admin"],
+      "delete": ["admin"]
```
sleidig, line 1039:
```suggestion
```
works as expected :+1: 

resetting this to allow everyone by default

src/app/core/config/config-fix.json
```diff
@@ -1054,5 +1058,12 @@
       {"name": "health_lastVitaminD", "schema": { "dataType": "Date" } },
       {"name": "health_lastDeworming", "schema": { "dataType": "Date" } }
     ]
+  },
+  "entity:School": {
+    "permissions": {
+      "create": ["admin"],
+      "update": ["admin"],
+      "delete": ["admin"]
```
sleidig, line 1066:
```suggestion
```

src/app/core/permissions/disable-entity-operation.directive.ts
```diff
@@ -0,0 +1,63 @@
...
+   * The operation property defines to what kind of operation a element belongs, e.g. OperationType.CREATE
+   * The entity property defines for which kind of entity the operation will be performed, e.g. Child
+   */
+  @Input("appDisabledEntityOperation") arguments: {
+    operation: OperationType;
+    entity: typeof Entity;
+  };
+
+  private wrapperComponent: ComponentRef<DisabledWrapperComponent>;
+  private text: string = "Operation disabled for current user";
```
sleidig, line 36:
```suggestion
  private text: string = "Your account does not have the required permission for this action.";
```

sleidig (approved):
looks and works well! :+1:

src/app/core/permissions/permissions.module.ts
TheSlimvReal: True that. I guess it is not required

PR #651, opened 2021-02-11
AttendanceMonth migration

sleidig:


sleidig:
by migrating 3972 AttendanceMonth records for ~150 students with this approach into events of two master activities, ~1000 new event Notes have been created.

src/app/child-dev-project/attendance/activity-attendance-section/activity-attendance-section.component.ts
```diff
@@ -80,17 +81,35 @@ export class ActivityAttendanceSectionComponent
...
-
     this.detailsComponent = {
       component: AttendanceDetailsComponent,
       componentConfig: { forChild: this.forChild },
     };
+
+    this.updateDisplayedRecords(false);
+  }
+
+  updateDisplayedRecords(includeRecordsWithoutParticipation: boolean) {
```
TheSlimvReal, line 96: This function could be tested easily

src/app/core/entity/schema-datatypes/datatype-number.ts
```diff
@@ -33,14 +33,14 @@ export const numberEntitySchemaDatatype: EntitySchemaDatatype = {
   name: "number",
 
   transformToDatabaseFormat: (value) => {
-    if (!value) {
+    if (value === undefined) {
```
TheSlimvReal, line 36: I dont think this check is necessary any more

src/app/core/entity/schema-datatypes/datatype-string.ts
```diff
@@ -33,14 +33,14 @@ export const stringEntitySchemaDatatype: EntitySchemaDatatype = {
   name: "string",
 
   transformToDatabaseFormat: (value) => {
-    if (!value) {
+    if (value === undefined) {
```
TheSlimvReal, line 36: Same as above

src/app/utils/expect-entity-data.ts
```diff
@@ -0,0 +1,80 @@
...
+) {
+  if (onlyExpected) {
+    expect(actualEntities.length).toBe(expectedEntities.length);
+  }
+
+  const cleanExpected = comparableEntityData(expectedEntities, ignoreId);
+  const cleanActual = comparableEntityData(actualEntities, ignoreId);
+
+  for (const data of cleanExpected) {
+    const passed = expect(cleanActual).toContain(data);
```
TheSlimvReal, line 51: I would recommend using [custom jasmine matchers](https://jasmine.github.io/tutorials/custom_matcher) for this. This would make the integration of the custom checks a little easier to understand.

src/app/utils/expect-entity-data.ts
```diff
@@ -0,0 +1,80 @@
...
+    if (!passed) {
+      console.warn("expected object not found:");
+      console.log("expected", data);
+      console.log("actual overall", cleanActual);
```
TheSlimvReal, line 55: Why is there additional logging required?

TheSlimvReal (changes requested):
I am wondering if in this migration also the `Note` entity should be renamed to an `Event` activity. Currently, the code is quite difficult to understand if you don't know that a note should be an activity and I would like to have this clarity in the code.

src/app/utils/expect-entity-data.ts
sleidig: I looked into this but there seems to be no clean and easy way to have good typescript support (probably need to patch the jasmine namespace to pass type checks in the spec files). I'd like to postpone this topic for now although I like the suggestion to follow a more standard approach here.

TheSlimvReal (approved):

PR #653, opened 2021-02-12
Fix datatype undefined

sleidig:
handle "undefined" values in schema transformations more consistently

### DIFF
--- a/src/app/child-dev-project/children/model/child.ts
+++ b/src/app/child-dev-project/children/model/child.ts
@@ -58,7 +58,7 @@ export class Child extends Entity {
    */
   @DatabaseField() photoFile: string;
 
-  @DatabaseField({ dataType: "load-child-photo" })
+  @DatabaseField({ dataType: "load-child-photo", defaultValue: true })
   photo: BehaviorSubject<SafeUrl>;
 
   get age(): number {
--- a/src/app/core/entity/schema/entity-schema.service.ts
+++ b/src/app/core/entity/schema/entity-schema.service.ts
@@ -96,6 +96,16 @@ export class EntitySchemaService {
   public transformDatabaseToEntityFormat(data: any, schema: EntitySchema) {
     for (const key of schema.keys()) {
       const schemaField: EntitySchemaField = schema.get(key);
+
+      if (data[key] === undefined) {
+        if (schemaField.defaultValue !== undefined) {
+          data[key] = schemaField.defaultValue;
+        } else {
+          // skip and keep undefined
+          continue;
+        }
+      }
+
       const newValue = this.getDatatypeOrDefault(
         schemaField.dataType
       ).transformToObjectFormat(data[key], schemaField, this, data);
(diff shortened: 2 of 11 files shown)


src/app/child-dev-project/children/model/child.ts
```diff
@@ -58,7 +58,7 @@ export class Child extends Entity {
    */
   @DatabaseField() photoFile: string;
 
-  @DatabaseField({ dataType: "load-child-photo" })
+  @DatabaseField({ dataType: "load-child-photo", defaultValue: true })
   photo: BehaviorSubject<SafeUrl>;
```
TheSlimvReal, line 62: Why is the default value for this field a boolean?

src/app/core/entity/schema/entity-schema.service.ts
```diff
@@ -140,12 +150,24 @@ export class EntitySchemaService {
...
+      let value = entity[key];
       const schemaField: EntitySchemaField = schema.get(key);
 
-      if (entity[key] !== undefined) {
-        data[key] = this.getDatatypeOrDefault(
-          schemaField.dataType
-        ).transformToDatabaseFormat(entity[key], schemaField, this, entity);
+      if (value === undefined) {
+        if (schemaField.defaultValue !== undefined) {
+          value = schemaField.defaultValue;
```
TheSlimvReal, line 158: Is this step necessary? When a entity is loaded the value is set and when a entity is saved the value is set.

src/app/core/entity/schema/entity-schema.service.ts
```diff
@@ -140,12 +150,24 @@ export class EntitySchemaService {
...
+        } else {
+          // skip and keep undefined
+          continue;
```
TheSlimvReal, line 161: And shouldn't then the undefined checks be removed from the datatype definitions for clarity?

TheSlimvReal (commented):
The change looks good but the changes to the `load-child-photo` datatype seem a little hacky to me.

src/app/child-dev-project/children/model/child.ts
sleidig: the important point for the `load-child-photo` is that a default is set at all - it actually ignores the value of it but this ensures the transformation is trigger even if no value was stored. Do you have a suggestion for a more clear value?

src/app/core/entity/schema/entity-schema.service.ts
sleidig: agreed, will remove them in the individual data-type implementations

src/app/core/entity/schema/entity-schema.service.ts
sleidig: not sure whether I understood your point. This ensures that the defaultValue is passed through the transformation rather than being directly set, to allow for some advanced use cases.

TheSlimvReal: My point is that in `transformEntityToDatabaseFormat` the default value is assigned to the attribute if it is undefined. This means no entity can be saved to the database withouth having the default value set. However, `transformDatabaseToEntityFormat` checks again if the value is not set and if a default value is available. Is there a reason why the default value is assigned in both transformations?

sleidig: this is somewhat redundant but a failsafe in case data somehow ends up in the database through an import from other channels - or the data model changes

TheSlimvReal (approved):

Issue #637, opened 2021-02-05
filter ChildrenList by School/Team


sleidig:
For some organizations an important distinction / grouping for children is by which School they attendance (i.e. their current `ChildSchoolRelation`s). As a user I need to filter the overall children list to view all students at a selected school.

Schools or teams are technically equivalent here as we use the same data model.

- [x] implement a new filter type for the ChildrenListComponent or the generic EntityListComponent that shows a dropdown of schools and filters by the selected one
- [x] implement a config interface to define such filters (?)

PR #671, opened 2021-02-23
School filter

TheSlimvReal:
see issue: #637

Currently the filter only works with the `schoolId` field. Due to the current status, I dont think a more general/flexible approach is required.

### Visible/Frontend Changes
- Added a filter for schools in the children list section

- Added a prebuilt filter for the `ChildrenListComponent`


src/app/child-dev-project/children/children-list/children-list.component.spec.ts
```diff
@@ -87,16 +101,31 @@ describe("ChildrenListComponent", () => {
         Angulartics2Module.forRoot(),
       ],
       providers: [
+        {
```
sleidig, line 104: when mocking childrenservice and entitymapperservice we do not need any provider for the database itself

sleidig (approved):

PR #677, opened 2021-02-24
fix(dashboard): faster calculation of note dashboard metrics

sleidig:


src/app/child-dev-project/children/children.service.ts
```diff
@@ -322,19 +323,40 @@ export class ChildrenService {
...
     const stats = await this.dbIndexing.queryIndexStats(
-      "notes_index/note_date_in_days_for_child"
+      "notes_index/note_child_by_date",
+      {
+        startkey: [startDay.year(), startDay.month(), startDay.date()],
+      }
     );
 
     const results = new Map();
+    const children = await this.getChildren().pipe(take(1)).toPromise();
```
TheSlimvReal, line 344: A little thought on this construction: https://stackoverflow.com/a/65506511/10713841
Is this the intended behavior or would `toPromise` be sufficient?.

TheSlimvReal (approved):
Looks very solid, just a small comment and the tests should be fixed

src/app/child-dev-project/children/children.service.ts
sleidig: I am intentionally using `take(1)` here, hoping that this will skip the additional computation ChildrenService does to add more information. We only need the list of child ids here.

I would agree that this is rather confusing, however. Probably `ChildrenService` shouldn't emit multiple times

Issue #532, opened 2020-11-17
Upgrade to Angular 11


sleidig:
A new version of Angular has been released: https://blog.angular.io/version-11-of-angular-now-available-74721b7952f7

That sets us two versions behind. We should upgrade to Angular 11 and also check our other dependencies to be upgraded.

Interesting in Angular 11 for us:
- switch from tslint to eslint
- Hot Module Reload (which would probably help when testing locally because you don't have to constantly log in again after if the page is not fully reloading)

PR #678, opened 2021-02-26
Upgrade to Angular 11

TheSlimvReal:
see issue: #532
Angular and all other `npm` packages have been upgraded to the latest compatible version

### Visible/Frontend Changes
--

- See my comments below for my manual changes


src/app/core/webdav/cloud-file-service.service.spec.ts
```diff
@@ -151,14 +149,4 @@ describe("CloudFileService", () => {
       "image"
     );
   });
-
```
TheSlimvReal, line 154: This test got removed because the assertion did not get triggered before and the function `doesFileExist` is not even called by the function. I think this test got committed by accident.

src/app/core/session/session-service/synced-session.service.ts
```diff
@@ -89,48 +89,47 @@ export class SyncedSessionService extends SessionService {
...
-        );
+          });
 
         // asynchronously check if the local login failed --> this happens, when the password was changed at the remote
         localLogin.then(async (loginState: LoginState) => {
           if (loginState === LoginState.LOGIN_FAILED) {
             // in this case: when the sync is completed, retry the local login after the sync
-            await syncPromise;
+            try {
+              await syncPromise;
```
TheSlimvReal, line 109: At least in some test scenarios this promised rejected what resulted in the later login process not being executed. The login process is now executed no matter what the promise returns. I am not 100% sure if this is the expected behavior.

src/app/core/session/session-service/synced-session.service.spec.ts
```diff
@@ -316,56 +317,39 @@ describe("SyncedSessionService", () => {
...
-      const localLogin = spyOn(localSession, "login").and.returnValue(
-        Promise.resolve(LoginState.LOGGED_IN)
-      );
-      const remoteLogin = spyOn(remoteSession, "login").and.returnValue(
-        Promise.resolve(ConnectionState.CONNECTED)
-      );
-      const syncSpy = spyOn(sessionService, "sync").and.returnValue(
-        Promise.reject()
-      );
+    it("behaves correctly when the sync fails and the local login succeeds", fakeAsync(() => {
```
TheSlimvReal, line 320: I changed the syntax of this and the following test because they initially failed due to some unhandled rejected promises.

angular.json
```diff
@@ -29,6 +29,18 @@
             ],
             "scripts": [
               "node_modules/marked/lib/marked.js"
+            ],
+            "allowedCommonJsDependencies": [
```
TheSlimvReal, line 33: This is the list of CommonJS libraries that we or some of our libraries use. They prevent the tree-shakeability of the compilation process and we should therefore consider replacing them (see [How CommonJS is making your bundles larger](https://web.dev/commonjs-larger-bundles/))

TheSlimvReal (commented):
These are the things I had to adjust manually

PR #682, opened 2021-03-01
feat(core): coming-soon component as placeholder and feedback page

sleidig:
introduce a generic "Coming Soon" page that can be used as a placeholder for features that are not available yet and automatically integrates with analytics to track users' interest.


TheSlimvReal:
When I open the Heroku deployment, I get a lot of erros `TypeError: t is null` and `TypeError: null has no properties`.

TheSlimvReal (changes requested):
Looks very good!
Only problem is the type error which I get on the Heroku deployment mentioned in the previous comment.

sleidig:
fixed, thanks for pointing this out.

Issue #618, opened 2021-01-26
Make Child.center a configurable enum


sleidig:
The "center" field should be a dropdown form element rather than free text.

- [x] change the `center` property of the Child entity to type configurable-enum
- [x] adapt the child details form config for the field
- [x] adapt the child list config for the column
- [x] adapt the child list filter toggle for centers
- [x] ~adapt the `ChildrenCountDashboardComponent` to support grouping by configurable-enum values~ (done in earlier PR)

Schottkyc137:
When initially opening the child-details component, the initial value of the center property is not set. Is this a bug or is there more to do from my side?

sleidig:
This sounds like a bug. For other enum fields it seems to load correctly for me. Is this for all existing children's details?

Schottkyc137:
I fixed it by adding the center to the child-entity attributes, so fortunately no bug

PR #685, opened 2021-03-02
Child center enum

Schottkyc137:
see issue: #618

### Visible/Frontend Changes
None

- [x] Child now is a configurable-enum, compared to a simple string
- [x] Updated config to reflect these changes 
- [x] Updated specs to reflect these changes
- [x] Updated Filter-selections in the select-group-children component to reflect these changes


sleidig (approved):
Looks and works well :+1:

Issue #681, opened 2021-03-01
User dropdown select component


sleidig:
Users should be able to select a `User` entity from a dropdown similar to the ChildSelect. This is required for assigning a user to a RecurringActivitiy or linking users to a Note.

also see #136 and #577

sleidig:
completed with PR #728

Issue #136, opened 2018-06-28
Make child-select component into generic entity-select component


sleidig:
Rather than only allowing a select list for Child entities, this should become generic to handle any kind of Entity types. Especially required for selection list of team members (`User`s) in NoteDetails.

sleidig:
For this a more general problem has to be solved:
How can we dynamically select `app-child-block`, `app-school-block` or some other component to be used in a template based on the specific object's type?

-> This may be an interesting lead: https://medium.com/@DenysVuika/dynamic-content-in-angular-2-3c85023d9c36

TheSlimvReal:
The `SchoolSelectComponent` was now added which also shares much of the code of the `ChildSelectComponent` (and #691). These components could be combined for a generic `EntitySelectComponent` which uses a two-way binding to either share entities or just their IDs. Also it should have a `[multi]` flag which determines whether many or just one element can be selected. If the array of the two-way binding contains some initial values, these entities or IDs should already be selected and can not be selected again.

sleidig:
see PR #728

PR #691, opened 2021-03-03
User dropdown

Schottkyc137:
see issue: #681 

### Visible/Frontend Changes
- [x] When selecting a user, a list of possibilities with all users will show up
- [x] Multiple authors can be chosen when looking at a note 

- [x] The note-details component uses the new field compared to the old one
- [x] Notes has an array of authors (strings)


src/app/core/user/user-select/user-select.component.ts
```diff
@@ -0,0 +1,61 @@
...
+import { AlertService } from "../../alerts/alert.service";
+import { LoggingService } from "../../logging/logging.service";
+
+@Component({
+  selector: "app-user-select",
+  templateUrl: "./user-select.component.html",
+  styleUrls: ["./user-select.component.scss"],
+})
+export class UserSelectComponent implements OnInit {
+  @Input() user: string;
```
sleidig, line 21: To make this more generally powerful we should probably use `User` entity objects (or maybe id) in the input and output. Then the code using this component can decide which fields are needed at a higher level.

We'll probably need to do a simple migration of old data, mapping names to ids and save user ids with the note, but that should not be a big issue.

src/app/core/user/user-select/user-select.component.ts
```diff
@@ -0,0 +1,61 @@
...
+  @Output() userChange = new EventEmitter<string>();
```
sleidig, line 22: That was not clearly specified in the issue, but we have cases where more than one user is selected so we will need to support multi-select somehow

src/app/core/user/user-select/user-select.component.ts
Schottkyc137: So to just to clarify, a Note should therefore not only have one author (which currently is a string) but multiple authors (which should be an array of users, resp. an array of user IDs)?

src/app/core/user/user-select/user-select.component.ts
sleidig: The Note "author" is actually misnamed currently, it is rather used as a description of staff members (= users) who have been part of the real-world interaction the Note describes. Therefore more than one user can be linked.

src/app/core/user/user-select/user-select.component.ts
sleidig: Yes, that should be the it.

And maybe we can abstract this (now or later) into a generic base component also, so the UX and implementation is shared between child and user select field (#136)

Schottkyc137:
Closed since this is in PR #728

Issue #632, opened 2021-02-03
clarify UX of "age" field


sleidig:
Some users are confused about the age being shown as "-1" in the Child Details. We could improve the usability here with some details:

- [x] keep age empty if no date of birth exists to calculate the age
- [x] automatically update the age as soon as a date of birth is selected (not only after saving the form) in order to show users how these are interrelated
- [x] add a tooltip to the "age" form field to explain that this is always read-only and the user should change the date of birth instead (and possibly use 1. Jan of a suitable year if they don't know the exact date)
- [x] fix the date select next to the age (and give it a label/placeholder) and remove the duplicate date of birth select field below it.

PR #697, opened 2021-03-04
Clarify ux of age field

TheSlimvReal:
see issue: #632 

The mentioned aspects in issue #632 have been taken care of.

The function calculateAge in form.component.html is very similar to get age() in child.ts . Maybe this could be generalized.


sleidig (approved):
:+1:

Issue #680, opened 2021-03-01
make label/placeholder of list filters configurable


sleidig:
Currently the label for dropdown style filters are using the id to display to the user. This should be independently configurable.

![image](https://user-images.githubusercontent.com/1682541/109502765-b4647700-7a99-11eb-9211-233678805a17.png)
(label, in screenshot "category" in lower case spelling)

PR #702, opened 2021-03-06
Configurable list filter label

Schottkyc137:
see issue:  #680

### Visible/Frontend Changes
- [x] Labels for dropdown-style filters have their user-friendly label instead of their id


sleidig (approved):
Awesome! This makes the UI feel much cleaner, even though it's just a detail.

PR #703, opened 2021-03-07
Angular 11

TheSlimvReal:
see issue: #532 

Generally updating all packages to their latest supported version


sleidig (approved):
As far as I can tell everything works well :+1: 

Thanks!

Issue #705, opened 2021-03-09
Tooltips in form.component.html


christophscheuing:
*Allowing configurable tooltips for different entry fields in form.component.html might be handy.*
The "placeholder"-fields from config-fix.json/ts are only shown with a limited number of characters. In some cases it might be useful to allow the configurator to display a longer explanation for a form field within a tooltip.

**Objective: Be more clear for the users entering data.**

**Proposed Solutions & Alternatives**
The tooltip-feature could be added for every field type in form.component.html, but maybe there is also a more generic way?

PR #706, opened 2021-03-09
Help info as tooltips for forms

christophscheuing:
see issue: #705

### Visible/Frontend Changes
yes

no


christophscheuing:
Does allowing such tooltips make sense to you?
I'm note sure how to set the [matTooltipDisabled]="!editing" via the config file.
Maybe we should also add a possibility for a delay for the tooltip to appear?

TheSlimvReal:
I guess providing the option for definition tooltips through the config makes sense. However it requires quite some text in the config if a tooltip for each field is provided and the delay also makes sense. 
Using `[matTooltipDisabled]="!editing"` should already be working, this should require anything special in the config file.

sleidig:
I also think tooltips can be an important addition and should be set in the config.

An alternative approach for the UI could be to show a (?) icon next to the component and only add the tooltip to that. That would make the help text more explicitly noticeable (especially on mobile) but might also clutter the interface a lot. Not sure what works best for us or what best practices are regarding this.

src/app/core/entity-components/entity-details/form/form.component.html
```diff
@@ -112,7 +112,10 @@
           </mat-form-field>
         </div>
 
-        <div *ngIf="row.input === 'select'">
+        <div *ngIf="row.input === 'select'"
+          [matTooltip]="row.tooltip"
+          [matTooltipPosition]="row.tooltipPosition"
```
sleidig, line 117: couldn't you add this to the parent div (row 9) and automatically make it available for all types?

christophscheuing: Yes, that would work; thank you for that input.

christophscheuing:
I now moved the tooltips to a question mark icon shown at the end of the field when editing. I don't know how you like this UI.

christophscheuing:
It also works (more or less) on mobile. On my iPad a simple tap on the question mark icon shows the tooltip. On the iPhone I have to do a long press on the icon.

What might be a little inconvenient is the fact that he tooltip is sometimes hidden behind to entries of the drop down menu on mobile. I don't know if there's a good solution to that.

What do you think?

src/app/core/entity-components/entity-details/form/form.component.scss
```diff
@@ -46,3 +46,8 @@
   padding-left: 10px;
   color: black;
 }
+
+.mat-icon{
+  font-size: 1em;
```
sleidig, line 51:
```suggestion
  font-size: 1.25em;
  color: grey;
```
let's make them a bit bigger to make it an easier target especially on mobile

sleidig:
Looks great to me!
I like the style with the question mark icon, this clearly shows the user that there is extra help available if needed.
Mostly works well on mobile as well. I don't think it's much of a problem that it is covered by the opened dropdown. Only sometimes I cannot reliably touch the icon in active fields, focusing the form field rather than displaying the tooltip.

Didn't you want to move up the help icon rather then duplicating it for every type?

sleidig (approved):

src/app/core/config/config-fix.ts
```diff
@@ -719,11 +719,13 @@ export const defaultConfig = {
                   [
                     {
                       "input": "age",
+                      "tooltip": "This field is read-only. Edit Date of Birth to change age. Select Jan 1st if you only know the year of birth.",
                       "id": "dateOfBirth",
                       "placeholder": "Date of Birth"
                     },
                     {
                       "input": "select",
+                      "tooltip": "This is a sample toopltip for a select field. Please select the biological sex of the person.",
```
sleidig, line 728:
```suggestion
```
cleanup before merge

src/app/core/config/config-fix.ts
```diff
@@ -733,6 +735,7 @@ export const defaultConfig = {
                     },
                     {
                       "input": "text",
+                      "tooltip": "This is a sample toopltip for a textfield. Please enter the native language.",
```
sleidig, line 738:
```suggestion
```
cleanup before merge

src/app/core/config/config-fix.ts
```diff
@@ -750,6 +753,7 @@ export const defaultConfig = {
                     },
                     {
                       "input": "configurable-enum-select",
+                      "tooltip": "This is a sample tooltip for a configurable-enum-select-field. Please choose Adahar status.",
```
sleidig, line 756:
```suggestion
```
cleanup before merge

Issue #679, opened 2021-03-01
Make matomo analytics configurable


sleidig:
Whether user behaviour will be tracked and to which matomo URL should be configurable at runtime in the config.json (or maybe even in the config in the database?)

Issue #595, opened 2021-01-10
Rework startup process


TheSlimvReal:
Currently, the startup process consists of many setup processes including the initialization of the app config, the setup of the database, the creation of routes and the navigation links. Working towards having a single application which serves any customer and only loads data from a different database, this initial setup needs to be adjusted.

When the website is opened for the first time on a device, only the `_user` DB should be available to check whether the user has valid login credentials. The `AppConfig` needs to be available because it holds the link to the database. 

When the user successfully logs in for the first time, the correct database can be fetched. This allows to retrieve the configuration of the application together with all the other entities. At this point the app should wait until the configuration was loaded from the database. After this is done, the router and the navigation can be set up.

For further logins from the same device, the login process can also work against the local database and all the steps that need the configuration from the database can already run before the user is logged in.

Steps to do:

- [ ] Evaluate which services need to be available for the initial login process
- [x] Make sure the session only uses the `_user` DB for the initial login
- [ ] Provide a hook which executes asynchronous tasks that have to be run after the user successfully logged in for the first time but before the application can be used (load config, init routing, init navigationbar, ...)
- [ ] Add the config and all dependent services to the previously implemented hook
- [ ] If the local DB is already present, the previously described tasks can already run before the user is logged in, **but** they still need to be finished before the application is ready to use

----
Cleanup from PR #594 
- [x] remove entityMapper parameter from ConfigService.save/load
- [x] remove EntityModule dependency from AdminModul and AdminComponent

TheSlimvReal:
Some things to consider:
-  `this.router.navigate([], { relativeTo: this.activatedRoute });` needs to be called **once** in the startup process after routes are initialized from the config
- The config currently always emits the `defaulConfig` on startup (config is a `BehaviorSubject`) this should be changed to the `defaultConfig` being provided through a demo data generator service to not interfere with the production mode 
- Maybe adjust the config service so it only emits new values when the new config actually differs to prevent unnecessary reloads/initializations
- create a clear concept of runtime dependencies (e.g. login -> load config -> init routes -> apply new routes) to cleanup app component and make the dependencies more visible.

PR #709, opened 2021-03-11
Added a matomo service

TheSlimvReal:
see issue: #679 

### Visible/Frontend Changes
--

- Added config for matomo url
- Setting up matomo only if config is set


src/app/core/analytics/matomo.service.ts
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * This class dynamically adds sets up everything for Matomo.
+ *
+ * The code is inspired by:
+ * https://github.com/Arnaud73/ngx-matomo/blob/master/projects/ngx-matomo/src/lib/matomo-injector.service.ts
+ */
+@Injectable({
+  providedIn: "root",
+})
+export class MatomoService {
```
sleidig, line 12: I think this `MatomoService` should be part of the existing `AnalyticsService`, whose purpose is to manage everything related to usage analytics (in our case with Matomo as the backend)

sleidig (approved):
works as intended, thanks! :+1: 

The warning "Piwik not found" seems almost unavoidable because we can only set up the matomo script when AppConfig is loaded - which is after Module imports. Maybe we can optimize this after working on #595 . Tracking is working, however.

src/app/core/analytics/analytics.service.ts
```diff
@@ -7,74 +7,128 @@ import { LoginState } from "../session/session-states/login-state.enum";
 
 const md5 = require("md5");
 
+/**
+ * Track usage analytics data and report it to a backend server like Matomo.
+ *
+ * This is automatically disable if the config.json does not specify "usage_analytics" settings.
```
TheSlimvReal, line 13:
```suggestion
 * This is automatically disabled if the config.json does not specify "usage_analytics" settings.
```

TheSlimvReal (commented):
Looks very good now, with a proper startup process the error should also disapear

PR #714, opened 2021-03-15
Fix importing backup files

TheSlimvReal:
When importing a backup file where documents have the `_rev` field, CouchDB thinks the document is updated rather than created and throws document conflicts. Because the document should be overwritten anyways, the `_rev` field is now removed before saving the document to the database.

### Visible/Frontend Changes
--

- Removing the `_rev` field of documents before saving it to database


sleidig (approved):
functionally tested after refactoring

Issue #354, opened 2019-10-23
Overview of BMI (height&weight) as a children list


sleidig:
As a social worker I need to follow up with all children who are malnourished, identified by a low/high BMI. For this I need to list all affected children to see them at a glance.
At the moment each child's height&weight is visible form its individual details page but there is no overall list from which to get an overview.

Lorsbyjm:
Hi @sleidig, @TheSlimvReal just took me a bit to understand the new design completely defined in the config files. 
The BMI is computed based on height and weight and no database field. Therfore following need to be done:
- [ ] ceate a new BMI table component
   - weight and height are not stored in the child entity. Call the newest health check entry. 
- [ ] add the BMI in the configuration file

sleidig:
welcome back @Lorsbyjm :smiley: 
The documentation around the config system is a work in progress, sorry about that.

This sounds like a good overview of the tasks. I guess the key parts are loading the latest health check entries and designing to table view. Developing the view component might be easier in isolation using storybook, I'm happy to help anyone getting started with that if necessary.

Lorsbyjm:
Hey,@sleidig . I added a component and changed the config file, but I cannot load Health information from the database. 

 this.childrenService
        .getHealthChecksOfChild(this.child.getId())
        .pipe(untilDestroyed(this))
        .subscribe((results) => {this.records = results
       //   .sort(
       //   (a, b) =>
       //     (b.date ? b.date.valueOf() : 0) - (a.date ? a.date.valueOf() : 0)
       // );
      });
      console.log(this.records)

always returns an empty array. 
If selecting a child the panel there is full.

sleidig:
is there an effect when you omit `.pipe(untilDestroyed(this))`?

Please push a code sample in a draft pull request, then I can have a look at the exact context

Lorsbyjm:
Already solved the problem with @TheSlimvReal ;)

TheSlimvReal:
How about creating a dashboard widget for this feature?

PR #716, opened 2021-03-15
Bmi widget for dashboard

Lorsbyjm:
see issue: #354

### Visible/Frontend Changes
- [x] widget on dashboard page displaying all children with an urgent BMI level


sleidig:
I like the approach. For usability a separate "full screen" view showing a table of all BMIs and maybe allowing to sort would be a great complement to this. Especially as the BMI is probably only updated a few times a year in bulk and users may rather do a detailed analysis at some point in time than pick up the most critical new case every day.

Looks like you have a few errors from `ng lint` left.

Lorsbyjm:
Thanks for the appreciation. Theroratically there is a full screen view inside the children list, but there the column is not sortable.

Lorsbyjm:
I added a test fot the component. Therefore I mocked the ChildService and the getChildren function. 
Now the tests fail beacuse the loading of the children called inside the component does not work. 
Error message is: cannot read property pipe of undefined.

src/app/child-dev-project/children/children-bmi-dashboard/children-bmi-dashboard.component.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+    fixture.detectChanges();
+  });
+
+  it("should create", () => {
+    expect(component).toBeTruthy();
+  });
+
+  it("should load the BMI data for the childs, but only display the unhealthy one", (done) => {
+    const testChild = new Child("testID");
+    const HealthCheck1 = new HealthCheck("hc1");
```
TheSlimvReal, line 39: It would make the test a little easier to understand if you call the variables according to what they represent, e.g. `childWithLowBmi`, `oldGoodHealthCheck`, `newBadHealthCheck`,... Because then the assertions in the end are easier to understand.

TheSlimvReal (approved):
Looks very good, just a small suggestion for the future.

TheSlimvReal:
@Lorsbyjm The tests failed, because on default the mocked `ChildrenService` returned `null` and therefore the call in the `ngOnInit` method failed because it called `.pipe` on a `null` object.

Lorsbyjm:
@TheSlimvReal this sound logic for the first test but in the second test the function is defined before it is called.

TheSlimvReal:
@Lorsbyjm I think the `detectChanges` method in the `beforeEach` block initially calls the `ngOnInit` function.

PR #718, opened 2021-03-17
Small fixes

TheSlimvReal:
Some small issues that came up with a recent setup process

### Visible/Frontend Changes
- `PreviousSchools` component now sets the `current school` correctly

- Added safeguard in case a activity does not have a category
- `PreviousSchools` is now configurable
- Changed `config.default.json` to use `mock` DB on default. Local DB works bad on Heroku


doc/compodoc_sources/concepts/configuration.md
```diff
@@ -343,6 +343,32 @@ The inner object can have the following but not always required fields: `"input"
   }
 ```
 
+#### The PreviousSchools Component
```
sleidig, line 346: We should find a different place to describe all the custom sections soon, otherwise this page will become unusably long

src/app/child-dev-project/previous-schools/previous-schools.component.spec.ts
```diff
@@ -76,4 +83,40 @@ describe("PreviousSchoolsComponent", () => {
...
     });
   });
+
+  it("should only show columns which are defined by the config", fakeAsync(() => {
+    const config: PanelConfig = {
+      entity: new Child(),
+      config: {
+        single: true,
+        columns: {
+          schoolId: "Team",
```
sleidig, line 93: It's not intuitively clear that these are assigning a label to the entity property key. An array of column objects (`[{ id: "schoolId", label: "Team" }]`) may be more explicit and extendable - although a bit more verbose as well.

In general there is a question here, as Andy also raised it, whether generalization (and its hard to understand ways of pushing around config) trumps some duplication with explicit, obvious implementations for teams and schools

src/app/child-dev-project/previous-schools/previous-schools.component.html
```diff
@@ -1,10 +1,10 @@
 <div
-  *ngIf="child.schoolId"
+  *ngIf="config.single && current"
   matTooltip="To edit the current school or class, please add to the school history below"
```
sleidig, line 3: The tooltip still makes this only usable for schools

src/app/child-dev-project/previous-schools/previous-schools.component.ts
```diff
@@ -44,6 +56,18 @@ export class PreviousSchoolsComponent
...
+
+  schoolMap: Map<string, School>;
+
+  public config: PreviousRelationsConfig = {
+    single: false,
+    columns: {
+      schoolId: "School",
+      start: "From",
+      end: "End",
+    },
```
sleidig, line 69: Shouldn't the default value here match the previous behaviour for the PreviousSchools? If users have a custom config but do not update it, backward compatibility can be preserved then

```suggestion
    single: true,
    columns: {
      schoolId: "School",
      schoolClass: "Class",
      start: "From",
      end: "End",
      "result": "Result",
    },
```

src/app/child-dev-project/previous-schools/previous-schools.component.spec.ts
TheSlimvReal: I this case I chose a little different approach for the config, because the options are quite limited and only the mentioned five are available. Theoretically, we could also make the `ChildSchoolRelation` a configurable entity and allow to display any number of columns in this view.

Regarding your second point: As long as we are using the exact same classes to model teams and schools, as well as the relations between the children and their teams/schools, I do not see a reason why we should use different components to display these relations. The visualization differences are only additional fields with no further special meaning. The only real difference, which we now also had in the case were classes were modeled through `School` objects, is whether a child can have many active relations or not. Therefore, I would not expect to have different components for teams and schools but rather single and multiple active relations which in this case is configured using the `single` flag.

src/app/child-dev-project/previous-schools/previous-schools.component.spec.ts
```diff
@@ -76,4 +85,69 @@ describe("PreviousSchoolsComponent", () => {
...
+    expect(columnNames).toContain("Class");
+    expect(columnNames).toContain("Result");
+  }));
+
+  it("should display errors for invalid fields", () => {
+    const entry = new ChildSchoolRelation();
+    let validation = component.formValidation(entry);
+    expect(validation.hasPassedValidation).toBeFalse();
+
+    entry.schoolId = "test school";
```
sleidig, line 138: if each of these would be a separate test with a title, it would be slightly easier to grasp to intended scenario that is tested

src/app/child-dev-project/previous-schools/previous-schools.component.ts
```diff
@@ -150,14 +172,16 @@ export class PreviousSchoolsComponent
     };
     if (!record.schoolId) {
       validationResult.validationMessage =
-        '"Name" is empty. Please select a school.';
+        '"Name" is empty. Please select a ' + this.schoolNaming;
```
sleidig, line 175:
```suggestion
        "Please select a " + this.schoolNaming;
```

src/app/child-dev-project/previous-schools/previous-schools.component.ts
```diff
@@ -77,52 +99,52 @@ export class PreviousSchoolsComponent
...
+        column.inputType = ColumnDescriptionInputType.SELECT;
+        column.selectValues = new Array(...this.schoolMap.values())
+          .sort((s1, s2) => s1.name.localeCompare(s2.name))
+          .map((t) => {
+            return { value: t.getId(), label: t.name };
+          });
+        column.valueFunction = (entity: ChildSchoolRelation) =>
+          this.schoolMap.get(entity["schoolId"]).name;
+        break;
+      case "percentageResult":
```
sleidig, line 138: we are adding layers and layers of configuration (special types here, transformed to more general column types ...) :sweat_smile: 

We need to revisit the configuration and table column systems some time and see whether we can generalize, merge and simplify some of this

sleidig (approved):
Looks good! :+1: 

(Although I am getting a bit scared in general about the myriad of configuration properties ...)

PR #719, opened 2021-03-17
Data reports

TheSlimvReal:
### Visible/Frontend Changes
- [x] Add a component for admins to download reports for a specified time

- [x] Create new reporting module and aggregation service
- [x] Count children based on their attributes
- [x] Count events in a timespan based on attributes of linked activities and schools
- [x] Count participants of events in a timespan based on attributes of linked activities and schools
- [x] Create config structure for report queries
- [x] Combine query results in a excel
- [x] Add `groupBy` option which disaggregates objects by the passed attribute


src/app/child-dev-project/reporting/aggregation.service.spec.ts
```diff
@@ -0,0 +1,253 @@
...
+import { Note } from "../notes/model/note";
+import moment from "moment";
+
+describe("AggregationService", () => {
+  let service: AggregationService;
+  const mockEntityMapper: jasmine.SpyObj<EntityMapperService> = jasmine.createSpyObj(
+    ["loadType"]
+  );
+
+  const loadTypeFake = (
```
sleidig, line 19: If you have complex test data, using MockDatabase may be easier than directly mocking EntityMapperService.

`MockDatabase.createWithData(arrayOfEntities)` let's you init the in-memory database with any entities and let EntityMapper handle the filtering based on types, etc.

src/app/child-dev-project/reporting/aggregation.service.spec.ts
```diff
@@ -0,0 +1,253 @@
...
+      maleChristianChild,
+      femaleChristianChild,
+      femaleMuslimChild,
+      maleChild,
+    ]);
+    service.loadData();
+
+    const maleChristianQuery = {
+      gender: "M",
+      religion: "christian",
```
sleidig, line 75: can we query for data where the property is not set, also?

src/app/child-dev-project/reporting/aggregation.service.ts
```diff
@@ -0,0 +1,130 @@
...
+  private entities = new Map<typeof Entity, Entity[]>();
+  private dataLoaded: Promise<any>;
+  constructor(private entityMapper: EntityMapperService) {}
+
+  public loadData(): void {
+    this.dataLoaded = Promise.all(
+      AggregationService.ENTITY_CLASSES.map((entityClass) =>
+        this.entityMapper
+          .loadType(entityClass)
+          .then((entities) => this.entities.set(entityClass, entities))
```
sleidig, line 35: `(entities) => this.entities` is confusing with the name duplication

src/app/child-dev-project/reporting/aggregation.service.ts
```diff
@@ -0,0 +1,130 @@
...
+    EventNote,
+  ];
+
+  private entities = new Map<typeof Entity, Entity[]>();
+  private dataLoaded: Promise<any>;
+  constructor(private entityMapper: EntityMapperService) {}
+
+  public loadData(): void {
+    this.dataLoaded = Promise.all(
+      AggregationService.ENTITY_CLASSES.map((entityClass) =>
```
sleidig, line 32: Do we have safeguards if the data should ever be larger than available memory? Are you testing this with (pouchdb) demo data of >1000 students, >5 years of activity data?

src/app/child-dev-project/reporting/reporting/report-row/report-row.component.ts
```diff
@@ -0,0 +1,25 @@
...
+import { ReportRow } from "../../reporting.service";
+
+@Component({
+  selector: "app-report-row",
+  templateUrl: "./report-row.component.html",
+  styleUrls: ["./report-row.component.scss"],
+})
+export class ReportRowComponent implements OnInit {
+  @Input() rows: ReportRow[] = [];
+  @Input() values: string[];
```
sleidig, line 11: can you add jsdoc comments here to all inputs? It's not fully clear what these `values` are

src/app/child-dev-project/reporting/reporting/reporting.component.ts
```diff
@@ -0,0 +1,64 @@
...
+  async calculateResults() {
+    this.reportingService.setAggregations(this.config.aggregationDefinitions);
+    this.results = await this.reportingService.calculateReport(
+      this.fromDate,
+      this.toDate
+    );
+
+    // The ReportRowComponents manipulates the values array of the elements
+    // Using a timeout the calculation waits for this manipulation to be done
+    setTimeout(() => (this.reportTable = this.getExportTable()));
```
sleidig, line 39: where does it change data? I can't see it (at least when [values] input is not used)

src/app/child-dev-project/reporting/reporting/reporting.component.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  private getExportTable(
+    rows = this.results
+  ): { label: string; result: any }[] {
+    const tableRows: { label: string; result: any }[] = [];
+    rows.forEach((result) => {
```
sleidig, line 46: in unit tests an error shows up in console: forEach on undefined

src/app/child-dev-project/reporting/reporting/reporting.component.ts
```diff
@@ -0,0 +1,64 @@
...
+  selector: "app-reporting",
+  templateUrl: "./reporting.component.html",
+  styleUrls: ["./reporting.component.scss"],
+})
+export class ReportingComponent implements OnInit {
+  config: ReportingComponentConfig;
+  results: ReportRow[];
+  fromDate: Date;
+  toDate: Date;
+  reportTable: { label: string; result: any }[];
```
sleidig, line 19: this is only used for export? rename to `resultsForExport` or something that is more clear?

src/app/child-dev-project/reporting/reporting.module.ts
```diff
@@ -0,0 +1,32 @@
...
+import { MatStepperModule } from "@angular/material/stepper";
+import { MatIconModule } from "@angular/material/icon";
+import { MatDatepickerModule } from "@angular/material/datepicker";
+import { MatFormFieldModule } from "@angular/material/form-field";
+import { FormsModule } from "@angular/forms";
+import { AdminModule } from "../../core/admin/admin.module";
+import { ReportRowComponent } from "./reporting/report-row/report-row.component";
+
+@NgModule({
+  declarations: [ReportingComponent, ReportRowComponent],
```
sleidig, line 17: shall we move this module into the `app/features/*` folder - or keep it as a part of the core framework?

src/app/child-dev-project/reporting/query.service.spec.ts
```diff
@@ -0,0 +1,384 @@
...
+        femaleMuslimChild,
+      ])
+    );
+  });
+
+  it("should return all children attending a school based on attributes", async () => {
+    const maleChildrenOnPrivateSchoolsQuery = `
+      ${School.ENTITY_TYPE}:toArray[*privateSchool=true]
+      :getRelated(${ChildSchoolRelation.ENTITY_TYPE}, schoolId)
+      :getActive.childId:addPrefix(${Child.ENTITY_TYPE}):unique
```
sleidig, line 226: this chain is really hard to understand (`:getActive.childId:addPrefix(${Child.ENTITY_TYPE}):unique`) can we change the phrasing to make it more intuitive?

src/app/child-dev-project/reporting/query.service.ts
```diff
@@ -0,0 +1,213 @@
...
+  private filterByObjectAttribute(
+    objs: any[],
+    attr: string,
+    key: string,
+    value: string
+  ): any[] {
+    const values = value.replace(new RegExp(" ", "g"), "").split("|");
+    return objs.filter((obj) => {
+      if (obj?.hasOwnProperty(attr)) {
+        return values.includes(obj[attr][key]?.toString());
```
sleidig, line 167: shouldn't this be an exact match rather than an "includes" here, when used for configurable-enum IDs?

sleidig: ah, never mind - the `value` parameter can have multiple options ...  maybe line 164 (the regex split) or the `values` variable can make that a bit more clear.

maybe add a jsdoc for the `value` parameter explaining `a value or a list of values separated by "|" (e.g. "ID_1|ID_2")`

src/app/child-dev-project/reporting/query.service.ts
```diff
@@ -0,0 +1,213 @@
...
+    );
+    return attendedChildren;
+  }
+
+  /**
+   * Returns the active child school relations from the input array
+   * @param relations an array of child school relations
+   * @returns an array of active child school relations
+   */
+  private getActive(relations: ChildSchoolRelation[]): ChildSchoolRelation[] {
```
sleidig, line 210: can and should this also be used for Child.isActive? The current demo report actually includes inactive children, thereby showing a different number than the dashboard widget, for example.

sleidig (commented):
Functionally tested, seems to work well :slightly_smiling_face: 
- needs a loading state (loading indicator + disabling the calculate button)
- let's allow multiple report configs and a simple dropdown in the UI to select one of them

Performance at 1500 Child, 25 School, IndexedDb:
- 60-75 seconds with 5 years of attendance events (in-memory: 12 seconds)
- 15 seconds with 1 year of attendance events (in-memory: 3 seconds)

I think this is acceptable for generating an extensive report - although the fact that very old data that is not even relevant for the reporting period is slowing things down _will_ become problematic in the long run.

Should we try to release memory after the report is finished? Calculating the report with the large dataset is taking 1-1.5 GB RAM on my machine. The memory remains in use even after seeing the result and leaving that page.

src/app/child-dev-project/reporting/query.service.spec.ts
TheSlimvReal: What would you find more intuitive? Maybe `getAttendingChildren` and `removeDuplicates` but for `addPrefix` I can not think of a more explanatory name

src/app/child-dev-project/reporting/query.service.ts
TheSlimvReal: Hmm I am not sure if `isActive` makes sense for all classes (why would a note be active or not) but I see that in some cases its reasonable. Concerning the dashboard the two numbers refer to a different entity (active children based on the dropout and active child-school-relations based on the end date) therefore the numbers differ in the case where a child that didnt dropout has not visited any events in the recent period.

src/app/child-dev-project/reporting/reporting.module.ts
TheSlimvReal: I dont consider this a core functionality because no other module is dependent on this feature. Therefore the features folder seams appropriate

src/app/child-dev-project/reporting/query.service.ts
```diff
@@ -0,0 +1,213 @@
...
+const jsonQuery = require("json-query");
+
+/**
+ * A query service which uses the json-query library (https://github.com/auditassistant/json-query).
+ */
+@Injectable({
+  providedIn: "root",
+})
+export class QueryService {
+  private static readonly ENTITY_CLASSES: typeof Entity[] = [
```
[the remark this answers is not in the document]
TheSlimvReal: currently both these places are rather wrong for the entity map. I would also target this in #764 to have a central place where all entities can be found.

TheSlimvReal:
The new _big_ changes:

- Loading data only from the given `from` date till today
- Directly passing the values of the groupBy`s down, so each row has a full list of values of the group it is in
- Moved module to the features folder

src/app/features/reporting/reporting.service.ts
```diff
@@ -0,0 +1,172 @@
...
+  private fromDate: Date;
+  private toDate: Date;
+
+  constructor(private queryService: QueryService) {}
+
+  public setAggregations(aggregations: Aggregation[]) {
+    this.aggregations = aggregations;
+  }
+
+  public calculateReport(from?: Date, to?: Date): Promise<ReportRow[]> {
```
sleidig, line 30: any benefit to persist the date range in the service rather than only directly pass it directly into calculation method?

src/app/features/reporting/reporting.service.ts
```diff
@@ -0,0 +1,172 @@
...
+        aggregation.query,
+        this.fromDate,
+        this.toDate,
+        data
+      );
+      if (aggregation.label) {
+        const newRow = {
+          header: {
+            label: aggregation.label,
+            values: additionalValues,
```
sleidig, line 54: does it make sense to call these "values"? They seem rather be grouping/filter criteria?

src/app/features/reporting/reporting.service.ts
```diff
@@ -0,0 +1,172 @@
...
+      if (!existingGroup) {
+        existingGroup = { value: currentValue, data: [] };
+        allGroups.push(existingGroup);
+      }
+      existingGroup.data.push(currentElement);
+      return allGroups;
+    }, new Array<{ value: ENTITY[PROPERTY]; data: ENTITY[] }>());
+  }
+
+  private getValueDescription(value: any, property: string): string {
```
sleidig, line 161: possibly something that should be in the component?

sleidig (approved):

Issue #554, opened 2020-12-11
Notes are not added directly


TheSlimvReal:
**Describe the bug**
The notes manager component allows to add new notes with a popup dialog.
The notes manager adds the note to the list of notes if it is returned when the dialog is closed.
This works if the `save` button is clicked in the first popup.
The `FormDialogService` opens a second popup if the first popup was closed by a click outside of the popup.
This second popup also allows to save the note.
However, when this is done, the notes manager is **not** notified about the saving of the note and can therefore not add the note to the list.
This results in the note not directly being visible in the list of notes.

**To Reproduce**
1. Go to 'Notes'
2. Click on 'add note'
3. Change some fields
4. Click outside of the popup
5. The popup should close and a new one should open which asks you if you want so save the entity
6. Click on 'save'
7. The saved entity is node visible in the list of notes

**Expected behavior**
The newly saved entity should directly be visible in the list of notes.

sleidig:
also see #401 - same problem occurs through the primary action (in general, the list is not refreshed)

Issue #695, opened 2021-03-04
Children that were added are not removed when cancelling


Schottkyc137:
When adding a child in the Note-details-component and one cancels the process, the child is still there in the list, and when re-opening the note

**To Reproduce**
1. Go to 'Notes'
2. Click on A Note
3. Add a Child (in the Children Section)
4. Click Cancel
5. The child is on the list

**Expected behavior**
The child should not be in the note's list when one cancels the process

**Assumed reason**
The copy of a chosen note is only shallow - the child-array is not copied but a reference to the original note.

sleidig:
Good find. I can confirm this :+1:

Issue #692, opened 2021-03-03
Make EntityMapper's return an Observable


Schottkyc137:
As of now, the way to load all objects of a type is using the entity-mapper's loadType-functionality. This works well when there is a constant amount of data to be loaded that can only change through one channel.
However, this does not work well, for example using the notes-list. Notes can both be added from an outside source (the primary action) as well as from the new-button. When adding this via the primary-action, it is very hard to route the way back to an open notes-list component (if there is any).

**Objective**
Find a way to load the data from the database but also enable continuous updates that can be seen on screen.

**Proposed Solutions & Alternatives**
A second function (for example `loadAll`) could be added, that - instead of a promise - would return an Observable. This would not emit all loaded values at once, but each value one-by one. The observable can be subscribed to - the subscription should stay active as long as the view is active. This way, notes or other data that is provided from an "outside" source can be added and seen on the screen directly.

**Additional context**
See the branch [entity-mapper-test](https://github.com/Aam-Digital/ndb-core/tree/entity_mapper_test) and especially the addition to [entity-mapper-service](https://github.com/Aam-Digital/ndb-core/blob/a1a11847ea3b7c697753864f39937f7a7c1dea9e/src/app/core/entity/entity-mapper.service.ts#L105) as an example. This function is not intended to replace old functionality, but add new one that can be used by several modules that need it.
See Issue #554

sleidig:
Thanks for picking up this broader topic (and sorry for the slow response ...)!

I think subscribing to updates will be a great addition. I am a bit undecided whether it would be easier to understand if we completely separate the loading of initial data and the subscribing to update events. The interface is still a bit confusing/complex.

We also need to check and test that we don't inadvertently trick Angular change detection by updating existing objects. As far as I know change detection usually only notices if the reference of an @Input is changed, not if elements within are updated. Not sure what that means for editing arrays as data sources for some of our components.

This is a great start! Let's discuss further details at a meeting.

Issue #401, opened 2020-01-30
New note not displayed when saving via the primary-action


Schottkyc137:
When a new note is being created via the primary-action component while in the note-manager component, it is only visible after a reload.
I'm not sure what the best approach here is. A general solution could be to add the functionality to subscribe to the EntityMapperService in order to get notified when a new Entity is being saved which can then lead to a reload of the data.

sleidig:
probably same cause as #554 , closing this issue as the newer one contains more information.

PR #723, opened 2021-03-24
Update Entities

Schottkyc137:
see issues: #692 #695 #554 #401

### Visible/Frontend Changes
- [x] New / updated notes will always be visible in the note-manager, no matter the method (fixes issues #554, #401)
- [x] Children  are only added to a note when save is clicked

- [x] Added the capability to copy entities to enable deep copies
- [x] One can subscribe to updates of entities of any kind
- [x] Notes are copied deeply when used in a dialog
- [x] Editing a note will no longer reflect live changes, but will only change the note when saving


src/app/core/entity/entity-mapper.service.ts
```diff
@@ -105,6 +136,24 @@ export class EntityMapperService {
    * @param entity The entity to be deleted
    */
   public remove<T extends Entity>(entity: T): Promise<any> {
+    this.sendUpdate(entity, "remove");
```
TheSlimvReal, line 139: I would consider using a enum for `remove`, `update` and `new`

src/app/core/entity/entity-mapper.service.ts
```diff
@@ -105,6 +136,24 @@ export class EntityMapperService {
...
     return this._db.remove(entity);
   }
+
+  /**
+   * publishes a new entity update to all subscribing listeners
+   *
+   * @param entity The entity to update
+   * @param type The type, see {@link UpdatedEntity#type}
+   */
+
```
TheSlimvReal, line 149:
```suggestion
```

src/app/core/entity/entity-update.ts
```diff
@@ -0,0 +1,56 @@
...
+export function update<T extends Entity>(
+  entities: T[],
+  next: UpdatedEntity<T>
+) {
+  if (!next || !next.entity) {
+    return entities;
+  }
+  switch (next.type) {
+    case "new":
+      return [next.entity].concat(entities);
```
TheSlimvReal, line 45: Will this mean that angular has to rebuild the view because the list of entities changes?

src/app/core/entity/entity-update.ts
```diff
@@ -0,0 +1,56 @@
+import { Entity } from "./entity";
+
+/**
+ * Interface that conveys an updated entity as well as information
+ * on the update-type
+ */
+
```
TheSlimvReal, line 7:
```suggestion
```

src/app/core/entity/entity-update.ts
```diff
@@ -0,0 +1,56 @@
...
+ * can either be a new entity (that should be inserted into the list), or an existing entity
+ * that should be updated or deleted.
+ * The given array will not be mutated but will be returned when the given new entity
+ * or type is illegal
+ * @param next An entity that should be updated as well as the type of update. This, as well as the entity
+ * may be undefined or null. In this event, the entities-array is returned as is.
+ * @param entities The entities to update, must be defined
+ * @return An array of the given entities with the update applied
+ */
+
```
TheSlimvReal, line 35:
```suggestion
```

src/app/core/entity/entity-update.ts
TheSlimvReal: And why exactly did you decide against mutating the original array? And for me word `update` suggests that the passed array is updated rather then a new array being built

src/app/core/entity/entity.spec.ts
```diff
@@ -99,4 +99,27 @@ describe("Entity", () => {
     expect(data.searchIndices).toContain("John");
     expect(data.searchIndices).toContain("Doe");
   });
+
+  it("can perform a shallow copy of itself", () => {
+    const id = "t1";
+    const entity: Entity = new Entity(id);
+    entity["value"] = 1;
+    const otherEntity = entity.copy();
+    expect(otherEntity).toEqual(entity);
```
TheSlimvReal, line 108: I think a assertion should also ensure, that the entity is copied rather than just assigned. A pointer (`otherEntity = entity`) would pass this test as well.

package-lock.json
```diff
@@ -5,7 +5,7 @@
   "requires": true,
```
TheSlimvReal, line 5: please remove this file from the pull request

src/app/child-dev-project/notes/model/note.ts
```diff
@@ -158,4 +158,16 @@ export class Note extends Entity {
 
     return false;
   }
+
+  /**
+   * Performs a deep copy of the note copying all simple data
+   * (such as the date, author, e.t.c.) as well as copying the
+   * child-array
+   */
+
```
TheSlimvReal, line 167:
```suggestion
```
Having the comment and the function connected makes it more clear that they belong together

TheSlimvReal (changes requested):
Really nice. It looks very good and clean. The only thing that stands out for me is that it took 10 additional (quite complex) lines to add this rather simple "update" functionality to the notes-manager.
But overall really expressive code and the tests allowed me to understand everything that is happening.

src/app/child-dev-project/notes/notes-manager/notes-manager.component.ts
```diff
@@ -80,8 +83,24 @@ export class NotesManagerComponent implements OnInit {
...
+      // This prevents edge-cases where updates are received
+      // before this is
+      this.notes = notes.concat(this.notes);
     });
+
+    this.entityMapperService
+      .receiveUpdates<Note>(Note)
+      .pipe(
+        untilDestroyed(this),
+        filter((note) => note !== null && note !== undefined),
```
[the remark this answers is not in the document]
Schottkyc137: This happens whenever the component initially subscribes to the observable and `null` is emitted due to the implementation of `BehaviorSubject`. I don't know if there is a better way (probably) but I couldn't find it.

src/app/core/entity/entity-mapper.service.spec.ts
```diff
@@ -172,4 +172,59 @@ describe("EntityMapperService", () => {
...
+    const testId = "t1";
+    receiveUpdatesAndTestTypeAndId("new", testId, done);
+
+    const testEntity = new Entity(testId);
+    entityMapper.save(testEntity, true);
+  });
+
+  function receiveUpdatesAndTestTypeAndId(
+    type: string | undefined,
+    entityId: string | undefined,
```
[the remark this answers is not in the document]
Schottkyc137: If it's undefined, the check won't happen (it will only test the defined values). That way one can test only the type (or id) with this one function. Will change to optional!

src/app/core/entity/entity-mapper.service.ts
Schottkyc137: The current type is a literal union type. Is there an advantage exchanging this for an enum?

src/app/core/entity/entity-update.ts
Schottkyc137: > Will this mean that angular has to rebuild the view because the list of entities changes?

Yes


> And why exactly did you decide against mutating the original array? And for me word `update` suggests that the passed array is updated rather then a new array being built

Mutating the original array wouldn't cause the view to rebuild itself. The name is confusing thou. I will change it to something more suitable. Maybe `updating`?

src/app/child-dev-project/notes/notes-manager/notes-manager.component.ts
```diff
@@ -27,6 +29,7 @@ import { EntityListConfig } from "../../../core/entity-components/entity-list/En
     ></app-entity-list>
   `,
 })
+@UntilDestroy()
```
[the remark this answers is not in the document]
Schottkyc137: Oops :P

package-lock.json
Schottkyc137: Should this actually be in .gitignore?

Schottkyc137:
> Really nice. It looks very good and clean. The only thing that stands out for me is that it took 10 additional (quite complex) lines to add this rather simple "update" functionality to the notes-manager.
> But overall really expressive code and the tests allowed me to understand everything that is happening.

Thanks a lot!
The reason that this is 10 lines long is because of the way colors are handled. If it wasn't for this, all would shrink down to only 5 lines.

src/app/child-dev-project/notes/notes-manager/notes-manager.component.ts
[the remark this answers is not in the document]
TheSlimvReal: So I googled around a little and found the concept of [hot and cold observables](https://stackoverflow.com/questions/32190445/hot-and-cold-observables-are-there-hot-and-cold-operators). A cold observable means, that every subscriber will get all the emitted values, even the ones that are emitted before they have subscribed. When subscribing to a hot observable, only the values which are emitted after the subscription happened, will trigger the subscriber. For this use case I believe, that subscribers should only be notified about updates that happen _after_ they subscribed (because the previous updates are already in the database), therefore a behavior subject is probably the wrong choice because it always emits one old value. Instead a simple `Observer` (which is a hot observable) should do the trick, that no old values are emitted and hence no null value should ever arrive.

One more thing to note: besides the `Subject`, all RxJS observers are lazy, which means that the first subscriber will get all values that have been emitted since the observable has been created. In this case I don't think, the laziness will be a problem, because the observers are only created when someone wants to subscribe to it.

package-lock.json
TheSlimvReal: No, the file is used by the CI to check whether new packages have been installed, in which case the build time will be much longer. But the file should only be committed if intentionally a new package has been installed or updated. Its a bit annoying that running `npm install` always changes something in this package, `npm ci` can be used to install the exact contents of the `package-lock.json` but for some users that resulted in problems.

src/app/child-dev-project/notes/notes-manager/notes-manager.component.ts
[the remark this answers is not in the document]
Schottkyc137: I opted for a `Subject` instead. This makes sure, more than one receiver can receive the values emitted, while not having the "behavioral" part. A simple `Observable` would require to think about the edge-case where more than one entity subscribes to updates of the same entity-type.

src/app/core/entity/entity-update.ts
TheSlimvReal: Maybe even something like `applyUpdate` because your passing the update and the array to which the update should be applied.

src/app/core/entity/entity.ts
```diff
@@ -201,4 +201,16 @@ export class Entity {
   public getWarningLevel(): WarningLevel {
     return WarningLevel.NONE;
   }
+
+  /**
+   * Shallow copy of the entity.
+   * The resulting entity will be of the same type as this
+   * (taking into account subclassing)
+   */
+
```
TheSlimvReal, line 210:
```suggestion
```
One last time 😄

TheSlimvReal (changes requested):
Looks very good now, just two more small fixes and we can merge it 👍

src/app/core/entity/entity-mapper.service.ts
```diff
@@ -93,6 +122,7 @@ export class EntityMapperService {
     const rawData = this.entitySchemaService.transformEntityToDatabaseFormat(
       entity
     );
+    this.sendUpdate(entity, entity._rev === undefined ? "new" : "update");
```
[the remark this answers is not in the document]
Schottkyc137: Good find! I did not think about that!
This opens a broader topic thou because the way it is currently implemented is the way it was before - at least when it worked (if I'm not mistaken). Maybe this should then not publish but instead send a notification to the user?

Issue #722, opened 2021-03-22
Put "All" at the beginning when filtering a list/table view by a boolean


christophscheuing:
When filtering a list/table view by a string variable, the "All"-option is displayed as the first one. When filtering by a boolean variable on the contrary, the "All"-option is shown as the last one:

<img width="547" alt="Bildschirmfoto 2021-03-22 um 10 01 06" src="https://user-images.githubusercontent.com/47225324/111965082-9ff22800-8af5-11eb-88b7-7310134939cc.png">

This might seem confusing to users not caring about a variable containing a string or a boolean value. We should harmonize the UI of the filters by putting the "All"-option to the beginning also for boolean variables.

PR #724, opened 2021-03-24
"All" option in same order for default and boolean filters

RichardNutt:
see issue: #722

### Visible/Frontend 
- [x] "All" option in same order for default and boolean filters

None


TheSlimvReal (approved):
I guess this makes sense to keep it in line with the other filters

PR #728, opened 2021-03-28
Generic select component

Schottkyc137:
see issue: #136, #681

### Visible/Frontend Changes
- [x] Minor changes concerning how the Child-Select and School-Select look
- [x] When selecting a user, a list of possibilities with all users will show up
- [x] Multiple authors can be chosen when looking at a note 
- [x] Multiple authors can be seen in the note-manager component
- [x] Fixes a bug where children can be removed in an event without being in edit-mode 

- [x] Removed Child-Select and School-Select component in favor of a more general select-component
- [ ] ~~This also makes the ChildMeetingNoteAttendance model obsolete~~
- [x] The note model uses an array of string-id's in favor of the author-name


src/app/core/user/user-list/user-list.component.ts
```diff
@@ -0,0 +1,57 @@
...
+  selector: "app-user-list",
+  template: `
+    {{ authorNames }}
+    <span *ngIf="_users.length > maxUserThreshold">
+      and {{ additionalUsers }} more
+    </span>
+  `,
+})
+export class UserListComponent implements OnInitDynamicComponent {
+  _users: User[] = [];
```
TheSlimvReal, line 17: As before, why the underscore if the variable is public

src/app/core/user/user-list/user-list.component.ts
```diff
@@ -0,0 +1,57 @@
...
+          );
+        })
+        .catch((error) => {
+          this.loggingService.error(error);
+        });
+    } else {
+      this._users = entities as User[];
+    }
+  }
+  @Input() inputType: "id" | "entity" = "id";
```
TheSlimvReal, line 35: Same as before, I think the input type can be taken from the type of input to the `set entities` function to reduce the amount of variables

TheSlimvReal (changes requested):
Looks very good and the code is well structured. I only noted small things that would be helpful before merging this into the master

src/app/child-dev-project/notes/model/note.ts
```diff
@@ -103,6 +104,7 @@ export class Note extends Entity {
    */
   removeChild(childId: string) {
     this.children.splice(this.children.indexOf(childId), 1);
+    this.children = [].concat(this.children); // TODO: hacky...
```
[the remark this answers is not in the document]
Schottkyc137: Correct. If this piece of code was to be removed, an automatic update wouldn't work anymore

src/app/core/entity-components/entity-select/entity-select/entity-select.component.ts
```diff
@@ -0,0 +1,241 @@
...
+  }
+  /**
+   * The (initial) selection. Can be used in combination with {@link selectionChange}
+   * to enable two-way binding to either an array of entities or an array of strings
+   * corresponding to the id's of the entities.
+   * <br>
+   * <b>Important:</b> If nothing is passed initially, then the type of this will be id's (in other
+   * words, id's will be emitted whenever something changes).
+   * If this behavior is not desired, the {@link selectionInputType} has to be set to "entity" explicitly
+   * These id's have to be legal id's for the given {@link entityType}
```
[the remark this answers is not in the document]
Schottkyc137: I forgot to change this comment. Initially, I experimented with automatic detection, as you suggested. However, I found no way to support this in a feasible way

src/app/core/entity-components/entity-select/entity-select/entity-select.component.ts
```diff
@@ -0,0 +1,241 @@
...
+   * @param sel The initial selection
+   */
+  @Input() set selection(sel: (string | E)[]) {
+    if (this.selectionInputType === "id") {
```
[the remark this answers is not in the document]
Schottkyc137: It could be if the selection is always non-empty. If it is empty, there is no way to find out if it's a string or an entity array (unless I'm missing something). This leads to problems when there is an empty initial array (for example when creating a new note) and then a new entity is added. Should this new entity be a string or an entity?

src/app/core/entity-components/entity-select/entity-select/entity-select.component.html
```diff
@@ -0,0 +1,42 @@
...
+      *ngFor="let entity of _selection"
+      class="chip">
+      <!-- This is going to be replaced with the dynamic content -->
+      <ng-container *ngTemplateOutlet="entityView; context: {entity: entity}"></ng-container>
+      <div
+        *ngIf="removable && !disabled"
+        class="fa fa-minus-square-o"
+        (click)="removeEntity(entity)">
+      </div>
+    </mat-basic-chip>
```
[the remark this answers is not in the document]
Schottkyc137: This is probably not planned when using a chip-list (https://material.angular.io/components/chips/overview). I will still try to see if this is feasible.

src/app/core/entity-components/entity-select/entity-select/entity-select.component.ts
```diff
@@ -0,0 +1,241 @@
...
+   * Additional filters
+   */
+  @Input() additionalFilter: (T) => boolean = (_) => true;
+  /**
+   * selects a given entity and emits values
+   * @param entity the entity to select
+   */
+  selectEntity(entity: E) {
+    this._selection.push(entity);
+    this._emitChange();
```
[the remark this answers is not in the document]
Schottkyc137: I used it where there are two variables that basically have the same name. Will refactor thou

src/app/core/entity-components/entity-select/entity-select/entity-select.component.ts
```diff
@@ -0,0 +1,241 @@
...
+      if (entity) {
+        this.selectEntity(entity);
+      }
+    }
+  }
+
+  private _filter(value: string): E[] {
+    const filterValue = value.toLowerCase();
+    return this.allEntities.filter((entity) =>
+      this.accessor(entity).toLowerCase().startsWith(filterValue)
```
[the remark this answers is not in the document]
Schottkyc137: The additional filter will only be applied to the entities that 'make' it through the original one.
This two-way filtering is needed since at the first stage either `_filter` is called, or all entities are being set (if the text is `null`). Moving additional filters here would mean that the filter only applies when a certain search-text is being set (would not apply when the selection is empty)

src/app/core/entity-components/entity-select/entity-select/entity-select.component.ts
```diff
@@ -0,0 +1,241 @@
...
+      )
+    );
+  }
+  /**
+   * The accessor used for filtering and when selecting a new
+   * entity.
+   * <br> Per default, this filters for the name. If the entity
+   * has no name, this filters for the entity's id.
+   */
+  @Input() accessor: accessorFn<E> = (e) => e["name"] || e.getId();
```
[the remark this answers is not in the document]
Schottkyc137: TSLint doesn't allow this. I think the reason is that the accessor as well as the additional filters are considered to be methods

src/app/app.module.ts
```diff
@@ -138,6 +140,8 @@ import { DashboardShortcutWidgetModule } from "./core/dashboard-shortcut-widget/
       ...DemoUserGeneratorService.provider(),
     ]),
     AttendanceModule,
+    MatChipsModule,
```
sleidig, line 143: Shouldn't we import the `EntitySelectModule` here instead of adding Material modules?

src/app/child-dev-project/notes/model/note.ts
sleidig, line 107:
```suggestion
    this.children = this.children.filter(c => c !== childId); 
```
this might make the purpose easier to understand

src/app/child-dev-project/notes/note-details/note-details.component.html
```diff
@@ -118,25 +118,53 @@ <h1>{{ entity.date?.toLocaleDateString() }}: {{ entity.subject }}</h1>
...
-        fxFlex
-      >
-      </app-note-presence-list>
+
     </div>
+
   </form>
 </app-form-dialog-wrapper>
+
+<ng-template
```
sleidig, line 157: could we use content projection for these templates as well? (maybe with "multi-slot" approach, if necessary https://imm9o.medium.com/angular-content-projection-the-complete-guide-45d3bc1692f2)

src/app/child-dev-project/notes/note-details/note-details.component.ts
```diff
@@ -17,12 +19,26 @@ export class NoteDetailsComponent implements ShowsEntity<Note> {
   @Input() entity: Note;
   @ViewChild("dialogForm", { static: true }) formDialogWrapper;
 
+  readonly Child: EntityConstructor<Child> = Child;
+  readonly User: EntityConstructor<User> = User;
+
   INTERACTION_TYPE_CONFIG = INTERACTION_TYPE_CONFIG_ID;
+  includeInactiveChildren: boolean = true;
```
sleidig, line 26:
```suggestion
  includeInactiveChildren: boolean = false;
```
inactive should be excluded by default

sleidig (commented):
Functionality looks great.

"meeting notes" list out all participants twice, however - in the entity-select and in the list below that includes their attendance status. In the previous implementation the entries in the select component were hidden to keep the UI simple. Is there a way to keep it like that?

src/app/child-dev-project/notes/model/note.ts
```diff
@@ -52,7 +52,8 @@ export class Note extends Entity {
   @DatabaseField() date: Date;
   @DatabaseField() subject: string = "";
   @DatabaseField() text: string = "";
-  @DatabaseField() author: string = "";
+  /** IDs of authors of this note */
+  @DatabaseField() authors: string[] = [];
```
sleidig, line 56: maybe we could use a getter here to automatically migrate the old "author" string to the new format here?

Or we will need a one-time migration script to convert the properties. That could be done semi-manually - triggered from the admin section like [AttendanceMigrationService](https://github.com/Aam-Digital/ndb-core/blob/master/src/app/child-dev-project/attendance/attendance-migration/attendance-migration.service.ts).

sleidig, line 56: maybe we could use a getter here to automatically migrate the old "author" string to the new format here?

Or we will need a one-time migration script to convert the properties. That could be done semi-manually - triggered from the admin section like [AttendanceMigrationService](https://github.com/Aam-Digital/ndb-core/blob/master/src/app/child-dev-project/attendance/attendance-migration/attendance-migration.service.ts).

src/app/child-dev-project/notes/note-details/note-details.component.html
Schottkyc137: something I had considered. Unfortunately, I don't see/know how to pass the entities to the underlying module (currently this is done using the 'let-entity' approach in the template). I am happy to exchange this approach for a template since that's cleaner and more intuitive. However, I don't know exactly how to do that. Maybe there is a way to use something similar to what angular uses for table cells?

src/app/core/entity-components/entity-select/entity-select/entity-select.component.ts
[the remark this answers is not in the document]
Schottkyc137: I changed this so there is only one filter-function. This should make things more clear

src/app/core/entity-components/entity-select/entity-select/entity-select.component.html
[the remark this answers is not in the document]
TheSlimvReal: Ah is see, I just find it rather uncomfortable to use. Maybe its sufficient to move the `<input>` outside the chip list.

src/app/core/entity-components/entity-select/entity-select/entity-select.component.html
```diff
@@ -0,0 +1,42 @@
...
+      [removable]="removable && !disabled"
+      *ngFor="let entity of _selection"
+      class="chip">
+      <!-- This is going to be replaced with the dynamic content -->
+      <ng-container *ngTemplateOutlet="entityView; context: {entity: entity}"></ng-container>
+      <div
+        *ngIf="removable && !disabled"
+        class="fa fa-minus-square-o"
+        (click)="removeEntity(entity)">
+      </div>
```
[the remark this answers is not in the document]
Schottkyc137: I changed it to where each chip has a distinct feeling. How do you like the UX of the new approach?

src/app/child-dev-project/notes/notes-migration/notes-migration.service.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+    nonMatchedUsers.forEach((userName, index) => {
+      const foundUsers = service.findUsers(userName);
+      expect(foundUsers.additional).toEqual(expected[index]);
+      expect(foundUsers.detectedUsers).toHaveSize(0);
+    });
+  });
+
+  it("migrates a note with existing users", () => {
+    const note = legacyNote("Peter L, Ursula & Jens");
+    const expectedUsers = users.slice(0, 3).map((u) => u.getId());
```
TheSlimvReal, line 96:
```suggestion
    const expectedUsers = [Peter, Ursula, Jens].map(u => u.getId());
```
This would make the tests more readable

src/app/child-dev-project/notes/notes-migration/notes-migration.service.ts
```diff
@@ -0,0 +1,118 @@
+import { Injectable } from "@angular/core";
+import { EntityMapperService } from "../../../core/entity/entity-mapper.service";
+import { Note } from "../model/note";
+import { User } from "../../../core/user/user";
+
+@Injectable({
+  providedIn: "root",
+})
+export class NotesMigrationService {
+  private static updateNoteText(note: Note, additionalUsers: string[]) {
```
TheSlimvReal, line 10: I know this function does not use any class variables, but defining it as static and making it the first function of the class confuses readers quite a lot. I would keep it a class variable and place it after the first usage in line 60.

Also please remove the whitespaces between the comments and the functions in this file

src/app/core/entity/mock-entity-mapper-service.ts
```diff
@@ -0,0 +1,123 @@
...
+    this.delete(entity);
+    return Promise.resolve();
+  }
+
+  receiveUpdates<T extends Entity>(
+    entityType: EntityConstructor<T>
+  ): Observable<UpdatedEntity<T>> {
+    return NEVER;
+  }
+}
```
TheSlimvReal, line 123: maybe also override the `saveDatabaseIndex` function and log a warning that it is not supported for the `mockEntityMapper`.

TheSlimvReal (changes requested):
Looks very good now, just a few small things:

When I want to select multiple users, I always have to click somewhere else on the screen to unselect the input field before the dropdown will open again. This could be fixed by automatically unselecting the input element once something has been selected.

Other than that, I really like the the styling of the selected components. It would just be good if the little remove button also indicates that its clickable, maybe by changing the cursor or showing a hover event.

src/app/core/entity/mock-entity-mapper-service.ts
Schottkyc137: This is not a function in the entity-mapper, but in the database. Or do you mean something else?

TheSlimvReal (approved):
Looks very good and works quite well. @sleidig I leave the migration to you

TheSlimvReal (approved):
I just realized that the user select component should also be used for the `RecurringActivities` (the assignedTo field). Currently this is also just a text field and should be replaced by an array(?) of users.

src/app/core/user/user-list/user-list.component.ts
```diff
@@ -0,0 +1,57 @@
...
+@Component({
+  selector: "app-user-list",
+  template: `
+    {{ authorNames }}
+    <span *ngIf="users.length > maxUserThreshold">
+      and {{ additionalUsers }} more
+    </span>
+  `,
+})
+export class UserListComponent implements OnInitDynamicComponent {
```
sleidig, line 16: There is already a component called UserList (the view displaying a table of all user accounts). This new component needs to be named differently

sleidig (approved):

TheSlimvReal (approved):
Works very  well, I can not find any problems

PR #731, opened 2021-03-30
Index Implementation for MockDB

TheSlimvReal:
Implemented the functionality to create and query indices for the mock database.

Currently `reduce` is not supported and the performance for the demo data is very bad due to the many usages of the index (`childBlockComponent`).

The performance could be improved by exchanging the sorting of the documents with a more sophisticated data structure that already sorts the elements when inserted.

- [x] `saveDatabaseIndex` and `query` implemented for the `MockDatabase`

### DIFF
--- a/src/app/core/database/mock-database.ts
+++ b/src/app/core/database/mock-database.ts
@@ -16,13 +16,7 @@
  */
 
 import { Database, GetAllOptions, QueryOptions } from "./database";
-import { Note } from "../../child-dev-project/notes/model/note";
-import { AttendanceMonth } from "../../child-dev-project/attendance/model/attendance-month";
-import { ChildSchoolRelation } from "../../child-dev-project/children/model/childSchoolRelation";
-import moment from "moment";
-import { RecurringActivity } from "../../child-dev-project/attendance/model/recurring-activity";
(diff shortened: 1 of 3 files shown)


src/app/core/database/mock-database.ts
```diff
@@ -37,6 +31,7 @@ export class MockDatabase extends Database {
     return instance;
   }
 
+  private views: { [key: string]: { map: string; reduce?: string } } = {};
```
Schottkyc137, line 34: That type confuses me. Can you maybe elaborate (comment)/make the type more clear?

Schottkyc137 (commented):
What about sorting the views once only and then re-sorting them on updates (resp. new/delete-commands)? Is that something that is feasible?

sleidig:
Do we want to keep mainting the "mock-database" implementation in the longer run? It is only relevant for demo purposes actually (and that mostly because the "local" mode with a real pouchdb is too slow with initial indexing). If we can use the [in-memory adapter](https://github.com/pouchdb/pouchdb/tree/master/packages/node_modules/pouchdb-adapter-memory) of pouch or something else we may not need this at all?

Issue #729, opened 2021-03-29
Un-descriptive title in certain events


Schottkyc137:
When all children are excused at a particular event, the description will say "attended (of 5 participants)". This should probably be something like "no events on this date" or "no participants". This also produces 'NaN' in the details-view, which should probably be something like 'N/A' or simply '-'

If this cannot happen in a real-world scenario, this is a bug of the demo-generator.

**To Reproduce**
1. Go to Recurring Activities
2. Click on a single activity and find an activity that is white, but still has an event
4. See the error when looking at the details

**Expected behavior**
it should be clear that no one participated at the events (without warning about 0% attendance).
If this is an error of the demo-data generator, such events should not be generated

**Screenshots**
<img width="434" alt="Bildschirmfoto 2021-03-29 um 19 59 22" src="https://user-images.githubusercontent.com/45085299/112879331-48d9ed80-90c9-11eb-9ac9-0afb9141f8bc.png">

sleidig:
good point. This happens in all modes. One more issue to be ironed out :blush:

PR #734, opened 2021-04-01
Changed descriptions to a more suitable format

Schottkyc137:
see issue: #729 

### Visible/Frontend Changes
These texts have been updated to a more readable format:
- [x] 'N/A' instead of 'NaN' in the activity details-component
- [x] 'All excused (out of x participants)' instead of 'attended (out of x participants)'
- [x] The attendance of a single child was in a format that was badly readable 

None


sleidig (approved):
Looks good. Thanks for fixing that.

PR #738, opened 2021-04-05
Preventing transfomation of null values

TheSlimvReal:
Currently setting a value to `null` can break the code because the schema transformation does not further check if the value is null. 
This is now prevented by explicitly not transforming `null` values of attributes.

### Visible/Frontend Changes
--

- Now transformation functions are not called with `null` values any more

### DIFF
--- a/src/app/core/entity/schema/entity-schema.service.spec.ts
+++ b/src/app/core/entity/schema/entity-schema.service.spec.ts
@@ -190,4 +190,31 @@ describe("EntitySchemaService", () => {
     expect(rawData.details.month).toEqual("2020-01");
     expect(rawData.details.otherStuff).toBeUndefined();
   });
+
+  it("should not transform null values", () => {
+    class Test extends Entity {
+      @DatabaseField({ dataType: "date-only" }) normalDate: Date;
+      @DatabaseField({ dataType: "date-only" }) nullDate: Date;
+    }
+    const testObject = new Test();
+    testObject.normalDate = new Date();
+    testObject.nullDate = null;
+    const rawData = entitySchemaService.transformEntityToDatabaseFormat(
+      testObject
+    );
+    expect(rawData.normalDate).toBeDefined();
+    expect(rawData.nullDate).toBeUndefined();
+  });
+
+  it("should not load null values", () => {
+    class Test extends Entity {
+      @DatabaseField({ dataType: "date-only" }) normalDate: Date;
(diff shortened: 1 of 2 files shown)


sleidig:
@TheSlimvReal can you explain a situation where this has become a problem? I agree that treating undefined and null differently often is just confusing, but there may be times when we want to distinguish between "the value was never set" (undefined) and "we explicitly set the value to 'nothing'" (null). A property can never be "null" by accident, only if we explicitly set it to null. And in that case it probably makes sense to not overwrite it with a default value (if we want the default value, we could have set the property to "undefined" explicitly, also)

TheSlimvReal:
> 
> 
> @TheSlimvReal can you explain a situation where this has become a problem? I agree that treating undefined and null differently often is just confusing, but there may be times when we want to distinguish between "the value was never set" (undefined) and "we explicitly set the value to 'nothing'" (null). A property can never be "null" by accident, only if we explicitly set it to null. And in that case it probably makes sense to not overwrite it with a default value (if we want the default value, we could have set the property to "undefined" explicitly, also)

I would prefer not to work too much with `undefined` and `null` values in the code.
However, in this case it breaks with the used `date-only` datatype because the transformation does not check whether a `null` value might be passed and this breaks the app if you remove a existing date from a existing `ChildSchoolRelation` in the `Previous Schools` section, because the input sets the value `null` and not `undefined` if no value is present. 
I would rather not check this problem in every datatype and instead treat `undefined` and `null` the same way (in both cases no value is set).

src/app/core/entity/schema/entity-schema.service.spec.ts
```diff
@@ -190,4 +190,31 @@ describe("EntitySchemaService", () => {
...
+  it("should not load null values", () => {
+    class Test extends Entity {
+      @DatabaseField({ dataType: "date-only" }) normalDate: Date;
+      @DatabaseField({ dataType: "date-only" }) nullDate: Date;
+    }
+    const rawData = { normalDate: "2021-05-04", nullDate: null };
+    const testEntity = new Test();
+    entitySchemaService.loadDataIntoEntity(testEntity, rawData);
+    expect(testEntity.normalDate).toBeDefined();
+    expect(testEntity.nullDate).toBeNull();
```
sleidig, line 218: The test claims "not to load null" but the expectation then is that the value is "toBeNull"?

PR #741, opened 2021-04-06
Event on wrong day bug

TheSlimvReal:
The datepicker creates dates like `Mon Apr 05 2021 00:00:00 GMT+0200 (Central European Summer Time)`. The problem is, that currently `.getISOString()` is used to get the date in format `YYYY-MM-DD` but this will transform the datepicker date to `2021-04-04` (the day before) because of the timezone. This results in the wrong events being displayed in the record attendance section.

### Visible/Frontend Changes
- Correct events are displayed in the `Recurring Activities`

- Using `moment` to format the date correctly

### DIFF
--- a/src/app/child-dev-project/attendance/add-day-attendance/roll-call-setup/roll-call-setup.component.ts
+++ b/src/app/child-dev-project/attendance/add-day-attendance/roll-call-setup/roll-call-setup.component.ts
@@ -52,12 +52,12 @@ export class RollCallSetupComponent implements OnInit {
         a.assignedTo === ""
     );
 
-    for (const activity of this.visibleActivities) {
-      const newEvent = await this.createEventForActivity(activity);
-      if (newEvent) {
-        this.existingEvents.push(newEvent);
-      }
-    }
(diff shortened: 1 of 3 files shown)


src/app/child-dev-project/attendance/add-day-attendance/roll-call-setup/roll-call-setup.component.ts
```diff
@@ -52,12 +52,12 @@ export class RollCallSetupComponent implements OnInit {
...
-      if (newEvent) {
-        this.existingEvents.push(newEvent);
-      }
-    }
+    const eventPromises = this.visibleActivities.map((activity) =>
+      this.createEventForActivity(activity)
+    );
+    let events = await Promise.all(eventPromises);
+    events = events.filter((event) => !!event);
+    this.existingEvents.push(...events);
```
TheSlimvReal, line 60: This change will show all the activities at once instead of appearing one by one. 
I made some performance comparisons and for my test setup, the previous method took around 4600ms and this method takes around 2600ms. The drawback is that the user does not get feedback about the loading of activities.

PR #743, opened 2021-04-06
Updated faker

TheSlimvReal:
Faker showed some deprecation notes.

### Visible/Frontend Changes
--

- Updated faker
- Removed deprecated calls


sleidig (approved):
Awesome!

I tried to do that before and failed to get the typings right :sweat_smile:

PR #749, opened 2021-04-08
Improved performance for roll call

TheSlimvReal:
This change will show all the activities at once instead of appearing one by one.
I made some performance comparisons and for my test setup, the previous method took around 4600ms and this method takes around 2600ms. The drawback is that the user does not get feedback about the loading of activities.

### Visible/Frontend Changes
- Activities on Roll Call load faster

- Loading all activities at once instead of one by one


TheSlimvReal:
Made part of #812

PR #750, opened 2021-04-08
Replace mockdatabase with PouchDB in-memory

TheSlimvReal:
Replaced the self-implemented `MockDatabase`  with the fully functional [in-memory adapter](https://pouchdb.com/adapters.html) of PouchDB. This allows to keep having small and independent tests but removes the burden of mocking indices etc.

### Visible/Frontend Changes
-- 

- [x] Remove `MockDatabase` implementation
- [x] Install `pouchdb-core`, `pouchdb-adapter-memory` and `pouchdb-mapreduce`
- [x] Add static function to create a `MockDatabase` with PouchDB in-memory
- [x] Chang all tests to correctly use new `MockDatabase`
- [x] Move tests for `MockDatabase` into test suite for `PouchDatabase`
- [x] <del>Create function to automatically wait for creation of all indices in `MockDatabase`
- [x] Replace usages of PouchDB browser with in-memory in all tests
- [x] In `mock` mode, the `New LocalSession` with the new `MockDatabase` is used
- [x] Remove `MockSession` and `EntityTestingModule` because both are not needed any more
- [ ] Querying a index should wait until the index is built


src/app/child-dev-project/attendance/add-day-attendance/roll-call-setup/roll-call-setup.component.spec.ts
```diff
@@ -59,24 +65,19 @@ describe("RollCallSetupComponent", () => {
     expect(component).toBeTruthy();
   });
 
-  it("generates a filled event Note from the RecurringActivity", fakeAsync(() => {
+  it("generates event notes with current user as author", fakeAsync(() => {
```
sleidig, line 68:
```suggestion
  it("generates event notes with current user as author and details implied from RecurringActivity", fakeAsync(() => {
```

the purpose of this test (although described in weird language) is actually to check that category and subject are taken from activities when generating events. the expectations do not ensure this anymore now

src/app/core/database/mock-database.ts
```diff
@@ -15,431 +15,51 @@
...
-          map: (doc) => { emit(doc.month); }
-        }
-      }
-    };
-     */
-
-    // tslint:disable-next-line:no-console
-    console.debug('MockDatabase does not implement "saveDatabaseIndex()"');
-    return Promise.resolve({});
+  public waitForIndexing(): Promise<any> {
```
sleidig, line 62: let's put this in the `PouchDatabase` parent class, so we can also wait for completion of its indices in performance tests?

src/app/core/database/pouch-database.ts
```diff
@@ -27,6 +28,13 @@ import { LoggingService } from "../logging/logging.service";
  * should be implemented in the abstract {@link Database}.
  */
 export class PouchDatabase extends Database {
+  static createWithInBrowserDB(
```
sleidig, line 31: couldn't this just be named `create()` for both classes as the type is implied by the class name already?

src/app/core/entity/entity.ts
```diff
@@ -22,13 +22,16 @@ import {
   WarningLevel,
   WarningLevelColor,
 } from "../../child-dev-project/warning-level";
+import { Note } from "../../child-dev-project/notes/model/note";
```
sleidig, line 25: unnecesssary import?

src/app/core/database/pouch-database.ts
TheSlimvReal: I know moved both static methods to `PouchDatabase`. Now the distinctive name makes sense.

src/app/child-dev-project/children/children.service.ts
```diff
@@ -124,64 +123,6 @@ export class ChildrenService {
     return from(promise);
   }
 
-  getAttendances(): Observable<AttendanceMonth[]> {
```
TheSlimvReal, line 127: I removed these functions because they are deprecated and not used anywhere

src/app/child-dev-project/children/children.service.ts
```diff
@@ -124,64 +123,6 @@ export class ChildrenService {
...
-      AttendanceMonth,
-      "attendances_index/by_month",
-      monthString
-    );
-
-    return from(promise);
-  }
-
-  /**
-   * @deprecated use AttendanceService instead. This can be removed after all AttendanceMigrationService tasks are completed.
```
sleidig, line 154: the whole `AttendanceMonth` implementation is deprecated (i.e. shouldn't be used anymore) but not irrelevant yet. Please leave this until all active user instances are migrated, parts of this may be needed for migration and troubleshooting during migration.

src/app/core/database/pouch-database.ts
sleidig: looks good :+1: 

Maybe we should call this `createWithIndexedDB`

src/app/child-dev-project/children/children.service.ts
TheSlimvReal: Yes `AttendanceMonth` is still used in the `AttendanceMigrationService`. But the function call in the `ChildrenService` are not used anywhere? I dont really see how this could be used by anyone. And currently the indices are built every time even though nobody uses it.

sleidig (approved):
:-D

PR #751, opened 2021-04-09
feat: sort lists automatically by first column

sleidig:
better usability, especially NotesManager, with reasonable sorting by default.


TheSlimvReal (approved):
Good idea and works nicely

PR #759, opened 2021-04-10
feat(attendance): show events of recurring activities in the notes list

sleidig:
show events of recurring activities (`EventNote` entities) in the notes list and make that option configurable


TheSlimvReal (changes requested):
Works very well, just a small renaming

PR #762, opened 2021-04-14
PerformanceAnalysisLogging annotation

sleidig:
easily annotate a method without changing them and see its running duration in the console


TheSlimvReal (approved):
Works nicely

Issue #707, opened 2021-03-09
Allow maximum option for pagination settings


christophscheuing:
*As a user I want to have a maximum option tin the pagination settings for a table in order to show all available entries at once (and maybe search through them via the browser).*

PR #782, opened 2021-04-20
Pagination all option

christophscheuing:
see issue: #707


christophscheuing:
It would be easy to implement a all/max(imum) option for the pagination with reference to the maximum number in the specific case (e.g. if there are 8 entries having "8" as the maximum option for pageSize).

However we agreed that having an option called "All" or "Max" would be more intuitive and therefore preferable. This seems to be harder to implement. I tried with a rather hacky solution for the entity-list.component. This UI works more or less reliable. Sometimes I still get displayed "2147483647" instead of "All", but I cannot quite reproduce this behavior. It probably has to do with where to put
"this.paginatorHtmlElement.nativeElement.querySelector('div.mat-select-value > span > span').innerHTML = this.paginator.pageSize == 2147483647? 'All': this.paginator.pageSize;"
inside the entity-list.component.ts. Are there better placed than at the end of ngOnChanges and onPaginateChange?

If you approve this solution, we would have to think of how to adopt it for the entity-subrecord.component as well. I couldn't quit get this hacky solution to work there. The architecture of the pagination feature in the entity-subrecord.component is different as it has the same pageSize options for all tables being shown via the entity-subrecord.component. I don't really know how to deal with that.

src/styles.scss
```diff
@@ -215,3 +215,53 @@ mat-card.dashboard-widget {
...
   color: black;
 }
+
+/* Replace the max java integer in the page size dropdown with the label 'All'  */
+mat-option[ng-reflect-value='2147483647'] > span {
+	display: none;
+}
+
+/* Replace the max java integer in the page size dropdown with the label 'All'  */
+mat-option[ng-reflect-value='2147483647']:last-child:after {
```
TheSlimvReal, line 225: Is this still relevant? I dont see the max integer being used

src/app/core/entity-components/entity-list/entity-list.component.ts
```diff
@@ -206,6 +212,46 @@ export class EntityListComponent<T extends Entity>
...
+    if (!this.showAllToggle) {
+      this.paginatorPageSizeBeforeToggle = this.paginatorPageSize;
+      this.paginatorPageSize = this.entityDataSource.data.length;
+    } else {
+      if (
+        this.paginatorPageSizeBeforeToggle <= this.entityDataSource.data.length
+      ) {
+        this.paginatorPageSize = this.paginatorPageSizeBeforeToggle;
+      } else {
+        const po = this.getPaginatorPageSizeOptions();
```
TheSlimvReal, line 245: I think you could also access the pagination session that is saved in the user entity.

src/styles.scss
```diff
@@ -215,3 +215,53 @@ mat-card.dashboard-widget {
...
+	content: 'All' !important;
+	visibility: visible !important;
+	display: block !important;
+	position: absolute !important;
+	background-color: transparent !important;
+}
+
+.mat-slide-toggle-thumb {
```
TheSlimvReal, line 233: Does this have to be defined here or could it be moved to `entity-list.component.scss`?

TheSlimvReal (changes requested):
It already works very well and I like the little gimmick with the icon on the toggle :)
After some cleanup I think this is ready to be merged

src/styles.scss
christophscheuing: Right; I've deleted that.

src/app/core/entity-components/entity-list/entity-list.component.ts
christophscheuing: How do you mean that? Do you think that I don't need the paginatorPageSizeBeforeToggle? But imho the pagination session that is saved in the user entity only holds the current value, not the historic one.

TheSlimvReal: Currently, the page size is only saved on the user entity if the dropdown is used but not with the toggle. I think the page size should be saved in both cases.

christophscheuing: I've added pdateUserPaginationSettings() at the end of clickShowAllToggle.

src/app/core/entity-components/entity-list/list-paginator/list-paginator.component.ts
```diff
@@ -0,0 +1,137 @@
...
+
+  // This key is used to save the pagination settings on the user entity
+  readonly paginatorKey: string;
+
+  constructor(
+    private sessionService: SessionService,
+    private entityMapperService: EntityMapperService,
+    private router: Router,
+  ) {
+    this.paginatorKey = getUrlWithoutParams(this.router);
```
TheSlimvReal, line 36: There might be a problem when we also use this in the entity subrecord. In case there are multiple tables on one page (e.g. child details) all of them would have the same `paginatorKey` and therefore overwrite each others settings. Can you think of a way to distinguish different tables?

src/app/core/entity-components/entity-list/list-paginator/list-paginator.component.ts
```diff
@@ -0,0 +1,137 @@
...
+  @Input() dataSource: MatTableDataSource<E>;
+
+  @ViewChild(MatPaginator) paginator: MatPaginator;
+
+
+  user: User;
+  paginatorPageSize: number = 10;
+  paginatorPageSizeBeforeToggle: number = 10;
+  paginatorPageIndex: number = 0;
+  showAllToggle: boolean = false;
```
TheSlimvReal, line 26: Maybe rename this property to `showingAll` otherwise I would expect that when `showAllToggle === false` then the toggle is hidden.

src/app/core/entity-components/entity-list/list-paginator/list-paginator.component.ts
```diff
@@ -0,0 +1,137 @@
...
+      console.log("AfterViewInit mit Timeout aufgerufen. Length ist " + this.dataSource.data.length);
+    });
+    console.log("AfterViewInit aufgerufen. Length ist " + this.dataSource.data.length);
+  }
+
+  onPaginateChange(event: PageEvent) {
+    this.paginatorPageSize = event.pageSize;
+    this.paginatorPageIndex = event.pageIndex;
+    this.updateUserPaginationSettings();
+    this.showAllToggle =
```
TheSlimvReal, line 76: Why does this have to be recalculated when the page is changed? Is this also triggered when the page size is increased?

src/app/core/entity-components/entity-list/list-paginator/list-paginator.component.ts
```diff
@@ -0,0 +1,137 @@
...
+    const ar = [3, 10, 20, 50].filter((n) => {
+      return n < this.dataSource.data.length;
+    });
+    ar.push(this.dataSource.data.length);
+    return ar;
+  }
+
+  getPaginatorPageSize(): number {
+    if (
+      this.dataSource.data.length &&
```
TheSlimvReal, line 91: what is this check necessary for?

src/app/core/entity-components/entity-list/list-paginator/list-paginator.component.ts
```diff
@@ -0,0 +1,137 @@
...
+      this.paginatorPageSize >= this.dataSource.data.length
```
TheSlimvReal, line 92: In which case can the page size be greater than the data length? Doesn't `getPaginatorPageSizeOptions` ensure that only options smaller or equal to the data length are available?

src/app/core/entity-components/entity-list/list-paginator/list-paginator.component.ts
```diff
@@ -0,0 +1,137 @@
...
+        this.paginatorPageSizeBeforeToggle <= this.dataSource.data.length
+      ) {
+        this.paginatorPageSize = this.paginatorPageSizeBeforeToggle;
+      } else {
+        const po = this.getPaginatorPageSizeOptions();
+        this.paginatorPageSize = po.length > 2 ? po[po.length - 2] : po[0];
+      }
+    }
+    this.paginator._changePageSize(this.paginatorPageSize);
+    this.showAllToggle = !this.showAllToggle;
```
TheSlimvReal, line 114: Doesnt the `[(ngModel)]` of the toggle already change the value of `this.showAllToggle`?

src/styles.scss
```diff
@@ -215,3 +215,33 @@ mat-card.dashboard-widget {
 .mat-select-disabled .mat-select-value {
   color: black;
 }
+
+/* configuration for the mat-slide-toggle feature, used e.g. in the entity-list.component for the "Show All"
+option within the pagination of the tables */
```
TheSlimvReal, line 220: Can this be moved to `list-paginator.component.scss`? Otherwise this might influence other toggles components

TheSlimvReal (changes requested):
Looks good in the new component. I have only a few points for clarification.

I run into a small bug: 

1. Go to the schools list
2. Set _3_ as the page size
3. Leave the school list and go back to it
4. _3_ will be the page size but the show all toggle will be set to true

src/styles.scss
christophscheuing: It is moved now to list-paginator.component.scss.

src/app/core/entity-components/entity-list/list-paginator/list-paginator.component.ts
christophscheuing: I have renamed it.

src/app/core/entity-components/entity-list/list-paginator/list-paginator.component.ts
christophscheuing: This is also triggered when the page size is changed or increased. I think that calling the update method is necessary because the changes might need to be saved to the user.

src/app/core/entity-components/entity-list/list-paginator/list-paginator.component.ts
christophscheuing: This check guarantees that getPaginatorPageSize() does't fail when the data is not yet loaded correctly.

src/app/core/entity-components/entity-list/list-paginator/list-paginator.component.ts
christophscheuing: getPaginatorPageSizeOptions only pushes the selectable options to the UI, but the initial paginatorPageSize is hard coded in list-paginator.component.ts and thus might be greater than the length of data.

src/app/core/entity-components/entity-list/list-paginator/list-paginator.component.ts
christophscheuing: I don't really know why, but this manual setting seems to be necessary.

src/styles.scss
christophscheuing: Yes, it has been moved.

src/app/core/entity-components/entity-list/list-paginator/list-paginator.component.ts
```diff
@@ -99,7 +103,7 @@ export class ListPaginatorComponent<E extends Entity>
       this.paginatorPageSize = po.length > 2 ? po[po.length - 2] : po[0];
     }
     this.paginator._changePageSize(this.paginatorPageSize);
-    this.allToggle = !this.allToggle;
```
christophscheuing: Why did you change this? For me this needs to be

this.allToggle = !this.allToggle;

instead of 

this.allToggle = this.paginatorPageSize >= this.dataSource.data.length;

for the component to work.

TheSlimvReal (approved):

Issue #786, opened 2021-04-23
Hide "All" filter of no options are available


TheSlimvReal:
**Describe the bug**
In the entity-list view, the a _All_ option for filters is shown even if no other options are available. This is confusing if people first use the app and because the filter does not have any descriptions, it is not clear what this _All_ button does.

**Solution**
To solve this problem, filters without options besides _All_ should be hidden. (This has already been implemented for the column groups)

TheSlimvReal:
Mergeed with #787

PR #787, opened 2021-04-23
Ux improvements

TheSlimvReal:
see issue: #786 

### Visible/Frontend Changes
- [x] School block component should not display the `uuid` of a school
- [x] Filters with only one (the default) option should not be shown
- [x] The schools in School filter of the children list view should be sorted by name


sleidig (approved):
:+1:

PR #795, opened 2021-04-27
fix: show attendance remarks in event/note details list

sleidig:


TheSlimvReal (approved):
Looks good, I removed some old commented code

PR #796, opened 2021-04-27
fix: let users skip a participant during recording attendance

sleidig:
let users skip a participant during recording attendance and make that more explicit than the "Skip" status among other status

### DIFF
--- a/src/app/child-dev-project/attendance/model/calculate-average-event-attendance.ts
+++ b/src/app/child-dev-project/attendance/model/calculate-average-event-attendance.ts
@@ -32,8 +32,7 @@ export function calculateAverageAttendance(
   return {
     average:
       stats.get(AttendanceLogicalStatus.PRESENT) /
-      (stats.get(AttendanceLogicalStatus.PRESENT) +
-        stats.get(AttendanceLogicalStatus.ABSENT)),
+      event.children.length,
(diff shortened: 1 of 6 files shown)


src/app/child-dev-project/attendance/model/calculate-average-event-attendance.ts
```diff
@@ -32,8 +32,7 @@ export function calculateAverageAttendance(
   return {
     average:
       stats.get(AttendanceLogicalStatus.PRESENT) /
-      (stats.get(AttendanceLogicalStatus.PRESENT) +
-        stats.get(AttendanceLogicalStatus.ABSENT)),
+      event.children.length,
```
TheSlimvReal, line 35: I changed the computation of the average attendance. Otherwise the information in the `AttendanceCalendarComponent` is rather misleading (e.g. _100% Attended (of 24 Participants)_ even when some of the participants are skipped or not existing  any more (see #792)).

TheSlimvReal (approved):
It works well, I only changed the computation for the average attendance, for details see my comment

src/app/child-dev-project/attendance/model/calculate-average-event-attendance.ts
sleidig: adapted the participant count displayed there

sleidig:
closed in favor of cleaned up and rebased version in PR #798

PR #798, opened 2021-04-28
Attendance patches

sleidig:
see commit messages (please do not squash)

### DIFF
--- a/src/app/child-dev-project/notes/model/note.ts
+++ b/src/app/child-dev-project/notes/model/note.ts
@@ -149,6 +149,9 @@ export class Note extends Entity {
       attendance = new EventAttendance();
       this.childrenAttendance.set(childId, attendance);
     }
+    if (!(attendance instanceof EventAttendance)) {
+      attendance = Object.assign(new EventAttendance(), attendance);
+    }
     return attendance;
   }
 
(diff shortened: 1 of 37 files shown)


src/app/child-dev-project/notes/model/note.ts
```diff
@@ -149,6 +149,9 @@ export class Note extends Entity {
       attendance = new EventAttendance();
       this.childrenAttendance.set(childId, attendance);
     }
+    if (!(attendance instanceof EventAttendance)) {
+      attendance = Object.assign(new EventAttendance(), attendance);
```
TheSlimvReal, line 153: This creates a bug. After recording the attendance once, it is not possible to change the attendance in the `NoteDetailsComponent` or the `RollCallComoponent`. I think because of the `Object.assign` call, the attendance object is a different one than in the `childrenAttendance` map and therefore editing this new object is not persisted in the entity.

TheSlimvReal (changes requested):
Looks good but the latest change added a bug with the attendance

TheSlimvReal (commented):
Some questions for clarification

TheSlimvReal (approved):

Issue #794, opened 2021-04-27
Make entity-select available in FormComponent (configurable forms)


sleidig:
The new entity-select component (#728) offers a powerful way to select entities from a combobox for any entity type. It can not be selected as a form field from config for the FormComponent yet, however. It would be good to make this available not only when writing custom components but also when simply configuring a normal form.

PR #808, opened 2021-05-04
Configurable forms

Schottkyc137:
see issue: #794 

### Visible/Frontend Changes
- [x] `assignedTo`-field now directly visible in a recurring activity and no longer inside participants-section
- [x] removed spacing between label and chips in Entity-select module when disabled  

- [x] `EntitySelect` can now be used inside the generic form-component


src/app/core/entity-components/entity-select/entity-select/entity-select.component.ts
```diff
@@ -56,6 +56,7 @@ export class EntitySelectComponent<E extends Entity> implements OnChanges {
    * @param sel The initial selection
    */
   @Input() set selection(sel: (string | E)[]) {
+    console.log(sel);
```
sleidig, line 59:
```suggestion
```

src/app/core/view/dynamic-components-map.ts
```diff
@@ -28,6 +28,11 @@ import { DashboardShortcutWidgetComponent } from "../dashboard-shortcut-widget/d
 import { UsersBlockComponent } from "../user/users-block/users-block.component";
 import { UserListComponent } from "../admin/user-list/user-list.component";
 import { HistoricalDataComponent } from "../../features/historical-data/historical-data/historical-data.component";
+import { SchoolBlockComponent } from "../../child-dev-project/schools/school-block/school-block.component";
+
+export function isRegisteredDynamicComponent(dynamicComponent: string) {
```
sleidig, line 33: Is this really adding much value, just being an alias?

src/app/core/entity-components/entity-select/entity-select.module.ts
```diff
@@ -22,6 +23,7 @@ import { MatTooltipModule } from "@angular/material/tooltip";
     MatInputModule,
     ReactiveFormsModule,
     MatTooltipModule,
+    ViewModule,
```
sleidig, line 26: it seems you are not using anything from this here, only in the FormEntitySelectComponent of the other module?

src/app/core/entity-components/entity-details/form-entity-select/form-entity-select.component.ts
```diff
@@ -0,0 +1,86 @@
...
+
+/**
+ * A view that shows an entity-select
+ * optimized so that it can be used in a configurable forms-component
+ */
+@Component({
+  selector: "app-form-entity-select",
+  templateUrl: "./form-entity-select.component.html",
+})
+export class FormEntitySelectComponent<E extends Entity> {
```
sleidig, line 16: Can we merge this `FormEntitySelectComponent` with the `EntitySelectComponent` implementation? Apart from (only slightly) different inputs the functionality they offer seems to be the same and this one would be more convenient to use everything, I think.

Schottkyc137: I chose not to, though I am not sure which way is better.
The reason I chose to make two different components instead of one is due to the single-responsibility principle. It felt wrong to add more inputs to the `EntitySelectComponent` since it's overloaded with quite a few already. The `FormEntitySelectComponents` only makes the inputs visible that are required when in a form; all other inputs should not be available (or should be added explicitly).

sleidig: I think the FormEntitySelectComponent is actually easier to use - the usual approach will be to have the Entity*Block as template rather than something custom, for example.  And having two implementations has the danger of getting confused which one to use where.

We can have a look at it together if it really bloats things up too much ... but this class doesn't seem to contain too much logic. Maybe the selection getter could even fit into the FormComponent template.

src/app/core/entity-components/entity-select/entity-select/entity-select.component.ts
```diff
@@ -27,13 +28,33 @@ export type accessorFn<E extends Entity> = (E) => string;
...
+  }
+  entityBlockComponent?: string;
   /**
    * The type of entity to load. This is required and will cause all
    * entities of the given type to be available in the selection
    * and auto-complete
    * @param type The type of the entity
    */
-  @Input() set entityType(type: EntityConstructor<E>) {
+  set entityType(type: EntityConstructor<E>) {
```
TheSlimvReal, line 57: Because its not a `@Input()` anymore , I think this can be just a normal function

TheSlimvReal (changes requested):
Works very well and looks like it made things a little easier.
I just found a small refactoring that would make things a little more clear

TheSlimvReal (approved):

src/app/core/entity-components/entity-select/entity-select/entity-select.component.ts
```diff
@@ -172,6 +143,26 @@ export class EntitySelectComponent<E extends Entity> implements OnChanges {
...
+      .loadType<E>(type)
+      .then((entities) => {
+        this.allEntities = entities;
+        this.loading.next(false);
+        this.formControl.setValue(null);
+      })
+      .catch((error) => {
+        this.loggingService.warn(error);
+        this.loading.next(false);
+      });
```
sleidig, line 164:
```suggestion
      });
```
There is no expected reason for this to throw an error. If it fails, staying indefinitely in loading mode seems as good as being empty. Errors thrown without catch are logged more explicitly with better stacktrace in sentry so I prefer not to have a catch here.

Issue #700, opened 2021-03-05
School not shown in Children List


sleidig:
**Describe the bug**
The school column in the children list does not show the current school of the child.

This seems to be an issue with the way the `schoolId` property is currently loaded in the ChildrenService.getChildren() method. This is done asynchronuously for better performance but the components don't recognize the change after the schoolId is set.

sleidig:
An approach to solve this could be to change the string properties to BehaviourSubject:

Child.ts
```
  schoolId = new BehaviorSubject<string>("");
  schoolClass = new BehaviorSubject<string>("");
```

and change the ChildrenService to update the schoolId instead of emitting the full array twice:
https://github.com/Aam-Digital/ndb-core/blob/b95457f4c843cd817864c9f20aaa92632340a84d/src/app/child-dev-project/children/children.service.ts#L50-L64
e.g. like this:
```
        loadedChild.schoolClass.next(childCurrentSchoolInfo.schoolClass);
        loadedChild.schoolId.next(childCurrentSchoolInfo.schoolId);
```

Then all the references to `child.schoolId` need to be updated because this is now an Observable<string> rather than string:
`<app-school-block [entityId]="child.schoolId | async">`
or in some cases if you only need the current value (disregarding updates) use `child.schoolId.value`

also see details of BehaviorSubject: https://www.learnrxjs.io/learn-rxjs/subjects/behaviorsubject

sleidig:
can't reproduce anymore. While the way we handle this is not very elegant and foolproof, the functionality seems ok.

Issue #694, opened 2021-03-04
Extend view to load Children and Schools at the same time


TheSlimvReal:
Currently, the `ChildrenService` loads the `ChildSchoolRelation` for each child manually. This could be made more efficient by using [view collations](https://docs.couchdb.org/en/stable/ddocs/views/joins.html#optimization-using-the-power-of-view-collation). 
The idea is to create a view that emits the `childId` for each `Child` object as well as for each `ChildSchoolRelation` object, which does not have an `endDate`. This way a child with all its current `ChildSchoolRelations` are loaded with one call and only the `schoolId` and `schoolClass` have to be extracted from the returned docs.

Issue #257, opened 2019-05-06
Improve overall performance of loading data


codingfabi:
For a overview, what's the performance on the live system currently like? @sleidig can you give us a quick update about that and the satisfaction level of the users?

Furthermore, could we list things that can be done to boost the performance like implimenting search indizes etc.?

-----
**Summary of (possible) performance issues:**
- Building database indices (pouchdb needs much longer for the first query after changing an index compared to later queries)
- loading data from database (especially if we need multiple queries of different entity types to combine more complex objects): #462
- creating/updating Angular components (our long lists are loading very slow, especially if they include a lot of subcomponents. this probably has to do with angular change detection)

-----
**Relevant tools and approaches:**
- do change detection profiling (e.g. with https://chrome.google.com/webstore/detail/angular-devtools/ienfalfjdbdpebioblfackkekamfmbnh)
- switch some components to the "OnPush" change detection strategy
- ...

![image](https://user-images.githubusercontent.com/1682541/126469505-4800b9d4-0bd6-4632-a04c-e6599b870735.png)

sleidig:
There haven't been any complaints about performance. Overall it is usable but some areas are noticeably slow. 

Some roughly stopped response times:
* "Children" (populating the list of all current children): 4-5 seconds
* "Notes" (loading the list of all notes): 3-4 seconds

The bottleneck is _not_ the server, as all this data is loaded from the PouchDB local database. But our naive implementation of the database requests may be an issue and the large number of related data (together with the children we load each child's school and recent attendance records). Also, I have heard that PouchDB is quite slow generally with larger loads.

One larger thing we could consider is implementing a caching system, that keeps for example the array of Child objects in memory rather than always loading them from disk through PouchDB. (or maybe there is even a PouchDB plugin for this?)

Lorsbyjm:
@TheSlimvReal  and @Sepper007  wrote a script to create demo data for the demo database (~500 children with 1500 records).  In the following we will dicuss the resolving performance/loading issues:

**Problem description (children-list):**
Right now we fetch all children with one database call and then load the school and attendance information with one call each. This results in over thousand database calls (500 children * (1+1 school + 1 attendance information).
School call: returns the id of the current school and then load this school in a second call


**Solution 1: Filter in Frontend**
Remove the two extra calls for the school. Therefore load all ChildSchoolRelations and Schools with the Childs. The mapping is then done in the frontend. 
[...]

Issue #693, opened 2021-03-04
Make ChildSchoolRelation part of Child entity


TheSlimvReal:
**Problem**
Currently, we have the `ChildSchoolRelation` objects which link children to schools in a M:N relationship. This is not really best practice for document based databases and also has bad performance (in the `ChildrenService`, a call needs to be made for each child to get the current school).

**Solution**
Make the `ChildSchoolRelation` part of the `Child` entity by adding a `schoolHistroy` property to this class. This property is an array of objects, which contain the ID of the visited school, the class and a start and end date. To make it possible to get all children that are currently visiting a certain school, a view needs to be built which for each child emits the `schooldID` for each entry in `schoolHistory` that does not have an end date.
A data migration for current users needs to be implemented.

**Outlook**
This solution could be implemented very flexible. Which would allow to define any kind of M:N relationship as a property. It could be evaluated whether this can be one with a moderate amount of work.

TheSlimvReal:
The 1:N relationships for the `Aser`, `EducationalMaterial` and `HealthCheck` entities could also be replaced by arrays on the respective `Child` object. Would this be desried?

TheSlimvReal:
The approach has the downside, that more write operations are performed on the `Child` object which might cause more conflicts. Alternatively, to query the children with their schools at the same time, [view collation](https://docs.couchdb.org/en/stable/ddocs/views/joins.html#optimization-using-the-power-of-view-collation) can be used.

TheSlimvReal:
There currently is still some but not much special logic still using the `ChildSchoolRelation`

- `NotesRelatedToEntityComponent` copies some information from an existing `ChildSchoolRelation` if a new `Note` is created for it
- `ChildrenService` indices load relations for a given `Child`/`School`. As data might not be that big in most cases, using the standard `RelatedTimePeriodComponent` without using an index for loading the data.
- `ChildrenService` enriches loaded `Child` entities with current school information. If the data lives on the `Child` entity, this enrichment is not needed anymore and data can be shown directly.

PR #812, opened 2021-05-10
Performance testing and improvements

TheSlimvReal:
see issue: #700, #257, #694

This pull requests shows a new helper test-function that allows to compare the execution time of two functions that return an identical result. This should be used in the future to re-implement bad performing functions while ensuring that the functionality is not affected. 

As an example I implemented two improved functions:

1. The `getChildrenImproved` function loads all children and school info at once, using a special index
2. The `getChildrenForSchoolImproved` function loads all children attending the current school with less db-accesses

The tests can be inspected by removing the `x` of `xdescribe` in the `performance-test.spec.ts` file. The tests will fail, but display the performance improvement of the improved functions. Later the tests should be removed and the old function should be replaced by the new one.

### Visible/Frontend Changes
--

- [x] Add helper test-function that compares the execution time and results of two functions
- [x] Implement index and function that loads children with school info at once and faster
- [x] Implement function that loads children currently attending a school faster


src/app/child-dev-project/children/children.service.ts
```diff
@@ -53,55 +53,37 @@ export class ChildrenService {
         const childCurrentSchoolInfo = await this.getCurrentSchoolInfoForChild(
           loadedChild.getId()
         );
-        await this.migrateToNewChildSchoolRelationModel(
-          loadedChild,
-          childCurrentSchoolInfo
-        );
-        loadedChild.schoolClass = childCurrentSchoolInfo.schoolClass;
-        loadedChild.schoolId = childCurrentSchoolInfo.schoolId;
+        loadedChild.schoolClass = childCurrentSchoolInfo.schoolClass || "";
```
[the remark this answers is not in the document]
TheSlimvReal: This was changed so both functions return  the **exact** same results. If we decide to change the current function we can decide on one behavior.

src/app/utils/utils.ts
```diff
@@ -29,17 +29,17 @@ export function getParentUrl(router: Router): string {
  * @param array A simple array to be grouped.
  * @param propertyToGroupBy The key of the property in the elements by whose value the result is grouped.
  */
-export function groupBy<T>(
-  array: T[],
-  propertyToGroupBy: keyof T
-): Map<string, T[]> {
+export function groupBy<ENTITY, PROPERTY extends keyof ENTITY>(
+  array: ENTITY[],
+  propertyToGroupBy: PROPERTY
```
sleidig, line 34: Why does PROPERTY need to be a generic var itself? what is the difference to the previous `keyof T` as type of propertyToGroupBy?

sleidig (commented):
For larger datasets (>1000 children) and slower CPUs it is taking very long to load the initial children list with the new method:
old: ~7 seconds
new: >30 seconds

while this may not be the most common setup we probably should not build on an approach that will become a bottleneck in the future. Maybe we need to directly go with #693

src/app/utils/utils.ts
TheSlimvReal: This allows to reference the correct (same) type for the return value. Because `PROPERTY` always refers to the property of the input value, whereas `keyof ENTITY` could be a different one than the input property.
E.g. when using `groubBy(<array of notes>, "date") the return type will be `Map<Date, Note[]>`.

TheSlimvReal:
> 
> 
> For larger datasets (>1000 children) and slower CPUs it is taking very long to load the initial children list with the new method:
> old: ~7 seconds
> new: >30 seconds
> 
> while this may not be the most common setup we probably should not build on an approach that will become a bottleneck in the future. Maybe we need to directly go with #693

I guess this is due to the fact that the index first have to be built, before the request can be made. So yes I agree

src/app/utils/performance-tests.spec.ts
```diff
@@ -17,85 +17,149 @@
...
+  expect(improvedResult).toEqual(
+    currentResult,
+    "current " +
+      JSON.stringify(currentResult) +
+      " improved " +
+      JSON.stringify(improvedResult)
+  );
+  return currentDuration - improvedDuration;
+}
+
```
sleidig, line 154: @TheSlimvReal  let's take this improved test setup into master and discard the other parts for now?

This PR is becoming more and more left behind and had some performance problems, so the load changes are not mergable at the moment.

TheSlimvReal:
Closed in favor of #958

Issue #761, opened 2021-04-13
"School" filter not available in Children list on first view


sleidig:
**Describe the bug**
The "Schools" filter for the Children List is sometimes not displayed at all.

**To Reproduce**
1. Use the app in "local" or "synced" mode (i.e. not just with the mocked in-memory database) set in your config.json
2. (re)load the app and go to "Children"
3. (!) The "School" filter element is not displayed
4. Navigate to another page within the app and then again to the "Children" list
5. now the "school" filter is available as expected

PR #816, opened 2021-05-12
Prebuilt filter fix

TheSlimvReal:
The prebuilt filters directly work on the config and currently this changes are also exported when downloading the config. This has some bad sideeffects when using this config again. Instead the exported config is now directly taken from the database.

### Visible/Frontend Changes
- [x] Prebuilt schools-filter is always present (not only after second entry of the page #761

- [x] Loading config from database before exporting
- [x] Retrigger `ngOnChanges` of entity list component after schools-filter is built

### DIFF
--- a/src/app/core/config/config.service.spec.ts
+++ b/src/app/core/config/config.service.spec.ts
@@ -2,6 +2,7 @@ import { fakeAsync, TestBed, tick } from "@angular/core/testing";
 import { ConfigService } from "./config.service";
 import { EntityMapperService } from "../entity/entity-mapper.service";
 import { Config } from "./config";
+import { defaultJsonConfig } from "./config-fix";
 
 describe("ConfigService", () => {
   let service: ConfigService;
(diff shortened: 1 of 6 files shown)


src/app/core/config/config.service.spec.ts
```diff
@@ -30,13 +31,15 @@ describe("ConfigService", () => {
   }));
 
   it("should use the default config when none is loaded", fakeAsync(() => {
-    const configBefore = service.getAllConfigs("");
-    entityMapper.load.and.throwError("No config found");
+    const defaultConfig = Object.keys(defaultJsonConfig).map((key) => {
+      defaultJsonConfig[key]._id = key;
```
sleidig, line 35: why should there be these _id on each config item? They are not present on the config we usually have so far.

sleidig (commented):
fixes the bug :+1: 

I don't understand the additional `_id`s on the config export, however.

src/app/core/config/config.service.spec.ts
TheSlimvReal: This is due to the fact that we set the `_id` field when using `getAllConfigs` [here](https://github.com/Aam-Digital/ndb-core/blob/master/src/app/core/config/config.service.ts#L61).
In this case I *fake* the side effect of the `getAllConfigs` function

sleidig: ok, that export of config previously already contained the _id properties. nevermind

sleidig (approved):

PR #819, opened 2021-05-17
Add Photo Datatype

TheSlimvReal:
The current workflow for loading photos of children is rather difficult to follow and included multiple attribute. This got changed to a single attribute. 

The changes are necessary in order to create a better encapsulation of the form component (currently changing the photo file includes changes to **two** attributes which is a problem). It should also open up the possibility to add photos to other entities.

The open question about integration a cloud file service remains.

### Visible/Frontend Changes
- [x] Add migration button for old photo paths

- [x] Replace the `photoPath` and `photo` attributes of the child entity with a combined `photo` attribute
- [x] Replace the `load-child-photo` with a `photo` datatype
- [x] Add migration script for old child entities
- [x] Register the `photo-datatype` **before** the demo data is generated in order to correctly save the values
- [x] Add migration service for photo paths


src/app/child-dev-project/children/child-photo-service/children-migration.service.ts
```diff
@@ -0,0 +1,26 @@
...
+import { Database } from "../../../core/database/database";
+import { Child } from "../model/child";
+
+@Injectable({
+  providedIn: "root",
+})
+export class ChildrenMigrationService {
+  constructor(private database: Database) {}
+
+  async migratePhotoFormat(): Promise<void> {
```
sleidig, line 11: as this is only a simple renaming: shall we leave the old name to avoid migration? This can potentially create a number of conflicts if not all users' local data have been synced just before the migration is executed

src/app/child-dev-project/children/child-photo-service/datatype-photo.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+    const entity = new TestEntity(id);
+    entity.photo = { path: "12345", photo: new BehaviorSubject<SafeUrl>(null) };
+
+    const rawData = entitySchemaService.transformEntityToDatabaseFormat(entity);
+    expect(rawData.photo).toEqual(entity.photo.path);
+  });
+
+  it("should set the default photo when loading", () => {
+    class TestEntity extends Entity {
+      @DatabaseField({ dataType: "photo", defaultValue: true })
```
sleidig, line 57: the way it's used in `Child` class, I would expect the `defaultValue` to become the value, not just trigger some processing in this case?

src/app/child-dev-project/children/child-photo-service/datatype-photo.ts
```diff
@@ -20,31 +20,42 @@ import { ChildPhotoService } from "./child-photo.service";
...
-  public readonly name = "load-child-photo";
+export class PhotoDatatype implements EntitySchemaDatatype {
+  public readonly name = "photo";
 
-  constructor(private childPhotoService: ChildPhotoService) {}
-
-  public transformToDatabaseFormat(value) {
-    return undefined;
+  public transformToDatabaseFormat(value: Photo, schema: EntitySchemaField) {
+    if (value.path === schema.defaultValue) {
```
sleidig, line 34: we could delete the old `photoFile` property here during save and thereby gradually migrate documents, whenever a user updates them

src/app/child-dev-project/children/child-photo-service/datatype-photo.ts
```diff
@@ -20,31 +20,42 @@ import { ChildPhotoService } from "./child-photo.service";
...
+      parent["photoFile"].trim() !== ""
+    ) {
+      value = parent["photoFile"];
     }
-
-    return this.childPhotoService.getImageAsyncObservable(childDummy);
+    return {
+      path: value,
+      photo: new BehaviorSubject<SafeUrl>(
+        ChildPhotoService.getImageFromAssets(value)
```
sleidig, line 57:
```suggestion
        // reactivate the integration of cloud file loading here after testing and performance improvements
        ChildPhotoService.getImageFromAssets(value)
```

src/app/child-dev-project/children/child-photo-service/photo.ts
```diff
@@ -0,0 +1,7 @@
+import { BehaviorSubject } from "rxjs";
+import { SafeUrl } from "@angular/platform-browser";
+
+export interface Photo {
+  path: string;
+  photo: BehaviorSubject<SafeUrl>;
+}
```
sleidig, line 7:
```suggestion
/**
 * simple object managing file path as well as full usable image path/data used in template
 */
export interface Photo {
  /** relative file path set and saved */
  path: string;
  /** observable containing actual full image path or binary data to be used as img src */
  photo: BehaviorSubject<SafeUrl>;
}
```

src/app/child-dev-project/children/child-photo-service/children-migration.service.ts
TheSlimvReal: Yes the migration is not totally necessary, but without it we have to leave the fallback in the photo datatype for ever because we never know if all of the organizations have the new format.

src/app/child-dev-project/children/child-photo-service/datatype-photo.ts
TheSlimvReal: Deleting is not necessary: The old `photoFile` attribute does not have a schema anymore and is therefore not saved to the database.

src/app/core/admin/services/child-photo-update.service.ts
```diff
@@ -42,7 +42,8 @@ export class ChildPhotoUpdateService {
       Child.generatePhotoPath(filename)
     );
     if (fileExists) {
-      child.photoFile = filename;
+      const currentPhoto = child.photo;
+      child.photo = { path: filename, photo: currentPhoto?.photo };
```
sleidig, line 46: why should `photo.photo` still be the "old" photo rather than the new filename here?

TheSlimvReal: To be honest, I dont really know what this function does and when it is supposed to be used but what I have done is equivalent to the current behavior where only the filename is updated but the `BehaviorSubject` is not changed.

sleidig (changes requested):
For some reason the PhotoDatatype transformation is pushing the BehaviourSubject rather than the path into the getImageFromAssets and creates errors in the console. I am still trying to figure out what's happening there ...

src/app/child-dev-project/children/child-photo-service/datatype-photo.spec.ts
TheSlimvReal: I changed it now to `defaultValue = ""` in both cases. The default value is really tricky here because no matter which value is set at the beginning `assets/child-photos/` will be appended. This means setting `"default.png"` as default value will end up as `"assets/child-photos/default.png"`. So currently this whole component is still a little weird because it is very dependent on the Child entity. Once we want to actually allow this for other entities as well, this will have to be further adapted.

sleidig (approved):

Issue #818, opened 2021-05-14
Cant delete dates in form component


TheSlimvReal:
**Describe the bug**
When removing an existing date from the form component, the date will be back once the save button is clicked.

**To Reproduce**
1. Go to `Children`
2. Click on a child
3. Click on `Edit`
4. Delete the entries of `Date of birth` and/or `Admission Date`
5. Click on `Save`
6. See the date being reset to the value before editing

**Expected behavior**
The date value should be empty after saving and also when visiting the page again (it should be saved to the database).

**Additional context**
The problem occurs due to [this null check in the form component](https://github.com/Aam-Digital/ndb-core/blob/master/src/app/core/entity-components/entity-details/form/form.component.ts#L131). It should be checked whether this null check is necessary and if possible removed or changed to make it possible to remove dates.

TheSlimvReal:
Fixed with #764

Issue #764, opened 2021-04-14
Connect datatypes with components


TheSlimvReal:
Each property of a entity is marked with a datatype in order to correctly save it to the database. This information is stored in a `schema` attribute of the class.

Currently, when configuring a list-view or a details-view, for each property, the component to display this property has to be named. Instead this information could be derived from the schema of the class. This should make the configuration file much smaller and less error-prone.
E.g. 
Instead of writing
```
{
  "input": "text",
  "id": "name",
  "placeholder": "Name"
}
```
Providing `name` is sufficient.

To achieve this, a connection between datatypes and components for entity-list and entity-details (form-component) needs to be defined. 

In same cases (attendance, age) it has to be evaluated, whether a new datatype should be added or if the component information is still added through the config.

PR #821, opened 2021-05-18
Remove falsy attributes from DB

RichardNutt:
see issue: #818

I suspect, that the null check was originally implemented to stop the creation of empty/null valued properties in the DB-json-files. As far as I can tell the easiest way to accomplish this would be by checking if a given value is falsy (keep 0 in case this makes sense somewhere) and remove them in that case.

- [x] remove checks from form component
- [x] add checks to entity schema

### DIFF
--- a/src/app/core/entity-components/entity-details/form/form.component.ts
+++ b/src/app/core/entity-components/entity-details/form/form.component.ts
@@ -128,8 +128,10 @@ export class FormComponent implements OnInitDynamicComponent, OnInit {
   private assignFormValuesToEntity(entity: Entity, form: FormGroup) {
     Object.keys(form.controls).forEach((key) => {
       const value = form.get(key).value;
-      if (value !== null) {
+      if (value || value === 0) {
         entity[key] = value;
+      } else {
+        delete entity[key];
       }
     });
   }
--- a/src/app/core/entity-components/entity-details/form/form.component.ts
+++ b/src/app/core/entity-components/entity-details/form/form.component.ts
@@ -128,11 +128,7 @@ export class FormComponent implements OnInitDynamicComponent, OnInit {
   private assignFormValuesToEntity(entity: Entity, form: FormGroup) {
     Object.keys(form.controls).forEach((key) => {
       const value = form.get(key).value;
-      if (value || value === 0) {
-        entity[key] = value;
-      } else {
-        delete entity[key];
-      }
+      entity[key] = value;
     });
   }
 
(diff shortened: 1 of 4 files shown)


sleidig:
@TheSlimvReal how does this relate to the previous PR #763 ?

TheSlimvReal:
> 
> 
> @TheSlimvReal how does this relate to the previous PR #763 ?

Not much. PR #763 solved a problem where the DB transformation failed in case a date is `null`. The issue for this PR was in the `form.component` not correctly saving a empty date.

src/app/core/entity-components/entity-details/form/form.component.ts
```diff
@@ -128,8 +128,10 @@ export class FormComponent implements OnInitDynamicComponent, OnInit {
   private assignFormValuesToEntity(entity: Entity, form: FormGroup) {
     Object.keys(form.controls).forEach((key) => {
       const value = form.get(key).value;
-      if (value !== null) {
+      if (value || value === 0) {
```
sleidig, line 131: checkboxes setting a value `false` should also be allowed. This check seems like something that will make us stumble in the future again.

Maybe we should allow any value to be set (at least if the form field has been changed) and let the schema datatypes handle special requirements (e.g. if null should not be written to the database)?

sleidig: btw.: good investigation of the problem, @RichardNutt !

RichardNutt: @sleidig I agree that this check seems like a bad idea. I just realized that we are already filtering undefined entries
https://github.com/Aam-Digital/ndb-core/blob/80538b02af7f8bd8ba0b072b98088e2f3bb82dcf/src/app/core/entity/schema/entity-schema.service.ts#L165-L172

Checking for null etc. in `ndb-core/src/app/core/entity/schema-datatypes/` seems better (and perhaps print some info to the console if values are omitted to make debugging easier). Should we just remove null/empty string/etc to keep the DB smaller?

TheSlimvReal: For now, there is not really a reason to distinguish between `null` and `undefined` in the database and I thing also empty strings or arrays. I can just imagine that this might interfere with the default values set in the classes when creating a new entity, e.g. in the child entity `status` is set to `""` on default. When the database now has `status === undefined` or `status === null` the `Object.assign` call in the schema transformation will overwrite the default value. If however the property is not present in the database, the value will not be overwritten. 

So I think it would make sense to delete all _empty_ values (`null`, `undefined`, `[]`, `""`) before writing a entity to the database, but make sure to not throw away zeros or `false` values.

sleidig:
Although we didn't merge the PR, the fix has been implemented as part of PR #764.
Thanks for figuring out the solution, @RichardNutt !

Issue #757, opened 2021-04-09
Better sorting for numbers in string properties


sleidig:
Some properties like schoolClass in the `PreviousSchoolsComponent` or projectNumber in the `ChildrenListComponent` are of type `string` but often contain numbers as well. These fields require the flexibility of strings to allow users adding a prefix or suffix or something completely different. At the same time, the alphabetical ordering feels "wrong":

- Actual string order: `["1", "10", "2", "3", "X"]`
- Expected order: `["1", "2", "3", "10", "X"]`

We had implemented a custom sort function for this, which (probably accidentally) was removed from active code:
https://github.com/Aam-Digital/ndb-core/blob/1ece1020023e0a252817882e153920c31ee2036d/src/app/child-dev-project/children/model/child.ts#L105-L121

Let's bring this back for the above mentioned columns (or maybe even for all string columns?) that can be sorted in tables.

sleidig:
@Schottkyc137 , @TheSlimvReal  can we assign columns to this sorting through config? (or do it for every string)
Project numbers should also be sorted with the numerical logic.

TheSlimvReal:
Currently, all strings are converted to a number if possible ([here](src/app/core/entity-components/entity-subrecord/entity-subrecord/sorting-accessor.ts) is the the exact transformation) . This means if project numbers are numbers, it will already work. if project numbers also contain string literals, it wont work.

TheSlimvReal:
The expected order is currently achieved in the app. Therefore, I consider this issue closed.

PR #839, opened 2021-06-05
Implement numeric sorting for strings

Schottkyc137:
see issue: #757 

### Visible/Frontend Changes
- [x] All values the `EntityListComponent` will sort according to their numeric values (e.g. "2" < "10")


TheSlimvReal (commented):
Wouldn't it be easier to just extends the existing [sorting-accessor](https://github.com/Aam-Digital/ndb-core/blob/master/src/app/core/entity-components/entity-list/sorting-accessor.ts) with this logic

Schottkyc137:
> Wouldn't it be easier to just extends the existing [sorting-accessor](https://github.com/Aam-Digital/ndb-core/blob/master/src/app/core/entity-components/entity-list/sorting-accessor.ts) with this logic

The sorting-accessor only gets the value that should be sorted. The current approach is a copy of (the source)[https://github.com/angular/components/blob/d84612eac7a1f0e75948e4a9749e929ca387c6e0/src/lib/table/table-data-source.ts#L115] with just a small modification

TheSlimvReal (approved):

Issue #820, opened 2021-05-18
Search bar is not working


sleidig:
Somehow search is not showing any results anymore.

Maybe also related to #753

PR #840, opened 2021-06-06
Modernize Search Bar

Schottkyc137:
see issue: #820 

### Visible/Frontend Changes
- [x] Search bar works again

- [x] Modernize `SearchBarComponent`


src/app/child-dev-project/notes/notes-of-child/notes-of-child.component.ts
```diff
@@ -49,9 +51,10 @@ export class NotesOfChildComponent
       visibleFrom: "md",
     },
     {
-      name: "author",
+      name: "authors",
       label: "SW",
-      inputType: ColumnDescriptionInputType.TEXT,
+      inputType: ColumnDescriptionInputType.READONLY,
+      asyncValueFunction: (note: Note) => this.getAuthorNames(note.authors),
```
TheSlimvReal, line 57: This approach is rather ineffective, for every entry, all the users have to be loaded and can than only be displayed. A faster approach would be to just load all of them once and then filter out the once needed in the value function.
However, the entity subrecord component is getting refactored and also this problem will be solved in #764.

TheSlimvReal (changes requested):
Very good that you took on this issue and also refactored some of the old components. 
It would be good if you also mention additional small fixes in the PR like the changes you made for the EntitySubrecord. This makes it easier to follow the modified files.

src/app/core/sync-status/background-processing-indicator/background-processing-indicator.component.ts
```diff
@@ -22,7 +22,7 @@ export class BackgroundProcessingIndicatorComponent implements OnChanges {
   @Input() backgroundProcesses: BackgroundProcessState[] = [];
 
   /** whether processes of with the same title shall be summarized into one line */
-  @Input() summarize: boolean = true;
+  @Input() summarize: boolean = false;
```
TheSlimvReal, line 25: I dont think this should go into the production mode because the information is not necessary for the final user

Schottkyc137: Funny side note: This was the actual reason that the search bar wasn't working (resp. the `summarizeProcesses` function) since it deleted something from the array which caused the `queryIndexRaw` function to wait forever)

src/app/core/ui/search/search.component.ts
```diff
@@ -161,90 +119,35 @@ export class SearchComponent implements OnInit {
...
-      resultEntity = new School(doc.entityId);
-    }
-
-    if (resultEntity) {
-      this.entitySchemaService.loadDataIntoEntity(resultEntity, doc);
-      return resultEntity;
-    } else {
-      return null;
-    }
+    return fromPromise(this.indexingService.createIndex(designDoc));
```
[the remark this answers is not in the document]
Schottkyc137: It's now used in the latest commit

src/app/core/sync-status/background-processing-indicator/background-processing-indicator.component.ts
```diff
@@ -1,81 +1,85 @@
...
+      startWith(0),
+      map((tc) => tc === 0)
+    );
+    this.taskCounterObservable
+      .pipe(untilDestroyed(this))
+      .subscribe((amount) => {
+        if (amount === 0) {
+          this.taskListDropdownTrigger.closeMenu();
+        } else {
+          this.taskListDropdownTrigger.openMenu();
```
TheSlimvReal, line 48: Personally I dont really like this behavior because it means that every time something is saved the popup opens. Most oft the users are probably quite irritated by that. I think just showing the sync icon with the potential to open it is sufficient

sleidig:
Thanks a lot for working on these improvements! Can you split these into separate PRs please? Then we can test (and discuss) the background process indicator separately, for example. That has been needing thorough testing for all kinds of edge cases and weird behaviour for a while already ...

TheSlimvReal (changes requested):
The code looks all fine, but I still see all the individual processes in the sync component. I think its because they are only combined if they have the same title but they actually each have different ones.

background_processes

Schottkyc137:
> The code looks all fine, but I still see all the individual processes in the sync component. I think its because they are only combined if they have the same title but they actually each have different ones.
> 
> background_processes

No, I simply forgot to set `summarize` to true 😑

TheSlimvReal:
This has been moved to #877

PR #842, opened 2021-06-07
Implement entity block component

Schottkyc137:
- [x] Implement a standalone generic `EntityBlockComponent`

The `EntityBlockComponent` looks like a registered block component when one is available (currently only the school block or child block). If none is available, the block component will display the entity's name, or its ID, if it has no name.


TheSlimvReal (commented):
Good idea. But I would leave this for now because a similar and a little more integrated version will be part of #764. That branch should be merged this week and then we can see if there is room for improvement

TheSlimvReal:
This is already part of #747 and will therefore be closed

PR #843, opened 2021-06-07
refactor sync status

### LINKED ISSUE #820: Search bar is not working

Schottkyc137:
See issue #820 

Modernizes the architecture for the sync status by taking advantage of `rxJs` and asynchronous processing of the data.
This fixes #820 by creating a new `BackgroundProcessState` instead of altering the given state which in turn alters internal data from the `DatabaseIndexingService` causing certain indices to look like they are not registered


TheSlimvReal (approved):
Works very well, thank you

Issue #701, opened 2021-03-05
Simplify and refactor the SessionService


sleidig:
The SessionService interface and the SyncedSessionService should be further simplified:

- [ ] `sync` should not be part of the public interface. All sync related logic should probably be handled by the SyncedSessionService internally?
- [x] the old `LocalSession` implementation used internally in the SyncedSessionService should be replaced and use the `NewLocalSessionService` instead, which can also be used as a stand-alone service if only local session is required
- [x] `StateHandler` implementation is rather similar to Observable. It would be easier to work with this if the widely used Observable interface replaces the custom StateHandler. Maybe we can simply use BehaviourSubject instead.
- [x] database.provider probably doesn't have to depend

sleidig:
@TheSlimvReal should we close this one - or are there any other ideas to simplify the code?

TheSlimvReal:
Yes, you are right. I think we did all we had in mind when doing the session/startup refactoring in #1259

PR #847, opened 2021-06-08
Replace state handler with Behavior Subject

Schottkyc137:
see issue: #701

- [x] Replace `StateHandler` with `BehaviorSubject`
- [x] Modernize some tests


src/app/core/session/session-service/synced-session.service.ts
```diff
@@ -38,6 +41,7 @@ import { LoggingService } from "../../logging/logging.service";
  * [Session Handling, Authentication & Synchronisation]{@link /additional-documentation/concepts/session-and-authentication-system.html}
  */
 @Injectable()
+@UntilDestroy()
 export class SyncedSessionService extends SessionService {
```
TheSlimvReal, line 45: Some changes in this file can be undone if the `isInitial` function of the local session keeps returning a promise

TheSlimvReal (changes requested):
Very good that you refactored these services, makes it a lot more understandable when the native strategies are used. Anyway, at some points I think its better to stay with the promise just for better readability and easier understanding. It would be great if you could undo these changes.

TheSlimvReal:
The changes are now part of #947. Thanks @Schottkyc137 for taking this on and sorry that it took so long to merge.

PR #849, opened 2021-06-09
Config migration

TheSlimvReal:
An automated migration for the old config format.
The new config is very compatible with the old format, only the fields and some input types have to be renamed

### Visible/Frontend Changes
- [x] Migrate config button in admin component

- [x] Migration service that applies the migration rules


sleidig (approved):
Looks clean and seems to work well :100:

Issue #846, opened 2021-06-08
UX: Highlight current menu entry


sleidig:
Guide users and show where they currently are by highlighting the selected menu entry. Ideally styling it similar to this draft to give the feel of a selected tab.

For details pages, the parent entry should still be highlighted (i.e. when viewing an individual child's details page).

![image](https://user-images.githubusercontent.com/1682541/121144920-b8ddc000-c83e-11eb-8592-3f9d23cf1611.png)

TheSlimvReal:
fixed with #1086

Issue #845, opened 2021-06-08
UX improvements for list view


sleidig:
The design of the list views (e.g. the "Children" list) has some aspects that are confusing to users. The following redesign is suggested:
![image](https://user-images.githubusercontent.com/1682541/121142663-8632c800-c83c-11eb-8d82-34a077ba2962.png)

- [x] **simplify top section:** remove the elevation and reorder the top filter/action section a bit. Move the "Add" button, move the "export" button into a three-dot menu to hide this less used option from cluttering the interface.
- [x] **table column groups through "tabs"**: Try to style the controls to display different column groups as tabs. This way it is more intuitive what the controls will do and no danger of confusing them with the filters.
- [x] **Less columns for demo**: some fields are extremely specific. The default config set in code and used for the demo should just offer a few fields for each tab as illustration and remove some details. (e.g. remove some health checkups and document status fields completely from the entity)
- [x] **Health Tab:** the columns can be reordered in a more informative way. For example. Name / Gender / Age / DoB / BMI / the rest of info. Obtain the basic info first and the specific after giving clarity.

TheSlimvReal:
Implemented with #1086

PR #851, opened 2021-06-10
Highlight menu entry

Schottkyc137:
see issue: #846

### Visible/Frontend Changes
- [x] The currently selected menu entry is highlighted

### DIFF
--- a/src/app/core/navigation/navigation/navigation.component.html
+++ b/src/app/core/navigation/navigation/navigation.component.html
@@ -20,13 +20,14 @@
     angulartics2On="click"
     angularticsCategory="Navigation"
     angularticsAction="app_navigation_link_click"
-    [angularticsLabel]="item.label"
+    [angularticsLabel]="item.name"
     *ngFor="let item of menuItems"
     [routerLink]="[item.link]"
+    [class]="activeElement === item.name ? 'highlighted' : ''"
   >
     <a
       ><mat-icon class="nav-icon" fontIcon="fa-{{ item.icon }}"></mat-icon
-      >{{ item.label }}</a
+      >{{ item.name }}</a
     >
     <mat-divider></mat-divider>
   </mat-list-item>
--- a/src/app/core/navigation/navigation/navigation.component.ts
+++ b/src/app/core/navigation/navigation/navigation.component.ts
@@ -23,8 +23,8 @@ import { RouterService } from "../../view/dynamic-routing/router.service";
 import { ViewConfig } from "../../view/dynamic-routing/view-config.interface";
 import { ConfigService } from "../../config/config.service";
 import { NavigationEnd, Router } from "@angular/router";
-import { UntilDestroy, untilDestroyed } from "@ngneat/until-destroy";
-import { filter, map, startWith, tap } from "rxjs/operators";
+import { UntilDestroy } from "@ngneat/until-destroy";
+import { filter, map, startWith } from "rxjs/operators";
 
 /**
  * Main app menu listing.
--- a/src/app/core/navigation/navigation/navigation.component.ts
+++ b/src/app/core/navigation/navigation/navigation.component.ts
@@ -65,12 +65,17 @@ export class NavigationComponent {
         } else if (items.length === 1) {
           this.activeElement = items[0].name;
         } else {
-          // If there are multiple matches (multiple elements start with the give link),
-          // use the element where the length is bigger. For example:
-          // '/attendance' matches both manage attendance ('/attendance') as well as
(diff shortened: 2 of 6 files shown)


src/app/core/navigation/navigation/navigation.component.ts
```diff
@@ -31,44 +34,66 @@ import { ConfigService } from "../../config/config.service";
...
+          this.activeElement = items[0].name;
+        } else {
+          // If there are multiple matches (multiple elements start with the give link),
+          // use the element where the length is bigger. For example:
+          // '/attendance' matches both manage attendance ('/attendance') as well as
+          // '/attendance/add/day' (add day attendance).
+          // If 'attendance/add/day' was the active link, then only this link
+          // would have been filtered
+          this.activeElement = items.reduce((i1, i2) =>
+            i1.link.length > i2.link.length ? i1 : i2
```
TheSlimvReal, line 75: I think the _shorter_ element should be used? In the example with `/attendance`, the `/attendance` menu item should be highlighted and not the `/attendance/add/day`.

src/app/core/navigation/navigation/navigation.component.ts
```diff
@@ -31,44 +34,66 @@ import { ConfigService } from "../../config/config.service";
...
-      }
-    }
+    this.menuItems = config.items.filter((item) => this.hasPermissionFor(item));
   }
 
   /**
    * Check whether the user has the required rights
    */
-  private checkMenuItemPermissions(link: string): boolean {
+  private hasPermissionFor(item: MenuItem): boolean {
```
TheSlimvReal, line 94: using this function (`this.hasPermissionFor(item)`) sounds a little bit like if the function would check if the current component has the required permissions. 
```suggestion
  private currentUserHasPermissionsFor(item: MenuItem): boolean {
```
or
```suggestion
  private currentUserIsPermitted(item: MenuItem): boolean {
```

src/app/core/navigation/navigation/navigation.component.html
```diff
@@ -20,13 +20,14 @@
     angulartics2On="click"
     angularticsCategory="Navigation"
     angularticsAction="app_navigation_link_click"
-    [angularticsLabel]="item.label"
+    [angularticsLabel]="item.name"
     *ngFor="let item of menuItems"
     [routerLink]="[item.link]"
+    [class]="activeElement === item.name ? 'highlighted' : ''"
```
TheSlimvReal, line 26: Did you check whether `[routerLinkActive]="highlighted"` already does the trick? [API Doc](https://angular.io/api/router/RouterLinkActive#routerlinkactive)

TheSlimvReal (changes requested):
Good that you cleaned up the component a little bit. But did you check if the "native" way works? That would make the whole process a little less complex.

sleidig:
The idea between that UX suggestion by Rosa is to give the selected menu item the feel like a tab that connects to the main area to the right, putting that selected menu item at the same elevation and background as the main view area. Can we tune the styling more in that direction?

src/app/core/navigation/navigation/navigation.component.ts
Schottkyc137: This is counterintuitive but correct. The function matches the `url` with the `link`. So suppose the `url` is `/attendance`, `/attendance/add/day` wouldn't be matched (since `attendance` doesn't start with `attendance/add/day`).
The reason to choose this approach (as compared to the approach where the `link` would be matched against the `url`) is that in the latter approach, the url `/child/7` (child -> details: childId = 7) would not match against the link `/child`, since `/child/7` is not a prefix of `/child`. You can have a look at the test cases to verify that this is correct.

Schottkyc137:
> The idea between that UX suggestion by Rosa is to give the selected menu item the feel like a tab that connects to the main area to the right, putting that selected menu item at the same elevation and background as the main view area. Can we tune the styling more in that direction?

We certainly can but the intention of this PR is only to make the tabs on the left selectable and highlight them. A later PR can then deal with the styling (which I believe is more related to #845). Thinking about this I would suggest using another branch though (maybe something like `ux-improvements`?) Where we can merge all related PR's and then finally merge everything into master.

src/app/core/navigation/navigation/navigation.component.ts
TheSlimvReal: Ok I understand it now, this makes sense. Thanks for the explanation.

TheSlimvReal (approved):
The highlighting works very well. I will merge this now into a general `ux-redesign` branch where we collect the redesign changes so we can publish them later as a single update.

Issue #813, opened 2021-05-10
Selecting day without event in attendance-calendar throws error


sleidig:
**Describe the bug**
Selecting day without event in attendance-calendar throws error

**To Reproduce**
1. Go to the details of a Recurring Activity
2. Open the "Events & Attendance" section and click on a date in the calendar that does not have an event
3. the console is showing error: "cannot read property 'children' of undefined"

**Expected behavior**
The UI looks okay, but no error should be thrown and logged.

Maybe we want to offer users the option to create an event for this day?

**Additional context**
https://sentry.io/share/issue/67d222b6b44640cbafdd5e81aa36732c/

PR #854, opened 2021-06-11
Attendance calendar fix

Schottkyc137:
see issue: #813
Also fixes further bugs

### Visible/Frontend Changes
- [x] The user has the option to create a new event when clicking on a field when there was no event before at a selected date
- [x] Events will update instantly when changed in the popover (previously this only happened when the user clicked twice
- [x] When no children attended, they will no longer be displayed as 'excused', but as '0% attended' 

- [x] The obscure error message no longer appears
- [x] unsubscribe from the observable in `EntitySelectComponent` to avoid memory leaks
- [x] use `filter` instead of `skipUntil` in the same component since that reflects better what is going on


src/app/child-dev-project/attendance/attendance-calendar/attendance-calendar.component.ts
```diff
@@ -143,6 +165,16 @@ export class AttendanceCalendarComponent implements OnChanges {
     await this.entityMapper.save(this.selectedEvent);
   }
 
+  newNote(): EventNote {
```
TheSlimvReal, line 168: To create a event for a activity the `createEventForActivity` function of the `AttendanceService` should be used.

src/app/child-dev-project/attendance/attendance-calendar/attendance-calendar.component.html
```diff
@@ -57,6 +57,9 @@ <h3 class="selected-title">
   </div>
 
   <div *ngIf="!selectedEvent" class="no-events-notice">
-    no events on this date
+    <p>no events on this date</p>
+    <button mat-stroked-button (click)="showEventDetails(newNote())">
+      New event here
```
TheSlimvReal, line 62:
```suggestion
      Add new event
```

src/app/child-dev-project/attendance/attendance-calendar/attendance-calendar.component.ts
```diff
@@ -143,6 +165,16 @@ export class AttendanceCalendarComponent implements OnChanges {
...
+    const note = new EventNote();
+    note.date = this.selectedDate.toDate();
+    note.children = this.activity.participants;
+    note.authors = this.activity.assignedTo;
+    note.category = this.activity.type;
+    note.subject = this.activity.title;
+    return note;
+  }
```
TheSlimvReal, line 176: The newly created events are not saved correctly.

To reproduce:

1. Select empty date
2. Create event
3. Add some attendance
4. Click `save`
5. Leave page and open activity again
6. No attendance is visible for the previously selected date

TheSlimvReal (changes requested):
I like the look and feel of the changes, but the events are not correctly saved to the database. Once this is changed, it can be merged.

TheSlimvReal (approved):

Issue #574, opened 2020-12-19
Generalize the edit button of detail forms


sleidig:
The edit button for details forms is already used in a lot of components. it would make sense to move it to a general component, so the styling and behavior is consistent.

_Originally posted by @TheSlimvReal in https://github.com/Aam-Digital/ndb-core/pull/560#discussion_r546124392_

TheSlimvReal:
This changes should move the `edit`, `save` and `cancel` button into a single component which provides the input parameter `entity` which is used to determine whether the current user is allowed to use the button as well as two output even emitters to notify about clicks on the `save` and `cancel` button. 
There might be a problem where the `save`  function runs some additional checks before actually saving the entity which eventually can fail. So maybe instead of using output parameters, it might make more sense to pass the function which should be run, once the save button is clicked, into the component.

sleidig:
I think this has become obsolete as the logic is now already centralized in the `EntityFormComponent` that is used everywhere for these kind of forms since PR #817.

PR #817, opened 2021-05-13
Connect datatypes with components

TheSlimvReal:
see issue: #764 #552 #180 #826 

### Visible/Frontend Changes
--

### Architectural/Backend Changes
- [x] `EntitySchema` holds information how an attribute can be displayed
- [x] `EntitySchema` holds information how an attribute can be edited
- [x] `EntitySchemaDatatype` provides a default component for displaying an data type
- [x] `EntitySchemaDatatype` provides a default component for editing an data type
- [x] `EntityListComponent` automatically fetches column information from schema if no other definition is found 
- [x] `FormComponent` automatically fetches form field information from schema if no other definition is found
- [x] `EntitySubrecord` automatically fetches display and edit information from schema if no other definition is found
- [x] `EntityList` uses `EntitySubrecord`
- [x] Automatically create filters for properties using schema information
- [x] Update documentation
- [x] Automatically create (popup) form for entities -> #834

PR #855, opened 2021-06-12
Implement edit button

Schottkyc137:
see issue: #574

- [x] Implement Generic edit button
- [x] Button is disabled when the user doesn't have the required rights


src/app/child-dev-project/attendance/activity-participants-section/activity-participants-section.component.spec.ts
```diff
@@ -11,9 +13,10 @@ describe("ActivityParticipantsSection", () => {
   beforeEach(
     waitForAsync(() => {
       TestBed.configureTestingModule({
-        imports: [AttendanceModule],
+        imports: [AttendanceModule, AppButtonsModule],
```
TheSlimvReal, line 16: I think the `AppButtonsModule` is imported by the `AttendanceModule` and therefore not needed additionally here

TheSlimvReal (changes requested):
What you implemented looks very good but the issue description did not really catch what we actually wanted. The _edit button_ was referring to the whole component where you click the edit button which is disabled or not but also the `save` and `cancel` buttons which replace the edit button once it is clicked.
I updated the [issue description](https://github.com/Aam-Digital/ndb-core/issues/574#issuecomment-862177001), I hope this makes it more clear.

sleidig:
I think this has become obsolete since PR #817 has made sure that all forms are based on the `EntityFormComponent`. This is now the only (and logical) place for the buttons.

Really sorry @Schottkyc137 to discard your work here! :cry: 
If you added anything beyond the refactoring, please let me know so we can still bring than in.

Issue #806, opened 2021-04-30
Create demo data for HistoricalData


TheSlimvReal:
Currently no historical data is created in the demo mode, but some columns are defined through the columns. Some demo data should be created for each child.

PR #856, opened 2021-06-12
generate demo for historical data

Schottkyc137:
see issue: #806 

### Visible/Frontend Changes
- [x] Generate demo data for a child's observations


TheSlimvReal (approved):
Thanks for taking this on. This will make it much more easy to understand

PR #866, opened 2021-06-28
Connect Datatype Review Improvements

TheSlimvReal:
This PR holds all the changes added after the review on the branches #817, #834 and #849


src/app/child-dev-project/aser/model/aser.spec.ts
```diff
@@ -77,15 +79,19 @@ describe("Aser", () => {
...
     const id = "test1";
     const entity = new Aser(id);
-    entity.english = Aser.ReadingLevels[0];
-    entity.math = Aser.MathLevels[1];
+    entity.english = readingLevels[1];
+    entity.math = readingLevels[2];
 
-    expect(entity.getWarningLevel()).toBe(WarningLevel.WARNING);
+    expect(entity.getWarningLevel()).toBe(
+      warningLevels.find((level) => level.id === "WARNING")
```
sleidig, line 94: As the Warning Level is quite deeply connected to business logic in code, maybe we should wrap this into its own object/enum rather than relying on string filtering

src/app/core/configurable-enum/configurable-enum-datatype/configurable-enum-datatype.ts
```diff
@@ -37,8 +37,9 @@ export class ConfigurableEnumDatatype
       enumId = CONFIGURABLE_ENUM_CONFIG_PREFIX + enumId;
     }
 
+    // The label check is used to also match enums that were automatically created
     return this.configService
       .getConfig<ConfigurableEnumConfig>(enumId)
-      ?.find((option) => option.id === value);
+      ?.find((option) => option.id === value || option.label === value);
```
sleidig, line 43: the label shouldn't match here, this makes the concept confusing and opens ways for bugs if stuff is renamed in a circular way over the years. if enums are auto-created, why not create the "correct" ID?

TheSlimvReal: The enums are auto created based on the configuration of how they are editable (e.g. through the config of entity subrecords). To fully migrate the instances, all the entities where the values are used have to be edited. The adaption here can be seen as a temporary workaround. Running the migration [here](https://github.com/Aam-Digital/ndb-core/pull/866/files#diff-96ec83c71bd1fe9599268c3a2bfc1352a59fae23c828e572497ecadf52cea94aR15) will automatically migrate all entities to the new format and afterwards this second option can be removed.

src/app/core/entity-components/entity-form/entity-form.service.ts
```diff
@@ -56,25 +60,35 @@ export class EntityFormService {
...
       }
     });
     return this.fb.group(formConfig);
   }
 
-  public saveChanges(form: FormGroup, entity: Entity): Promise<Entity> {
+  /**
+   * This function applies the changes of the formGroup to the entity.
+   * If the form is invalid or the entity does not pass validation after applying the changes, an error will be thrown.
+   * In order to not edit the initial entity until the new one is saved, call entity.copy() on the input entity.
```
sleidig, line 77: do the copy here to encapsulate this?

src/app/core/entity-components/entity-form/entity-form/entity-form.component.spec.ts
```diff
@@ -92,21 +92,25 @@ describe("EntityFormComponent", () => {
...
     spyOnProperty(component.form, "invalid").and.returnValue(true);
 
-    return expectAsync(component.save()).toBeRejected();
+    await component.save();
+
+    expect(alertService.addWarning).toHaveBeenCalled();
   });
 
-  it("logs error when saving fails", () => {
+  it("logs error when saving fails", async () => {
```
sleidig, line 105:
```suggestion
  it("shows error when saving fails for valid form", async () => {
```

src/app/child-dev-project/notes/notes-manager/notes-manager.component.ts
```diff
@@ -44,18 +44,25 @@ export class NotesManagerComponent implements OnInit {
...
   notes: Note[] = [];
 
+  private readonly urgentLevel = warningLevels.find(
+    (level) => level.id === "URGENT"
+  );
+  private readonly warningLevel = warningLevels.find(
+    (level) => level.id === "WARNING"
+  );
+
   private statusFS: FilterSelectionOption<Note>[] = [
```
sleidig, line 54: replace with simple configurabe-enum filter for all options?

Issue #868, opened 2021-06-30
Pagination Size remains at 1 if in some filter only 1 element visible


sleidig:
**Describe the bug**
Pagination page size in the list behaves a bit weird, which can feel like data is lost

**To Reproduce**
1. Go to "Notes" list
2. see a long list of notes (e.g. 10 entries if that page size is selected)
2. Filter the list, so that only 1 or 2 entries are visible (e.g. using the "urgent" or the "category" filter)
3. Remove the filter --> expect to see full list of entries again
4. But: list continues to show only 1 or 2 entries
5. Change pagination page size to a larger value manually in order to see the full list (but this is very unintuitive)

It took me quite a while to figure out what's going on even though I know the system well. Page size should not shrink even if less than a full page of entries is available at the moment.

sleidig:
After discussion, the expected behavior should be:

when "All" toggle active:
- switch to filter where more results available --> show all results
- turn "all" toggle off manually --> a fixed page size selected and shown

when "All" toggle inactive:
- switch to filter with less than page size results (e.g. page size: 20; available entries: 15) --> show all results, leave all toggle inactive and keep fixed page size (when showing more results, cap it at 20 again)

PR #881, opened 2021-07-07
Fix paginator all issue

TheSlimvReal:
see issue: #868 

The paginator now works as follows:

- The all button is always enabled
- Enabling the all button will keep the all option enabled for all views on this table
- The other options (10, 20 50) are always available and call always be selected and will be maintained throughout the views of a table
- The all option is not automatically enabled once a page size is selected that is bigger than the amount of items.

### Visible/Frontend Changes
- [x] The paginator defaults to bigger page-sizes if available, instead of keeping the smallest one that was visited

--


sleidig (changes requested):
mostly works great, one edge case bug remains:
1. create a filtered list with exactly 10 results (e.g. a total of 10 inactive children)
2. activate that filter to show these 10 results
3. select the fixed page size = 10 from the paginator dropdown
4. unexpectedly, "all" toggle is now active; expected: page size set to 10, remains at 10 when switching to another filter with more results

sleidig (approved):

Issue #838, opened 2021-06-04
Upgrade to Angular 12 and other latest dependencies


sleidig:
Angular 12 is out and some other dependencies also have new versions and should be updated

TheSlimvReal:
#908 should also be considered

PR #883, opened 2021-07-13
Angular and other dependecies updated

TheSlimvReal:
see issue: #838 

### Visible/Frontend Changes
--

- [x] Upgraded Angular to version 12
- [x] Upgraded all but [rxjs](https://www.npmjs.com/package/rxjs) to the latest version (rxjs 7 is not supported yet [source](https://github.com/angular/angular/issues/41897#issuecomment-848018116))
- [x] Remove [tslib](https://www.npmjs.com/package/tslib) and [reveal](https://www.npmjs.com/package/reveal) from `package.json` because it is not used directly
- [x] Replace [crypto-js](https://www.npmjs.com/package/crypto-js) with [crypto-es](https://www.npmjs.com/package/crypto-es) (as suggested [here](https://angular.io/guide/build#configuring-commonjs-dependencies))
- [x] Replace [stream](https://www.npmjs.com/package/stream) with [stream-browserify](https://www.npmjs.com/package/stream-browserify) because its much more pppular (200k vs 12million downloads)
- [x] Replace [tslint](https://www.npmjs.com/package/tslint) with [eslint](https://www.npmjs.com/package/eslint) because it is deprecated
- [x] Removed `patch-webpacke.js` file because of the [changes in webpack 5](https://webpack.js.org/configuration/node/). Tnstead the polyfills are now added in `tsconfig.json` under `path`
- [x] Added allowed commonJS Dependencies in `angular.json`
- [x] Run `ng lint ndb-core --fix` to apply new ESLint rules
- [x] Changed some namings to be in line with ESLint (especially [no on in event names](https://angular.io/guide/styleguide#style-05-16))


TheSlimvReal:
Only supported with angular 13 and therefore part of #883

Issue #892, opened 2021-07-15
Include inactive not set correctly


TheSlimvReal:
**Describe the bug**
The 'include inactive' checkbox is not working correctly on note-details

**To Reproduce**
1. Go to 'Notes'
2. Click on a note
3. Click on 'Add Participants'
4. Click on 'Include Inactive Children'
5. The inactive children will be included but the checkbox is not checked.
6. Click on 'Include Inactive Children' again
7. The checkbox will be checked but the inactive children are not included

**Expected behavior**
The checkbox should be checked immediately.

PR #902, opened 2021-07-20
Fixing "Include inactive" not being set correctly

christophscheuing:
see issue: #892

In the HTML of note-details.component.html instead of `(clicked)` we should rather use `(change)` and we should do this in the `mat-checkbox` and not the `mat-option` tag.


TheSlimvReal (approved):
Beautiful fix, thank you!

Issue #898, opened 2021-07-20
Undoing delete doesnt work


TheSlimvReal:
**Describe the bug**
When deleting a school, a child or a recurring activity, a little snack bar opens that allows to undo the deletion. Pressing the undo button however does not bring the deleted entity back.

**To Reproduce**
1. Go to a child
2. Click on 'Delete'
3. Click on 'Yes'
4. Click on 'Undo'
5. The details page opens, but it does not load correctly.
6. Click on 'Children'
7. The child is not in the children overview and cant be found through the search.

**Expected behavior**
The deleted entity should be restored correctly.

PR #904, opened 2021-07-20
Fixing the bug that undoing the deletion of an entity doesn't work

christophscheuing:
see issue: #898 

The bug probably is due to the fact that when trying to dave an entity to the couchDB that has the _rev field set, couchDB tries to load this entity from the database. This fails because the entity has been deleted from the database.

Setting the _rev field of the concerned entity to undefined gets the EntityMapperService to save it a new database entry without trying to load a non existing entry before.

This seems to fix the bug. But I'm not sure if setting the _rev field to undefined is OK or has any side effects. Maybe a more sophisticated thing could be done with the _rev field.


src/app/core/entity-components/entity-details/entity-details.component.ts
```diff
@@ -142,6 +142,7 @@ export class EntityDetailsComponent {
           { duration: 8000 }
         );
         snackBarRef.onAction().subscribe(() => {
+          this.entity._rev = undefined;
           this.entityMapperService.save(this.entity, true);
```
TheSlimvReal, line 146: We are using the `forceOverwrite` flag. In case the saving of the entity fails (like in this case due to the wrong rev), a conflict resolution is started which loads the existing entity and tries to solve the conflict. In case for a deleted document I think the function throws an error because the `await this.get(newObject._id)` fails ([see here](https://github.com/Aam-Digital/ndb-core/blob/master/src/app/core/database/pouch-database.ts#L192)). So I think you could add your solution to that function to make it more generally available.

sleidig (approved):

PR #907, opened 2021-07-23
ci: new gitflow-like release process

sleidig:
switching to a new release process where stable releases are explicitly released onto the `official-release` branch


TheSlimvReal (approved):
Looks good. Let's see it in action on this PR 😄

Issue #887, opened 2021-07-13
Display bug with the progress bars in the Progress Dashboard Component


christophscheuing:
In the Progress Dashboard Component the progress bars aren't shown correctly:

<img width="433" alt="Bildschirmfoto 2021-07-13 um 20 54 58" src="https://user-images.githubusercontent.com/47225324/125509262-329bfd17-1e22-49a7-a752-dd78eebf7d36.png">

PR #909, opened 2021-07-23
Progress widget fix

kirtijadhav:
see issue: #887

### Visible/Frontend Changes
- [x] Progress bar bug fix
- [ ] 

- [x] 
- [ ]


src/styles.scss
```diff
@@ -107,7 +107,10 @@ mat-card.dashboard-widget {
   padding: 10px;
 }
 
-
+.align-dashboard-table-cell{
+  padding-right: 10px;
+  text-align: right;
+}
```
sleidig, line 113: styles should usually be added to the .scss file of the component for which they are needed. That way they are easy to find during development and encapsulated, i.e. they do not accidentally change style of completely different components because the component stylesheet file only affects its own component.

Please move this to `progress-dashboard.component.scss`

sleidig (changes requested):
There is still some screen size where a horizontal scroll bar appears and the numbers on the right are cut off:
image

sleidig (approved):

Issue #836, opened 2021-06-03
Update or remove Help page


sleidig:
The integrated "Help" page in the app is out of date as the attendance system has changed recently. Also, with the configurable interface now the available features and their titles are custom to each instance, so the static help page is not feasible anymore. If there is no suitable solution we may need to remove the page completely.

At a later time we could think about adding a virtual guide/tour (e.g. something like https://daftmonk.github.io/angular-tour/) in a flexible way directly as part of each feature module.

sleidig:
As I am removing the existing Help Page, the content is archived here: https://docs.google.com/document/d/11t1-bB8gdnc9gSZZ2rnwTg_8fIR_ALBlxThPAEtkDFY/edit?usp=sharing

PR #912, opened 2021-07-26
Update help page and generic markdown page module

sleidig:
see issue: #836 

### Visible/Frontend Changes
- [x] "Help" page only shows support contact details (as different customers have very different UI nowadays)
- [x] i18n German version shows translated help page (through config pointing to different markdown file)

- [x] transformed HelpModule into a generic MarkdownPageModule that displays any markdown file passed through dynamic config (and can therefore be reused in different contexts also)


src/app/core/config/config-fix.ts
```diff
@@ -256,7 +256,10 @@ export const defaultJsonConfig = {
     "lazyLoaded":  true
   },
   "view:help": {
-    "component": "Help"
+    "component": "MarkdownPage",
+    "config": {
+      "markdownFile": $localize`:Filename of markdown help page:assets/help/help.en.md`,
```
Schottkyc137, line 261: Maybe there should be a small note that this is not supposed to be 'translated' other than by someone that actually knows where this file is?

Schottkyc137 (changes requested):

src/app/core/config/config-fix.ts
sleidig: done :+1:

Schottkyc137 (approved):

Issue #891, opened 2021-07-15
Tooltip doesnt open on mobile


TheSlimvReal:
**Describe the bug**
When editing a form with descriptions for a field, the description tooltip cannot be seen on a mobile device.

**To Reproduce**
1. Open the app in mobile view
2. Go to a child
3. Click on 'Observations'
4. Click on '+'
5. Click on the '?'
6. No hint will open

**Expected behavior**
A little too tooltip should open describing this field

kirtijadhav:
For mobile screen, the issue arises while observing the changes made to the tooltip is - when you long press on '?', then only the description tooltip appears.

PR #932, opened 2021-08-02
Tooltip does not open on mobile

kirtijadhav:
see issue: #891 

### Visible/Frontend Changes
- [x] fix child observation tooltip click

- [ ] 
- [ ]


package-lock.json
```diff
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "ndb-core",
```
TheSlimvReal, line 8: Can you please undo these changes.

src/app/core/entity-components/entity-form/entity-form/entity-form.component.html
```diff
@@ -18,6 +18,13 @@
           }"
         >
         </ng-template>
+        <mat-icon
+          (click)="tooltipElement.show()" #tooltipElement="matTooltip"
```
TheSlimvReal:
```suggestion
          #tooltipElement="matTooltip"
          (click)="tooltipElement.show()"
```

TheSlimvReal (changes requested):
Looks good so far, please also remove the tooltip from the following components:

- [ ] `EditLongText`
- [ ] `EditText`

Besides that I think this is mostly done.

package-lock.json
kirtijadhav: Yes I can. Thanks

kirtijadhav: When I remove the tooltip component from EditLongText and EditText, position of the tooltip changes.
Please have a look on the attached image -

tooltip-error

TheSlimvReal: Yes it looks like the `<ng-template>` takes up to much space. You should add some css rules so all the form elements have the same length and the tooltip icon does not go into the next line.

kirtijadhav: Okay I will do that!

TheSlimvReal (changes requested):
On some screen sizes the tooltip icon is still in a new line 
tooltip_displayed

kirtijadhav:
> On some screen sizes the tooltip icon is still in a new line
> tooltip_displayed

Okay I will try to modify it!

TheSlimvReal (approved):
Works perfect now, good job!

Issue #469, opened 2020-07-15
BackupService uses invalid mix of json


sleidig:
The `BackupService` currently creates a list of json objects resulting in an overall result that is not valid json (one object per row but not a proper json array). To allow reuse and follow standards this should be changed to export and import fully valid json.

PR #933, opened 2021-08-03
Complex exports (part 1)

sleidig:
_see commit messages_


src/app/core/export/export-service/export.service.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+    test.propertyOne = "Hello";
+    test.propertyTwo = "World";
+
+    const result = service.createJson([test, test]);
+
+    let expected = JSON.stringify({
+      propertyOne: "Hello",
+      propertyTwo: "World",
+    });
+    expected += BackupService.SEPARATOR_ROW;
```
TheSlimvReal, line 58: Why is the `SEPARATOR_ROW` of the backup service and not the export service used

src/app/core/export/export-service/export.service.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+    expected += JSON.stringify({ propertyOne: "Hello", propertyTwo: "World" });
```
TheSlimvReal, line 59: Is there still a reason why we dont use plain JSON instead of this mix of JSON and rows?

src/app/core/export/export-service/export.service.ts
```diff
@@ -0,0 +1,63 @@
...
+import { entityListSortingAccessor } from "../../entity-components/entity-subrecord/entity-subrecord/sorting-accessor";
+
+/**
+ * Prepare data for export in csv format.
+ */
+@Injectable({
+  providedIn: "root",
+})
+export class ExportService {
+  /** CSV row separator */
```
TheSlimvReal, line 13: Update comment, this is also used in the JSON export

src/app/core/export/export-service/export.service.spec.ts
```diff
@@ -91,22 +91,41 @@ describe("ExportService", () => {
...
-    const test = new TestClass();
-    test._id = 1;
-    test._rev = 2;
-    test.propOne = "first";
-    test.propTwo = "second";
-    const expected =
-      '"_id","_rev","propOne","propTwo"\r\n"1","2","first","second"\r\n"1","2","first","second"';
-    const result = service.createCsv([test, test]);
-    expect(result).toEqual(expected);
+  it("should export all properties", async () => {
```
TheSlimvReal, line 94: Better: "It should export all properties if no config is provided"

TheSlimvReal (commented):
Looks very solid so far. Looking at the JSON export I have the feeling that it is very deprecated and not really used any more. Maybe the JSON export can be removed for now.

src/app/core/export/export-service/export.service.spec.ts
sleidig: yes, I'll fix this here now (see #469 )

sleidig:
> Looking at the JSON export I have the feeling that it is very deprecated and not really used any more. Maybe the JSON export can be removed for now.

may be required for a full backup (we omit/transform some csv columns), also I wanted to just do a refactoring not a larger change

src/app/core/export/export-service/export.service.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+
+    const csvExport = await service.createCsv([testEntity]);
+
+    const rows = csvExport.split(ExportService.SEPARATOR_ROW);
+    expect(rows).toHaveSize(1 + 1); // includes 1 header line
+    const columnValues = rows[1].split(ExportService.SEPARATOR_COL);
+    expect(columnValues).toHaveSize(3 + 1);
+    expect(columnValues).toContain('"' + testEnumValue.label + '"');
+    expect(columnValues).toContain(new Date(testDate).toISOString());
+    expect(columnValues).toContain('"1"'); // true => "1"
```
TheSlimvReal, line 135: I find it a little weird that `true` is exported as `1` but `false` is exported as `false`

TheSlimvReal (approved):
The exporting works very well with the usage of the data accessor and the option to define which values to include. We could soon add some good default values for exporting so people are not confused with internal properties like `_rev`.

Issue #765, opened 2021-04-14
Refactor authentication and session


TheSlimvReal:
The current session is build around having one single database. In order to be able to implement #767 this has to be changed. The following steps should be done:

- [x] Remove usage of `pouchdb-authentication` because it enforces authentication against a single database (done in #934)
- [x] Rework local session to save and authenticate against user objects in the local storage, the user objects should use the values from the [CouchDB user objects](https://docs.couchdb.org/en/latest/intro/security.html?highlight=_users#users-documents) in order to allow local login
- [x] Remove passwords from `User` entities
- [x] Change usage of `User` entity towards usage of `DatabaseUser` (CouchDB format)
- [x] Migrate `isAdmin` flag to CouchDB role `admin`
- [x] Decouple the local database from the session
- [x] Merge `NewLocalSession` and `LocalSession`
- [x] Merge `OnlineSession` and `RemoteSession`
- [x] Remove connection state and rather use the local and the remote login state
- [ ] (to be discussed) move synchronization to `PouchDatabase` class (rather than do it inside the session)

PR #934, opened 2021-08-03
Rework remote session

TheSlimvReal:
Part of #765.
Removing dependency to `pouchdb-authentication` and preparing for a one-database-per-entity scenario.

### Visible/Frontend Changes
--

- [x] Refactor remote session to use basic HTTP requests
- [x] Remove `pouchdb-authentication` dependency

### DIFF
--- a/src/app/core/session/session-service/remote-session.ts
+++ b/src/app/core/session/session-service/remote-session.ts
@@ -99,7 +99,7 @@ export class RemoteSession {
       await this.httpClient
         .post(
           `${AppConfig.settings.database.remote_url}_session`,
-          { name: "demo", password: "pass" },
+          { name: username, password: password },
           { withCredentials: true }
         )
         .toPromise();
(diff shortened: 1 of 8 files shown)


sleidig (approved):
Code looks good and clean.

Remote auth expires after 10 minutes of inactivity (see #935) but this is also the case in master, so an independent issue.

src/app/core/session/session-service/remote-session.ts
```diff
@@ -129,8 +119,12 @@ export class RemoteSession {
   /**
    * Logout at the remote database.
    */
-  public logout(): void {
-    this.database.logout();
+  public async logout(): Promise<void> {
+    await this.httpClient
+      .delete(`${AppConfig.settings.database.remote_url}/_session`, {
```
TheSlimvReal, line 124:
```suggestion
      .delete(`${AppConfig.settings.database.remote_url}_session`, {
```

PR #938, opened 2021-08-04
feat: allow exports from multiple related entities in one row

sleidig:


src/app/core/export/export-service/export.service.ts
```diff
@@ -34,34 +35,100 @@ export class ExportService {
...
-        if (value?.toString().match(/\[object.*\]/) !== null) {
-          // skip object values that cannot be converted to a meaningful string
-          continue;
-        }
-
-        exportableObj[label] = value;
-        allFields.add(label);
+        Object.keys(dataRow).map(
+          (key) => ({ query: key } as ExportColumnConfig)
+        );
```
TheSlimvReal, line 49: I feel like it would make things much easier if we also refer to the property schemas of the entities here. The labels of the properties can be reused and we dont have to worry about entities that are missing some properties, so the `rowConfig` and `columns` dont have to be updated for every `dataRow`. 
For the export buttons on the table, it will always be just one entity and for the general export we would just have to iterate over every entity and create the export for the entity.

src/app/core/export/export-service/export.service.ts
```diff
@@ -34,34 +35,100 @@ export class ExportService {
...
+  private async transformDataRowToExportRow(
+    element: any,
+    config: ExportColumnConfig[]
+  ) {
+    const exportableObj = {};
+
+    for (const columnConfig of config) {
+      const label = columnConfig.label ?? columnConfig.query;
+      let value;
+      if (isQuery(columnConfig.query)) {
```
TheSlimvReal, line 81: `isQuery(columnConfig.query)` looks a little suspicious now. Its not very self explaining that `query` might be something else than a query. Maybe `valueAccessor` is more suitable.

src/app/core/export/export-service/export.service.ts
```diff
@@ -34,34 +35,100 @@ export class ExportService {
...
+    function isQuery(queryKey) {
+      return queryKey.startsWith(".") || queryKey.query.startsWith(":");
+    }
+  }
+
+  private extendIntoMultipleRows(
+    exportableElement: {},
+    rowsToExtend: ExportColumnConfig[]
+  ): any[] {
+    const multiElementColumns = rowsToExtend.filter(
```
TheSlimvReal, line 111: I think this should be moved outside of this function because it only makes sense on the first step of the recursion

src/app/features/reporting/query.service.ts
```diff
@@ -41,13 +41,19 @@ export class QueryService {
     query: string,
     from: Date = null,
     to: Date = null,
-    data: any = this.entities
+    data?: any
   ): Promise<any> {
-    if (!data || (data === this.entities && from < this.dataAvailableFrom)) {
+    if (from === null) {
```
TheSlimvReal, line 46:
```suggestion
    if (!from) {
```
Just to be bullet proof

TheSlimvReal (changes requested):

src/app/core/export/export-service/export.service.spec.ts
```diff
@@ -5,18 +5,48 @@ import { ConfigurableEnumValue } from "../../configurable-enum/configurable-enum
...
+import { ChildSchoolRelation } from "../../../child-dev-project/children/model/childSchoolRelation";
+import { ExportColumnConfig } from "./export-column-config";
+import { defaultAttendanceStatusTypes } from "../../config/default-config/default-attendance-status-types";
 
 describe("ExportService", () => {
   let service: ExportService;
+  let db: PouchDatabase;
+
+  beforeEach(async () => {
+    db = await PouchDatabase.createWithInMemoryDB("export-service-tests");
```
TheSlimvReal, line 28:
```suggestion
    db = PouchDatabase.createWithInMemoryDB("export-service-tests");
```

src/app/core/export/export-service/export.service.ts
```diff
@@ -34,34 +35,155 @@ export class ExportService {
...
+
     const exportableData = [];
+    for (const dataRow of data) {
+      const extendedExportableRows = await this.generateExportRows(
+        dataRow,
+        config
+      );
+      exportableData.push(...extendedExportableRows);
+    }
+
```
TheSlimvReal, line 54: Maybe move the `entitySortingDataAccessor` call here so its applied to all fields just before they are exported.

src/app/core/export/export-service/export.service.ts
```diff
@@ -34,34 +35,155 @@ export class ExportService {
...
+    const rowsOfRows: ExportRow[][] = additionalExportRows.map((addRow) =>
+      exportRows.map((row) => Object.assign({}, row, addRow))
     );
+    // return flattened array
+    return rowsOfRows.reduce((acc, rowOfRows) => acc.concat(rowOfRows), []);
   }
 }
+
+interface ExportRow {
+  [key: string]: any;
```
TheSlimvReal, line 188: Can we get to a point where the type is `[key: string]: string | number`?

TheSlimvReal (changes requested):
Todo

- [ ] Fix config (`key` -> `query`)
- [ ] Add documentation

src/app/core/export/export-service/export.service.ts
TheSlimvReal: Or use `entitySortingDataAccessor` whenever assigning to a `ExportRow` and ensure the return type is `string | number`

TheSlimvReal (approved):
I added some documentation and cleaned up code but no big changes. The feature works very well and I think we can get quite far with that. Good job!

PR #939, opened 2021-08-04
fix(changelog): remove irrelevant versions and details

### LINKED ISSUE #925: Changelog Popup should not show pre-releases
With the new release process master builds are tagged as "pre-release" versions as well but should not be listed in the popup shown to users:

image

also, it would be good to filter changelog lines applying to generic features, maybe any line starting with "-" to allow easy marking when managing release notes

sleidig:
excludes prerelease versions
removes lines marked as hidden by starting with .

fixes #925

### DIFF
--- a/src/app/core/latest-changes/latest-changes.service.spec.ts
+++ b/src/app/core/latest-changes/latest-changes.service.spec.ts
@@ -47,6 +47,13 @@ describe("LatestChangesService", () => {
       body: "A",
       published_at: "2018-01-01",
     },
+    {
+      name: "prerelease 1",
+      tag_name: "1.0-rc.1",
+      prerelease: true,
+      body: "A",
+      published_at: "2018-01-01",
+    },
   ];
 
   beforeEach(() => {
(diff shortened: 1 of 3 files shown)


src/app/core/latest-changes/latest-changes.service.spec.ts
```diff
@@ -122,4 +129,56 @@ describe("LatestChangesService", () => {
...
+      done();
+    });
+  });
+
+  it("should remove irrelevant details from release changelog body", (done) => {
+    const testRelease = {
+      name: "test with notes",
+      tag_name: "3.0",
+      body:
+        "* fix ([e03dcca](https://github.com/Aam-Digital/ndb-core/commit/e03dcca7d89e584b8f08cc7fe30621c1ad428dba))",
```
TheSlimvReal, line 173: Is it correct, that the links to the GitHub issues are kept in the changelogs?

TheSlimvReal (approved):
Looks good, I am just not sure if its correct to keep the GitHub issues in the changelogs

src/app/core/latest-changes/latest-changes.service.spec.ts
sleidig: fair point, that is debatable. maybe would be better to hide them from normal users. But let's keep it as is for now.

Issue #869, opened 2021-06-30
Improve separation of columns listing multiple entities


sleidig:
User names as well as child blocks are not separated very well when displayed with multiple values in a list column:
![image](https://user-images.githubusercontent.com/1682541/123974818-38355e00-d9bd-11eb-91df-7817e1a029eb.png)

More spacing or some other separation between elements?

sleidig:
The child-blocks are missing some margin to have proper spacing between the name of the first and the photo of the second element.

Moreover, for "RecurringActivities" with their assigned users (that are just displayed as text), separating one element from another is even harder for users:
![image](https://user-images.githubusercontent.com/1682541/128210349-ffd8eccd-e39d-4f9d-9123-23c398b280cb.png)

PR #940, opened 2021-08-06
- fix separation of column

kirtijadhav:
see issue: #869 

### Visible/Frontend Changes
- [x] fix margin listing multiple entities in notes
- [x] fix separation of multiple entities in recurring activities

- [ ] 
- [ ]


TheSlimvReal (approved):
Looks better now

PR #941, opened 2021-08-07
Local session refactor

TheSlimvReal:
part of: #765 

### Visible/Frontend Changes
--

- [x] `LocalSession` saves and loads user from local storage
- [x] Refactor `SyncedSessionService`
- [x] Replace `NewLocalSession` with `LocalSession`
- [x] Remove password from `User` entity


src/app/core/session/session-service/remote-session.ts
```diff
@@ -125,6 +142,29 @@ export class RemoteSession {
...
+    this.getLoginState().setState(LoginState.LOGGED_OUT);
+  }
+
+  getCurrentDBUser(): DatabaseUser {
+    return this.currentDBUser;
+  }
+
+  checkPassword(username: string, password: string): boolean {
+    // Cannot be checked against CouchDB due to cookie-auth
+    return false;
```
sleidig, line 156: throw error (or log warning) to avoid accidental use here and in getCurrentUser?

src/app/core/user/user-account/user-account.service.ts
```diff
@@ -1,72 +1,59 @@
...
     }
 
     userResponse["password"] = newPassword;
-    user.setNewPassword(newPassword);
     try {
-      await Promise.all([
-        this.saveNewPasswordToCouchDB(user, oldPassword, userResponse),
-        this.entityMapper.save<User>(user),
-      ]);
+      await this.saveNewPasswordToCouchDB(username, oldPassword, userResponse);
```
sleidig, line 33: shouldn't this also update the local storage?

src/app/core/session/session-service/remote-session.spec.ts
```diff
@@ -20,54 +29,85 @@ describe("RemoteSessionService", () => {
...
-    const unauthorizedError = new Error();
-    unauthorizedError.name = "unauthorized";
-    mockHttpClient.post.and.returnValue(throwError(unauthorizedError));
+    await service.login(TEST_USER, "wrongPassword");
 
-    await service.login("", "");
-
-    expect(service.connectionState.getState()).toBe(ConnectionState.REJECTED);
+    expect(service.getConnectionState().getState()).toBe(
+      ConnectionState.REJECTED
```
sleidig, line 87: Could we merge ConnectionState + LoginState?

Maybe something like this could work (with some adjustments)?
- SyncedSession LoginState = LocalSession LoginState
- SyncedSession ConnectionState (if needed internally?) = RemoteSession LoginState

sleidig (commented):
This is becoming a lot more readable. Awesome!
Just a few remarks in the code but I think it is good to go.

I still have to do the functional testing tomorrow

src/app/core/session/session-service/local-session.spec.ts
```diff
@@ -32,11 +39,78 @@ describe("LocalSessionService", () => {
...
+  it("should login a previously saved user with correct password", async () => {
+    expect(localSession.getLoginState().getState()).toBe(LoginState.LOGGED_OUT);
+
+    await localSession.login(TEST_USER, TEST_PASSWORD);
+
+    expect(localSession.getLoginState().getState()).toBe(LoginState.LOGGED_IN);
+  });
+
+  it("should fail login with correct username but wrong password", async () => {
+    await localSession.login(TEST_USER, "wrong password");
```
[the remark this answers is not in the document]
TheSlimvReal: Good question, I mimicked the old/current behavior here which never rejects but only returns `LoginState.LOGGED_IN` or `LoginState.LOGIN_FAILED`.

src/app/core/session/session-service/remote-session.spec.ts
```diff
@@ -20,54 +29,85 @@ describe("RemoteSessionService", () => {
...
     });
+
+    // Remote session allows TEST_USER and TEST_PASSWORD as valid credentials
+    dbUser = { name: TEST_USER, roles: ["user_app"] };
+    mockHttpClient.post.and.callFake((url, body) => {
+      if (body.name === TEST_USER && body.password === TEST_PASSWORD) {
+        return of(dbUser as any);
+      } else {
+        return throwError(
+          new HttpErrorResponse({ statusText: "Unauthorized" })
```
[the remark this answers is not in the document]
TheSlimvReal: Your right, the status code should be a saver method

src/app/core/user/user-account/user-account.service.ts
TheSlimvReal: I was thinking about that too. But I didn't want to make the assumption that there is a `LocalSession` because the password change also works if only the `RemoteSession` is used where nothing should be written to the local storage. In this case, the `UserAccountComponent` executes a login after successful password change (see [here](https://github.com/Aam-Digital/ndb-core/pull/941/files#diff-662c352c412ac6bcc03b47a13ef7fcfee16b1c324141c7e7fb402cb75a5bbb2fR108)) which will then write the new credentials to the local storage

sleidig:
Functional Testing:

empty local + correct login --> error in UI is `{"status":404,"name":"not_found","message":"missing","error":true,"reason":"missing","docId":"User:demo"}`
Should be less technical

Otherwise all cases work well!
https://docs.google.com/document/d/1VBx_Zsk0d7KdfcXyUhf77a-nWHArHvsKmST3Ss0jr_k/edit?usp=sharing

src/app/core/session/session-service/local-session.spec.ts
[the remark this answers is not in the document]
sleidig: I realized this afterwards. As we have more than just a fail/pass here, this implementation makes sense. We should probably add an `expect` for the return value though?

TheSlimvReal:
Deployed to https://test-deployment-pr-941.herokuapp.com/

src/app/core/session/login/login.component.ts
```diff
@@ -45,64 +44,61 @@ export class LoginComponent {
...
-              $localize`Username or password incorrect!
-               You might also face this problem because you are currently offline.
-               Please connect to the internet to synchronize the latest user data.`
+              $localize`:LoginError:Username and/or password incorrect`
+            );
+            break;
+          default:
+            throw new Error(
+              $localize`:LoginError:Unexpected login state: ${loginState}`
             );
```
sleidig, line 79: why localize the error object when it is caught and a more user-friendly message is set below?

src/app/core/session/login/login.component.ts
```diff
@@ -45,64 +44,61 @@ export class LoginComponent {
...
-            this._sessionService.getSyncState().getState() ===
-              SyncState.ABORTED &&
-            this._sessionService.getConnectionState().getState() ===
-              ConnectionState.OFFLINE
-          ) {
+        switch (loginState) {
+          case LoginState.LOGGED_IN:
+            this.onLoginSuccess();
+            break;
+          case LoginState.UNAVAILABLE:
```
sleidig, line 66: with the LocalSession making a difference between username wrong (--> UNAVAILABLE) and password wrong (--> FAILED) we shouldn't handle the two states differently for the user in the UI - otherwise this potentially exposes what usernames exist to unauthenticated attackers?

sleidig (approved):
Looks good and works :+1: 

Feel free to merge (squashed please!) after reviewing my last remark about login error messaging

src/app/core/session/login/login.component.ts
TheSlimvReal: I added this for a better feedback when offline. This effect is only visible when offline and it should help to show the user whether just the password was misspelled or if the user is generally not available offline. Which user is available offline can be easily found out by an attacker by looking at the local storage.

PR #954, opened 2021-08-20
fix: added unsubscribe calls to observables

TheSlimvReal:
sometimes tests are throwing `Object unsubscribe error` this might fix the problem


sleidig (approved):
Looks good! Thanks for moving this to its own PR :blush:

PR #959, opened 2021-08-30
docs(storybook): fix and clean up storybook

sleidig:
- removed generic storybook samples
- avoid import from a .spec file (`expect` not available in storybook and fails)
- reorder stories in their categories and some additional args


TheSlimvReal (approved):
Good start for cleaning up storybook

PR #960, opened 2021-08-31
chore(demo): use the new export config for a custom event export

sleidig:
export format config as a sample

### DIFF
--- a/src/app/core/config/config-fix.ts
+++ b/src/app/core/config/config-fix.ts
@@ -256,7 +256,6 @@ export const defaultJsonConfig = {
                 { "label": "participant_id", "query": "_id" },
                 { "label": "participant", "query": "name" },
                 { "label": "gender", "query": "gender" },
-                { "label": "area", "query": "area" },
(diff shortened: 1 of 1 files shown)


src/app/core/config/config-fix.ts
```diff
@@ -240,6 +240,32 @@ export const defaultJsonConfig = {
...
+        {
+          "query": ":getAttendanceArray",
+          "subQueries": [
+            {
+              "query": ".participant:toEntities(Child)",
+              "subQueries": [
+                { "label": "participant_id", "query": "_id" },
+                { "label": "participant", "query": "name" },
+                { "label": "gender", "query": "gender" },
+                { "label": "area", "query": "area" },
```
TheSlimvReal, line 259:
```suggestion
```
This is not a property of the child in the demo configuration

TheSlimvReal (changes requested):
Looks like a useful export that probably all organizations could use.
Just a minor adjustment

TheSlimvReal (approved):

Issue #952, opened 2021-08-18
Filters in "Record Attendance" setup view


sleidig:
*As a user assigned to many activities I want to filter the list of activities in the "Record Attendance" screen in order to quickly find the relevant activity and not scroll through a long list.*

Some users have 20+ recurring activities assigned to them, so the simple filter to only show activities linked to the current user is not sufficient to narrow down the list. In that case dropdown filters for school/team and activity type are needed to help them find the right activity.
The filters could be placed next to the date selection. Maybe hide them if less then 5 results are shown?

![image](https://user-images.githubusercontent.com/1682541/129872541-d2b6eeab-e5ce-4080-87a8-0118116b4c9c.png)

PR #967, opened 2021-09-03
Roll call setup filters

sleidig:
see issue: #952


src/app/core/entity-components/entity-list/filter-generator.service.ts
```diff
@@ -98,7 +114,7 @@ export class FilterGeneratorService {
     filter: BooleanFilterConfig
   ): FilterSelectionOption<T>[] {
     return [
-      { key: "all", label: filter.all, filterFun: () => true },
+      { key: "", label: filter.all, filterFun: () => true },
```
TheSlimvReal, line 117: This creates a problem with the configurable enums where we sometimes have the `NONE` option with a empty string as key.

src/app/core/filter/filter/filter.component.ts
```diff
@@ -0,0 +1,76 @@
...
+ * and providing a filter function based on the selection as output.
+ */
+@Component({
+  selector: "app-filter",
+  templateUrl: "./filter.component.html",
+  styleUrls: ["./filter.component.scss"],
+})
+export class FilterComponent<T> {
+  /** special filter option value/key for displaying all data without applying a filter */
+  readonly FILTER_VALUE_ALL = "";
```
TheSlimvReal, line 16: Should also be used in the `FilterGeneratorService`

src/app/core/entity-components/entity-list/list-filter/list-filter.component.html
```diff
@@ -1,30 +1,8 @@
...
-        {{ option.label }}
-      </mat-option>
-    </mat-select>
-  </mat-form-field>
-</div>
+<app-filter
+  (selectedOptionChange)="selectOption($event)"
+  [selectedOption]="selectedOption"
+  [label]="filterConfig.label || filterConfig.name"
+  [data]="filterConfig.options"
```
TheSlimvReal, line 5: Here the data is the possible options of the filter but in the JSDoc of the component it says that this should be the data which should be filtered?

src/app/core/filter/filter/filter.component.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  @Input() selectedOption: string;
+  @Output() selectedOptionChange = new EventEmitter<string>();
+  @Output() selectedFilterFunction = new EventEmitter<(T) => boolean>();
+
+  constructor() {}
+
+  private updateFilterOptions() {
+    const propertyId = this._filterProperty ?? "key";
+    const propertyLabel = this._filterProperty ?? "label";
```
TheSlimvReal, line 51: what are these fallback values for?

src/app/core/filter/filter/filter.component.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    this.filterOptions = this._data
+      .map((element) => ({
```
TheSlimvReal, line 54: is this now an alternative to the `EntityFilterService` and the `FilterSelection` class?

TheSlimvReal (changes requested):
I see that the filters need a general refactoring but this new component makes it just more complex for me because it creates a second alternative system that is somewhat hidden in the old approach. Lets rather move the refactoring to a second topic with concrete specifications.

TheSlimvReal (approved):

Issue #896, opened 2021-07-19
Query params not applied when reloading


TheSlimvReal:
**Describe the bug**
The filter settings which are written to the query params are not correctly applied when reloading the page.

**To Reproduce**
1. Go to 'Schools'
2. Click on 'Hindi'
3. Reload the page
4. The filter 'Hindi' will not be applied automatically but `?medium=hindi` is written to the query params.

**Expected behavior**
The filter settings which are written to the query params should be applied whenever the page is visited.

PR #983, opened 2021-09-12
Fix: Filter selections are considered when reloading the page

Schottkyc137:
See issue #896

**Backend changes / Fix**
When the filters are initialized, the activated route is being used to initialize any filter selections.
Additionally, the (filtered) data is only shown after the filter are properly initialized

### DIFF
--- a/src/app/core/entity-components/entity-list/entity-list.component.ts
+++ b/src/app/core/entity-components/entity-list/entity-list.component.ts
@@ -10,7 +10,7 @@ import {
   ViewChild,
 } from "@angular/core";
 import { MediaChange, MediaObserver } from "@angular/flex-layout";
-import { ActivatedRoute, Router } from "@angular/router";
+import { ActivatedRoute, Params, Router } from "@angular/router";
 import {
   ColumnGroupsConfig,
   EntityListConfig,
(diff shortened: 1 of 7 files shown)


src/app/core/entity-components/entity-list/entity-list.component.ts
```diff
@@ -109,10 +112,9 @@ export class EntityListComponent<T extends Entity>
       this.displayColumnGroup(this.defaultColumnGroup);
     }
     if (changes.hasOwnProperty("allEntities")) {
-      this.filteredEntities = this.allEntities;
-      this.initFilterSelections();
+      await this.initFilterSelections();
+      this.applyFilterSelections();
     }
-    this.loadUrlParams();
```
TheSlimvReal, line 115: There is a slight problem with this change. If you visit the children list and don't click anything the schools are never shown but only once a different view is selected. This line was responsible for reloading the view once the schools are fully loaded. This also allows to load the component with a schools filter selected in the params.

src/app/core/entity-components/entity-list/entity-list.component.ts
```diff
@@ -215,6 +215,7 @@ export class EntityListComponent<T extends Entity>
       this.entityConstructor,
       this.allEntities
     );
+    this.loadUrlParams(this.activatedRoute.snapshot.queryParams);
```
TheSlimvReal, line 218: Adding to the previous comment, this should go back to the `ngOnChanges` function.

TheSlimvReal (changes requested):
Very good, nice to see how small changes can fix this easily. Just a small problem regarding the (annyoing) schools.

TheSlimvReal (approved):

Issue #982, opened 2021-09-10
Improve UX of "Add Child" in school details view


sleidig:
A few aspects about the form to add a child to a school from the school details view could still be polished further:

- The child select does not open a dropdown when clicking into it, only when starting to type (in other forms this is opening, e.g. RecurringActivity details to add participants)
- The purpose of "From" and "To" fields has often been unclear to users, we should add tooltip explanations.
- Can we use the same component in the school details that we use in the child details, also offering to remove a child, etc.?
- Is there a way to label to button more generically (for some customers we talk of "participants" instead of children), maybe related to #886

TheSlimvReal:
Regarding the button: The name can be changed by providing a different label for the `childId` property of the `ChildSchoolRelation` entity.

TheSlimvReal:
Regarding the deleting: We can't use the basic logic of the `EntitySubrecord` there because currently we are actually displaying the children, not the `ChildSchoolRelation` objects, which means the delete button would delete the child and not just the child's participation at the school. To make a more similar look and feel to the component in the child details view, we could instead display the `ChildSchoolRelation`, which would then only show the `ChildBlockComponent` (not the age as a additional column), but it would allow to use all the default behavior of the `EntitySubrecordComponent` (creating , editing, deleting out of the box).
Otherwise we would need to make the delete behavior _overwritable_

sleidig:
True. I think for logical consistence doing it the same way as in the PreviousSchoolsComponent in the Child Details would make sense (i.e. displaying the school block and ChildSchoolRelation infos)

TheSlimvReal:
- [ ] Select single entity should automatically open the dropdown when selected
- [ ] Add description for from and to button e.g. _Date at which the child started at the school_
- [ ] Make the `ChildrenOverviewComponent` more similar to the `PreviousSchoolsComponent`. It should display the columns `childId`, `schoolClass`, `start`, `end`, `result` and the `Add Child` button should be removed and instead the `add` button from the `EntitySubrecord` should be used.

PR #988, opened 2021-09-16
Refactor edit single entity component

Schottkyc137:
see issue: #982

### Visible/Frontend Changes
- [x] Update UX of edit single entity component


src/app/core/entity-components/entity-utils/dynamic-form-components/edit-single-entity/edit-single-entity.component.html
```diff
@@ -1,23 +1,37 @@
...
     [fxHide]="!formControl.enabled"
     [placeholder]="placeholder"
     [formControl]="entityNameFormControl"
     [matAutocomplete]="autoSuggestions"
+    (keyup.enter)="submit()"
+    (focusout)="submit()"
   />
   <mat-autocomplete
     #autoSuggestions="matAutocomplete"
+    (optionSelected)="select($event)"
```
TheSlimvReal, line 15: Good fix! Currently selecting a option with the 'enter' button results in an error because `formControl.setValue` is never called. This fixes the problem.

src/app/core/entity-components/entity-utils/dynamic-form-components/edit-single-entity/edit-single-entity.component.ts
```diff
@@ -52,7 +63,40 @@ export class EditSingleEntityComponent
...
+      (e) => e.toString().toLowerCase() === event.option.value.toLowerCase()
+    );
+    if (entity) {
+      this.selectedEntity = entity;
+      this.editingSelectedEntity = false;
+      this.formControl.setValue(entity.getId());
+    }
+  }
+
+  submit() {
```
TheSlimvReal, line 85: This function is very similar to the `select` function. I guess the function parameter `event` could be changed to `entityString: string` and then it just has to be called with either `event.option.value` or `entityNameFormControl.value`.

src/app/core/entity-components/entity-utils/dynamic-form-components/edit-single-entity/edit-single-entity.component.ts
```diff
@@ -52,7 +63,40 @@ export class EditSingleEntityComponent
...
+    if (entity) {
+      this.selectedEntity = entity;
+      this.editingSelectedEntity = false;
+      this.formControl.setValue(entity.getId());
+    }
+  }
+
+  editSelectedEntity() {
+    this.editingSelectedEntity = true;
+    this.formControl.setValue(null);
```
TheSlimvReal, line 97: Why ist this set to `null`?

src/app/core/entity-components/entity-utils/dynamic-form-components/edit-single-entity/edit-single-entity.component.ts
```diff
@@ -52,7 +63,40 @@ export class EditSingleEntityComponent
...
     if (selectedEntity) {
+      this.selectedEntity = selectedEntity;
+      this.editingSelectedEntity = false;
       this.entityNameFormControl.setValue(selectedEntity.toString());
+    } else {
+      this.entityNameFormControl.setValue("");
     }
   }
+
+  select(event: MatAutocompleteSelectedEvent) {
```
TheSlimvReal, line 74: I guess some tests wouldn't hurt 😄

TheSlimvReal (changes requested):
I like the look and feel of it! Thank you for taking this up. I requested some changes but I can also implement this myself.

src/app/core/entity-components/entity-utils/dynamic-form-components/edit-single-entity/edit-single-entity.component.ts
Schottkyc137: There could be the case where the user edits the selected entity, enters an invalid entity, and exits the form field. If the value wasn't set to `null`, the wrong value would still be valid.

TheSlimvReal (approved):
Thanks for the final adjustments. Works very well now

PR #991, opened 2021-09-17
Usage tracking improvements

sleidig:
see commit messages


src/app/core/dashboard-shortcut-widget/dashboard-shortcut-widget/dashboard-shortcut-widget.component.html
```diff
@@ -17,6 +17,7 @@
         *ngFor="let entry of shortcuts"
         class="dashboard-table-row clickable"
         [routerLink]="[entry.link]"
+        angulartics2On="click" angularticsCategory="Dashboard" angularticsAction="shortcut-widget_click"
```
TheSlimvReal, line 20:
```suggestion
        angulartics2On="click" angularticsCategory="Dashboard" angularticsAction="shortcut-widget_click" [angularticslabel]="entry.link"
```
To also see which shortcuts are used and which ones not.

src/app/core/entity-components/entity-subrecord/entity-subrecord/entity-subrecord.component.html
```diff
@@ -97,12 +100,18 @@
               class="table-action-icon"
               aria-label="cancel"
               fontIcon="fa-times-circle"
+              angulartics2On="click"
```
TheSlimvReal, line 103: Is there a reason why in this case the angulartics are attached to the icon and not to the button like in the other two cases?

src/app/app.component.ts
```diff
@@ -56,28 +56,36 @@ export class AppComponent implements OnInit {
...
+
+  private async initBasicServices() {
     // If loading the config earlier (in a module constructor or through APP_INITIALIZER) a runtime error occurs.
     // The EntityMapperService needs the SessionServiceProvider which needs the AppConfig to be set up.
     // If the EntityMapperService is requested to early (through DI), the AppConfig is not ready yet.
     // TODO fix this with https://github.com/Aam-Digital/ndb-core/issues/595
-    configService.loadConfig(entityMapper);
+    await this.configService.loadConfig(this.entityMapper);
+
+    this.analyticsService.init();
```
TheSlimvReal, line 69: Can we make sure to not use the analytics in development mode? I guess it would be quite misleading if all locally running apps are also sending analytics information. Angular has some ways to [check whether app is in prod or dev mode](https://stackoverflow.com/a/39535462/10713841)

TheSlimvReal (changes requested):
Very nicely structured PR and commits!

It might be useful to have a little guide or guideline for how the categories, labels and actions should be named/organized.

src/app/core/entity-components/entity-subrecord/entity-subrecord/entity-subrecord.component.html
sleidig: good find, this was just messily done

src/app/app.component.ts
sleidig: This is actually done here: https://github.com/Aam-Digital/ndb-core/blob/c17829bff957bf703662e172564f4d154c61be85/src/app/app.module.ts#L91-L93

sleidig: but we can save the setup as well, will add your suggested flag

TheSlimvReal (approved):

Issue #889, opened 2021-07-13
Update FontAwesome


TheSlimvReal:
The currently installed [font-awesome](https://www.npmjs.com/package/font-awesome) package is quite old (last update 5 years ago). It also produces some error messages since the update to Angular 12. Therefore, the packages should be removed and replaced by a later version (e.g. [Angular FontAwesome](https://www.npmjs.com/package/@fortawesome/angular-fontawesome)).
Maybe there is now even a different approach which could be checked in the [Angular Material Documentation](https://material.angular.io/components/icon/overview#font-icons-with-css).

PR #992, opened 2021-09-17
Update to Font-Awesome 5

Schottkyc137:
see issue: #889 

Additionally to upgrading to fa 5, this also introduces the change that the concept of an `icon` in the config is decoupled from the concept of an `icon` in the real app. The implementation detail (the internal name of an item) shouldn't be reflected in the config. This enables the usage of new icons or even a completely new icon library without having to change the config for every active user.

### Visible/Frontend Changes
- [x] The app is enhanced with a new icon-set

- [x] Every instance of `mat-icon` or similar is replaced with the `fa-icon` element
- [x] The concept of an `icon` in the config is decoupled from the concept of an `icon` in the real app


src/app/core/view/dynamic-components/fa-dynamic-icons.ts
```diff
@@ -0,0 +1,33 @@
...
+  ["child", faChild],
+  ["university", faUniversity],
+  ["calendar", faCalendar],
+  ["calendar-check", faCalendarCheck],
+  ["table", faTable],
+  ["notes", faFileAlt],
+  ["wrench", faWrench],
+  ["user", faUser],
+  ["question", faQuestionCircle],
+  ["line-chart", faChartLine],
```
TheSlimvReal, line 32: I am not sure if this is a desirable approach. This would mean whenever we want to use a new icon (e.g. a football when setting up a instance for a soccer club) we first have to make code changes and create a release for it. What was the reason to add this layer of abstraction? 
Maybe a solution would be to have a fallback handling where if no alias can be found in this map, the `IconDefinition` is fetched from font awesome using the exact name from the config? So we would have the chance to define aliases but at the same time don't have to make code changes to use new icons.

TheSlimvReal (changes requested):
Thank you for taking this up. Finally we can use all the new icons and it also looks much cleaner in the code now.

Just a small problem with the static definition of the icons. Is this necessary or could we have a more dynamic handling there?

src/app/core/view/dynamic-components/fa-dynamic-icons.ts
Schottkyc137: There is definitively room for discussion here. From my perspective, having 'hardcoded' icons in the config means that whenever the icon definitions change (as was the case when upgrading to the new font awesome icon set; e.g. there are no longer icons with an `-o` suffix) the config also has to be updated in order to reflect these changes. In my thinking, this means that we have to redefine the whole config for each instance when switching from one icon set to another (or, as in this case, from one version to another).
The other thing is that for the new font awesome version, icons can also prefixed with either `far` or `fas` indicating either regular or solid icons.

TheSlimvReal: Good point so it's either that or a little config migration script. Let's figure this out tomorrow.

src/app/core/view/fa-dynamic-icon/fa-dynamic-icon.component.ts
```diff
@@ -0,0 +1,88 @@
...
+   */
+  @Input() set icon(icon: string) {
+    let definition = faDynamicIcons.get(icon);
+    if (!definition) {
+      // Fallback if the icon is not available: search through the icon definitions
+      definition = this.iconLibrary.getIconDefinition("far", icon as IconName);
+    }
+    if (!definition) {
+      // Fallback if the icon is neither in the map nor a registered icon
+      definition = FaDynamicIconComponent.fallbackIcon;
```
TheSlimvReal, line 77: Maybe add a log here to show that something might be misconfigured

src/app/core/view/fa-dynamic-icon/fa-dynamic-icon.component.ts
```diff
@@ -0,0 +1,88 @@
...
+  static fallbackIcon = faQuestionCircle;
+
+  /**
+   * Sets the dynamic icon by name.
+   * You should make sure that the icon is registered inside the {@link faDynamicIcons}-map,
+   * or put it into this map if it isn't there.
+   * <br>
+   * If for some reason the icon is not inside the map or cannot be inserted into the map,
+   * the icon name has to match the <em>exact</em> icon name as provided by font-awesome.
+   * Prefixed icon names are not supported and the prefix is always assumed to be `far`
```
TheSlimvReal, line 59: I guess `far` refers to the font awesome regular library. Which icons belong to the font awesome solid map and are therefore no that easily accessible?

src/app/core/view/fa-dynamic-icon/fa-dynamic-icon.component.ts
```diff
@@ -0,0 +1,88 @@
...
+export class FaDynamicIconComponent {
+  /** The fallback icon if the given icon is neither known (inside the internal map)
+   * nor registered as a font-awesome icon
+   */
+  static fallbackIcon = faQuestionCircle;
+
+  /**
+   * Sets the dynamic icon by name.
+   * You should make sure that the icon is registered inside the {@link faDynamicIcons}-map,
+   * or put it into this map if it isn't there.
```
TheSlimvReal, line 55: What is the benefit of always registering the icon instead of 'just' using it with the normal name?

TheSlimvReal (commented):
I like the new component. This makes it very easy to use. Just some questions for clarification and I think we can merge this.

src/app/core/view/fa-dynamic-icon/fa-dynamic-icon.component.ts
Schottkyc137: Most of these icons are regular icons that have a "filled" look and thus exist both with `far` and `fas`. I had a quick look at how many icons exist in both definitions and I think that it would actually be better to use the solid icons since there are way more of them

src/app/core/view/fa-dynamic-icon/fa-dynamic-icon.component.ts
Schottkyc137: To have it decoupled from the implementation. 
When it comes to change icon definitions, it is a trade-of between a regular update or a config-migration that needs to be executed. I have chosen the former one.

sleidig: This adds quite a bit of complexity. If it requires code changes when adding some non-standard custom icon for an individual customer, I think this would be a barrier rather than a help for our current state of operations

src/app/core/view/fa-dynamic-icon/fa-dynamic-icon.component.ts
sleidig: we should be consistent, although I am not sure that the number of choices is the only important aspect. Let's briefly rope in Rosa and compare the look and feel of solid vs. outlined icons.

TheSlimvReal (commented):
I have cleaned up some imports of the test files. Let's discuss the final open questions later

TheSlimvReal (approved):
Very good work. This was more than necessary

PR #1001, opened 2021-09-21
Child list - UX improvement

kirtijadhav:
see issue: #982 

### Visible/Frontend Changes
- [x]  Select single entity should automatically open the dropdown when selected
- [x]  Add description for from and to button e.g. Date at which the child started at the school
- [x]  Make the ChildrenOverviewComponent more similar to the PreviousSchoolsComponent. It should display the columns childId, schoolClass, start, end, result and the Add Child button should be removed and instead the add button from the EntitySubrecord should be used.


TheSlimvReal (approved):
Looks good now, very nice. Thank you!

Issue #973, opened 2021-09-06
(1) Dedicated ImportModule and Component


sleidig:
1. _As a project coordinator I want to have a simple "Import" page in the app where I can select a (csv) file, upload and add new entities from the file in order to import existing data that I have from other systems._

**Additional Info**
- For this first iteration the CSV file uploaded by the users must contain an _id column (including the correct "entity prefix" and a unique identifier, e.g. "Child:0c3d6032-edb6-490c-a5c2-5bf358191840") and column titles that exactly match the entity property names. That way no data transformation is necessary.
- have a look at the existing rudimentary implementation at https://github.com/Aam-Digital/ndb-core/blob/master/src/app/core/admin/admin/admin.component.ts (and its services)
- implement this as a separate, isolated Angular Module that exports a component that we can use as a page.
- set up storybook stories for the component to document the design and have a base for further extending it later

PR #1006, opened 2021-09-23
Hwr/data import

Spikeone:
see issue: #973 

### Visible/Frontend Changes
- none

- moved dataimport functionality from admin component into separate module
- minor enhancements to corresponding unit tests


src/app/features/data-import/data-import/data-import.component.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+
+describe("DataImportComponent", () => {
+  let component: DataImportComponent;
+  let fixture: ComponentFixture<DataImportComponent>;
+  let mockDataImportService: jasmine.SpyObj<DataImportService>;
+  let mockCsvFile: Blob = new Blob(["1;2;3"]);
+
+  beforeEach(
+    waitForAsync(() => {
+      AppConfig.settings = {
```
TheSlimvReal, line 16: I think these configurations are not needed for this component test.

src/app/features/data-import/data-import/data-import.component.ts
```diff
@@ -0,0 +1,24 @@
...
+@Injectable({
+  providedIn: "root",
+})
+export class DataImportComponent implements OnInit {
+
+  constructor(
+    private dataImportService: DataImportService
+  ) {  }
+
+  ngOnInit(): void {
```
TheSlimvReal, line 18: this function can be removed

package.json
```diff
@@ -9,6 +9,7 @@
     "build-localized": "ng build --configuration=production,localized",
     "test": "ng test",
     "test-ci": "ng test --code-coverage --karmaConfig=build/karma-ci.conf.js ",
+    "test-cov": "ng test --code-coverage --karmaConfig=karma.conf.js ",
```
TheSlimvReal, line 12:
```suggestion
```
I don't think this is needed (the `package.json` already has so many scripts). If you want to run tests with coverage locally you can also use
`npm test -- --code-coverage`, everything after the first `--` is passed to the `ng test` command.

TheSlimvReal (changes requested):
The changes are very good so far, it looks like you are already very familiar with the code. For more details see the comments.

Maybe you saw that the tests failed. This is because `npm lint` is failin. You can see there error messages [at the bottom of the log](https://github.com/Aam-Digital/ndb-core/pull/1006/checks?check_run_id=3690132827#step:5:909) or run the command locally. We are using [prettier](https://prettier.io/) for our code linting. 
In Webstorm you can automatically fix all the lint errors while coding by downloading the prettier plugin and apply prettier on-save. Otherwise prettier can also be used in the command line with `prettier --write <relative path to file or folder>`. Most of the errors can be fixed that way automatically.

TheSlimvReal (approved):

PR #1007, opened 2021-09-24
Implement entity-remove service

Schottkyc137:
Removing an entity or the row of a table currently is done inside the components and uses the `EntityMapperService`. This leads to duplicate code and sometimes even inconsistent behavior since there always is a confirmation dialog and an option to undo this change.
This PR includes a service that makes this consistent and avoids duplicate code.

This also fixes a bug where the app won't navigate back when pressing 'undo'

### Visible/Frontend Changes
None

- [x] Implemented `EntityRemoveService`


kirtijadhav (approved):
Now looks very clean and neat!

PR #1018, opened 2021-10-01
Creating app with single service worker

TheSlimvReal:
Angular automatically builds a service worker for each language/locale that is available. This means each locale is a separate PWA that can be installed. By combining the individual service workers, the whole app with all locales becomes one PWA that can be installed. 

### Visible/Frontend Changes
-- 

- [x] Merging service workers in Dockerfile
- [x] Adjusted scope to be absolute and the root folder

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -2,67 +2,12 @@
 # To use it only Docker needs to be installed locally
 # Run the following commands from the root folder to build, run and kill the application
 # >> docker build -f build/Dockerfile -t aam/digital:latest .
-# >> docker run -p=80:80 aam/digital:latest
+# >> docker run -p=80:80 aam/digital:latest:
 # >> docker kill aam-digital
-FROM node:15.11.0-alpine3.13 as builder
-WORKDIR /app
-
-COPY package*.json ./
-COPY patch-webpack.js .
-RUN npm ci --no-progress
-
-RUN $(npm bin)/ng version
-
-COPY . .
-
-ARG APP_VERSION="UNKNOWN"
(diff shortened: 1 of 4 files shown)


TheSlimvReal:
Current problem

- [ ] Changing language somehow breaks the cache e.g. install app -> change language -> go offline -> reload app -> **does not load**

src/assets/icons/flags/de.svg
```diff
@@ -0,0 +1,5 @@
+<svg xmlns="http://www.w3.org/2000/svg" id="flag-icon-css-de" viewBox="0 0 640 480">
```
sleidig, line 1: where do we get additional flags when adding locales? Does it really make sense to manage these files ourselves instead of through a library?

sleidig (approved):
Update also seems to work and offline has no issues.

Switching language while offline only possible if previously visited that language while online. But I think that's a very reasonable assumption.

src/assets/icons/flags/de.svg
TheSlimvReal: The problem which I ran into with the library was that we either have to explicitly tell the service worker which flags to cache or it caches all flags which is around 30mb of storage. So I guess its not too much of a burden just download the library and copy the flag which is required when adding a new language.

sleidig: I think it is somewhat significant effort - and bears the risk that someone will add flags from a different source, making the design inconsistent.

how about instead setting the serviceworker to "lazy" for all flag resources (only cache once explicitly the file is requested)?

Similar could be an optimization for the double loading of es5 and es2018 script variants - but they do advice against it as es5 is anyway gonna be removed: https://github.com/angular/angular/issues/31256#issuecomment-506507021

PR #1040, opened 2021-11-16
Strict type check in Angular compiler

TheSlimvReal:
This PR adds the [strictTemplates](https://angular.io/guide/template-typecheck#strict-mode) compiler option for the Angular/Typescript compiler. This performs further static analysis on the types used inside the app. This should help us to make more use of Typescript and detect errors earlier.

### Visible/Frontend Changes
-- 

- [x] Enabled `strictTemplates` for the Angular compiler
- [x] Fixed type errors that were in the system

### DIFF
--- a/src/app/core/entity-components/entity-utils/dynamic-form-components/edit-component.spec.ts
+++ b/src/app/core/entity-components/entity-utils/dynamic-form-components/edit-component.spec.ts
@@ -1,19 +1,23 @@
 import { EditComponent } from "./edit-component";
 import { FormControl, FormGroup } from "@angular/forms";
 
-export const FORM_CONTROL_NAME = "testProperty";
-
 /**
  * A simple helper class that sets up a EditComponent with the required FormGroup
  * @param component that extends EditComponent
+ * @param propertyName (optional) the name of the property for which the edit component is created
  */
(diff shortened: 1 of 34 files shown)


src/app/core/entity-components/entity-utils/dynamic-form-components/edit-component.spec.ts
```diff
@@ -0,0 +1,19 @@
+import { EditComponent } from "./edit-component";
+import { FormControl, FormGroup } from "@angular/forms";
+
+export const FORM_CONTROL_NAME = "testProperty";
+
+/**
+ * A simple helper class that sets up a EditComponent with the required FormGroup
+ * @param component that extends EditComponent
+ */
+export function setupEditComponent<T>(component: EditComponent<T>): FormGroup {
```
sleidig, line 10: can we make the FORM_CONTROL_NAME as a parameter of this function? Would make it a bit more clear (no depending on the other constant) and flexible - although in line 12 currently the name is hardcoded?

not super essential, just an idea. Otherwise maybe rename to `setupEditComponentForTestProperty`?

sleidig (approved):
Looks great! :-)

PR #1044, opened 2021-12-01
fixes #753 'Search broken for imported data'.

### LINKED ISSUE #753: Search broken for imported data | Importing data does not generate searchIndices
**Describe the bug**
In the Admin section data can be imported from a csv or json file. This data is parsed and directly written to the database by the `BackupService` without using the `EntityMapperService`. Therefore, the schema is not enforced. In particular the `searchIndices` property is not generated, therefore search is only working after someone edits the entities and saves them again within the app.

Paschild:
see issue: #753 

guess its more like a work around for now but looking forward to your opinions


src/app/features/data-import/data-import.service.ts
```diff
@@ -32,7 +34,9 @@ export class DataImportService {
           delete record[propertyName];
         }
       }
-
+      if (record["name"] !== undefined) {
+        record["searchIndices"] = record["name"].split(" ");
```
TheSlimvReal, line 38: I like the simple approach. 

There is only the problem, that some subclasses of `Entity` override the `searchIndices` function. E.g. the `Child` class includes the `projectNumber`.
To correctly support that you can use the `_id` field of the `record` and the `DynamicEntityService.getEntityConstructor` function to access the entity's constructor and then using something like `Object.assign(new entityContructor(), record).searchIndices`.

Let me know if you need help with that. We can also do this in pair programming!

TheSlimvReal (commented):
Great! This is exactly what I had in mind and it works perfectly!
I merged the latest master into the branch.

It would be great if you could add a test, but then this is ready to be merged. Maybe the code could even become more easy once this is combined with the other changes to the import component

src/app/features/data-import/data-import.service.ts
sleidig: I added a test for it now, as we want to get this bugfix release.

sleidig (approved):
Works as expected. Great :+1:

PR #1059, opened 2021-12-02
Hwr/data import type select

Osbie-Feel:
see overarching summary issue: #1135


src/app/features/data-import/data-import.service.ts
```diff
@@ -18,12 +18,47 @@ export class DataImportService {
...
+
+      const record = parsedCsvFile.data[0];
+
+      // check all properties, if there is an _id, make sure it fits
+      for (const propertyName in record) {
+        if (propertyName !== "_id") {
+          continue
+        }
+
+        if (!record[propertyName].startsWith(entityType)) {
```
TheSlimvReal, line 40: This check is not working. I am getting `record[propertyName] === undefined` even when I have provided IDs. I would suggest you to write tests that ensure the code is correct (Test-Driven-Development!).

TheSlimvReal (changes requested):
Functionality wise it already looks quite fine, even though the `validateCsvFile` is having some bugs. 
Most of my comments are more focused on the usability which in this case is crucial if this component should not just be used by admins. What is therefore missing are understandable error messages, a better user guidance through disabling buttons etc. and a final overview where user can check that everything they entered is correct. Let me know if you need any help with that. I also merged the latest master into the app.

TheSlimvReal:
Deployed to [https://test-deployment-pr-1059.herokuapp.com/](https://test-deployment-pr-1059.herokuapp.com/)

src/app/features/data-import/data-import.service.ts
```diff
@@ -74,20 +74,23 @@ export class DataImportService {
...
+
+    const dialogRef = this.confirmationDialog.openDialog(
+      refTitle,
+      refText
     );
 
     dialogRef.afterClosed().subscribe(async (confirmed) => {
       if (!confirmed) {
-        return;
+        return Promise.resolve(undefined);
```
TheSlimvReal, line 123: You dont need to to return a promise in a async funciton. This will automatically return a promise.

src/app/features/data-import/data-import/data-import.component.ts
```diff
@@ -67,12 +73,18 @@ export class DataImportComponent implements OnInit {
 
   async importSelectedFile(): Promise<void> {
     if (this.csvFile === undefined) {
-      return;
+      return Promise.resolve(undefined);
```
TheSlimvReal, line 77: Also not necessary, see above

src/app/features/data-import/data-import.service.ts
```diff
@@ -55,17 +57,45 @@ export class DataImportService {
...
+
+    for (const record of existingRecord) {
+      await this.db.remove(record);
+    }
+
+    let hasIdProperty = true;
+
+    if (!parsedCsv.meta.fields.includes("_id")) {
+      hasIdProperty = false;
+      parsedCsv.meta.fields.push("_id");
```
TheSlimvReal, line 78: In case the `_id` property is specified, shouldn't the `db.remove(record)` step be skipped? Because in this case the `transactionId` is irrelevant?

src/app/features/data-import/data-import.service.ts
```diff
@@ -55,17 +57,45 @@ export class DataImportService {
     });
   }
 
-  async importCsvContentToDB(csv: string): Promise<void> {
+  async importCsvContentToDB(csv: string, importMeta: ImportMetaData): Promise<void> {
```
TheSlimvReal, line 60: Is there a reason why you dont just work with 3 function parameters?

TheSlimvReal (changes requested):
Looks already quite straight forward. I noted down some questions

src/app/features/data-import/data-import/data-import.component.html
```diff
@@ -1,13 +1,49 @@
...
-  >
-    Import (.csv)
-</button>
-<input
-  #csvImport
-  type="file"
-  style="display: none"
-  (change)="importCsvFile($event.target.files[0])"
-/>
+<link href="https://fonts.googleapis.com/icon?family=Material+Icons"rel="stylesheet">
```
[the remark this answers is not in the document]
Spikeone: Material uses those icons for displaying the "done" markers (that pen which is shown after a step is done). It is however somehow possible to override icons and display custom icons (or at least there was a feature request) when using material steppers. 
I did try to override them but it didn't work out so I moved on for now.

src/app/features/data-import/data-import/data-import.component.html
```diff
@@ -1,13 +1,49 @@
...
+          <input #csvImport type="file" style="display: none" (change)="setCsvFile($event.target.files[0])"/>
+        </div>
+        <div>
+          <button mat-button matStepperPrevious>Back</button>
+          <button mat-button matStepperNext>Next</button>
+        </div>
+      </form>
+    </mat-step>
+    <mat-step>
+      <ng-template matStepLabel>Excecute</ng-template>
```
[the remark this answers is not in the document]
Spikeone: Agreed, got that one in my presentation that this is missing, but desinging it would be nice anyway.
- Entity Type: <Type>
- Import File: <Filename>
- Records in File: <Number>
- Transaction ID: <ID>

src/app/features/data-import/data-import/data-import.component.ts
```diff
@@ -9,10 +12,61 @@ import { DataImportService } from "../data-import.service";
...
+        this.secondFormGroup.setValue({ secondCtrl: ''});
+      } else {
+        this.csvFile = file;
+        this.secondFormGroup.setValue({ secondCtrl: file.name});
+      }
+    }
+
+    async importSelectedFile(): Promise<void> {
+      if(this.csvFile === undefined) {
+        return;
```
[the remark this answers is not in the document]
Spikeone: I don't think that this is possible - but then again I'm not 100% sure how the file input exactly works if the stored file can become undefined somewhat, else you have to jump back to step 1, invalidate the file and then you can't jump to the last step

src/app/features/data-import/data-import/data-import.component.ts
```diff
@@ -9,10 +12,61 @@ import { DataImportService } from "../data-import.service";
...
+      }
+      await this.dataImportService.handleCsvImport(this.csvFile);
+    }
+
+  importCsvFile(file: Blob): Promise<void> {
```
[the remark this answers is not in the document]
Spikeone: I removed that one, didn't I? The view is a bit strange here

src/app/features/data-import/data-import.service.ts
Spikeone: Good question, if the transaction ID is expected to be unique, the search shouldn't find anything anyway - so calling the getAll shouldn't return anything, thus not deleting anything.

To change this, the workflow needs to be changed. The file needs to be parsed before it's passed to the import so one knows if there is an _id or not, if there is one, no transaction ID needs to be generated (which is currently always generated). So simply checking if there is an _id in the file and then not deleting something isn't correct either I think. What about having a file with _id and giving the transaction ID?

So changing the flow here is required to fully improve this.

src/app/features/data-import/data-import.service.ts
Spikeone: I did, but I expected further information which are passed later. E.g. the parsed file, line numbers, if the file as the _id property etc. which then is encapsulated in the ImportMetaData - Felt like the correct way to pass those information

src/app/features/data-import/data-import/data-import.component.ts
```diff
@@ -9,10 +12,61 @@ import { DataImportService } from "../data-import.service";
...
+  @Input() firstFormGroup: FormGroup;
+  @Input() secondFormGroup: FormGroup;
 
-  importCsvFile(file: Blob): void {
-    this.dataImportService.handleCsvImport(file);
+  csvFile: Blob = undefined;
+
+  constructor(
+    private dataImportService: DataImportService,
+    private _formBuilder: FormBuilder,
```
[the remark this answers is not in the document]
Spikeone: Actually, because that was copied from the Material Stepper example :-/

src/app/features/data-import/data-import/data-import.component.ts
```diff
@@ -9,10 +12,61 @@ import { DataImportService } from "../data-import.service";
...
+    private dynamicEntityService: DynamicEntityService) {}
+
+    ngOnInit() {
+      this.firstFormGroup = this._formBuilder.group({
```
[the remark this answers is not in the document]
Spikeone: Hm, actually same as before, this was copied from the example, so I didn't question is but expected that as the "way to go", gotta research that

src/app/features/data-import/data-import/data-import.component.ts
```diff
@@ -9,10 +12,61 @@ import { DataImportService } from "../data-import.service";
...
+      this.secondFormGroup = this._formBuilder.group({
+        secondCtrl: ['', Validators.required],
+      });
+    }
+
+    getEntitiesMap(): Map<string, EntityConstructor<Entity>> {
+      return this.dynamicEntityService.EntityMap;
+    }
+  
+    get hasValidFile(): boolean {
```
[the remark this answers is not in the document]
Spikeone: As discussed, possibly not that easy to pass a file to the validator, keeping it as is for now - if you don't object

src/app/features/data-import/data-import/data-import.component.ts
```diff
@@ -9,10 +12,61 @@ import { DataImportService } from "../data-import.service";
...
+      return this.csvFile !== undefined;
+    }
+
+    entitySelectionChanged(): void {
+      // whenver the selection changes, the file can't be valid (if there was one)
```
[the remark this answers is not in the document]
Spikeone: Currently this is checked by the file validation, but there is no message to the user yet, except the file is not used (silent fail).

src/app/features/data-import/data-import/data-import.component.html
```diff
@@ -1,13 +1,49 @@
...
+          <button mat-button matStepperNext>Next</button>
+        </div>
+      </form>
+    </mat-step>
+    <mat-step [stepControl]="secondFormGroup" [editable]="true" [completed]="hasValidFile">
+      <form [formGroup]="secondFormGroup">
+        <ng-template matStepLabel>Import File</ng-template>
+        <mat-form-field>
+          <mat-label>Selected File: </mat-label>
+          <input matInput formControlName="secondCtrl" readonly placeholder="No file selected" required/>
```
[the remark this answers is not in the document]
Spikeone: Good point, this was before I used the ```[completed]``` Input

src/app/features/data-import/data-import.service.ts
```diff
@@ -18,12 +18,47 @@ export class DataImportService {
...
   ) {}
 
+    async validateCsvFile(file: File, entityType: string): Promise<boolean> {
+      const csvData = await readFile(file);
+      const parsedCsvFile = this.parseCsvFile(csvData);
+
+      // an empty csv file is not valid
+      if (parsedCsvFile.data.length === 0) {
+        // TODO: Either open a popup here which seems missplaced
+        // Better: Have validation results, so any component can handle the result
```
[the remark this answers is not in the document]
Spikeone: using alert service for now

src/app/features/data-import/data-import/data-import.component.ts
```diff
@@ -9,10 +12,61 @@ import { DataImportService } from "../data-import.service";
 @Injectable({
   providedIn: "root",
 })
-export class DataImportComponent {
-  constructor(private dataImportService: DataImportService) {}
+export class DataImportComponent implements OnInit{
+  @Input() firstFormGroup: FormGroup;
```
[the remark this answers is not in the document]
Spikeone: Have renamed the variables, I think a ```FormControl``` would suffice as well here

src/app/features/data-import/data-import/data-import.component.html
```diff
@@ -1,13 +1,49 @@
...
-/>
+<link href="https://fonts.googleapis.com/icon?family=Material+Icons"rel="stylesheet">
+<mat-vertical-stepper linear #stepper>
+    <mat-step [stepControl]="firstFormGroup" [editable]="true">
+      <form [formGroup]="firstFormGroup">
+        <ng-template matStepLabel>Import Type</ng-template>
+        <mat-form-field>
+            <mat-label>Entity Type</mat-label>
+            <mat-select formControlName="firstCtrl" required (selectionChange)="entitySelectionChanged()">
+                <mat-option *ngFor="let entity of getEntitiesMap() | keyvalue" [value]="entity.key">
```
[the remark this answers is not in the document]
Spikeone: changed this to use a member instead

src/app/features/data-import/data-import.service.ts
Spikeone: Hm, my newly written tests did what I expected them to do in each case

src/app/features/data-import/data-import.service.ts
Spikeone: fixed up

src/app/features/data-import/data-import/data-import.component.html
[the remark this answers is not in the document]
TheSlimvReal: I think [this section](https://material.angular.io/components/stepper/overview#overriding-icons) and [this example](https://material.angular.io/components/stepper/examples#stepper-states) might be helpful. 
You will probably just have to override the `edit` and `done` icons like so: 
```
<ng-template matStepperIcon="done">
   <fa-icon icon="check-circle"></fa-icon>
</ng-template>
```

src/app/features/data-import/data-import/data-import.component.ts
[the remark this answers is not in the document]
TheSlimvReal: On the next button you could say `[disabled]='csvFile === undefined'` and once a file is selected the whole validation is ran and on success `this.csvFile` is set.

src/app/features/data-import/data-import.service.ts
TheSlimvReal: Ok did you try this manually then? Upload a CSV file with a entity type in the `_id` column?

Spikeone: Yeah I actually did, well to the point the import would start which still requires the file to be validated.
I have a file with _id (and 122 records) and one without _id (and 1 record), maybe we can have a look at this together

src/app/features/data-import/data-import/data-import.component.ts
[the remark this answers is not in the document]
Spikeone: Makes sense, boils down to the same thing: The file should be parsed after uploading it, and the parsed result kept instead of the raw file. That way one could check those things and react properly to it and set the information in the importMetaData object

src/app/features/data-import/data-import.service.spec.ts
```diff
@@ -152,4 +153,48 @@ describe("DataImportService", () => {
...
+  it("should reject empty csv without _id", async () => {
+    const mockFileReader = createFileReaderMock("value");
+    const valResult = await service.validateCsvFile(null, "Child");
+
+    expect(mockFileReader.readAsText).toHaveBeenCalled();
+    expect(valResult.status).toEqual(CsvValidationStatus.ErrorEmpty);
+  });
+
+  it("should reject invalid csv", async () => {
+    // I'm unable to break the csv import
```
TheSlimvReal, line 198:
```suggestion
   spyOn(TestBed.inject(Papa), "parse").and.throwError(new Error());
```

TheSlimvReal: That would fail the `this.papa.parse(...)` call so you can work with the error case.

src/app/features/data-import/csv-validation-result.type.ts
```diff
@@ -0,0 +1,6 @@
+import { CsvValidationStatus } from "./csv-validation-Status.enum";
+
+export interface CsvValidationResult {
+    status: CsvValidationStatus,
+    resultMessage: string
```
TheSlimvReal, line 5:
```suggestion
    message: string
```

src/app/features/data-import/data-import.service.ts
```diff
@@ -18,12 +22,57 @@ export class DataImportService {
...
   ) {}
 
-  async importCsvContentToDB(csv: string): Promise<void> {
-    const parsedCsv = this.papa.parse(csv, {
+  async validateCsvFile(file: File, entityType: string): Promise<CsvValidationResult> {
+    const csvData = await readFile(file);
+    const parsedCsvFile = this.parseCsvFile(csvData);
+
+    if (parsedCsvFile === undefined || parsedCsvFile.data === undefined) {
+      return {status: CsvValidationStatus.ErrorNoData, resultMessage: 'The file provided is invalid, parsing was not possible!'};
```
TheSlimvReal, line 30: Why did you decide against throwing error messages?

TheSlimvReal (changes requested):
to be continued

src/app/features/data-import/data-import.service.spec.ts
Spikeone: Thanks for the hint, gotta try that out :)

src/app/features/data-import/data-import.service.ts
Spikeone: @TheSlimvReal For two reasons: In most languages using "throw" is used when evaluating something that should never even happen - but is checked to make sure one can gracefully continue. So using it here to show a message box simply felt wrong. The other idea I had behind this approach is being able to use the module and functionality in other components if required, maybe like in an initial setup dialogue where you can import multiple files etc., then one would like to validate the files one after another and show another message, or no message at all. Thats why I didn't add it to the function.

But yeah, as in most other cases, I'm not that experienced in sofware engineering yet but it felt like a good idea :D

TheSlimvReal:
TODO
- [ ] When importing a config, additional fields from a file are discarded

src/app/features/data-import/data-import.service.ts
```diff
@@ -20,69 +32,154 @@ export class DataImportService {
...
+  ): Promise<void> {
+    for (const row of csv.data) {
+      const entity = this.createEntityWithRowData(row, importMeta);
+      this.createSearchIndices(importMeta, entity);
+      if (!entity["_id"]) {
+        entity["_id"] = `${importMeta.entityType}:${
+          importMeta.transactionId
+        }-${uuid().substr(9)}`;
+      }
+      await this.db.put(entity, true);
```
sleidig, line 133: Couldn't we instantiate a new entity and save that with EntityMapper instead of raw db.put? That would also take care of things like searchIndices out of the box

sleidig (commented):
Code looks pretty good overall! :+1: 

In the future, when this grows further, we can consider
- extracting the import component steps into separate sub-components
- moving the UI logic of confirmation dialog and undo out of the service and into one of the sub-components

The UI could still use a bit of polishing with regards to button styles, margins, additional descriptions, etc.   I can take up that point over the next days.

src/app/features/data-import/data-import.service.ts
TheSlimvReal: Lets do this together with more a more advanced column mapping when we bring the property schemas into action. I note it in #1135

sleidig (approved):
I still have some ideas to simplify the UI of this but let's get this merged as a foundation and then work further on it in new issues/PR.

Great feature. Thanks for the good work, everyone! :smiley:

TheSlimvReal (approved):

Issue #984, opened 2021-09-13
Report component throws error when no config is defined


TheSlimvReal:
**Describe the bug**
When the `config` attribute is not set in the configuration for the reporting component a error message is logged.

**Expected behavior**
The error should fail silently because the reporting component shows a default landing page if no reports are configured

PR #1064, opened 2021-12-07
Report Component no longer shows an error when config is undefined

Schottkyc137:
see issue: #984 

- [x] Error is no longer shown in the console


TheSlimvReal (approved):
Great work, so simple and still so correct. How did you do this?

Issue #1050, opened 2021-12-02
Opening notes details should not focus on date


sleidig:
When opening an existing (or new) Note the form field in focus should not be the date (which opens the calendar) but maybe the subject which is more relevant to be edited.

PR #1065, opened 2021-12-07
Set initial focus to the subject of a note

Schottkyc137:
see issue: #1050

### Visible/Frontend Changes
- [x] Subject of a note is initially focused


TheSlimvReal (approved):
So small so good, thank you!

Issue #1057, opened 2021-12-02
form field validation rules (configurable, regex?)


sleidig:
Some fields require specific formats of input, e.g. Aadhar card numbers in India are 12 digits only (not less/not more, always 12 digit numeric).

Allow in the config definition of properties to add a validation logic that will trigger an error when trying to save the form if it is not matching.

PR #1080, opened 2022-01-11
Configurable validators

Schottkyc137:
see issue: #1057 

### Visible/Frontend Changes
None

- [x] Validation rules can be defined via the config or an entity-field's schema


TheSlimvReal (changes requested):
Currently, there is also a problem with the number inputs. 
When selecting `type="number"` no validation for the input is possible and any letters are omitted when saving. When selecting `type="string"` the up and down buttons are missing but it is possible to display a error message in case a invalid character has been written.

TheSlimvReal (approved):

Issue #908, opened 2021-07-23
Update compodoc


TheSlimvReal:
**Describe the bug**
Compodoc is currently failing due to [this issue](https://github.com/compodoc/compodoc/issues/1102). It has been fixed in [this commit](https://github.com/compodoc/compodoc/commit/40870b387300c27c0ca1be983159cbfc7518de00), but it has not been released yet. Once a later version than `1.1.14` has been released [here](https://github.com/compodoc/compodoc/releases) it should be updated.
Once compodoc works again make sure the following things are correct (they might have been changed in the meantime due to compodoc not working):

- in `package.json` it should say `"storybook": "npm run docs:json && start-storybook -p 6006",`
- in `.storybook/preview.js` the following lines should be present (and not commented out):
```
import { setCompodocJson } from "@storybook/addon-docs/angular";
import docJson from "../documentation.json";
setCompodocJson(docJson);
```

TheSlimvReal:
Somehow even in version `1.1.15` `npm run compodoc` is still failing

TheSlimvReal:
It looks like there has been [another bug](https://github.com/compodoc/compodoc/issues/1102#issuecomment-930064763) which is fixed in [this commit](https://github.com/compodoc/compodoc/commit/33aa835ccdadb032ba7ffef0f38fc6033392640b) and should be part of release 1.1.16.

TheSlimvReal:
Lets also do this as part of #838

TheSlimvReal:
closing this in favor of #838

PR #1081, opened 2022-01-11
Update Compodoc

Schottkyc137:
see issue: #908 

- [x] Update Compodoc to version 1.1.18


TheSlimvReal:
Hmm I am not sure what exactly is the problem. To not put too much time into that we can make this part of #838 (and hopefully also do this one day).

Issue #797, opened 2021-04-28
Attendance of notes are directly edited


TheSlimvReal:
**Describe the bug**
When editing the attendance information of a note in the notes manager, the changes remain even when the editing is canceled.

**To Reproduce**
1. Go to 'Notes'
2. Click on 'a note'
3. Change the attendance of a child (e.g. change status, add remarks, ..)
4. Click on 'cancel`
5. Open the same note again
6. The changes from before are still visible

**Expected behavior**
When clicking cancel, all changes in the `NoteDetailsComponent` should be undone.

**Additional context**
The problem occurs because no deep copy of the note is passed to the `NoteDetailsComponent` it is possible to add `note.childrenAttendance = new Map(this.childrenAttendance);` in the `copy()` function of the note entity, but this will only generate a shallow copy of the map. If existing `EventAttendance` objects are edited like in the scenario above, the bug remains. Either a deep copy of the map has to be generated or the `NoteDetailsComponent` should not directly edit the input note.

sleidig:
In connection to this we should also think about replacing the "NoteDetailsComponent" or basing it on the EntityDetailsComponent that is usually used

Schottkyc137:
Is this issue still in the app? I can't seem to replicate the bug

PR #1084, opened 2022-01-18
Attendance of notes are no longer directly edited

Schottkyc137:
see issue: #797

### Visible/Frontend Changes
- [x] When changing the attendance of a child and canceling the change is no longer visible

- [x] Copy a note really deep


TheSlimvReal (approved):
Nice fix, thanks

Issue #1091, opened 2022-01-27
Note category on mobile out of screen (new design)


sleidig:
See screenshot
![IMG-20220127-WA0006.jpg](https://user-images.githubusercontent.com/1682541/151414187-9424eec8-8a57-44c7-a747-9b5bc0b5c6f6.jpg)

PR #1094, opened 2022-02-01
Use adaptive grid instead of custom implementation

Schottkyc137:
see issue: #1091 

### Visible/Frontend Changes
- [x] Menus should no longer overflow


sleidig (approved):
great :+1:

PR #1099, opened 2022-02-04
fix problem with deleted children in exports

TheSlimvReal:
### Visible/Frontend Changes
--
- [x] If a query is returning undefined, this result is not used for further sub-queries, which currently results in errors with the default notes export


TheSlimvReal:
- [ ] Apply the same fix for reports

sleidig (approved):

Issue #568, opened 2020-12-16
Replace fixed component maps with a service to dynamically register them


sleidig:
Currently we are keeping a few hard-coded Maps of components that can be used in configurable parts of the app. To make adding components more straightforward, decouple the code and enable tree shaking to remove unused parts we should move towards a dynamic registration system, where each module calls a central service to register the components it provides. That gets us closer to a plugin like system where the core does not need to know about every single feature.

I think it is probably sufficient to just wrap the map in an Angular service and use that from different parts of the app?

Fixed maps currently:

- EntityDetails -> Entity map
- Router -> component map
- Dynamic Component map

TheSlimvReal:
Maybe we could introduce a rule that modules in core are not allowed to have dependencies to other folders.

sleidig:
> Maybe we could introduce a rule that modules in core are not allowed to have dependencies to other folders.

good point :+1:   there is even tooling supporting to enforce such constraints: https://nx.dev/latest/angular/workspace/structure/monorepo-tags

PR #1102, opened 2022-02-08
Dynamic registry

Schottkyc137:
see issue: #568 

### Visible/Frontend Changes
None

- [x] All static maps now follow a general registry concept
- [x] Removed the `DynamicEntityService`


src/app/core/view/dynamic-components/dynamic-component.decorator.ts
```diff
@@ -0,0 +1,15 @@
+import { DynamicRegistry } from "../../registry/DynamicRegistry";
+import { OnInitDynamicComponent } from "./on-init-dynamic-component.interface";
+
+export function DynamicComponent<
```
sleidig, line 4: short jsDoc summary please

sleidig: and we should update the documentation: https://aam-digital.github.io/ndb-core/documentation/additional-documentation/concepts/configuration.html#views

src/app/core/entity-components/entity-utils/view-components/display-entity-array/display-entity-array.component.ts
```diff
@@ -25,9 +30,7 @@ export class DisplayEntityArrayComponent extends ViewDirective {
     const entityIds: string[] = this.entity[this.property] || [];
     if (entityIds.length < this.aggregationThreshold) {
       const entityType = this.entity.getSchema().get(this.property).additional;
-      const entityConstructor = this.dynamicEntityService.getEntityConstructor(
-        entityType
-      );
+      const entityConstructor = this.entityRegistry.get(entityType);
```
TheSlimvReal, line 33: maybe this step is not necessary if we can later pass the `entityType` string directly to the `entityMapper.load` function

TheSlimvReal (changes requested):
to be continued

src/app/core/registry/dynamic-registry.ts
```diff
@@ -0,0 +1,66 @@
...
+    return this.map.get(key);
+  }
+
+  public has(key: string) {
+    return this.map.has(key);
+  }
+}
+
+export type EntityRegistry = Registry<EntityConstructor>;
+export const ENTITIES = new InjectionToken<EntityRegistry>(
```
TheSlimvReal, line 42: So there is a trick for not having to use this injection token that I found in the CASL library ([source](https://casl.js.org/v5/en/package/casl-angular#type-script-support)).
It would look something like the following
```suggestion
export const EntityRegistry = Registry as new (beforeAddCheck?: (key: string, mapping: EntityRegistry) => void) => EntityRegistry;
```
Its rather hacky but we can write a readable helper type for it.
The service can be provided like this:
```javascript
{ provide: EntityRegistry, useValue: entityRegistry }
```
and injected like this:
```javascript
constructor(private: entityRegistry: EntityRegistry) {}
```
What do you think @Schottkyc137 ? I can also implement it if you like the approach.

src/app/core/view/dynamic-components/dynamic-component.decorator.ts
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * {
+ *   view: "DoSomething"
+ * }
+ * ```
+ * @param aliases Aliases that the component is also known under
+ * @constructor
+ */
+export function DynamicComponent<
+  Constructor extends new (...args: any[]) => OnInitDynamicComponent
```
TheSlimvReal, line 27: We could generalize the `EntityConstructor` type towards a general `Constructor` that could be used here like `Constructor<OnInitDynamicComponent>` to make this more readable.
Then the `EntityConstructor` could only be the extension with `Constructor<T extends Entity = Entity> & typeof Entity`

src/app/core/registry/dynamic-registry.ts
```diff
@@ -0,0 +1,66 @@
+import { Entity, EntityConstructor } from "../entity/model/entity";
+import { InjectionToken } from "@angular/core";
+import { ComponentType } from "@angular/cdk/overlay";
+import { OnInitDynamicComponent } from "../view/dynamic-components/on-init-dynamic-component.interface";
+
+export class Registry<T> {
```
TheSlimvReal, line 6: Did you consider extending the Map instead of using one? This way you wouldn't have to implement functions like `has()`.

TheSlimvReal (changes requested):
Very clean solution and good code. This cleans up a lot of things. Thank you for taking this up. 

The general comments I made are mostly about 
- [x] Moving Regsitry, Provider and Decorator to the same location
- [x] Typing & Dependency Injection cleanup
- [x] Common error/logging behavior

Please let me know if you want to discuss something further, I am also happy to help with implementing something

src/app/core/registry/dynamic-registry.ts
Schottkyc137: If this is working I like the idea. But will this work with multiple registries? If you provide the `EntityRegistry` and a `RouteRegistry` for example, how can angular distinguish between the two?

TheSlimvReal (approved):

TheSlimvReal:
Unfortunately Angular uses uglify in productino (`ng serve --configuration production`) to *mangle* variables, class names etc. this results in wrong names when running `viewRegistry.add(ctor.name.replace("Component", ""), ctor);`
Mangling can be turned of but only with rather hacky solutions e.g. [this](https://stackoverflow.com/questions/44160894/angular-cli-how-to-ignore-class-names-from-being-minified).
Instead we will just provide the actual strings without the fallback to the constructor name.

PR #1117, opened 2022-02-22
Additional query functions

TheSlimvReal:
### Visible/Frontend Changes
--

- [x] Add optional `school` info to `getAttendanceArray` query function
- [x] Add query function to get a aggregate attendance report for each participant
- [x] Add extra page where exports can be done based on any query, using the `ExportService`
- [x] Fix problem where the `ExportService` omits rows where for example after `getAttendanceArray` no `school` exists and then a subquery is run on the `school` property.


src/app/features/reporting/query.service.spec.ts
```diff
@@ -455,25 +464,94 @@ describe("QueryService", () => {
...
+      status: new EventAttendance(absentAttendanceStatus),
+    });
+    expect(attendanceResult).toContain({
+      participant: "maleChristianChild",
+      status: new EventAttendance(presentAttendanceStatus),
+    });
+  });
+
+  it("should create a attendance report with percentages", async () => {
+    const reportQuery = `${EventNote.ENTITY_TYPE}:toArray:getAttendanceArray:getAttendanceReport`;
```
sleidig, line 528: is there any feasible way to set up the data here in the unit test itself? with a 500+ lines test file it becomes hard to match what the "expect()" result should be

sleidig (commented):
- [x] exporting properties of school (e.g. school:toEntities(School).name) causes all rows that do not have school to be completely omitted from the export file

src/app/core/export/export-service/export.service.ts
```diff
@@ -32,25 +32,70 @@ export class ExportService {
   /**
    * Creates a CSV string of the input data
    *
-   * @param data an array of elements
+   * @param data (Optional) an array of elements. If not provided, the first query in `config` will be used to get the data.
```
sleidig, line 35:
```suggestion
   * @param data (Optional) an array of elements. If not provided, the queries in `config` will be used to get the initial data.
```

src/app/features/reporting/reporting/reporting.component.html
```diff
@@ -2,91 +2,25 @@
...
-        Download report
-      </ng-container>
-    </button>
-  </div>
-
-  <mat-progress-bar mode="indeterminate" *ngIf="loading"></mat-progress-bar>
-</div>
-
-<app-report-row *ngIf="results?.length > 0" [rows]="results"></app-report-row>
+<!--using [hidden] here to ensure that MatSort is correctly initialized-->
```
sleidig, line 14: "hidden" means the whole component is initialized and updated invisibly though, doesn't it? That would unnecessarily drain performance?

sleidig: how about moving the whole table into an <app-export-data> component - would that allow us to handle sort correctly even when using ngIf on our custom wrapper component?

TheSlimvReal: yes, it's a little less performant but updating only happens when data is assigned to the datasource and this only happens when something should actually be displayed. So I don't think performance is an issue here.

TheSlimvReal: > how about moving the whole table into an component - would that allow us to handle sort correctly even when using ngIf on our custom wrapper component?

that could work though

sleidig (approved):
Looks and works well :+1: 

If you like, maybe try to extraction of the export-data component, otherwise good to merge

Issue #1119, opened 2022-02-23
Record attendance should not be possible without a date


sleidig:
**To Reproduce**
1. Go to 'Record Attendance'
2. Delete the Date which is set by default (when empty, the field displays red with an error
3. Click on an activity to start recording attendance (this works despite the red, empty date field)
4. -- roll call starts and results in an event without a date

**Expected behavior**
Roll call should not start without a date. Maybe jump back focus to the date field when trying to click an activity.

PR #1137, opened 2022-03-11
Fix: Record Attendance without a date can no longer be edited

Schottkyc137:
see issue: #1119 

### Visible/Frontend Changes
- [x] When the date is not set or otherwise invalid, a user can no longer record an attendance

None


TheSlimvReal (approved):
Looks good. I just made it translatable

Issue #1140, opened 2022-03-14
Changes to Progress Dashboard are not saved


sleidig:
Updating tasks of the ProgressDashboard widget in the app appear in the UI but are not saved to the database. They dissappear after reloading the page.

PR #1141, opened 2022-03-14
Progress is saved correctly after closing the dialog

Schottkyc137:
see issue: #1140 

The progress dashboard component saves data after changes


TheSlimvReal (approved):
Nice, thanks for the quick fix!

Issue #1142, opened 2022-03-16
EditSingleEntityComponent without selected Entity looks very bad in default Form


TheSlimvReal:
**Describe the bug**
When having a field as `DisplaySingleEntity` in the default forms and no entity is selected yet, the form element looks very bad.

![image](https://user-images.githubusercontent.com/33730997/158588414-26f0879b-4875-428b-a5df-33eb1c8f047e.png)

What should be done is

- [ ] Form element should looks similar to neighboring elements when nothing is selected
- [x] Name of attribute should be visible if nothing is selected
- [x] Element should be correctly marked as disabled

When a entity is selected, the form incorrectly shows an edit button when editing is disabled.

![image](https://user-images.githubusercontent.com/33730997/158589737-273d8d8b-568a-455b-be90-e1145481fd1a.png)

This should also not be visible and usable.
- [x] The edit button should only be shown when `edit` is clicked in the form component

PR #1144, opened 2022-03-16
fix: edit single entity correctly works in forms

TheSlimvReal:
see issue: #1142 

### Visible/Frontend Changes
- [x] EditSingleEntityComponent is properly displayed in entity form component

--


src/app/core/entity-components/entity-utils/dynamic-form-components/edit-single-entity/edit-single-entity.component.ts
```diff
@@ -23,7 +28,10 @@ export class EditSingleEntityComponent extends EditComponent<string> {
 
   @ViewChild("inputElement") input: ElementRef;
 
-  constructor(private dynamicEntityService: DynamicEntityService) {
+  constructor(
+    private dynamicEntityService: DynamicEntityService,
+    private changeDetection: ChangeDetectorRef
```
sleidig, line 33: do we really, really need to handle change detection manually here?

TheSlimvReal: Unfortunately the asynchronous function makes this necessary

sleidig (approved):
If we can find a solution without overwriting change detection in the future that'd be great but for now this seems to work alright :+1:

Issue #1122, opened 2022-02-25
Sorting should be case-insensitive


TheSlimvReal:
**Describe the bug**
When sorting a table, lower case characters are put after upper case characters

**Expected behavior**
The sorting should be case-independent

**Screenshots**
![image](https://user-images.githubusercontent.com/33730997/155799107-addfea3d-82b5-44d3-8df1-c9062dc17a08.png)

**Additional Info**
This can probably be fixed in the `sortingDataAccessor` function.

PR #1147, opened 2022-03-21
fix: sort is not case sensitive

TheSlimvReal:
see issue: #1122 

### Visible/Frontend Changes
- [x] Sorting in tables is not case sensitive anymore

--

### DIFF
--- a/src/app/core/entity-components/entity-subrecord/entity-subrecord/table-sort.ts
+++ b/src/app/core/entity-components/entity-subrecord/entity-subrecord/table-sort.ts
@@ -1,4 +1,4 @@
-import { entityListSortingAccessor } from "./sorting-accessor";
+import { getReadableValue } from "./sorting-accessor";
 
 export function tableSort<OBJECT, PROPERTY extends keyof OBJECT>(
   data: OBJECT[],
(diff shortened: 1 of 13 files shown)


src/app/core/entity-components/entity-subrecord/entity-subrecord/table-sort.ts
```diff
@@ -0,0 +1,58 @@
...
+    active,
+  }: { direction: "asc" | "desc" | ""; active: PROPERTY | "" }
+): TableRow<OBJECT>[] {
+  if (direction === "" || !active) {
+    return data;
+  }
+  data.sort((objA, objB) => {
+    const valueA = getComparableValue(objA.record, active);
+    const valueB = getComparableValue(objB.record, active);
+    return compareValues(valueA, valueB);
```
sleidig, line 24: nice, this is so easy to read - perfect example of clean code :heart_eyes:

src/app/core/entity-components/entity-subrecord/entity-subrecord/table-sort.ts
```diff
@@ -0,0 +1,58 @@
...
+    return value + "";
+  } else {
+    return value;
+  }
+}
+
+function compareValues(a, b) {
+  let result = 0;
+  if (a === b) {
+    // keep result 0
```
sleidig, line 49: wouldn't it be easier to just do return statements in each if branch? Or is code climate complaining about too many exit paths then :see_no_evil:

sleidig (approved):
Works well and looks very clean :100:

src/app/core/entity-components/entity-subrecord/entity-subrecord/table-sort.ts
sleidig: ok, looks like CC is complaining about the complexity of this either way ... nevermind

PR #1151, opened 2022-03-22
fix: improve performance of download button

TheSlimvReal:
see issue: #XX

### Visible/Frontend Changes
--

- [x] Query service is not used when clicking download button and there is no `exportConfig`. This greatly increases the performance when working with reports that only work on a timespan


sleidig (approved):
Great :+1:   Speeds up the download incredibly

PR #1157, opened 2022-03-24
fix: export service can handle the count query

TheSlimvReal:
The export service currently breaks if the `:count` function is used in a query because it always expects the result to be an array.

### Visible/Frontend Changes
--

- [x] Exports with `:count` work


sleidig (approved):
:+1:

Issue #1168, opened 2022-03-28
Try batch create for demo data generation


TheSlimvReal:
PouchDB offers a function to save/create multiple documents in a single operation using the `bulkDocs` api ([docs](https://pouchdb.com/api.html#batch_create));
This might have a signification impact on the demo data generation where it takes around 30 seconds to create all demo data (without indexing) with the local session and ~3 seconds in the in memory session. 
A very simple prototype should be sufficient where the `entityMapper` is extended with a bulk save function and the required function for the database is added as well.

PR #1173, opened 2022-03-31
Implement batch save

Schottkyc137:
see issue: #1168 

### Visible/Frontend Changes
None

- [x] Implement `saveAll` in the entity mapper with the `bulkSave` option of PouchDB
- [x] When generating demo data, save multiple entities at once

I have also made a few tests to analyze the performance when saving multiple entities:
- The performance is much higher when batching. Using the `PerformanceAnalysisLogging` tool I found that generating demo data for about 1000 children was approximately twice as fast (around 5-6 seconds compared to around 11 seconds).
- Saving heterogeneous or homogeneous arrays where the schema only has to be extracted once does not impact the performance notably. The timing was around 4-6 seconds which seems within the margin of error
- flat-mapping the entities and then saving them instead of saving each generator separately does not improve performance as well. Timings were in the neighborhood of 5 seconds as well.


TheSlimvReal:
Nice work, good to see that the changes required quite little code.

I just tested it with the IndexedDB (`"session_type": "local" in `config.json`) and the demo data generation went down from ~35s in the current master to ~6s with the batch create. That is pretty cool!

src/app/core/entity/entity-mapper.service.ts
```diff
@@ -141,6 +141,29 @@ export class EntityMapperService {
...
+      this.entitySchemaService.transformEntityToDatabaseFormat(e)
+    );
+    entities.forEach((entity) => {
+      this.sendUpdate(entity, entity._rev === undefined ? "new" : "update");
+    });
+    const results = await this._db.putAll(rawData);
+    results
+      .filter((res) => res?.ok)
+      .map((res, idx) => {
+        entities[idx]._rev = res.rev;
```
TheSlimvReal, line 162: Let's move the `sendUpdate` call here so only entities that have been correctly saved/created are emitted.

src/app/core/database/pouch-database.ts
```diff
@@ -135,6 +141,19 @@ export class PouchDatabase extends Database {
     });
   }
 
+  putAll(objects: any[], options?: PutAllOptions): Promise<any> {
+    if (options?.force) {
+      objects.forEach((obj) => (obj._rev = undefined));
+    }
+    return this._pouchDB.bulkDocs(objects, options).catch((err) => {
+      if (err.status === 409) {
```
TheSlimvReal, line 149: I guess its not too relevant for the demo data generation but what happens in case only some of the elements fail to be saved (e.g. wrong `_rec`). Besides that I think the call to `resolveConflicts` does not make sense here as this function can only handle single documents.

src/app/core/database/database.ts
```diff
@@ -48,6 +48,13 @@ export abstract class Database {
    */
   abstract put(object: any, forceUpdate?: boolean): Promise<any>;
 
+  /**
+   * Save a bunch of documents at once to the database
+   * @param objects The documents to be saved
+   * @param options PouchDB options object as in the normal PouchDB library
+   */
+  abstract putAll(objects: any[], options?: any): Promise<any[]>;
```
TheSlimvReal, line 56:
```suggestion
  abstract putAll(objects: any[], options?: PutAllOptions): Promise<any[]>;
```

src/app/core/entity/entity-mapper.service.ts
```diff
@@ -141,6 +141,29 @@ export class EntityMapperService {
...
+    const rawData = entities.map((e) =>
+      this.entitySchemaService.transformEntityToDatabaseFormat(e)
+    );
+    entities.forEach((entity) => {
+      this.sendUpdate(entity, entity._rev === undefined ? "new" : "update");
+    });
+    const results = await this._db.putAll(rawData);
+    results
+      .filter((res) => res?.ok)
+      .map((res, idx) => {
```
TheSlimvReal, line 161:
```suggestion
      .forEach((res, idx) => {
```

TheSlimvReal (changes requested):
Really clean code. I think we can finish this up quickly.

src/app/core/entity/entity-mapper.service.ts
```diff
@@ -152,14 +152,13 @@ export class EntityMapperService {
     const rawData = entities.map((e) =>
       this.entitySchemaService.transformEntityToDatabaseFormat(e)
     );
-    entities.forEach((entity) => {
-      this.sendUpdate(entity, entity._rev === undefined ? "new" : "update");
-    });
     const results = await this._db.putAll(rawData);
     results
       .filter((res) => res?.ok)
       .forEach((res, idx) => {
```
TheSlimvReal, line 158: I think there will be a problem here with the `idx` as you might filter out some entities before.
I would recommend adding a test case where some entities fail to be saved.

src/app/core/database/pouch-database.ts
Schottkyc137: I have implemented a solution that might do the trick. Can you have a look?

TheSlimvReal (approved):

PR #1176, opened 2022-04-01
jasmine and test updates

Schottkyc137:
Implement custom matchers to test functionalities easier.
This PR is purely aimed at making testing easier. The following categories of matchers have been implemented:

- Entity matchers:
  old: `expect(entity.getId()).toBe('id')`   new: `expect(entity).toHaveId('id')`
  old: `expect(entity.getType()).toBe('EntityType')`   new: `expect(entity).toHaveType('EntityType')`
- Form matchers:
   old: `expect(someForm.valid).toBeTrue()`  new: `expect(someForm).toBeValidForm()`
   old `expect(someForm.hasErrors('err')).toBeTrue()`  new: `expect(someForm).toContainFormError('err')`
  and two more
- Several utility matchers
  old: `expect(someArray).toHaveSize(0)`   new: `expect(someArray).toBeEmpty()`
  old: - no equivalent -  new: `expect(someNumber).toBeFinite()`
  old: `expect(someObject.hasOwnProperty('prop')).toBeTrue()`  new: `expect(someObject).toHaveOwnProperty('prop')`
- Map matchers
  old: `expect(someMap.has('key')).toBeTrue()`  new: `expect('key').toBeKeyOf(someMap)`
- RxJs Observable matchers
  old: `await expectAsync(observable.pipe(first()).toPromise()).toBeResolvedTo(something)`  new: `await expectObservable(observable).first.toBeResolvedTo(something)`
  old: `await expectAsync(observable.pipe(toArray()).toPromise()).toBeResolvedTo([a, b])`  new `await expectObservable(observable).inSequence.toBeResolvedTo([a, b])`

Additionally, this PR cleans up some tests to make them more readable and use best practices (for example replace `expect(a.length).toBe(5)` with `expect(a).toHaveSize(5)`)


src/app/test-utils/entity-matchers.spec.ts
```diff
@@ -0,0 +1,38 @@
+import { Entity } from "../core/entity/model/entity";
+
+const entityMatchers: jasmine.CustomMatcherFactories = {
+  toHaveId: (util) => {
+    return {
+      compare: (entity: Entity, id: string) => {
+        const result = { pass: false, message: "" };
+        if (entity.getId() === id) {
+          result.pass = true;
```
TheSlimvReal, line 9: Generally, the functions are missing the negative (`.not...`) error messages. This should definitely be added for all of them otherwise error messages become less readable than they currently are.

TheSlimvReal: See https://jasmine.github.io/tutorials/custom_matcher#section-Failure_Messages

src/app/test-utils/form-matchers.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+    };
+  },
+  toHaveValue: (util) => {
+    return {
+      compare: (form: AbstractControl, expected: any) => {
+        const result = { pass: false, message: "" };
+        if (util.equals(form.value, expected)) {
+          result.pass = true;
+        } else {
+          result.message = "Form does not contain value " + expected;
```
TheSlimvReal, line 24:
```suggestion
          result.message = `Form ${util.pp(form.value)} does not contain value ${expected}`;
```

src/app/test-utils/form-matchers.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+    };
+  },
+  toBeEnabled: () => {
+    return {
+      compare: (form: AbstractControl) => {
+        const result = { pass: false, message: "" };
+        if (form.enabled) {
+          result.pass = true;
+        } else {
+          result.message = "Expected form to be enabled";
```
TheSlimvReal, line 50:
```suggestion
          result.message = `Expected form ${util.pp(form.value)} to be enabled`;
```

src/app/test-utils/generic-matchers.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+      ) => {
+        const result = { pass: false, message: "" };
+        const expectedDate = new Date(expected);
+        const actualDate = new Date(actual);
+        if (util.equals(expectedDate, actualDate, matchers)) {
+          result.pass = true;
+        } else {
+          result.message = `Expected date ${util.pp(
+            expectedDate
+          )} to be the same date as ${actualDate}`;
```
TheSlimvReal, line 57:
```suggestion
          )} to equal ${actualDate}`;
```
As the `equal` function is also used inside.

src/app/test-utils/map-matchers.spec.ts
```diff
@@ -0,0 +1,21 @@
+export const mapMatchers: jasmine.CustomMatcherFactories = {
+  toBeKeyOf: (util) => {
```
TheSlimvReal, line 2:
```suggestion
  toHaveKey: (util) => {
```
Lets turn this matchers around to be more in line with other matchers like `toContain` or `toHaveOwnProperty`

src/app/utils/performance-analysis-logging.spec.ts
```diff
@@ -30,7 +30,7 @@ class TestClass {
   }
 
   @PerformanceAnalysisLogging
-  other(obj?: any) {}
+  other(_obj?: any) {}
```
TheSlimvReal, line 33: Any special reason for this and similar changes before?

TheSlimvReal (changes requested):
Good idea to make the tests more readable.
Generally, in order to still have good error messages in the future, all the matchers should also have a error message for the `.not.` case (see comment)
Besides that the code generally looks quite redundant.
Maybe a small factory method could clean this up for most of the simple cases (e.g. `passed: boolean, positiveErrorMessage: string, negativeErrorMessage:string`).
Besides that I left some smaller notes regarding consistent error messages

src/app/utils/performance-analysis-logging.spec.ts
Schottkyc137: Sonarcloud suggested this. The choice was to remove the argument completely or add an underscore before the name. I chose the latter because I think that the semantics matter in this case.

TheSlimvReal: I see now, it's the problem with the unused variables

TheSlimvReal (approved):
Great, thanks for implementing the requested changes. Maybe you can present this in one of the following session @Schottkyc137 so people will start using these matchers?

Issue #1171, opened 2022-03-30
Make Aser color coding more flexible


TheSlimvReal:
Currently the `Aser` color coding is bound to the attributes `english`, `hindi`, `bengali` and `math` and displays green if they have the highest (or no) value. 
This should be changed to a more flexible way where any property that is a configurable enum can have a `passed` (or similar?) property on the enum which is then used to check whether all the properties have passed.
This should allow to the Aser entity to be freely configurable while still showing the correct colors in a table.

PR #1179, opened 2022-04-02
Flexible aser

TheSlimvReal:
see issue: #1171 
The configurable enums for Aser results now have an optional `passed: boolean` property which indicates if the subject/skill is passed. If all set properties are passed, then the warning level is OK.

Also very unrelated some updates on translation files.

### Visible/Frontend Changes
--

- [x] Aser warning level/color coding can also be used with other properties
- [x] Removed translations from admin components


src/app/core/admin/admin/admin.component.ts
```diff
@@ -111,8 +111,8 @@ export class AdminComponent implements OnInit {
     const newData = await readFile(this.getFileFromInputEvent(inputEvent));
 
     const dialogRef = this.confirmationDialog.openDialog(
-      $localize`Overwrite complete database?`,
-      $localize`Are you sure you want to restore this backup? This will
+      `Overwrite complete database?`,
```
sleidig, line 114: Is this still (and if yes, how) translated now?

And the general reminder: review, discussion and the final changelog / git history are easier if any unrelated things are in separate PRs, even when it feels ridiculous for a few lines of change only ;-)

src/app/child-dev-project/children/aser/model/aser.ts
```diff
@@ -58,40 +34,44 @@ export class Aser extends Entity {
...
       return WarningLevel.WARNING;
     }
   }
+
+  private hasPassedEverything(): boolean {
+    const configurableEnum = new ConfigurableEnumDatatype(undefined).name;
+    const schema = this.getSchema();
+    return Object.keys(this)
+      .filter((key) => schema.get(key)?.dataType === configurableEnum)
+      .every((key) => !!(this[key] as SkillLevel).passed);
```
sleidig, line 75: the previous definition used to ignore subjects that were still empty (e.g. if for some kids "bengali" is not a relevant language). The new implementation seems the require all in an explicit pass state?

src/app/child-dev-project/children/aser/model/aser.ts
```diff
@@ -58,40 +34,44 @@ export class Aser extends Entity {
...
   })
-  hindi: ReadingLevel;
+  hindi: SkillLevel;
   @DatabaseField({
     label: $localize`:Label of the Bengali ASER result:Bengali`,
     dataType: "configurable-enum",
     innerDataType: "reading-levels",
   })
-  bengali: ReadingLevel;
+  bengali: SkillLevel;
```
sleidig, line 43: Is this a good time to move the explicit subject properties out of the class and into config?

In general, this would be a good case to be merged with the HistoricalEntityData module once we can define entity types completely in config.

TheSlimvReal: The aser is also where the HistoricalEntityData was inspired by. However, the color coding is unique to the aser.

src/app/core/admin/admin/admin.component.ts
TheSlimvReal: Yes I definitely agree to the second point. Have to be more strict to myself :D 

But no this and also other things that were changed are not translated anymore as they are only internals and confuse people when the translation to a new language is done (at least this was the case for the french translation where I had to tell which ones are irrelevant).

src/app/child-dev-project/children/aser/model/aser.ts
sleidig: I know. Now that color coding is configurable maybe we wouldn't need the specific ASER implementation anymore?

Issue #1203, opened 2022-04-14
Avoid skipping participant when double clicking an attendance status in roll call


sleidig:
**Describe the bug**
Clicking on a status button multiple times during the roll call workflow causes the next participant in the list to be skipped. Instead, the system should always show the next participant, no matter how often someone clicks on a status. (The old participant is still displayed for a short duration after clicking, in order to give the user feedback about the successfully selected status. That's why multiple clicks are possible)

**To Reproduce**
1. Go to 'Record Attendance' and select an activity to begin the roll call that shows one participant after the other to track attendance
2. Double click on 'Present' for the first participant
3. The next participant is skipped (but that should not happen) and only the one after that is shown

PR #1204, opened 2022-04-14
Fix roll call

TheSlimvReal:
see issue: #1203

### Visible/Frontend Changes
- [x] Clicking same attendance multiple times in roll call does not skip next participants
- [x] Reduced transition time between participants from 750ms to 500ms 

--


sleidig (approved):
:100:

PR #1208, opened 2022-04-14
changes feed to detect entity updates

TheSlimvReal:
Instead of manually creating entity updates in the entity mapper, these are now source directly from PouchDB's changes feed which allows to also get updates about changes from the server (sync).

### Visible/Frontend Changes
- [x] Also changes received through synchronization can be shown directly (not only after navigation)

- [x] Use [changes feed](https://pouchdb.com/api.html#changes) of PouchDB to detect entity updates


src/app/core/entity/entity-mapper.service.ts
```diff
@@ -102,22 +102,27 @@ export class EntityMapperService {
    *
    * <em>Important:</em> Loading via the constructor is always preferred compared to loading via string. The latter
    * doesn't allow strict type-checking and errors can only be discovered later
-   * The first update that one will receive - immediately after subscribing - is <code>null</code>.
-   * The <code>update</code>-function takes this into account.
    * @param entityType the type of the entity or the registered name of that class.
    */
   public receiveUpdates<T extends Entity>(
```
TheSlimvReal, line 107: This could be extended so you can receive updates for **all** entity types or a list of entity types.

src/app/core/database/database.ts
```diff
@@ -103,6 +105,8 @@ export abstract class Database {
    * Closes all open connections to the database base and destroys it (clearing all data)
    */
   abstract destroy(): Promise<any>;
+
+  abstract changes(prefix: string): Observable<any>;
```
sleidig, line 109:
```suggestion
  /**
   * Listen to changes to documents which have an _id with the given prefix
   * @param prefix for which document changes are emitted
   * @returns observable which emits the filtered changes
   */
  abstract changes(prefix: string): Observable<any>;
```

sleidig (approved):
:100: 
nice, this actually makes our previous system even cleaner

PR #1216, opened 2022-04-21
update query caches on entity updates

sleidig:
Some downloads/exports have been extremely slow because the QueryServices loads all Notes if no date range is defined. This change makes the caching of entities for queries smarter and more reliable in case of changes through database syncing.

### DIFF
--- a/src/app/features/reporting/query.service.ts
+++ b/src/app/features/reporting/query.service.ts
@@ -10,6 +10,7 @@ import { ChildSchoolRelation } from "../../child-dev-project/children/model/chil
 import { ChildrenService } from "../../child-dev-project/children/children.service";
 import { AttendanceService } from "../../child-dev-project/attendance/attendance.service";
 import { EventAttendance } from "../../child-dev-project/attendance/model/event-attendance";
+import { UpdatedEntity } from "../../core/entity/model/entity-update";
 
 const jsonQuery = require("json-query");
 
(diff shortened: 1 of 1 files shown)


src/app/features/reporting/query.service.ts
```diff
@@ -75,35 +120,25 @@ export class QueryService {
...
+    if (!this.entities || Object.keys(this.entities).length === 0) {
+      for (const entityType of this.entityTypesPrimary) {
+        const loadedEntities = await this.entityMapper.loadType(entityType);
+        this.setEntities(entityType, loadedEntities);
+      }
     }
 
-    await Promise.all(dataPromises);
+    if (
+      this.entityTypesTimebound.some((t) => query.includes(t[0].ENTITY_TYPE)) &&
```
sleidig, line 132: same filter could be applied to "primary" entities as well

sleidig:
TODO: unit tests?

src/app/features/reporting/query.service.ts
```diff
@@ -20,14 +21,59 @@ const jsonQuery = require("json-query");
...
+      ...this.entityTypesTimebound.map((e) => e[0]),
+    ]) {
+      this.entityMapper
+        .receiveUpdates(entityType)
+        .subscribe((u) => this.updateCachedData(u));
+    }
+  }
+
+  private updateCachedData(u: UpdatedEntity<Entity>) {
+    const cachedEntities = this.entities[u.entity.getConstructor().ENTITY_TYPE];
```
TheSlimvReal, line 65:
```suggestion
    const cachedEntities = this.entities[u.entity.getType()];
```

src/app/features/reporting/query.service.ts
TheSlimvReal: But its a bit more challenging as we would also need to check for some (at least one) of the more specific functions: e.g. calling `getAttendanceArray(true)` needs the `ChildSchoolRelation` to be cached.

TheSlimvReal (commented):
Looks quite good already we might have to split up the query server and this entity cache though as its getting quite complex. (Maybe even with the `MockEntityMapper` which also allows to request entities synchronously).

sleidig:
also see #1697

TheSlimvReal:
Part of #1711

PR #1226, opened 2022-04-25
Edit entity label fix

TheSlimvReal:
Currently there is a bad UX when editing a single entity:
image

This is due to the label being linked to the wrong input element.

### Visible/Frontend Changes
- [x] Input element and label behave correctly

--


sleidig (changes requested):
seems like there still is something off (School History in Child Details):
image

sleidig (approved):

PR #1230, opened 2022-04-27
removed interaction types from event index

TheSlimvReal:
Since the permission changes, the indices are created too early and therefore a wrong config (the default config) was used to create the event index `by_date` which resulted in a wrong behavior in the roll call setup.

### Visible/Frontend Changes
- [x] Non-default interaction types are not lost in event index

- [x] All `EventNote` entities are included in the `by_date` index (not just the ones with a `isMeeting` interaction type)


sleidig (approved):
great that you found this! :relieved:

PR #1245, opened 2022-05-04
fix: Array values are correctly exported

TheSlimvReal:
Currently having a multi-select component with configurable enum values this is exported as `[Object object], [Object object]`. This should be changed similar to the single value case where the label is exported.

### Visible/Frontend Changes
- [x] Array values are correctly exported (especially configurable enum)

--

### DIFF
--- a/src/app/child-dev-project/children/child-block/child-block.component.html
+++ b/src/app/child-dev-project/children/child-block/child-block.component.html
@@ -3,11 +3,11 @@
   (mouseenter)="showTooltip()"
   (mouseleave)="hideTooltip()"
   (click)="showDetailsPage()"
-  [ngClass]="{ inactive: !entity.isActive, clickable: !linkDisabled }"
+  [ngClass]="{ inactive: !entity.isActive }"
   class="truncate-text"
 >
   <img [src]="entity.photo?.photo?.value" class="child-pic" i18n-alt="Alternative text for small picture" alt="Picture of a child"/>
-  <span class="underline-on-hover mat-body-1">{{ entity?.name }}</span>
+  <span class="mat-body-1">{{ entity?.name }}</span>
   <span style="font-size: x-small" *ngIf="entity?.projectNumber"
     > ({{ entity?.projectNumber }})</span
   >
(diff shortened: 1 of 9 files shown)


src/app/core/entity-components/entity-subrecord/entity-subrecord/value-accessor.ts
```diff
@@ -8,17 +8,17 @@
...
+    return value;
   }
 }
 
-function isConfigurableEnum<OBJECT, PROPERTY extends keyof OBJECT>(
-  data: OBJECT,
-  key: PROPERTY
-): boolean {
-  return typeof data[key] === "object" && data[key] && "label" in data[key];
+function isConfigurableEnum(value: any): boolean {
```
Schottkyc137, line 22:
```suggestion
function isConfigurableEnum(value: any): boolean {
function isConfigurableEnum(value: any): value is ConfigurableEnumValue {
```

See [type predicates](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates)

src/app/core/entity-components/entity-subrecord/entity-subrecord/value-accessor.ts
```diff
@@ -8,17 +8,17 @@
...
   key: PROPERTY
-): OBJECT[PROPERTY] | string {
-  if (isConfigurableEnum(data, key)) {
-    return (data[key] as any).label;
+): any {
+  const value = data[key];
+  if (isConfigurableEnum(value)) {
+    return (value as any).label;
+  } else if (Array.isArray(value)) {
+    return (value as Array<any>).map((v) => getReadableValue({ v }, "v"));
```
Schottkyc137, line 16:
```suggestion
    return (value as Array<any>).map((v) => getReadableValue({ v }, "v"));
     return value.map((v) => getReadableValue({ v }, "v"));
```

This should already be type-asserted

src/app/core/entity-components/entity-subrecord/entity-subrecord/value-accessor.ts
TheSlimvReal: Oh that is pretty cool. Thanks for pointing it out!

Issue #1244, opened 2022-05-04
Attendance Status labels not displayed with custom config


sleidig:
**Describe the bug**
On the event/note details popup the attendance list shows the attendance status of each participant correctly with the color bubble but the dropdown select field showing the label remains empty instead of showing the status label.

**To Reproduce**
reproducible not in demo mode but when a config is in the database and the `enum:attendance-status` is changed from the default. see https://dev.aam-digital.com/de/note for an example

PR #1248, opened 2022-05-04
MatSelect fix

TheSlimvReal:
see issue: #1244 

Background: Angular compares the value of a `MatSelect` with the options that are available in order to find the one that is currently selected. In our case both these values are loaded from the config but in some cases the config changes and therefore the objects might have different references. In this case Angular is not able to match the selected value with the options and does not display anything. `[compareWith]` allows to overwrite this behavior which we can use to compare the enum values more deeply ([source](https://angular.io/api/forms/SelectControlValueAccessor#customizing-option-selection)).

### Visible/Frontend Changes
- [x] Labels for `MatSelect` with configurable enum are correctly displayed when page has just loaded.

- [x] Using `compareWith` to correctly match options to values in `MatSelect`


sleidig (approved):
:100:

Issue #1182, opened 2022-04-07
Error message when entering invalid characters in DatePicker


TheSlimvReal:
**Describe the bug**
When entering a string or symbol into the datepicker, error messages are shown in the console.

**To Reproduce**
1. Go to 'Children'
2. Click on 'add new'
3. Enter a letter into 'Date of birth'
4. Open the console
5. See error

**Expected behavior**
The default behavior of the date picker

**Additional context**
The function `descriptionForValidator` cannot find the appropriate description for the `matDatePicker` because this validator is not defined through the config. This needs to be adapted in order to still show an appropriate error message.

PR #1251, opened 2022-05-05
fix: Showing error message when entering a invalid date

TheSlimvReal:
see issue: #1182 

### Visible/Frontend Changes
- [x] Entering a invalid date shows a error message in the datepicker

--


Schottkyc137 (approved):
This looks like a clean one-line solution.
However, we should probably think of a better way to deal with builtin validators like this because the current approach will fail on all predefined validators. Maybe a simple `Field Invalid` with a log  is better here?

TheSlimvReal:
@Schottkyc137 A general solution would be great but to have a user friendly error message at the same time will be difficult. A log is not really necessary as this should mostly point users towards input errors and not necessarily a problem on our side

PR #1253, opened 2022-05-05
Animated roll call

Schottkyc137:
### Visible/Frontend Changes
- [x] When selecting the next participant, the roll-call is now animated


TheSlimvReal (changes requested):
Pretty cool, thank you so much for the great work already! Love playing around with it.
I also like the gimmick with the swiping, works surprisingly well!

So far I only realize on the phone that there is always a little jump/flash when the transition starts which is hover not too critical (would be nice if this could be removed though)

On the computer it has a slightly more visible problem at the end of the transition:
grafik
This for sure should be fixed as this will also be what we are showing in demos and therefore it should look very clean there.

TheSlimvReal:
The animation now looks very good on my phone!

I think we should disable the horizontal scrolling in this whole component otherwise it can start to look weird:
image

Also I discovered that on the phone horizontal scrolling only works outside of the attendance list (so at the top and on the side of the component). This was also not the case before.

sleidig (commented):
I love it! Looks really good on desktop and Android Chrome and makes the automatic skipping intuitively clear.

Some last minor glitches:

There is a scroll to an empty space right of the current participant now:
image

The completion screen's forward/backward arrows are left aligned instead of symmetrically at the left + right edge:
image

sleidig (approved):
functionality looks great now

TheSlimvReal (approved):

PR #1256, opened 2022-05-05
fix: allow navigation to entities from disabled entity-select

sleidig:
allow navigation to entities from disabled entity-select and correctly visualize when navigation is disabled


TheSlimvReal (changes requested):
Nice, thank you for cleaning this up!

TheSlimvReal (approved):

PR #1259, opened 2022-05-07
Reworked Startup Process

TheSlimvReal:
see issue: #595 

### Visible/Frontend Changes
--

- [x] `AppConfig` is initialized before any modules are created
- [x] `Config` is provided through `DemoDataService` instead of the static file
- [x] `Config` updates are only emitted with actual config and not repeatedly with the same config
- [x] Reload app after logout
- [x] Routing should be kept (e.g. visiting `/child/1` directly)
- [ ] How to handle permissions?
- [ ] Fix demo-mode logout (not synced session)


src/app/core/demo-data/demo-data-initializer.service.ts
```diff
@@ -28,7 +28,7 @@ export class DemoDataInitializerService {
 
   constructor(
     private demoDataService: DemoDataService,
-    private sessionService: SessionService,
+    private localSession: LocalSession,
```
sleidig, line 31: Why does this require specifically the LocalSession?

src/app/core/session/session-service/synced-session.service.ts
```diff
@@ -291,18 +289,20 @@ export class SyncedSessionService extends SessionService {
...
+  public async logout() {
     this.cancelLoginOfflineRetry();
     this.cancelLiveSync();
+    this.localSession.logout();
+    await this.remoteSession.logout();
+    this.location.reload();
     this.loginState.next(LoginState.LOGGED_OUT);
-    this._localSession.logout();
-    this._remoteSession.logout();
+    this.syncState.next(SyncState.UNSYNCED);
```
sleidig, line 306: shouldn't this syncState update come from `cancelLiveSync()` maybe?

sleidig (changes requested):
I did some thorough testing in all session modes now:

- both config.json and falling back to config.default.json works smoothly to bootstrap core setup
- demo mode seems to work well, consistently for mock and local session
- some glitches in local/mock mode
- some minor glitches in synced mode

Let's discuss whether we handle the local session issues or not. We certainly should tackle the few remaining UX problems in synced mode.

-----
**Local / Mock**

- logout + reload = logged in again automatically
- admin route + logout + login as non-admin = shows page | only after reload shows 404
- ~ list is empty after switching from demo to admin user for first time directly on /child route
- ~ quickly switching from demo to demo-admin user allows login before all data is synced to demo-admin and errors like "Could not load user entity for demo-admin" at /child can occur (but is solved after all data is available)

-----
**Synced Session**

- if no config in db, no default config is picked up and a completely empty setup is shown (is this intended?)
- during initial sync, the 404 page is shown in background
- after reload/login always showing "Loading initial database" (previously the normal sync only ran in background)

TheSlimvReal:
> logout + reload = logged in again automatically

Should we add some url-flag that allows to reload without login in the demo-mode? I am not sure when this would be useful, maybe in demo mode but then you can also just turn of demo-mode

> admin route + logout + login as non-admin = shows page | only after reload shows 404

Good find. For the conflict resolution page we actually never checked for permissions. I fixed this now and the app automatically routes to the dashboard in case of missing permissions.

> ~ list is empty after switching from demo to admin user for first time directly on /child route

> ~ quickly switching from demo to demo-admin user allows login before all data is synced to demo-admin and errors like "Could not load user entity for demo-admin" at /child can occur (but is solved after all data is available)

This was the reason why we went for the page reload for the normal mode. I don't see any easy solution to these kind of problems. As they only affect special cases in the dev mode I would leave it out for now.

> if no config in db, no default config is picked up and a completely empty setup is shown (is this intended?)

As the default config is now part of the demo-mode this is intended (it's not easy to find out whether a config exists in a connected database or not).

> during initial sync, the 404 page is shown in background

Unfortunately, I did not find a solution where the current route is kept, but the 404 page is not shown. Even [disabling initial navigation](https://angular.io/api/router/InitialNavigation) still resulted in the route being lost.
We could somehow build a workaround: until config is read we show a empty page, after that we swap out this empty page with the 404 component.

src/main.ts
```diff
@@ -20,9 +20,15 @@ import { platformBrowserDynamic } from "@angular/platform-browser-dynamic";
...
   enableProdMode();
 }
 
-platformBrowserDynamic().bootstrapModule(AppModule);
+/**
+ * Loading AppConfig before bootstrap process (see {@link https://stackoverflow.com/a/66957293/10713841})
+ */
+AppConfig.load().then(() =>
+  platformBrowserDynamic().bootstrapModule(AppModule)
+);
```
sleidig, line 34:
```suggestion
await AppConfig.load();
platformBrowserDynamic().bootstrapModule(AppModule);
```

can't we use a top level `await` here? :blush:

TheSlimvReal: No, await can only be used inside (`async`) functions. We could move the whole block into a async function and then call this function.

```
async function startApplication() {
  await AppConfig.load();
  platformBrowserDynamic().bootstrapModule(AppModule);
}

startApplication();
```

sleidig: https://typescript.tv/new-features/top-level-await-in-typescript-3-8/

src/app/core/view/dynamic-routing/empty/empty.component.ts
```diff
@@ -0,0 +1,6 @@
+import { Component } from "@angular/core";
+
+@Component({
+  templateUrl: "empty.component.html",
+})
+export class EmptyComponent {}
```
sleidig, line 6: should rename this now, not empty anymore.
LoadingApplicationComponent or LoadingConfigComponent or BuildingAppComponent maybe?

Issue #953, opened 2021-08-18
Allow users to easily edit existing event from "Record Attendance" view


sleidig:
*As a user I want to quickly edit an event I recorded in order to correct mistakes or update attendance of a single participant.*

**Objective**
Quickly open the edit popup of an event from the "Record Attendance" view listing all the events or activities. This is where users first look when they want to edit an event.

To be discussed where to add the option in the UI: this should be accessible but not overbearing, it is only an advanced feature used occasionally.

Issue #1258, opened 2022-05-06
Cancelling to create a new note still adds it, no matter whether you say yes or no in confirmation


sleidig:
**To Reproduce**
1. Go to notes list
2. Click on 'add new'
3. Give the note some title but then click "Cancel"

Bug A:
4. The app asks you whether you want to discard changes, say "Yes"
5. --> A new empty note is still added to the notes list nevertheless (would expect no new note to be saved)
Bug B:
4. repeat 1.-3. and then say "No" to discarding changes
5. --> A new empty note is added (but the title you gave it seems to be discarded)

TheSlimvReal:
The app asks you whether you want to **save** the changes.

Bug A: Cannot reproduce
Bug B: Will be fixed with #1261

PR #1261, opened 2022-05-10
feat(attendance): allow reviewing/editing of an event after roll call

sleidig:
also see #953


TheSlimvReal:
@sleidig 
I have made the following changes to have the expected behavior:

- Clicking `Review details` does not automatically navigate back to the roll-call setup, otherwise the event notes are fetched from the database before the changes from the popup are saved
- Removed copy of entity in `RollCallComponent` and added correct deep copy inside `DialogWrapperComponent` to have correct `cancel` behavior
- Restructured `RollCallComponent` to not _cache_ the attendance statuses but instead always take them from the (Event-)Note

sleidig:
Thanks a lot, @TheSlimvReal !  Reviewed and tested your additions and confirming the fix for #1258 . From my side this is good to go :100:

TheSlimvReal (approved):

PR #1266, opened 2022-05-11
fix: index.html in root is correctly cached

TheSlimvReal:
Currently our offline functionality still has some limits. To reproduce:

1. Clear browser cache (at least for demo.aam-digital.com)
2. Visit https://demo.aam-digital.com/en-US/
3. Turn off internet
4. Reload page -> should work
5. Visit https://demo.aam-digital.com/
6. Does **not** work (not even `loading...` is shown)

This is because the root `index.html` (e.g. `https://demo.aam-digital.com/index.html`) is not actively cached by the service worker. 
This PR changes this and adds correct handling of requests to root without locale.
Try the same with https://test-deployment-pr-1266.herokuapp.com/en-US/ and https://test-deployment-pr-1266.herokuapp.com/ and it should work.

### Visible/Frontend Changes
--

- [x] Currently the `index.html` on root (e.g. `demo.aam-digital.com`) is not defined in the `ngsw.json` and only sometimes cached. This is no changed to `prefetch` which means it will actively be loaded and cached by the service worker

### DIFF
--- a/build/default.conf
+++ b/build/default.conf
@@ -15,7 +15,7 @@ server {
     }
 
     # Service Worker requests should not be changed
-    location ~* ^/ngsw- {
+    location ~* ^/ngsw {
         try_files $uri =404;
     }
 
--- a/build/default.conf
+++ b/build/default.conf
@@ -14,6 +14,11 @@ server {
         try_files $uri /assets/$1 =404;
     }
 
+    # Service Worker requests should not be changed
+    location ~* ^/ngsw- {
+        try_files $uri =404;
+    }
+
     location /de {
         index  index.html index.htm;
         try_files $uri $uri/ /de/index.html;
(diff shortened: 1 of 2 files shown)


build/default.conf
```diff
@@ -14,6 +14,11 @@ server {
         try_files $uri /assets/$1 =404;
     }
 
+    # Service Worker requests should not be changed
+    location ~* ^/ngsw {
+        try_files $uri =404;
+    }
```
sleidig, line 20: is that a manual request by user to ngsw/state or some internal ones by services?

sleidig (approved):
seems to work as described :+1:

build/default.conf
TheSlimvReal: Its the ones for the `ngsw-worker.js` (so the actual service worker which we request in the `app.module.ts`) and the request for the `ngsw.json` which tracks the cached files and indicates updates which the service worker itself requests.

PR #1279, opened 2022-05-18
fix(*): Student overview correctly shows assigned students

TheSlimvReal:
Currently the view that is used to load the currently active students on the school details page is bugged. It uses the current time to create the view which results in different students being displayed depending on the time at which the view was created.
This has been changed to use a in-memory filtering with the current date instead.

### Visible/Frontend Changes
- [x] Students listed in the `ChildrenOverview` component are always the same (and the active ones)

- [x] Removed unneeded view for `ChildSchoolRelation`
- [x] Removed some unnecessary functions

### DIFF
--- a/src/app/child-dev-project/children/children.service.spec.ts
+++ b/src/app/child-dev-project/children/children.service.spec.ts
@@ -154,7 +154,7 @@ describe("ChildrenService", () => {
   });
 
   it("should get the relations for a child in sorted order", async () => {
-    const relations = await service.queryRelationsOf("child", "3", false);
+    const relations = await service.queryRelationsOf("child", "3");
 
     expect(relations).toHaveSize(2);
     expect(relations[0].start.getTime()).toBeGreaterThanOrEqual(
@@ -163,7 +163,7 @@ describe("ChildrenService", () => {
   });
 
   it("should get all relations for a school", async () => {
-    const relations = await service.queryRelationsOf("school", "1", false);
+    const relations = await service.queryRelationsOf("school", "1");
 
     expect(relations).toHaveSize(2);
     const relation1 = relations.find((relation) => relation.getId() === "1");
@@ -177,7 +177,7 @@ describe("ChildrenService", () => {
     todayRelation.schoolId = "3";
     todayRelation.start = new Date();
     await entityMapper.save(todayRelation);
-    const relations = await service.queryRelationsOf("school", "3");
+    const relations = await service.queryActiveRelationsOf("school", "3");
     expectEntitiesToMatch(relations, [todayRelation]);
   });
 
@@ -187,7 +187,7 @@ describe("ChildrenService", () => {
       .filter((rel) => rel.isActive && rel.childId === "3")
       .sort(sortByAttribute("start", "desc"));
 
-    const result = await service.queryRelationsOf("child", "3");
+    const result = await service.queryActiveRelationsOf("child", "3");
     expect(result).toEqual(activeRelations);
   });
 });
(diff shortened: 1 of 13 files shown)


src/app/child-dev-project/children/children.service.ts
```diff
@@ -79,70 +76,46 @@ export class ChildrenService {
         by_child: {
           map: `(doc) => {
             if (!doc._id.startsWith("${ChildSchoolRelation.ENTITY_TYPE}")) return;
-            emit(doc.childId);
-            }`,
+            emit([doc.childId, new Date(doc.start || '3000-01-01').getTime()]);
```
sleidig, line 79: why should the fallback start date be in the future rather than 0, considering these as active by default?

src/app/child-dev-project/children/children.service.ts
```diff
@@ -79,70 +76,46 @@ export class ChildrenService {
...
-              return date >= tomorrow;
-            }
-            }`,
-        },
-        by_date: {
           map: `(doc) => {
             if (!doc._id.startsWith("${ChildSchoolRelation.ENTITY_TYPE}")) return;
-            let timestamp = (new Date(doc.start || '3000-01-01')).getTime().toString().padStart(14, "0");
-            emit(doc.childId + '_' + timestamp);
+            emit([doc.schoolId]);
```
sleidig, line 85: why not include the date here as well, so that by child/school can be used in the same way?

src/app/child-dev-project/children/children.service.spec.ts
```diff
@@ -72,30 +72,6 @@ describe("ChildrenService", () => {
 
   // TODO: test getAttendances
 
-  it("should find latest ChildSchoolRelation of a child", async () => {
```
sleidig, line 75: why are these tests not relevant anymore? Is there any other test case covering to check what will happen if 0, 1 or more relations are found and whether the correct one is picked for the child?

src/app/child-dev-project/schools/children-overview/children-overview.component.spec.ts
```diff
@@ -55,7 +55,8 @@ describe("ChildrenOverviewComponent", () => {
 
     expect(mockChildrenService.queryRelationsOf).toHaveBeenCalledWith(
       "school",
-      school.getId()
+      school.getId(),
+      true
```
sleidig, line 59: should we switch this param, so that the default (undefined or false) will result in the previous behaviour of excluding inactive ones?

src/app/child-dev-project/children/children.service.ts
TheSlimvReal: The idea is that they will be sorted first, compared to the other relations. This has no effect on the `isActive` calculation

src/app/child-dev-project/children/children.service.ts
TheSlimvReal: There is no use case for it yet (the tables filter anyway again based on the first column).

src/app/child-dev-project/children/children.service.ts
```diff
@@ -93,15 +93,15 @@ export class ChildrenService {
...
+    return this.queryRelationsOf("child", childId).then((relations) =>
       relations.length > 0 ? relations[0] : undefined
     );
   }
 
   async queryRelationsOf(
     queryType: "child" | "school",
     id: string,
-    onlyActive = false
+    onlyActive = true
```
sleidig, line 104: wouldn't it make slightly more sense to call this `includeInactive` instead and swap the logic? Sorry for digging at this again and again :see_no_evil:

sleidig (approved):
:+1:

src/app/child-dev-project/children/children.service.ts
TheSlimvReal: Extracted another function as this boolean parameter was kind of confusing anyways

Issue #1262, opened 2022-05-10
remove placeholder text on images


sleidig:
If a child photo can't be loaded, the alt text is messing up the layout: 
![image](https://user-images.githubusercontent.com/1682541/167574172-23de28c9-6a66-4a62-9e91-ebad9c513b33.png)

We should only display the "broken image" image without any additional text.

PR #1287, opened 2022-05-20
fix(*): Removed alt text from images

TheSlimvReal:
see issue: #1262

### Visible/Frontend Changes
- [x] 
- [ ] 

- [x] 
- [ ]


sleidig (approved):
not sure what accessibility guidelines recommend for such a case but we are not optimized for screen readers yet, anyway

Issue #1274, opened 2022-05-16
Participant list of event details not usable on small screen


sleidig:
on small screens, names next to the attendance status of the event/note details screen are not visible, making it almost impossible to use for reviewing the attendance list:
![image](https://user-images.githubusercontent.com/1682541/168623423-8b1a77d3-4888-449f-a816-ef3e365577c5.png)

PR #1290, opened 2022-05-25
fix(*): Improved layout of note details in mobile view

TheSlimvReal:
see issue: #1274 

### Visible/Frontend Changes
- [x] Attendance list components wrap on smaller screens

--


Schottkyc137 (approved):
Loogs good. I also like the design idea!

src/styles.scss
```diff
@@ -274,3 +274,11 @@ $standard-icon-size: 1.5em;
 .margin-bottom-small {
   margin-bottom: $standard-margin-small;
 }
+
+/**
+ * Flex
+ */
+
+.flex-grow {
```
sleidig, line 282: isn't this something FlexLayoutModule can provide?
@Schottkyc137, wanna keep an eye on this while refactoring and cleaning up the styles?

Schottkyc137: I actually do believe that this is a better way of doing things (I am using a similar approach while refactoring). I actually think that we should get rid of FxFlex altogether. We will probably have a discussion about this topic when I'm ready for review but the basic reasoning is that CSS flex classes are in most cases (all that I have encountered in the app) just as potent and – at least from my point of view – more clear than fxFlex. The latter works using "magic" by using styles that are added dynamically via scripts. This is almost never needed.

sleidig (changes requested):
The layout still looks unclean as the participant names expand to various widths based on the length of name:
image


On mobile it would be good to get the colored dot on the same line as the status dropdown and align the trash icon with something - maybe forced on a line with remarks?
image

Issue #1238, opened 2022-04-29
Dashboard widget: Notes needing follow-up


sleidig:
*As a social worker I want to immediately see the issues that need my attention from the dashboard without navigating to other screens like the notes list.*

**Objective**
Users can give notes a status of "Needs follow-up" or "Urgent" to flag them as needing attention. This allows them to be filtered for in the Notes list. To draw more attention to these notes, a dashboard widget to list these was requested. This could look similar to other widgets listing cases with absences.

**Proposed Solutions & Alternatives**
Dashboard widget showing count of notes with urgent/follow-up (allow setting via config) status. List all notes with date, title in the paginated widget details.
Allow (optionally, via config) to only display notes assigned to current user.

TheSlimvReal:
This would also go very well with user permissions where a user only has read access for the notes to which he/she is tagged.

PR #1297, opened 2022-05-28
Important notes (Dashboard Widget)

Schottkyc137:
see issue: #1238

### Visible/Frontend Changes
- [x] Add a dashboard widget that shows important notes
- [ ] Make the notes configurable to a user


src/app/child-dev-project/notes/dashboard-widgets/important-notes/important-notes.component.html
```diff
@@ -0,0 +1,58 @@
...
+          <td *matCellDef="let note" class="subject-cell">
+            {{ note.subject }}
+          </td>
+        </ng-container>
+
+        <tr
+          mat-row
+          *matRowDef="let row; columns: ['date', 'title'];"
+          class="dashboard-table-row"
+          (click)="openNote(row)"
```
TheSlimvReal, line 38: Mouse should indicate that row is clickable (changing the cursor style)

TheSlimvReal (changes requested):

sleidig (commented):
Only review UI so far: I like how this looks and works overall :smiley: 

Are the notes sorted by anything? Maybe by date makes sense?
Possibly could also indicate somehow whether it's "Urgent" or only "Needs Followup"

Schottkyc137:
All issues should be fixed now. I have also re-introduced the color as background since different notes can have a different color (red or orange)

TheSlimvReal (changes requested):
Looks solid now @sleidig can you give a last UX review? I am still struggling a little with the very strong red color on all the warning notes.

sleidig:
Very much like the widget overall. I do agree about the color being a bit overwhelming though. How about using the less heavy styling we have on the "record attendance" view, something like this:
image
(for reference, this is `.row-view { border-left: 8px solid orange; }`)

@TheSlimvReal I think you did the technical code review? I didn't look closer at that yet.

Schottkyc137:
> Very much like the widget overall. I do agree about the color being a bit overwhelming though. How about using the less heavy styling we have on the "record attendance" view, something like this: image (for reference, this is `.row-view { border-left: 8px solid orange; }`. Doesn't look perfectly clean yet without spacing, maybe we should try with an extra, empty but colored td instead)
> 
> @TheSlimvReal I think you did the technical code review? I didn't look closer at that yet.

I very much like the idea of having a more subtle hint instead of changing the whole background color. What do you think about the following?
Bildschirmfoto 2022-08-30 um 18 22 26

Though this is not more than a potential alternative; I don't have any strong opinions on either design.

TheSlimvReal (approved):
Nice work, I like where we got with this and I think a lot of organisations can use this

PR #1298, opened 2022-05-29
Add typings for jsonQuery and md5

Schottkyc137:
We have two dependencies that are not typed, i.e. had to be imported using the `const x = require("x")` style. Typing definitions exist for both of these dependencies that were added

- [x] Added typings for md5
- [x] Added typings for jsonQuery


sleidig (approved):
nice. Thanks for raising our level on code quality! :-)

PR #1299, opened 2022-05-29
Performance upgrades

Schottkyc137:
see issue: #257 

Issues of the form
> Warning: X depends on Y. CommonJS or AMD dependencies can cause optimization bailouts.

are often ignored. Each module that has an ES6-compatible replacement option was therefore replaced. This reduces the overall bundle size drastically.

Before: 

build_prev

After:

build_post

These results are generated by running `npm run build` before and after the changes.

- [x] Replace `crypto-js` with `crypto-es`
- [x] Replace `lodash` with `lodash-es`
- [x] Replace imports from `webdav` with imports from `webdav/web`
- [x] Remove `faker.ts` and replace every import with `faker/locales/en_IND`. Outsource the two utlity functions.

The changes reduce the bundle size by almost 2 MB (1.8 MB), which means that the bundle is almost 30% smaller.


src/app/child-dev-project/notes/demo-data/demo-note-generator.service.ts
```diff
@@ -3,7 +3,7 @@ import { DemoDataGenerator } from "../../../core/demo-data/demo-data-generator";
 import { Injectable } from "@angular/core";
 import { Child } from "../../children/model/child";
 import { Note } from "../model/note";
-import { faker } from "../../../core/demo-data/faker";
+import faker from "faker/locale/en_IND";
```
sleidig, line 6: can we do this with an intermediate wrapper file? Would be nice to be able to change the locale with one line instead of across dozens of files.

TheSlimvReal (approved):
Thanks for taking this up. Quite nice that we can reduce the bundle size this way.

Issue #841, opened 2021-06-07
remove config.default.json


sleidig:
Sometimes assets/config.default.json cannot be loaded (probably due to slow network?), according to sentry logs. It's not clear what happens in that case but our default config values should probably be inbuilt in the app in a failsafe way (--> environment.ts?). We may completely get rid of config.json even, if we move the settings to the config object in the db and a relative path (??) for the database connection.

- [x] move `site_name` to the database config (config-fix, where already `logo_path` etc. are defined)
- [ ] persist database url in local storage and load it from there instead of app-config
- [ ] delete all data before changing database url in local storage, to prevent accidental cross-sync
- [ ] allow passing the database url as url parameter (?db=localhost:4200/db) to enable proxy forwarding of existing instances and easy testing
- [ ] also implement session_type and demo_mode config to be handled through local storage in a similar way
- [ ] create simple UI component to allow setup of database url on new instance ?

also see #595

PR #1319, opened 2022-06-24
Relative DB URLs

TheSlimvReal:
solves #841
On default the app will now be in **demo_mode** with a **in-memory DB**. Do change this `DEMO="false"` needs to be passed to the Docker image on start.

### Visible/Frontend Changes
--

- [x] DB is only accessed through relative URL which makes the `database_url` in the `config.json` file unnecessary
- [x] Replace `database_name` with default `app`
- [x] Demo mode and session type can be configured through environment or query params
- [x] Dockerfile allows to add a default redirect to demo mode for all requests
- [x] Site name has been move to config in database


sleidig:
For me, one of the main objectives of removing the config.json is to have a multi-tenant system (one hosted app for all projects). Do we have an idea yet how to route to different DBs in the long run?

And if we remove the url, we not also move the name to the config document in couchdb and immediately delete the whole json file?

TheSlimvReal:
@sleidig 

Regarding the multi tenant system: 
I would expect the URL to the DB of a user would be provided through the authentication process. E.g. keycloak would send the DB URL after a successful login.

Regarding the full deletion of the `config.json` file:
It still holds the information for the session and demo mode. For this I don't have a clear solution yet how we would otherwise configure this as it cannot come from the database.

src/app/core/app-config/app-config.model.ts
```diff
@@ -43,13 +43,6 @@ export interface IAppConfig {
   database: {
```
sleidig, line 43: also adapt config.default.json?

TheSlimvReal:
- [ ] language select not working

build/Dockerfile
```diff
@@ -71,6 +70,8 @@ ENV PORT=80
 ENV WEBDAV_URL="http://localhost"
 # The url to the CouchDB database
 ENV COUCHDB_URL="http://localhost"
+# Whether the app should be started in demo mode. Everything else then "true" is considered as "false".
+ENV DEMO="true"
 # variables are inserted into the nginx config
-CMD envsubst '$$PORT $$WEBDAV_URL $$COUCHDB_URL' < /etc/nginx/templates/default.conf > /etc/nginx/conf.d/default.conf &&\
+CMD envsubst '$$PORT $$WEBDAV_URL $$COUCHDB_URL $$DEMO' < /etc/nginx/templates/default.conf > /etc/nginx/conf.d/default.conf &&\
```
sleidig, line 76: shall we remove $WEBDAV_URL as it got kicked out of the Angular code already?

build/default.conf
```diff
@@ -13,7 +13,22 @@ server {
...
+    }
+
     location / {
+        # Demo mode params are added to URL when not yet present and DEMO is set to true
+        if ($arg_demo != "true") {
+            set $demo "${DEMO}true";
+        }
+        if ($demo = "truetrue") {
+            # This first part is required because Heroku does some internal redirects which break the rewrite otherwise
+            # This does not work in localhost as it enforces HTTPS
```
sleidig, line 28: do we have to rewrite to https here or could we just leave it at whatever was used before?

It would make more sense to have a separate block that rewrites any mode to use https (easier to understand if separated - and applying to all modes). But I don't know whether that's doable in nginx, e.g. multiple rewrite blocks applying.

build/default.conf
```diff
@@ -13,7 +13,22 @@ server {
...
 
+    # Files should be directly accessed
+    location ~* \.[a-zA-Z]+$ {
+        try_files $uri =404;
+    }
+
     location / {
+        # Demo mode params are added to URL when not yet present and DEMO is set to true
+        if ($arg_demo != "true") {
+            set $demo "${DEMO}true";
```
sleidig, line 24: Can we make this value a bit more intuitive? I got stuck why we append "true" to "true" ... why not `set $demo "demo:${DEMO}"` and `if ($demo = "demo:true")` for example?

src/app/core/app-config/app-config.ts
```diff
@@ -34,46 +35,55 @@ import { IAppConfig } from "./app-config.model";
...
-
-  /** fallback file location of the config that is part of the project already if the "real" config file isn't found */
-  private static readonly DEFAULT_CONFIG_FILE = "assets/config.default.json";
+  /** Path for the reverse proxy that forwards to the database - configured in `proxy.conf.json` and `default.conf` */
+  static readonly DB_PROXY_PREFIX = "/db";
+  /** Name of the database that is used */
+  static readonly DB_NAME = "app";
+  /** Demo and session can be persisted in local storage */
+  static readonly DEMO_LOCAL_STORAGE_KEY = "demo_mode";
+  static readonly SESSION_LOCAL_STORAGE_KEY = "session_type";
```
sleidig, line 44: these could be private to this class only?

src/app/core/app-config/app-config.ts
```diff
@@ -34,46 +35,55 @@ import { IAppConfig } from "./app-config.model";
...
+    if (demoParam) {
+      localStorage.setItem(this.DEMO_LOCAL_STORAGE_KEY, demoParam);
+      return demoParam === "true";
+    } else if (demoStorage) {
+      return demoStorage === "true";
+    }
+    return environment.demo_mode;
+  }
+
+  private static getSession(params: URLSearchParams): SessionType {
```
sleidig, line 78: we could generalize this and have only one method "loadSettingFromUrlLocalOrEnvironment(params, key)". If we use the same key for the query param, local storage and environment.ts variable that might make it easier to follow anyway.

src/app/core/config/config-fix.ts
```diff
@@ -17,8 +17,9 @@ import { ratingAnswers } from "../../features/historical-data/model/rating-answe
 export const defaultJsonConfig = {
   "appConfig": {
     "displayLanguageSelect": true,
-    "default_language": "en-US",
-    "logo_path": null
+    "logo_path": null,
+    "site_name": "Aam Digital - DEMO",
+    "default_language": "en-US"
```
sleidig, line 22:
```suggestion
    "default_language": "en-US",
    "logo_path": null,
    "site_name": "Aam Digital - DEMO"
```

src/app/core/app-config/app-config.ts
```diff
@@ -34,46 +35,55 @@ import { IAppConfig } from "./app-config.model";
  * // no need for dependency injection here
  */
 export class AppConfig {
```
sleidig, line 37: possibly take the chance to rename this to get away from the very similar, confusing "AppConfig" (here) and "ConfigService" (config from the db)?

I just can't think of anything much clearer ... maybe "CoreConfig" or "PlatformConfig" (in contrast to the ConfigService that contains use case details). If we can't come up with something better, we can also leave it as is, though.

src/assets/config.default.json
```diff
@@ -3,10 +3,6 @@
 
   "session_type": "mock",
```
sleidig, line 4: delete config.default.json completely

src/environments/environment.ts
```diff
@@ -28,4 +30,6 @@ export const environment = {
   appVersion: "0.0.0", // replaced automatically during docker build
   repositoryId: "Aam-Digital/ndb-core",
   remoteLoggingDsn: undefined, // only set for production mode in environment.prod.ts
+  demo_mode: true,
```
sleidig, line 33: environment.ts and AppConfig with its static properties is kinda serving the same purpose, which might make it confusing for people where to put a new option. We could consider overwriting environment.ts properties during bootstrap and throughout the app always import and use environment.ts only. But in the end this might be too much hidden magic ... just a crazy thought I wanted to put out here.

src/app/core/app-config/app-config.ts
sleidig: also see my comment below on environment.ts somewhat related to this

sleidig (changes requested):
Wow, I really like how this cleans up some long standing loose ends :smiley: 
Just a bunch of clean code ideas in my comments but this looks pretty good overall.


Functionally I noticed one thing:
When I switch to "local" session and then do a hard reload (Ctrl+F5) the app gets the queryParam "session=mock" in the URL and switches back to "mock" session. Looking at the code I see that queryParam takes precedence over localStorage by design.

My intuitive expectation was that I will remain in the previous (-> local) mode. What should we have as the expected behaviour here?

build/default.conf
TheSlimvReal: Normally it definitely works the way you are asking for the problem is that it does not work with Heroku. Somehow in there a internal redirect is happening before this one which leads to a wrong URL in this rewrite (in particular loosing the `https` and adding a wrong port). This workaround is not needed for our own servers. There we could just use:

```
rewrite ^ $uri?demo=true&session=mock permanent;
```

sleidig (approved):
looks awesome! :heart_eyes:

Issue #1310, opened 2022-06-09
Entity subrecord view not updating on changes in record values stored in an array


christophscheuing:
**Describe the bug**
Usually when editing a record via the entity subrecord component (inline or in the popup), the changes are automatically shown in the table view of the entity subrecord component. However this is not the case for certain values of a records, probably all those stored not as a direct property of the record but in an array (like authors of a note or linked groups (schools) to a recurring activity).

**To Reproduce**
1. Go to 'children' in the navigation menu on the left, click on a child to view its details and click on "Notes & Reports".
2. Click on a note to view and edit its details in a popup.
3. Change the author of the note (either to another author or by adding a second author).
4. After clicking on "Save" the new author value(s) a not shown immediately in the "Notes & Reports" table, but only after reloading that child.

**Expected behavior**
The changes made to fields stored in an array (like author) should instantly take effect in the table view.

**Additional context**
This problem also affects e.g. the linkedGroups of a recurring activity (in the component yet to be built showing the recurring activities that a specific school is linked to in the details view of that school).

PR #1327, opened 2022-06-25
fix: updates of list properties correctly update in tables

TheSlimvReal:
see issue: #1310

The component was changed to not load each entity that is to be displayed individually but instead it loads all of that type and displays the ones that are found in the array.

### Visible/Frontend Changes
- [x] Tables correctly update

--


TheSlimvReal:
Same problem also occurs for `DisplayEntity`. Maybe the component should be re-structured so we dont have to handle this in every single view-component.

TheSlimvReal:
Maybe try to work with `row.record[col.id]` to have a binding on the actual value (instead of the whole entity) but this might not work for the array values

sleidig (approved):

PR #1343, opened 2022-07-05
Quick fix: Align the entity correct in the entity select component

Schottkyc137:
Due to PR #1290, the position of the displayed entity is misaligned with the remove button. This PR fixes this by giving the entity-select a `display` of `inline-block`


TheSlimvReal (approved):

TheSlimvReal:
Very nice, thanks for the quick action!

PR #1349, opened 2022-07-07
Implement a utility to impose a total ordering on configurable enums

Schottkyc137:
As discussed today, this PR implements the option to impose a total ordering on an array of configurable enums. This is useful when one wants to compare these enums, for example, to sort them.

Rationale
===

It does not make sense to impose an ordering on any enum implicitly. For example, the notion of one gender being 'less than' or 'greater than' another gender does not make sense. For other enum values, like our warning level, an ordering does make sense. It should therefore be possible to sort an array of `WarningLevel`s from the lowest warning level to the highest.

When desired, a user can call the function `imposeTotalOrdering` in the namespace `OrderedEnums` on an array of all the enum values to attach a `_ordinal` field. Later, the user can sort a 'scrambled' array of (for example) notes where the note with the highest warning level appears first.

All relevant methods are located under the namespace `EnumOrdering`. This reduces confusion with widely used names such as `compare` and `sort`

Alternatives
===

At the moment, the ordering either has to be set implicitly (via the `_ordinal` field) or using the `imposeTotalOrdering` method. This might not be a future-proof concept. Dynamic loading could be implemented for all configurable enums that require total ordering, but there does not seem to be a clean way that avoids config migration.

One unclean way is to make a `ConfigurableEnum` itself an object where the `values` are just a field of that object. Such an object could then have another field describing whether or not it can be totally ordered and the `ConfigService` could automatically add the `_ordinal` field.


src/app/core/configurable-enum/configurable-enum-ordering.ts
```diff
@@ -0,0 +1,138 @@
...
+   * ordinal value of each enum value matches the position of the element in the array.
+   *
+   * Note that this should be used 'early' in the pipeline, i.e. when the `ConfigurableEnumConfig` is just
+   * created. It is technically not incorrect to use this function on any array that contains configurable
+   * enum values, but it doesn't make sense to apply this function to 'scrambled' arrays, i.e. arrays where the
+   * ordering of values is _not_ the natural order of these values.
+   *
+   * @param values The values to impose a total ordering on
+   */
+  export function imposeTotalOrdering<T extends ConfigurableEnumValue>(
```
TheSlimvReal, line 49: Is this still needed? I thought we will just have the two cases: 

1. No ordinal value 
2. Ordinal value is already defined in config

src/app/core/configurable-enum/configurable-enum-ordering.ts
```diff
@@ -0,0 +1,138 @@
...
+   * Compare two elements that have an `_ordinal` property and thus are assumed to can sensibly
+   * be compared.
+   *
+   * - When `first._ordinal` is greater than `second._ordinal`, returns `1`
+   * - When `first._ordinal` is less than `second._ordinal`, returns `-1`
+   * - When the two ordinal values are equal, returns `0`
+   * @param first The first holder of an ordinal value
+   * @param second The second holder of an ordinal value
+   */
+  export function compare(first: HasOrdinal, second: HasOrdinal): number {
```
TheSlimvReal, line 69: When would this and the following functions be used? As far as I see now, all the processing of `_ordinal` happens in the built-in sort function?

src/app/core/entity-components/entity-subrecord/entity-subrecord/entity-subrecord.component.spec.ts
```diff
@@ -133,10 +131,10 @@ describe("EntitySubrecordComponent", () => {
 
   it("should sort non-standard objects", () => {
```
TheSlimvReal, line 132: I think this should rather be a new test: sort for normal objects **and** sort with `_ordinal` value present.

src/app/child-dev-project/notes/notes-manager/notes-manager.component.spec.ts
```diff
@@ -78,16 +79,17 @@ describe("NotesManagerComponent", () => {
...
-      id: "HOME_VISIT",
-      label: "Home Visit",
-    },
-    {
-      id: "GUARDIAN_TALK",
-      label: "Talk with Guardians",
-    },
-  ];
+  const testInteractionTypes: InteractionType[] =
+    EnumOrdering.imposeTotalOrdering([
```
TheSlimvReal, line 83: Is this required?

src/app/child-dev-project/notes/note-details/note-details.component.spec.ts
```diff
@@ -16,6 +16,7 @@ import { LoginState } from "../../../core/session/session-states/login-state.enu
 function generateTestNote(forChildren: Child[]) {
   const testNote = Note.create(new Date(), "test note");
   testNote.category = {
+    _ordinal: 0,
```
TheSlimvReal, line 19: Why is this required?

src/app/child-dev-project/notes/model/note.spec.ts
```diff
@@ -61,37 +62,34 @@ describe("Note", () => {
...
-      ] = testStatusTypes;
-
-      entitySchemaService = new EntitySchemaService();
-      entitySchemaService.registerSchemaDatatype(
-        new ConfigurableEnumDatatype(createTestingConfigService(testConfigs))
-      );
-    })
-  );
+  const testInteractionTypes: InteractionType[] =
+    EnumOrdering.imposeTotalOrdering([
```
TheSlimvReal, line 66: Is this required?

TheSlimvReal (changes requested):
I like the new approach and the simple changes that were required. 

Let's better remove all the parts that are not really required at the moment to keep this change less complex and add at least a simple test in the entity-subrecord component.

Then this should be ready for a merge.

src/app/core/configurable-enum/configurable-enum-ordering.ts
Schottkyc137: This is merely a convenience method that can be used when we define a configurable enum via some external ts-file

src/app/child-dev-project/notes/notes-manager/notes-manager.component.spec.ts
Schottkyc137: This (and the two occurrences below) are required because the type of the enum is declared as ordered.

TheSlimvReal (approved):
Seems to work very seamless now. Ready to go from my side! Thanks for the work

PR #1355, opened 2022-07-10
fix(core): using correct timezone when parsing dates

TheSlimvReal:
To test the issue on the current version:

1. Set your computers time to a timezone **before** UTC, e.g. `UTC-0600`
2. Create a new `ChildSchoolRelation` (child -> school history)
3. See the start date and the end date are one day before what you selected (even though the right time is saved to the database)

This is because `new Date("2022-01-01")` is interpreted as **UTC time**. Converting this to `UTC-0600` results in 6pm the day before (`2021-12-31`).
The very simple fix to this is to instead use a different date constructor: `new Date(2022, 0, 1)` is interpreted as first of January 2022 in the **local** time.

### Visible/Frontend Changes
--

- [x] datatypes `date-only` and `month` are interpreted with the local timezone

### DIFF
--- a/src/app/core/entity/schema-datatypes/datatype-date-only.spec.ts
+++ b/src/app/core/entity/schema-datatypes/datatype-date-only.spec.ts
@@ -2,9 +2,23 @@ import { dateOnlyEntitySchemaDatatype } from "./datatype-date-only";
 
 describe("Schema data type:Date", () => {
   it("should not fail on null values", () => {
-    const nullDateRes = dateOnlyEntitySchemaDatatype.transformToDatabaseFormat(
-      null
-    );
+    const nullDateRes =
+      dateOnlyEntitySchemaDatatype.transformToDatabaseFormat(null);
     expect(nullDateRes).toBeUndefined();
+  });
+
+  it("should correctly transform values", () => {
+    const date = new Date(2022, 0, 1);
+
+    const dbFormat =
+      dateOnlyEntitySchemaDatatype.transformToDatabaseFormat(date);
+    expect(dbFormat).toBe("2022-01-01");
+
+    const objFormat: Date =
+      dateOnlyEntitySchemaDatatype.transformToObjectFormat(dbFormat);
+    expect(objFormat).toBeInstanceOf(Date);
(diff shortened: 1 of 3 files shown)


src/app/core/entity/schema-datatypes/datatype-date-only.spec.ts
```diff
@@ -2,9 +2,23 @@ import { dateOnlyEntitySchemaDatatype } from "./datatype-date-only";
...
   });
+
+  it("should correctly transform values", () => {
+    const date = new Date(2022, 0, 1);
+
+    const dbFormat =
+      dateOnlyEntitySchemaDatatype.transformToDatabaseFormat(date);
+    expect(dbFormat).toBe("2022-01-01");
+
+    const objFormat: Date =
```
sleidig, line 17: maybe we should consider writing a test for different timezones? https://stackoverflow.com/questions/56261381/how-do-i-set-a-timezone-in-my-jest-config

src/app/core/entity/schema-datatypes/datatype-month.ts
```diff
@@ -40,12 +40,14 @@ export const monthEntitySchemaDatatype: EntitySchemaDatatype = {
...
 
-  transformToObjectFormat: (value) => {
-    value = value
+  transformToObjectFormat: (value: string) => {
+    const values = value
       .toString()
       .replace(/-(\d)-/g, "-0$1-")
-      .replace(/-(\d)$/g, "-0$1");
-    const date = new Date(value);
+      .replace(/-(\d)$/g, "-0$1")
```
sleidig, line 47: why do we still need the regex for leading zeros here if we work with the different constructor anyway?

src/app/core/entity/schema-datatypes/datatype-date-only.spec.ts
TheSlimvReal: Unfortunately, I did not find a way of achieving the same with `jasmine`.

TheSlimvReal: But if you set your computer to a different timezone and execute the test which I wrote, it will test for this bug/issue.

PR #1370, opened 2022-07-15
Display entity improvements

TheSlimvReal:
The usage of the `ChildBlockComponent`, `SchoolBlockComponent` and `DisplayEntityComponent` currently leads to inconsistent behavior for things that look the same. Sometimes there is a hover, sometimes the cursor becomes a pointer, sometimes it's clickable.

The approach here is to generally use the `DisplayEntityComponent` and also let it handle the click, hover and pointer logic.
The ...Block components are only dynamically inserted based on the entity which is displayed.

### Visible/Frontend Changes
- [x] All shown entities-blocks have the same click, hover and pointer behavior - if not disabled

- [x] Renamed routes `/users` and `/users/:id` to `/user` and `/user/:id` to be consistent with other entity types. **MIGRATION REQUIRED**

### DIFF
--- a/src/app/core/entity-components/entity-utils/view-components/display-entity/display-entity.component.html
+++ b/src/app/core/entity-components/entity-utils/view-components/display-entity/display-entity.component.html
@@ -1,4 +1,7 @@
-<span [ngClass]="{ clickable: !linkDisabled }">
+<div
+  [ngClass]="{ clickable: !linkDisabled }"
+  (click)="showDetailsPage()"
+>
   <ng-container
     *ngIf="entityBlockComponent && entityToDisplay"
     [appDynamicComponent]="{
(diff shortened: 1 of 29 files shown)


src/app/core/entity-components/entity-utils/view-components/display-entity/display-entity.component.html
```diff
@@ -1,4 +1,7 @@
-<span [ngClass]="{ clickable: !linkDisabled }">
+<div
```
sleidig, line 1: I think this was "<span>" for a reason - otherwise it can't be displayed inline properly, which we do at some occasions? Admittedly, I didn't find any UI issue in the PR deployment now, so maybe this is not relevant any more. But I do remember I explicitly changed this to span a while back.

sleidig (approved):
Nice generalization

src/app/core/entity-components/entity-utils/view-components/display-entity/display-entity.component.html
TheSlimvReal: I guess the recent style improvements made this not required any more

PR #1371, opened 2022-07-15
fix: tables adjust to mobile view

TheSlimvReal:
Tables like the `ChildrenList` or the `NoteDetails` currently don't automatically switch to the mobile view because of a wrong ordering of events.

### Visible/Frontend Changes
- [x] Opening a table on the mobile automatically shows the mobile column groups

--


src/app/core/entity-components/entity-list/entity-list.component.ts
```diff
@@ -289,9 +288,9 @@ export class EntityListComponent<T extends Entity>
   }
 
   private displayColumnGroupByName(columnGroupName: string) {
-    const selectedColumnIndex = this.getSelectedColumnIndexByName(
-      columnGroupName
-    );
+    console.log("displaying column group", columnGroupName);
```
sleidig, line 291:
```suggestion
```

Issue #1340, opened 2022-07-05
Use Keycloak for authentication


TheSlimvReal:
The current system of authenticating via the CouchDB or the ReplicationMS should be replaced by a centralized keycloak that mirrors the current behavior.
This includes

- [x] Users can login offline
- [x] User are auto-logged in if they still have a valid JWT
- [ ] Users from the existing CouchDB `_users` database can be migrated to Keycloak using the REST API and migrating the `salt` and the `derived_key` to Base64
- [x] Users can be manually added
- [x] Add repo for keycloak setup with `docker-compose.yml` file
- [x] Add JWT support in backend -> see [PR](https://github.com/Aam-Digital/replication-backend/pull/81)

PR #1381, opened 2022-07-19
Keycloak integration

TheSlimvReal:
see issue: #1340 

### Visible/Frontend Changes
- [x] Password change now happens through keycloak
- [x] Cleanup account component
- [x] Adapt Keycloak styling to match with Aam Digital corporate design (orange/white)

- [x] Login credentials are send to keycloak for validation
- [x] Keycloak provides a JWT which is used as Bearer token in all requests (PouchDB and others)
- [x] Token is automatically refreshed before its expiry while app is running
- [x] On application open, if a valid refresh token is available, user is automatically signed in
- [x] Make URL of Keycloak server configurable
- [x] Make Keycloak realm configurable (can be done through a static config file see [here](https://www.keycloak.org/docs/latest/securing_apps/index.html#_javascript_adapter))
- [x] (optional) make Keycloak client configurable
- [x] CouchDB auth is still available in the code but currently without a runtime option to actually use it

## Migration requirements
By adding `"authenticator": "couchdb"` to the config.json, the app can be updated without the need to use `keycloak`.

Otherwise the following steps need to be done.
- Users have to be migrated from CouchDB to Keycloak
- `keycloak.json` file needs to be placed in each subfolder
- **User roles have to be adapted so the order does not matter anymore**


src/app/core/user/user-account/user-account.component.html
```diff
@@ -32,115 +32,23 @@
             />
           </mat-form-field>
         </p>
-
-        <form [formGroup]="passwordForm" class="form-container">
```
TheSlimvReal, line 36: should we keep this in a component?

TheSlimvReal:
- [ ] reload on subroutes not working

src/assets/keycloak.json
```diff
@@ -0,0 +1,10 @@
+{
+  "realm": "ndb-dev",
```
sleidig, line 2: is there any danger of having this checked into git and probably built into the docker image by default. if someon forgets to provide a custom file during deployment? I guess worst case the instance will try to auth with our dev server, which is not a big issue?

sleidig (commented):
:-)

src/app/core/session/auth/auth.service.ts
```diff
@@ -0,0 +1,42 @@
...
+   * Authenticate a user without credentials based on a still valid session.
+   * @returns Promise that resolves with the user if the session is still valid, rejects otherwise.
+   */
+  abstract autoLogin(): Promise<DatabaseUser>;
+
+  /**
+   * Add headers to requests send by PouchDB if required for authentication.
+   * @param headers the object where further headers can be added
+   */
+  abstract addAuthHeader(headers: HttpHeaders);
```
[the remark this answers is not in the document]
TheSlimvReal: Unfortunately, this only works when the `HttpClient` is used for the requests. This means, we can for sure use it for any other service that we end up integrating, but `PouchDB` uses its own implementation for the requests (`PouchDB.fetch`) and it is not very straight forward to wire this through the `HttpClient`.

So yes, adding an intercepter is a way to go for the future, but at the moment we only need the tokens for `PouchDB` and there the interceptor does not help.

src/app/core/session/auth/keycloak-auth.service.ts
```diff
@@ -0,0 +1,120 @@
...
+  private keycloakReady = this.keycloak.init({});
+  private refreshTokenTimeout;
+
+  constructor(private httpClient: HttpClient) {
+    super();
+  }
+
+  authenticate(username: string, password: string): Promise<DatabaseUser> {
+    return this.keycloakReady
+      .then(() => this.credentialAuth(username, password))
```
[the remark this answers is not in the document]
TheSlimvReal: I found it more readable/less complex when not having to assign variables or using `async/await`.

src/app/core/session/session-service/remote-session.spec.ts
```diff
@@ -1,94 +1,52 @@
...
-
-  it("should disconnect after logout", async () => {
-    await service.login(TEST_USER, TEST_PASSWORD);
-
-    await service.logout();
-
-    expect(service.loginState.value).toBe(LoginState.LOGGED_OUT);
-  });
-
-  it("should assign the current user after successful login", async () => {
```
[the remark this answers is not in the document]
TheSlimvReal: I removed these tests here because they were basically testing the same things that are already covered by the abstract tests in `session-service.spec.ts`.

src/assets/keycloak.json
TheSlimvReal: Yes, I think it is fine. The keycloak instance is public and secured. It is the same as we have the `dev.aam-digital.com` as the checked-in default database. We could think about having a dev and a production keycloak though.

ngsw-config.json
```diff
@@ -50,6 +50,7 @@
     "!/**/*__*",
     "!/**/*__*/**",
     "!/db/",
-    "!/db/**"
+    "!/db/**",
```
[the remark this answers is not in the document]
TheSlimvReal: We could do this programatically in the `router-service` but as it's still (these changes were undone) just the `/db` path I think we are fine.

src/environments/environment.ts
```diff
@@ -33,4 +33,5 @@ export const environment = {
   /** The following settings can be overridden by the `config.json` if present, see {@link AppSettings} */
   demo_mode: true,
   session_type: SessionType.mock,
+  authenticator: "keycloak", // set to "couchdb" if auth should happen with CouchDB directly
```
TheSlimvReal, line 36: should we make the default the other way round -> `CouchDB` on default and `Keycloak` has to be activated? This would make the migration with `config.json` caching issues easier.

src/app/core/session/auth/auth.service.ts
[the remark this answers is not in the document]
sleidig: as this is rather specific for pouchdb then - how about reversing it:
just have a `getAccessToken()` here, and let the SyncedSessionService do the header setting?

On the other hand, this really amounts to the same thing. I am totally fine with you implementation here.

src/app/core/session/session-service/remote-session.spec.ts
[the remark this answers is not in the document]
sleidig: ah, this is tested through the call to `testSessionServiceImplementation()` then. Great :+1:

src/assets/keycloak.json
sleidig: yes, in fact we were talking about splitting servers also. Let's plan for that and use a different tld (maybe .net) for testing and dev

src/app/core/session/auth/auth.service.ts
```diff
@@ -6,6 +6,7 @@ import { DatabaseUser } from "../session-service/local-user";
  * Implement this for different authentication providers.
  */
 export abstract class AuthService {
+  someProp = "this is some value";
```
sleidig, line 9: ?

src/app/core/user/user-account/user-account.component.ts
```diff
@@ -33,38 +34,33 @@ export class UserAccountComponent implements OnInit {
...
+  couchdbAuthService: CouchdbAuthService;
 
   constructor(
     private sessionService: SessionService,
     public authService: AuthService
-  ) {}
+  ) {
+    if (this.authService instanceof CouchdbAuthService) {
+      this.couchdbAuthService = this.authService;
+    }
```
sleidig, line 47: wouldn't it be more straight forward to let the password component inject AuthService itself through DI and then simply disable itself in case it's not the right type, instead of leaking all this into the user account component?

src/environments/environment.ts
sleidig: yes, let's do that :+1: 
Also in general I think it's nice to have the most lightweight setup as default.

sleidig (approved):
Nice cleanups. This looks great.

PR #1385, opened 2022-07-30
Release 3.9.0

TheSlimvReal:
MIGRATION REQUIRED:

- [x] move `site_name` form `config.json` to db config
- [x] set `default_language` in default config if different than `en-US`
- [x] add `"demo_mode": false` to all `config.json` files that don't run in demo mode


TheSlimvReal:
- [ ] Current issue: When having the service workers installed for a old version, the demo mode is not activated in this version. This is because the demo mode is activated through a redirect which does not happen if the index file is served through the service worker.
This can be solved two ways:

1. Make a code change, that automatically de-registers the service worker for the `demo.aam-digital.com` url once
2. Replace the service worker file with the `safety-worker.js` to achieve the same

Question is, how long do we keep this active? In 1. we could base it on the current version.

- [ ] `DEMO=false` needs to be added to the default `docker-compose.yml` file for deployments and existing instances with a database need to get this setting too.

PR #1387, opened 2022-08-02
fix: permissions are correctly applied in popup form

TheSlimvReal:
Currently the row details component evaluates `update` permissions on the input entity. When clicking the `+` button on a table a new entity is created which can result in the popup being in disabled mode (e.g. a entity needs to have a certain value for the user to have `update` permissions). 
This is fixed by detecting, whether the entity is currently being created or updated.

Also, I discovered a small but in the `ActivitiesOverviewComponent`:

1. Create a new activity in the component
2. Edit this activity
3. The activity disappears from the table

This happens because the newly created entity is not present in the `records` array of the `ActivitiesOverviewComponent`.

### Visible/Frontend Changes
- [x] `RowDetailsComponent` is never disabled when creating a entity

--

### DIFF
--- a/src/app/child-dev-project/schools/activities-overview/activities-overview.component.ts
+++ b/src/app/child-dev-project/schools/activities-overview/activities-overview.component.ts
@@ -6,6 +6,7 @@ import { EntityMapperService } from "app/core/entity/entity-mapper.service";
 import { Entity } from "app/core/entity/model/entity";
 import { DynamicComponent } from "app/core/view/dynamic-components/dynamic-component.decorator";
 import { OnInitDynamicComponent } from "app/core/view/dynamic-components/on-init-dynamic-component.interface";
+import { delay } from "rxjs";
 
 @DynamicComponent("ActivitiesOverview")
 @Component({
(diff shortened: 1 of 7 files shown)


src/app/child-dev-project/schools/activities-overview/activities-overview.component.ts
```diff
@@ -24,30 +24,31 @@ export class ActivitiesOverviewComponent implements OnInitDynamicComponent {
...
-    ).filter((activity) => activity.linkedGroups.includes(this.entity.getId()));
+    this.initializeRelatedActivities();
     this.entityMapper
       .receiveUpdates(RecurringActivity)
       .subscribe((updateEntity) => {
         if (updateEntity.type === "update") {
-          this.records = this.records.filter((activity) =>
-            activity.linkedGroups.includes(this.entity.getId())
-          );
+          this.initializeRelatedActivities();
```
sleidig, line 41: seems a bit of an overkill to reload every entity of that type - we might be able to just replace (i.e. update) the entity that was updated? Could we use the same as in "EducationalMaterials" for example: `this.records = applyUpdate(this.records, update);`?

But otherwise I think the approach here is an acceptable tradeoff.

sleidig (approved):

PR #1388, opened 2022-08-04
Clarify comment

Schottkyc137:
A comment in the `EditSingle EntityComponent` didn't actually explain the double `inputs` and was confusing


TheSlimvReal (approved):
Good clarification. Thanks for taking the time!

PR #1395, opened 2022-08-11
`config.json` rollback

TheSlimvReal:
Nginx creates some problem when visiting a subroute without demo mode active.
This is probably due to the issues mentioned [here](https://www.nginx.com/resources/wiki/start/topics/depth/ifisevil/)

### Visible/Frontend Changes
--

- [x] Re-added `config.default.json` and connected handling, only `demo_mode` and `session_type` are defined in it

### DIFF
--- a/src/app/core/app-config/app-settings.ts
+++ b/src/app/core/app-config/app-settings.ts
@@ -30,29 +30,14 @@ export class AppSettings {
   /** file location of the config file to be created by the administrator */
   private static readonly CONFIG_FILE = "assets/config.json";
 
-  /** fallback file location of the config that is part of the project already if the "real" config file isn't found */
-  private static readonly DEFAULT_CONFIG_FILE = "assets/config.default.json";
-
   /**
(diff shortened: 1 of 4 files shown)


src/app/core/app-config/app-settings.ts
```diff
@@ -28,59 +27,22 @@ export class AppSettings {
...
-    const localStorageValue = localStorage.getItem(key);
-    if (paramValue) {
-      localStorage.setItem(key, paramValue);
-      return paramValue;
-    } else if (localStorageValue) {
-      return localStorageValue;
-    }
+  static async initRuntimeSettings() {
+    const res = await fetch(this.CONFIG_FILE)
+      .catch(() => fetch(this.DEFAULT_CONFIG_FILE))
```
sleidig, line 43: if we keep the environment.ts as fallback/default values, can't we do without the config.default.json? (i.e. only override environment if config.json exists

sleidig (commented):
The test build is not loading for me, so I couldn't check it functionality wise. Code looks good though

PR #1399, opened 2022-08-16
Refactor Custom Tooltips + Enable customizable tooltips

Schottkyc137:
Currently, there are two places where we have custom tooltips (i.e. a tooltip that is not just text):

1) The child block
2) The attendance block

The logic (and code) that shows and hides these tooltips is very similar. This PR aims to simplify and refactor tooltips while giving developers better access to create custom tooltips.

### Visible/Frontend Changes
- [x] The tooltip visible when hovering over a child has been visually updated
- [x] The tooltip visible when hovering over an attendance has been visually updated

- [x] The logic to create a custom tooltip has been simplified, refactored into its own component and directive and applied to the examples mentioned above


src/app/child-dev-project/children/child-block/child-block.component.html
```diff
@@ -1,44 +1,15 @@
 <span
   *ngIf="entity"
-  (mouseenter)="showTooltip()"
-  (mouseleave)="hideTooltip()"
-  [ngClass]="{ inactive: !entity.isActive }"
-  class="truncate-text container"
```
TheSlimvReal, line 6: Why did you remove this container class? It seems like this creates the vertical alignment of the name and the project number

src/app/core/common-components/customizable-tooltip/customizable-tooltip.directive.ts
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  @HostListener("mouseenter")
+  onMouseEnter() {
+    this.show();
+  }
+
+  @HostListener("mouseleave")
+  onMouseLeave() {
+    this.hide();
```
TheSlimvReal, line 118: There is a small problem with that. When you open the attendance-block tooltip at the bottom end of the screen, the tooltip is placed on top of the block component. This itself is not a problem but it results in this function being executed because the mouse leaves the block-component. This makes the tooltip flickering.

Any idea on how to solve this? It looks like we need to detect whether the mouse moved at all or make sure that the `mouseenter` event is executed on the tooltip.

TheSlimvReal (changes requested):
I really like the refactoring and the nice abstraction that you created. Especially the tests look amazing! Thanks for the effort. 
I left some smaller remarks as well as a glitch that I found.

TheSlimvReal (approved):
Very nice and clean work. Good job!

Issue #1357, opened 2022-07-11
EditSingleEntity in EntityForm looks bad


TheSlimvReal:
**Describe the bug**
When using the edit single entity component in a `EntityForm` (e.g. child-details) it looks very bad

![image](https://user-images.githubusercontent.com/33730997/178205935-5996b55f-338a-45a1-bfde-e18c7352b79d.png)

It should not take that much vertical space and look similar to the when it is used in the `RowDetails` popup (e.g. child -> education).

sleidig:
fixed with #1434

Issue #1105, opened 2022-02-14
Improve layout of tooltip icons


TheSlimvReal:
**Describe the bug**
The icons to display the tooltips are placed in a rather bad way. It is not clear to which form element they belong. They should appear in the same line as a form element.
The problem with the tooltip is the quite bad native behavior on mobile devices: You have to longpress the icon in order to see the additional information. This is very unintuitive. Therefore, we created our own tooltip icon that show the tooltip on-click and on-hover, however the styling of this is not very good yet.

**Screenshots**
![image](https://user-images.githubusercontent.com/33730997/153941525-e30b3a4e-7772-4256-9766-8c3249338f16.png)
![image](https://user-images.githubusercontent.com/33730997/153941549-e3508310-2898-4b0a-8b55-338df9945e88.png)

Schottkyc137:
Can you give an example of where this behavior can be found in the demo app?

sleidig:
I can't find any but maybe we should include one in the demo in general. If you work on this, please update child.ts, adding a `description`:
```
  @DatabaseField({
    label: $localize`:Label for the project number of a child:Project Number`,
    labelShort: $localize`:Short label for the project number:PN`,
    description: $localize`:Tooltip explaining project number field:You can assign a unique id to your beneficiaries if necessary`
  })
  projectNumber: string;
```

This will then show a tooltip on the child details view.

PR #1402, opened 2022-08-18
CSS Refactoring (Or: UI Redesign 2 )

### LINKED ISSUE #1375: Menu width and margins
The fixed bottom menu section (logout, etc.) is missing margins between icons and labels (e.g. icon - - "Profile") and not growing the menu width properly to keep longer labels in view (e.g. in the German version, see screenshot)

image

Schottkyc137:
see issues: #1375, #1357

This PR includes several minor updates to

### The UI as such

The main intent here is to iron out some rough edges that were left out by the UI Redesign. Examples include:
- [x] Limiting the width of long texts. This improves readability
- [x] Improving the general UI of the support component and the changelog component
- [x] Small improvements to the activity card component (subjective)
- [x] Small improvements to the whole `UI` component which makes `position: absolute` obsolete and fixes #1375 

### The Architecture
- [x] Remove `flex-layout`. Since CSS flex is flexible enough (pun intended) this is no longer needed, messes with styles, and is an extra dependency.
- [x] Remove empty (and now empty) CSS Files – SonarCloud will complain otherwise
- [x] Remove many containers as they mainly pollute the DOM with `div`s that are mostly not necessary (the `:host` pseudo-selector can do magic here)
- [x] (Rarely) replace `div`s or other generic containers with more specific HTML elements 
- [x] Refactoring of the `AlertService` so that there is no issue using `Alert`s and `MatSnackBar` together. Use the default styling so it is consistent

### Creation of several Helper-Components
- [x] `app-dialog-close` is a generic close button used in pop-ups. Also added this helper to elements that didn't have it but should have
- [x] A utility to simplify highlighting the left border of an element (which is common throughout the app)
- [x] A `pill` element which simply places a pill-shaped background around an element. This replaces usages of `mat-chip`s and `mat-chip-list`s

### The handling of styles
This PR redefines how styles are used around the app. The following has been changed:
- [x] Removed a lot of common styles and made them global styles.
- [x] Used variables to make most magic numbers obsolete (but kept them where variables are too burdensome)
- [x] Replace the usage of `@import` with `@use` (`@import is deprecated`)
- [x] Fix the import of `@angular/material` and replace it with `@use`


src/app/core/entity-components/entity-subrecord/row-details/row-details.component.html
```diff
@@ -1,26 +1,22 @@
-<fa-icon
-  mat-dialog-close
-  class="cancel-icon"
-  icon="times"
-  mat-ripple
->
-</fa-icon>
+<app-dialog-close mat-dialog-close></app-dialog-close>
```
TheSlimvReal, line 1: Somehow this button is always focused on opening the dialog

TheSlimvReal (changes requested):
Nice work on cleaning this all up. I definitely support this approach of having just one flex system (css flex) and also using a more general approach to styling.

UX-wise I have the following open question, maybe @sleidig can also have a look

- [x] Child details > attendance: Especially the bold header makes this very flashy (there is no other bold header on the page) and I think I preferred the old, more gentle style
- [x] Recurring activities > attendance: The color highlighting in the calendar is missing
- [x] Record attendance > completed: The green color of the completed box is very strong and also the horizontal margin between the box and the button below is missing
- [ ] Is there a reason why you created the `pill` component? E.g. in entity select it is still using the `mat-chip` component. This creates an inconsistent design for the entity blocks.
- [ ] A lot of translations are lost for now. This is probably because the html code changed. Let's see how many can't be merged and we will have to re-enter.
- [x] recurring activity > attendance > monthly attendance popup: This component is missing the dialog close button
- [x] note details > 3 dot menu: The icon for download is black and white. Is this intended?
- [x] login component: The input form does not adjust to smaller screen sizes
- [x] On mobile/iPhone: If I scroll/drag on the top menu bar or the quick-action button, I start scrolling the overlay and the menu bar becomes invisible (can provide screenshots if necessary). Not very critical though and only slightly worse than before 😄 
- [x] On mobile child > attendance: Margin between overall attendance and monthly attendance table is very small
- [x] Attendance > record attendance: The mat cards dont properly adjust to screensize on a width between 350 and 400 px
- [x] Attendance > record attendance > roll call: The progress number is not horizontally aligned with the skip/back buttons
- [x] on mobile profile: The password change form is much small than it needs to be, screensize is not used properly
- [x] on mobile latest changes: The popup buttons are missing horizontal margins (old problem but as you were redesigning this component we might as well fix this too)

Sorry for the long list of issues 😅 
On the other hand I did not write down the things that are now **much better** than before which would definitely be the longer list. Especially on my phone a lot of things are now looking way better/are finally properly visible. Thank you soo much for taking this up. I hope that these points are not too complicated to fix.

TheSlimvReal:
Does this also solve #1105?

TheSlimvReal:
I am not 100% sure about the issue in the test but it seems like the `sort.sort` call always goes to the next sorting direction (none -> asc -> desc -> none). As the sort somehow gets executed a few times, it always ended up switching to desc at this test. I solved it now but simply reseting the sort state to `none` before calling sort.

TheSlimvReal:
Regarding the E2E tests, I can take care of fixing them once the code is stable.

Schottkyc137:
> I am not 100% sure about the issue in the test but it seems like the `sort.sort` call always goes to the next sorting direction (none -> asc -> desc -> none). As the sort somehow gets executed a few times, it always ended up switching to desc at this test. I solved it now but simply reseting the sort state to `none` before calling sort.

Thanks a lot for the fix!



> Regarding the E2E tests, I can take care of fixing them once the code is stable.

I wouldn't worry about it too much. At this moment there's still a lot that needs fixing and testing so there will be a lot of changes at some point

TheSlimvReal (changes requested):
I really like what you did with the reporting component!


Further UX Issues that I found:

- [x] '+' Button in entity subrecord has a uncentered border (e.g. child -> notes)
- [x] When entity subrecord table is empty, the '+' button is in the middle of the screen (e.g. school -> activities)
- [ ] On my phone, part the navigation bar is higher than the screen size, so the bottom part is a little cut off. I am happy to provide screenshots
- [x] On my phone, the form component has a horizontal scroll, form fields are cut off on the side.
- [x] On my phone, the pages are missing a little bit of a bottom margin (e.g. the screen ends directly with the white part of the entity subrecord, instead a small margin should be at the end of the component to show that its really the end of the screen)

Besides that, a lot of really nice improvements to the general look and feel. I am looking forward to merging this!

sleidig:
Adding a few more random observations :blush: 

- dialog close button seems automatically focused (and thereby visually highlighted when opening a popup (maybe that's wanted but for me it draws too much attention to that corner)
- ... and the highlight should have at least a slight margin visually from the boundary
image

Shall we give a fixed width to the colored enum displays? I think this screen looks very cluttered at the moment:
image

The title of activities on the "record attendance" screen feels quite heavy to me. Maybe it's just being used to something else. But I couldn't find any other place where we use bold text of that size otherwise in the app?
image

src/app/core/ui/ui/ui.component.scss
```diff
@@ -65,8 +65,12 @@ body {
...
   height: 100vh;
 
-  /* mobile viewport bug fix */
-  min-height: -webkit-fill-available;
+  /* mobile viewport bug fix. */
+  // The viewport height on iOs is not necessarily the visible screen size.
+  // There might be a bar on the top of the screen. The following line
+  // fixes this, but since this is no official css feature, this line should
+  // be replaced as soon as a nicer solution exists.
+  height: -webkit-fill-available;
```
TheSlimvReal, line 73: This seems to break the height on normal devices. At least on my computer the navigation bar is far too long and needs to be scrolled

sleidig: And probably related: the "Record Attendance" screen is now too short (i.e. white space at bottom half of screen)

sleidig: and the primary action (blue bubble button on the bottom left) is not fixed but scrolling along other content

TheSlimvReal (changes requested):
Really nice changes. I went through the code now and I am impressed by all the generalisations that you achieved. Good job. I only have marked a few small issues, they are not too important.

Generally, you added some nice documentation to the global CSS classes which is very helpful, but I think it would be great if you could also give a little overview of the most relevant changes once this is merged. Important topics would be, what classes are very useful, where do we find them and what should we consider when designing/layouting a new component?

src/app/core/ui/ui/ui.component.html
```diff
@@ -16,50 +16,40 @@
...
+  </div>
 </mat-toolbar>
 
-<app-primary-action *ngIf="isLoggedIn()"></app-primary-action>
-
 <!-- MAIN NAVIGATION + CONTENT -->
 <mat-sidenav-container
-  class="app-content mat-typography primary-background"
   (backdropClick)="closeSidenavOnMobile()"
-  autosize
```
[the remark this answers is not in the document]
TheSlimvReal: Seems to work 👍

TheSlimvReal (approved):

PR #1403, opened 2022-08-22
updated all dependencies and removed casl angular overwrite

TheSlimvReal:
see issue: #XX

### Visible/Frontend Changes
- [x] 
- [ ] 

- [x] 
- [ ]


TheSlimvReal:
Currently this [lint issue](https://github.com/nrwl/nx/issues/11653) makes the build fail. It looks like a fix should be available soon.

sleidig (approved):
looks good, although I can't completely oversee where these dependencies might have further effects (according to semantic versioning they really shouldn't create problems, though) :+1:

Issue #1374, opened 2022-07-15
Save button in details view not visible when scrolling on mobile


sleidig:
When editing / creating a record (e.g. a Child) on a small screen, the forms often require scrolling. The save button is on top of the page but not fixed on screen. It disappears from view when scrolling, which makes it hard to find it again for users.

![image](https://user-images.githubusercontent.com/1682541/179273400-3fc92ef1-1867-4a14-851e-038aed25cef6.png)

In the Notes view (popup) we have a better UX. The save and other buttons are fixed, always visible and only the form container inside is scrollable.

Apart from keeping the save buttons always in view, it may be useful to also keep the heading and tab visible at all times.

sleidig:
done with PR #1404

PR #1404, opened 2022-08-25
Made the form overflow instead of whole page

MattKNichols:
see issue: #1374


src/app/core/entity-components/entity-form/entity-form/entity-form.component.scss
```diff
@@ -21,6 +21,9 @@ $max-screen-width: $min-block-width + $margin-main-view-right + $margin-main-vie
 .columns-wrapper {
   @include adaptive-grid($min-block-width, $max-screen-width);
   align-items: start;
+  position: relative;
+  height: calc(100vh - 240px);
```
sleidig, line 25: because the 240px depend on the space taken up by our heading of the view above, this will stop working (i.e. start showing a second scroll bar) when we change things about other components outside of this.

The idea of CSS certainly is to lay out things well stacked and use relative measures ... but with all our components inside each other this is a huge pain in the a**. So let's go with this for now :wink:

sleidig: @Schottkyc137 should we consider reworking things from the root inwards, so that the overall views always exactly fill the screen and inside components like this one take up relative space and display their own scrolling?

sleidig (approved):
tested this in various screen sizes and it seems to work reliably. Thanks, this is a big UX improvement! :tada: 

only on mobile there is a slight glitch - the component is still too long and scrolling to the very bottom, the title (name of beneficiary above the tabs) gets scrolled out. This is not a huge issue, though.

src/app/core/entity-components/entity-form/entity-form/entity-form.component.scss
Schottkyc137: > @Schottkyc137 should we consider reworking things from the root inwards, so that the overall views always exactly fill the screen and inside components like this one take up relative space and display their own scrolling?

At some point, I think that this is the way to go. That being said, our large HTML tree certainly complicates this (because a lot of elements need to have their height set appropriately) but in the end, it might be worth it.

Issue #1341, opened 2022-07-05
Password reset option


TheSlimvReal:
A password reset option in the login screen should allow to reset a password through a verified email address.

- [ ] Add password reset option in login screen which allows to enter a email and on submit, sends a password reset email.
- [ ] Add UI component for adding emails to user accounts
- [ ] Add emails to existing user accounts

PR #1413, opened 2022-08-31
Set email and password forget options

TheSlimvReal:
see issue: #1341 
All changes are only visible if Keycloak is used as `authenticator`.

### Visible/Frontend Changes
- [x] Forget password button on login screen
- [x] Form to send password reset email
- [x] Form to add an email to the user account

- [x] Further functions that use the new [account-backend](https://github.com/Aam-Digital/account-backend)


src/app/core/session/auth/keycloak/auth-interceptor.service.ts
```diff
@@ -0,0 +1,23 @@
...
+  HttpRequest,
+  HttpHandler,
+  HttpEvent,
+  HttpInterceptor,
+} from "@angular/common/http";
+import { Observable } from "rxjs";
+import { AuthService } from "../auth.service";
+
+@Injectable()
+export class AuthInterceptor implements HttpInterceptor {
```
sleidig, line 12: JSDocs? Also maybe briefly mention where/why the addAuthHeader function comes in as an alternative to be directly called?

src/app/core/session/login/login.component.html
```diff
@@ -16,7 +16,7 @@
   -->
 
 <mat-card class="login-panel centered mat-elevation-z2">
-  <mat-card-title i18n="Sign in title"> Please Sign In </mat-card-title>
+  <mat-card-title i18n="Sign in title"> Please Sign In</mat-card-title>
```
sleidig, line 19: somewhat unrelated but can we just trim spaces from the username as part of this PR as well? We recently had user support cases struggling there again.

sleidig (commented):
Code looks good :+1: 

I did not do a ux review or functional testing yet.

sleidig (changes requested):
Code looks good (had already reviewed most of it earlier).

Two last things we should cover before merging and declaring this completely done:
- adapt templates for emails to be more user friendly and specific to us
- display error msg when changing email in profile fails (and fixing the permissions to allow email change)

PR #1426, opened 2022-09-06
fix: labels in note details can be configured

TheSlimvReal:
Currently, all the labels in the note details component are hard coded. To make this more in-line with the other labels that are used throughout the app the component now uses the (configurable) labels of the respective properties. This means the labels are the same everywhere in the app.

### Visible/Frontend Changes
- [x] Labels in note details can be configured and match the other usages
- [ ] 

- [x] 
- [ ]


sleidig (approved):
:+1:    also works in translations

two small glitches:
- "Groups" not translated"
- placeholder in entity-select cut off
image

Issue #1405, opened 2022-08-26
allow preview of typed password


sleidig:
_As a user failing to log in I want to be able to see the password I typed into the login form in clear text to check for typos._

Many apps offer a small "eye" icon in the password field that switches the typed password from **** to clear text. That way a user can check whether she really typed the correct password. We should implement something like this as well.

PR #1435, opened 2022-09-08
Implemented show password toggle

christophscheuing:
see issue: #1405

### Visible/Frontend Changes
- Added a "Show/Hide password toggle" icon.
<-- I'm not sure about the color for this icon. Should this be primary, accent or warn or even just black?
<-- I'm also not quite sure about the logic of this icon: Should the icon display the current state or the state that can be achieved by clicking on the icon?


christophscheuing:
Not sure why some of the tests are still failing on GitHub. Maybe @TheSlimvReal can have a look at this.

TheSlimvReal:
It seems like the world out there is also quite undecided how the icon should look like:

Twitter shows the icon of the state that can be achieved:
Screenshot 2022-09-20 at 09 45 40

Facebook shows the icon for the current state:
Screenshot 2022-09-20 at 09 47 34

I feel like the the behaviour of twitter and the one that you have currently implemented is more intuitive.

src/app/core/session/login/login.component.scss
```diff
@@ -29,3 +32,7 @@
   color: red;
   font-weight: bold;
 }
+
+.password-icon {
+  color: mat.get-color-from-palette($ndb-app-primary);
```
TheSlimvReal, line 37: I think it would look cleaner if the color would be the same as the other elements in the form. So something black-/gray-ish

christophscheuing: Yes, of course; I'm very fine with that.

TheSlimvReal (approved):
Nice work, thanks

Issue #1429, opened 2022-09-07
Date parsing does not work with different locales


TheSlimvReal:
**Describe the bug**
When changing the locale of the app, it is not possible to write dates according to this locale.

**To Reproduce**
1. Switch language to 'de'
2. Click on 'Kind kinzufügen'
3. In 'Geburtsdatum' enter '15.3.1999'
4. See error 'Bitte ein korrektes Datum eingeben'
5. If the same date is selected through the date picker, it is displayed in the input as '15.3.1999'

**Expected behavior**
The date input should support the same date format as the one that is displayed when picking a date through the date picker.

**Additional context**
This is because the `MatNativeDateAdapter` does not support different parsing schemes ([source](https://material.angular.io/components/datepicker/overview#choosing-a-date-implementation-and-date-format-settings)). A solution would be to use a different date adapter, e.g. Moment, Luxor or fsn. As we would like to switch away from moment.js (#1062 ) it would probably make more sense to use a different one. Maybe it is also possible to only change the adapter for the datepicker, but keep using the normal date object in the rest of the application. At the moment just setting the date adapter to `MatMomentDateAdapter` results in errors because some functions expect a plain `Date` object.

sleidig:
Some additional hint to reproduce this: use a day larger than 12 (e.g. 15 as above). lower dates actually work just fine for me, with the exception that it switches day and month after typing any change :see_no_evil:

PR #1437, opened 2022-09-15
Allow dates to be entered in different locales

Schottkyc137:
see issue: #1429

- [x] Dates can now be entered in any locale supported by moment (a lot)

This change uses the `MomentDateAdapter` from `moment.js`. At some point, we want to get rid of `moment.js` altogether and replace it with something that is long-term supported. However, this change is very tedious and lots of caution needs to be taken to correctly replace the `moment.js` dependency with something different. Therefore, I propose to use this date-adaptor and when the time comes to replace moment, we can also replace the adaptor together with moment.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19,6 +19,7 @@
         "@angular/forms": "^14.2.0",
         "@angular/localize": "^14.2.0",
         "@angular/material": "^14.2.0",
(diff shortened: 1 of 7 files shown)


TheSlimvReal:
Solved with #1440

Issue #1317, opened 2022-06-22
automatically mark/select entity when only one matching


sleidig:
In the entity select combobox you can type the name and the suggested options narrow down while typing. To make multiple entries easier, the component could automatically select the option if only one is matching, so that the user only has to press enter to add that child.

PR #1457, opened 2022-09-22
Automatically mark/select entity when only one matching

christophscheuing:
see issue: #1317

Implemented this feature for the edit-single-entity-component and the entity-select-component.

Currently there is no hint to the user indicating that she/he can select the uniquely typed entity by only hitting enter. We could add a tooltip for that, but continuously showing the tooltip might be annoying to the user. As long as there is no way of keeping track if the user has already taken notice of this feature, we might just keep it silent. At least the automatic selection of the uniquely typed entity (by adding a gray background) is an unobtrusive hint to the user that she/he might try hitting enter to accept the selected entity.


christophscheuing:
As suggested by @TheSlimvReal I've reverted most of the proprietary changes in favor of using the attribute directive autoActiveFirstOption of the mat-autocomplete component.

This implies that always the first option of the autocomplete list is selected and taken when hitting enter, i.e. not only if there is only one unique option left. However this seems to be a reasonably UX approach.

src/app/core/entity-components/entity-utils/entity-select/entity-select.component.spec.ts
```diff
@@ -154,29 +152,22 @@ describe("EntitySelectComponent", () => {
...
-    expectedLength = 4;
+
     component.formControl.setValue(null);
     expectedLength = 3;
     component.formControl.setValue("A");
     expectedLength = 3;
     component.formControl.setValue("Ab");
     expectedLength = 1;
     component.formControl.setValue("Abc");
+    expect(component.filteredEntities.length).toEqual(expectedLength);
```
TheSlimvReal, line 167: This test needs to be adjusted a little bit. `component.filteredEntities.length` should be checked after each call of `component.formControl.setValue(...)` to verify the expected length is correct. At the moment this only makes one check at the end.

src/app/core/entity-components/entity-utils/entity-select/entity-select.component.spec.ts
```diff
@@ -87,14 +87,12 @@ describe("EntitySelectComponent", () => {
...
   }));
 
-  it("should suggest all entities after an initial load", (done) => {
-    subscription = component.filteredEntities.subscribe((next) => {
-      expect(next.length).toBe(testUsers.length);
-      done();
-    });
+  it("should suggest all entities after an initial load", fakeAsync(() => {
     component.entityType = User.ENTITY_TYPE;
+    tick();
```
TheSlimvReal, line 92: Is the tick necessary?

TheSlimvReal (changes requested):
Looks very good and I think the grey brackground is already enough of a hint.

src/app/core/entity-components/entity-utils/entity-select/entity-select.component.spec.ts
christophscheuing: I've updated this test. Does it look OK now?

src/app/core/entity-components/entity-utils/entity-select/entity-select.component.spec.ts
christophscheuing: It seems to be necessary. At least this test fails without the tick.

PR #1463, opened 2022-09-27
fix: import ui working again

TheSlimvReal:
- [x] fix error when uploading a sheet where UUID's are set, but not the entity type
- [x] fix error when datatype is written with capital letter in schema


sleidig:
This doesn't work for me (anymore). Let's scrap this and get #1418 out of the door instead. That is working functionally but unit tests still have to catch up with the changes

PR #1464, opened 2022-09-27
fix: schema service does not mutate the database object

TheSlimvReal:
When the same entity is matching the same query string, a error is thrown because the in-place database-to-entity transformation breaks.

### Visible/Frontend Changes
- [x] 
- [ ] 

- [x] Database transformation does not happen in-place anymore


sleidig (changes requested):
code looks good. However, when testing I still see an error in the console:

1. create new Child named "foo foobar" and give it some date of birth
2. search "foo"
3. console shows: `TypeError: Cannot read properties of null (reading 'toString')`

Issue #771, opened 2021-04-14
Create & edit users


TheSlimvReal:
App administrators should be able to create new users and edit permissions of users and user groups. 

- [x] Admins can invite (new) users to their Aam Digital instance via email
- [x] Admins can assign roles to users
- [x] Admins can withdraw users access to the application

TheSlimvReal:
When creating new users, it should also be possible to directly assign new roles to them. How do we display (and especially explain) which roles are available. Keycloak provides an endpoint for which roles are available, but then we should remove some of the default rules (e.g. `uma_authorization`...). Another option would be to only allow the rules mentioned in the `Config:Permissions` file.

TheSlimvReal:
A general question about the user roles. If we show all available roles, users might start giving themselves roles, which allow them to see the _technical_ components (admin, import, ...) which we at least at the moment should rather hide. 

So how do we keep users from assigning themselves these roles? In the short term we could filter out one (some?) predefined roles in the backend, in the long run we might not need these technical components anymore.

sleidig:
> When creating new users, it should also be possible to directly assign new roles to them. How do we display (and especially explain) which roles are available. Keycloak provides an endpoint for which roles are available, but then we should remove some of the default rules (e.g. `uma_authorization`...). Another option would be to only allow the rules mentioned in the `Config:Permissions` file.


Why don't we combine the two: take roles from Keycloak API and filter these to only show the ones matching roles actually used in our config?

I guess there is a permission to restricting who can add roles? Beyond that, I would allow any role to be assigned - we should replace in-app UI modules that are used only by tech support as soon as possible, anyway.

TheSlimvReal:
done in #1472

PR #1472, opened 2022-09-30
feat: create and update user accounts

TheSlimvReal:
see issue: #771 

### Visible/Frontend Changes
- [x] added section in user-details for assigning & updating the email and the roles of a user
- [x] users can be de-activated

- [x]


src/app/core/session/session-service/synced-session.service.spec.ts
```diff
@@ -191,22 +195,6 @@ describe("SyncedSessionService", () => {
     flush();
   }));
 
-  it("Remote succeeds, local fails", fakeAsync(() => {
```
TheSlimvReal, line 194: I dont know what this was supposed to test. The test case where the local initially fails, the remote succeeds in case of a changed password or new user is covered later. I don't see a practical reason where the local login would always fail.

src/app/core/session/auth/keycloak/keycloak-auth.service.ts
```diff
@@ -144,6 +144,33 @@ export class KeycloakAuthService extends AuthService {
       }
     );
   }
+
+  createUser(username: string, email: string, roles: Role[] = []) {
```
Schottkyc137, line 148:
```suggestion
  createUser(username: string, email: string, roles: Role[] = []): Observable<any> {
```

src/app/core/session/auth/keycloak/keycloak-auth.service.ts
```diff
@@ -144,6 +144,33 @@ export class KeycloakAuthService extends AuthService {
...
+    return this.httpClient.post(`${environment.account_url}/account`, {
+      username,
+      email,
+      roles,
+    });
+  }
+
+  updateUser(userId: string, user: any) {
```
Schottkyc137, line 156:
```suggestion
  updateUser(userId: string, user: any): Observable<any> {
```

src/app/core/session/auth/keycloak/keycloak-auth.service.ts
```diff
@@ -144,6 +144,33 @@ export class KeycloakAuthService extends AuthService {
...
+    return this.httpClient.put(
+      `${environment.account_url}/account/${userId}`,
+      user
+    );
+  }
+
+  getUser(username: string) {
```
Schottkyc137, line 163:
```suggestion
  getUser(username: string): Observable<{ id: string; email: string; roles: Role[] }> {
```

or, alternatively, create an interface describing the returned object

src/app/core/session/auth/keycloak/keycloak-auth.service.ts
```diff
@@ -144,6 +144,33 @@ export class KeycloakAuthService extends AuthService {
...
+    return this.httpClient.get<{ id: string; email: string; roles: Role[] }>(
+      `${environment.account_url}/account/${username}`
+    );
+  }
+
+  getRoles() {
```
Schottkyc137, line 169:
```suggestion
  getRoles(): Observable<Role[]> {
```

src/app/core/user/user-security/user-security.component.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+  beforeEach(async () => {
+    mockHttp = jasmine.createSpyObj(["get", "put", "post"]);
+    mockHttp.get.and.returnValue(of([assignedRole, notAssignedRole]));
+    mockHttp.put.and.returnValue(of({}));
+    mockHttp.post.and.returnValue(of({}));
+
+    await TestBed.configureTestingModule({
+      imports: [UserModule, MockedTestingModule],
+      providers: [
+        { provide: AuthService, useValue: new KeycloakAuthService(mockHttp) },
```
Schottkyc137, line 51: Is it maybe better to provide the `HttpClient` with the mocked version and provide the `AuthService` in a normal way? Currently, new constructor parameters in the `AuthService` automatically require one to rewrite the value

src/app/core/user/user-security/user-security.component.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+  });
+
+  it("should load existing account data", fakeAsync(() => {
+    mockHttp.get.and.returnValue(of(keycloakUser));
+
+    component.onInitFromDynamicConfig({ entity: user });
+    tick();
+
+    expect(component.userId).toBe("userId");
+    expect(component.form.get("username").value).toBe(user.name);
```
Schottkyc137, line 71:
```suggestion
    expect(component.form.get("username")).toHaveValue(user.name);
```

Same with the two `expects` below.

Alternatively, you should also be able to check the whole form:

```
expect(component.form).toHaveValue({ username: user.name, email: ... });
```

with a potential `jasmine.objectContaining({...})` if the form has more values that you don't want to check

src/app/core/user/user-security/user-security.component.ts
```diff
@@ -0,0 +1,104 @@
...
+    Object.keys(this.form.controls).forEach((control) =>
+      this.form.get(control).pristine ? delete update[control] : undefined
+    );
+    this.process(
+      this.keycloak.updateUser(this.userId, update),
+      $localize`:Snackbar message:Successfully updated user`
+    );
+  }
+
+  private process(obs: Observable<any>, message: string) {
```
Schottkyc137, line 93: This function seems a bit bland, is there a better name – both for the function as well as for `obs`?

src/app/core/user/user-security/user-security.component.html
```diff
@@ -0,0 +1,43 @@
+<form [formGroup]="form">
+  <mat-form-field>
+    <mat-label i18n="label of username input">Username</mat-label>
+    <input matInput type="text" formControlName="username">
+  </mat-form-field>
+  <mat-form-field>
```
Schottkyc137, line 6: UX-wise this might be confusing (why are there two emails for a single user?) I suppose it's not that big of an issue since (correct me if I'm wrong) this is only seen by admins who should know what they are doing.

Schottkyc137 (commented):
Nice work! I really like the exhaustive tests and the easy-to-understand functions. I have left some nitpicky comments but most of them will probably reflect my subjective opinion so don't feel obligated to implement these. 
I could not quite test the functionality since I do not believe that we can use Keycloak in our demo apps. Is there another way to test this?

src/app/core/user/user-security/user-security.component.ts
```diff
@@ -0,0 +1,104 @@
...
+} from "../../session/auth/keycloak/keycloak-auth.service";
+import { AuthService } from "../../session/auth/auth.service";
+import { User } from "../user";
+import { MatSnackBar } from "@angular/material/snack-bar";
+import { Observable } from "rxjs";
+import { PanelConfig } from "../../entity-components/entity-details/EntityDetailsConfig";
+
+@DynamicComponent("UserSecurity")
+@Component({
+  selector: "app-user-security",
```
sleidig, line 17: I am slightly thrown off by the name - security sounds more technical and sophisticated to me than the account credentials form that this actually implements. But I can live with this also :blush:

src/app/core/user/user-security/user-security.component.ts
```diff
@@ -0,0 +1,104 @@
...
+  keycloak: KeycloakAuthService;
+  availableRoles: Role[] = [];
+  userId: string;
+
+  constructor(
+    private fb: FormBuilder,
+    authService: AuthService,
+    private snackbar: MatSnackBar
+  ) {
+    if (authService instanceof KeycloakAuthService) {
```
sleidig, line 35: I think the textbook, cleaner way here would be to have every AuthService having all functions (if only as dummy implementations) and maybe an "allowsEditing". So that the component here does not know or care what auth service is active.

what other AuthService implementations do we have?

sleidig (commented):
Code looks pretty good! :+1: 

I couldn't test functionality yet either, as the setup on dev.aam-digital.com seems to be missing something (maybe permissions)

src/app/core/user/user-security/user-security.component.html
TheSlimvReal: Can you explain what you mean with the two emails? There should only be one email field with a label connected to it.

src/app/core/session/auth/keycloak/keycloak-auth.service.ts
```diff
@@ -144,6 +144,33 @@ export class KeycloakAuthService extends AuthService {
       }
     );
   }
+
+  createUser(username: string, email: string, roles: Role[] = []) {
+    return this.httpClient.post(`${environment.account_url}/account`, {
```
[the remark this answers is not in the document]
TheSlimvReal: The service is not necessarily Keycloak specific. We could implement the same functionality with CouchDB. That is why choose a more generic name.

src/app/core/ui/ui.module.ts
```diff
@@ -76,4 +77,6 @@ import { PwaInstallModule } from "app/core/pwa-install/pwa-install.module";
   declarations: [SearchComponent, UiComponent, PrimaryActionComponent],
   exports: [UiComponent],
 })
-export class UiModule {}
+export class UiModule {
+  static dynamicComponents = [UserSecurityComponent];
```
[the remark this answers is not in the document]
TheSlimvReal: This is a static link that we need for all dynamic components that have no hard reference in the code. If a component has no hard link anywhere, then the compiler throws away the component. How this is called or where we have it is irrelevant, we only need this link somewhere.

src/app/core/user/user-security/user-security.component.ts
TheSlimvReal: I am happy for any suggestions!

When coming up with the name, I had the following things in mind:

1. We already have a component that is called `UserAccountComponent` (to edit a users own details)
2. This component handles security aspects as it determines who can access the app and what permissions are given

src/app/core/user/user-security/user-security.component.ts
TheSlimvReal: We also have the `CouchdbAuthService` with limited functionality (only edit password, no password reset, ...).

I don't see a benefit of pulling the functions up into the `AuthService` interface and then providing mock implementations in the `CouchdbAuthService`. In this case we would have to check for each function individually whether it is implemented or not (in order to adapt the UI) and just adding a `allowsEditing` flag is very vague to me, because it does not say which functions this does involve. I am not super happy with the design either but this approach at least gives us type-safety (you cannot call `createUser` on the `AuthService` unless you explicitly type-cast it).

src/app/core/user/user-security/user-security.component.html
TheSlimvReal: Ah no, these are actually the same email. In the `UserAccountComponent` you can see the email for the currently logged in user (you) which everybody can see and change for themselves. The `UserSecurityComponent` (I am happy for any better name suggestion) shows the same email, but for any user that is registered to the system (only seen by managers with the required permissions).
So the email and the meaning is exactly the same - the email with which a user is registered and can reset the password.

Schottkyc137: Ah, I understand now. In this case, the two fields should, however, always show the same information. Currently, at least in the demo app, the email under `Security` is empty (while the username isn't) which has given me the impression that the two emails are different to begin with.

TheSlimvReal: Ah now I get it, yes the demo data has to be adjusted 😄 good point, thanks!

sleidig (changes requested):
Code looks good :+1: 

Functional testing:
- security tab: no edit/save buttons --> should be consistent with other screens
- when changing email of an account from user details (not own profile), the verification mail is sent to the previous, not the new email id
- latest version (with disable/enable) not deployed yet?

src/app/core/user/user-security/user-security.component.ts
```diff
@@ -56,72 +56,100 @@ export class UserSecurityComponent implements OnInitDynamicComponent {
...
     }
     this.form.setErrors({});
-    obs.subscribe({
-      next: () => this.snackbar.open(message, undefined, { duration: 5000 }),
-      error: ({ error }) => this.form.setErrors({ failed: error.message }),
-    });
+    return this.form.getRawValue();
+  }
+
+  private showSnackbar(message: string) {
```
sleidig, line 152: In order to have consistent UX it's probably better to use `AlertService` instead of Snackbar directly. That way things like the duration of messages is not randomly decided for every component separately.

sleidig (approved):
successfully tested functionally. Enable/Disable and account creation works as expected.

Committed a small improvement to the security component's UI.

TheSlimvReal:
- [ ] improve text when user account is not created yet
- [ ] disable form and buttons when user does not have permission (account-manager)

sleidig (approved):

PR #1477, opened 2022-10-08
Remove deprecated code

Schottkyc137:
Remove deprecated code and replace it with non-deprecated versions

Deprecated code includes:

- `Observable#toPromise()` is now deprecated
- `of()` with no arguments is deprecated
- `String#substr(...)` should not be used, instead either `String#substring(...)` or `String#slice(...)` should be used


src/app/child-dev-project/children/children.service.ts
```diff
@@ -57,17 +56,14 @@ export class ChildrenService {
...
-      .then((loadedChild) => {
-        return this.queryLatestRelation(id).then((currentSchoolInfo) => {
-          loadedChild.schoolClass = currentSchoolInfo?.schoolClass;
-          loadedChild.schoolId = currentSchoolInfo?.schoolId;
-          return loadedChild;
-        });
+  getChild(id: string): Promise<Child> {
+    return this.entityMapper.load<Child>(Child, id).then((loadedChild) => {
+      return this.queryLatestRelation(id).then((currentSchoolInfo) => {
+        loadedChild.schoolClass = currentSchoolInfo?.schoolClass;
```
sleidig, line 62: as far as I understand, as we don't emit something new after adding those properties the change detection doesn't fire and the school will not appear e.g. on the children list.

Apparently at least in in-memory mode this still works. Can you tell me why? And did you test it with indexeddb as well?

sleidig (commented):
Other code seems straight forward refactoring only :heavy_check_mark:

src/app/child-dev-project/children/children.service.ts
TheSlimvReal: In this case only one value is emitted, which is the child *after* the current school info has been resolved (so already including everything). 
What you refer to is what has been implemented in the `getChildren` function, where the children are emitted multiple times (once initially and once with current school info).
The change here is very trivial as it just removed the transformation of a promise into an observable.

TheSlimvReal (approved):
Thanks for cleaning this up (and even remove some code)!

PR #1480, opened 2022-10-17
fix: sentry only sends error messages for real deployments

TheSlimvReal:
The `whitelistUrls` option has been replaced by the `allowUrls` option as the first one does not exist anymore. Anyway, we can also configure in the sentry UI from which domains we receive error messages so this setting is rather redundant. Should we just remove it in favor of not having the Aam Digital domain in the source code?


sleidig:
yes, let's do that in sentry UI only :+1:

Issue #1448, opened 2022-09-21
Allow custom date ranges for notes filter


TheSlimvReal:
In the notes manager component (under *Notes*) there is a default filter which shows the notes from this week, the ones since last week and all the notes. This filter is hard coded and can only be enabled/disabled. 

This hard-coded filter should be extended with the option to create custom date ranges, e.g. last 7 days, last 30 days, this week, this month, ...
The functionality should be configurable through the `config-fix.ts`.
It probably makes sense to have a list of possible date ranges that should be available.

sleidig:
this is somewhat related to #1383 - making a filter to select concrete from-to range available to be selected by users

Issue #1383, opened 2022-07-26
date-range filter for Entity Lists


sleidig:
For some entity lists (e.g. Notes, overview of all HistoricalObservation records, etc.) it would be useful as a user to define a custom date range for which results should be displayed. The UI could look similar to the period selector for our reporting module.

TheSlimvReal:
Inspiration 
![Screenshot 2022-11-03 at 19 42 45](https://user-images.githubusercontent.com/33730997/199807677-cf00b840-1495-44ed-897d-a3ec47ce95ba.png)

PR #1487, opened 2022-10-20
Configurable date range filter for entity lists

christophscheuing:
Date range filters can now be configured via the `config-fix.ts`, as described in #1448.

Configuration via UI, as described in #1383 is also implemented.


src/app/core/entity-components/entity-list/filter-generator.service.ts
```diff
@@ -107,6 +111,89 @@ export class FilterGeneratorService {
     }
   }
 
+  private createDateRangeFilterOptions<T extends Entity>(
+    filterConfig: DateRangeFilterConfig
+  ): FilterSelectionOption<T>[] {
+    if (filterConfig.startingDayOfWeek) {
+      let configStartingDayOfWeek;
+      switch (filterConfig.startingDayOfWeek.toLowerCase()) {
```
TheSlimvReal, line 119: lets use a enum for this, so we don't need a switch-case here.

TheSlimvReal (commented):
Let's try to make more use of moments flexibility so we don't have to predefine all the units ourselves. Besides that looks good. Especially with the tests!

christophscheuing:
The test for the indexOfCorrespondingDateRange is quite incomplete and inadequate, but apart from that this PR could probably soon be reviewed, @TheSlimvReal.

TheSlimvReal (approved):
Great work. The UI and functionality is very clean. I am looking forward to releasing this!

PR #1494, opened 2022-10-21
fix: GitHub API can be used again

TheSlimvReal:
Currently the GitHub API requests fail because we sent a Bearer token with the request which is not valid for GitHub. 
This change allows to disable the setting of the Bearer token for special requests.

### Visible/Frontend Changes
- [x] Latest changes are shown again

- [x] Auth Interceptor can be disabled

### DIFF
--- a/src/app/core/session/auth/keycloak/auth-interceptor.service.ts
+++ b/src/app/core/session/auth/keycloak/auth-interceptor.service.ts
@@ -4,9 +4,21 @@ import {
   HttpHandler,
   HttpEvent,
   HttpInterceptor,
+  HttpContextToken,
 } from "@angular/common/http";
 import { Observable } from "rxjs";
 import { AuthService } from "../auth.service";
+
+/**
+ * This context can be used to prevent the Bearer token to be set by this interceptor.
+ * This can be useful when a request goes to a 3rd party service e.g. GitHub.
+ *
+ * Usage:
+ * ```javascript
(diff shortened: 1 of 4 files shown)


src/app/core/session/auth/keycloak/auth-interceptor.service.ts
```diff
@@ -4,10 +4,22 @@ import {
...
+/**
+ * This context can be used to prevent the Bearer token to be set by this interceptor.
+ * This can be useful when a request goes to a 3rd party service e.g. GitHub.
+ *
+ * Usage:
+ * ```javascript
+ * this.httpClient.get(`...`, { context: new HttpContext().set(AUTH_ENABLED, false) });
+ * ```
+ */
+export const AUTH_ENABLED = new HttpContextToken(() => true);
```
sleidig, line 21: From a big picture perspective, wouldn't it be better that developers don't have to actively disable something they are not even aware of being added? (i.e. default auth = false + maybe switching the default to true for requests to any subdomains of the own host)

I can live with this also but probably we will always release a broken version before thinking of putting the parameter in here?

sleidig (approved):

src/app/core/session/auth/keycloak/auth-interceptor.service.ts
TheSlimvReal: Let's see if we start using more external APIs which do not require our auth. If this will be the case then we can look for a more suitable solution.

PR #1495, opened 2022-10-21
Config from csv

sleidig:
branched off from import-ui refactoring as to use some of the components from there

- [x] generate entity config from csv
- [x] generate enum from comma-separated lists in csv
- [x] generate list view configs
- [x] generate details view configs
- [ ] save the generated config to db (do we merge or overwrite existing?)
- [x] integrate view into app (--> find it through Admin view)
- [x] allow field groups in details to be generated ("Overview:Family" --> tab "Overview", group header "Family"; "Overview:1" --> tab "Overview", group 1 without title)
- [x] allow a field to be included in multiple tabs of Details View (comma-separated: "Overview, Parent Details" --> included in two tabs)
--


src/app/core/config-setup/config-import-parser.service.spec.ts
```diff
@@ -0,0 +1,242 @@
...
+          schema: {
+            dataType: "string",
+            label: "date of birth",
+          },
+        },
+      ]
+    );
+  });
+
+  it("should generate sensible ids from labels", () => {
```
TheSlimvReal, line 165: good idea

src/app/core/config-setup/config-import-parser.service.spec.ts
```diff
@@ -0,0 +1,242 @@
...
+          },
+        },
+      ]
+    );
+
+    expect(parsedConfig["enum:hometown"]).toEqual([
+      { id: "Berlin", label: "Berlin" },
+      { id: "Hongkong", label: "Hongkong" },
+      { id: "New York", label: "New York" },
+      // TODO: should we rename the enum ids for distinction from labels? (--> NEW_YORK)
```
TheSlimvReal, line 238: Maybe also use the `generateIdFromLabel` function.

src/app/core/config-setup/config-import-parser.service.ts
```diff
@@ -0,0 +1,144 @@
...
+})
+export class ConfigImportParserService {
+  enumsAvailable: Map<string, ConfigurableEnumConfig> = new Map();
+  existingEnumHashmap: Map<string, string> = new Map();
+
+  constructor(private configService: ConfigService) {
+    const enums = this.configService.getAllConfigs(
+      CONFIGURABLE_ENUM_CONFIG_PREFIX
+    );
+    // TODO: how to get the id?
```
TheSlimvReal, line 24:
```suggestion
    enums.forEach((enum) => this.enumsAvailable.set(enum._id, enum));
```

src/app/core/config-setup/config-import-parser.service.ts
```diff
@@ -0,0 +1,144 @@
...
+        .map((field) => this.parseFieldDefinition(field)),
+    };
+
+    const generatedConfig: GeneratedConfig = {
+      ["entity:" + entityName]: entity,
+    };
+
+    // add enum configs
+    for (const [key, enumConfig] of this.enumsAvailable) {
+      generatedConfig["enum:" + key] = enumConfig;
```
TheSlimvReal, line 43:
```suggestion
      generatedConfig[key] = enumConfig;
```
If the approach from above is used

TheSlimvReal (commented):
Just had a look at the code so far but it looks very good. Simple approach and already very helpful.

src/app/core/config-setup/config-import-parser.service.ts
sleidig: but enum configs don't have _id, do they? they are just an array of options

src/app/core/config-setup/config-import-parser.service.spec.ts
sleidig: The advantage of leaving the IDs as given could be that during data imports, the id probably automatically matches the imported value without any matching logic from our code.

src/app/core/config-setup/config-import-parser.service.ts
```diff
@@ -0,0 +1,319 @@
...
+      defaultJsonConfig["view:admin/conflicts"]
+    );
+    configs.set(
+      "view:admin/config-import",
+      defaultJsonConfig["view:admin/config-import"]
+    );
+    configs.set("view:import", defaultJsonConfig["view:import"]);
+    configs.set("view:user", defaultJsonConfig["view:user"]);
+    configs.set("view:user/:id", defaultJsonConfig["view:user/:id"]);
+    configs.set("view:help", defaultJsonConfig["view:help"]);
```
sleidig, line 300: Should we maybe add those views more centrally, not only through config, because some of them are (almost) hard-wired?

TheSlimvReal: View like `user`, `user/:id` and `help` still allow configuration. Therefore it still makes sense to have them in the config. For the `import` it might be relevant to assign permissions.

src/app/core/config-setup/config-import-parser.service.ts
TheSlimvReal: The function `getAllConfigs` automatically sets this `_id` property.

TheSlimvReal: Ok probably I am wrong, the enums are arrays and I actually have no idea what happens with the assignment of the `_id` property of that array. At least it doesn't throw an error 😅

TheSlimvReal (approved):

Issue #1136, opened 2022-03-11
Feature: File Uploads


sleidig:
_As a user I want to upload files from my local device into the app directly. This can be to update participants' photos or attach documents, etc._

---
This is a summary issue to discuss our general approach of handling file uploads into Aam Digital. How do we store and secure files? Do we use integrations with systems like Nextcloud (and how do we integrate, #472 , #423 ). Specific tasks and user-facing features get their own separate issues to be implemented and are listed here.

Features:
- allow users to attach documents to notes: #89

TheSlimvReal:
Files that end up on a server should be checked for viruses/converted to a safe format.

Issue #89, opened 2017-08-02
Document Upload / Storing files


sleidig:
*As a social worker I want to upload and download documents in Aam Digital in order to have all relevant information in one place.*

- Files to be uploaded could include contracts, scanned school results or photos (the profile photo as well other photos from events, etc.). 
- Files should be attachable to any entity (e.g. a student's record or a note).
- Options to upload exactly one file or as many files as the user wants.

**Proposed Solutions & Alternatives**
- implement as a new dataType for Entity properties: `file`, `file-multiple`
- implement edit/display components to upload / download the file(s) (like we implemented entity-select etc.)
- ? store/retrieve files using a light-weight microservice on our server, auth via user's existing keycloak token

**User Stories**
- [x] Upload a file from my computer through a file-upload form field
- [x] Download an uploaded file through a file-upload field.
- [x] Delete an uploaded file from the record and the server overall.
- [x] ? Replace a uploaded file with a new file in a single-file upload field, which results in a warning to the user and of confirmed, deletes the previous file before adding the new one.
- [ ] ... ?

sleidig:
there might be a solution for local file access in the browser coming up: https://www.heise.de/developer/artikel/Native-File-System-API-Dateisystemzugriff-fuer-Websites-und-PWAs-im-Anflug-4505949.html?wt_mc=rss.developer.beitrag.atom

sleidig:
blocked by #472

TheSlimvReal:
A simple file server with REST API https://github.com/jveverka/file-server

Unfortunately it doesn't provide integration with an existing auth provider (Keycloak)

TheSlimvReal:
Another very easy approach would be to use the [CouchDB](https://docs.couchdb.org/en/3.2.2-docs/intro/api.html?highlight=attachments#attachments)/[PouchDB](https://pouchdb.com/guides/attachments.html) attachment features.
This would have the benefits, that we do not need to implement any further services (huge benefit) and can use the existing permission feature.

Some considerations:
We should probably not make the normal replication with attachments but only sync the attachment meta data. We can then make a *on-demand* request for the attachment through the [CoucDB API](https://docs.couchdb.org/en/3.2.2-docs/api/document/common.html?highlight=attachments#retrieving-attachments-content). For this we might need to implement the respective endpoint in the replication backend.

Suggestion from a CouchDB developer:

> I wouldn’t go >10MB of attachment(s) per doc. Ideally keep it <1MB

[...]

PR #1502, opened 2022-10-26
File upload

TheSlimvReal:
see issue: #1136 , #89

### Visible/Frontend Changes
- [x] Upload and show file button 
- [x] Upload progress indicator
- [x] indicate that a file has been uploaded

- [x] Update [client_max_body_size](https://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size) for Docker image and nginx proxy on server. -> [ndb-setup](https://github.com/Aam-Digital/ndb-setup/pull/10)
- <strike>Implement changes in permission backend</strike> -> later


src/app/core/entity/schema-datatypes/datatype-file.ts
```diff
@@ -0,0 +1,10 @@
+import { EntitySchemaDatatype } from "../schema/entity-schema-datatype";
+
+export const fileEntitySchemaDatatype: EntitySchemaDatatype = {
```
sleidig, line 3:
```suggestion
/**
 * Datatype to make a file attachment property available in an entity.
 * Attachment (file) itself is not stored as part of the database document but uploaded separatedly.
 */
export const fileEntitySchemaDatatype: EntitySchemaDatatype = {
```

src/app/core/file/edit-file/edit-file.component.ts
```diff
@@ -0,0 +1,86 @@
...
+import { DynamicComponent } from "../../view/dynamic-components/dynamic-component.decorator";
+import { FileService } from "../file.service";
+import { EntityMapperService } from "../../entity/entity-mapper.service";
+import { HttpEventType } from "@angular/common/http";
+import { ProgressSpinnerMode } from "@angular/material/progress-spinner";
+import { AlertService } from "../../alerts/alert.service";
+import { LoggingService } from "../../logging/logging.service";
+import { ConfirmationDialogService } from "../../confirmation-dialog/confirmation-dialog.service";
+
+@DynamicComponent("edit-file")
```
sleidig, line 12:
```suggestion
@DynamicComponent("EditFile")
```
other components have used CamelCase

src/app/core/file/file.service.ts
```diff
@@ -0,0 +1,74 @@
...
+export class FileService {
+  private attachmentsUrl = `${AppSettings.DB_PROXY_PREFIX}/${AppSettings.DB_NAME}-attachments`;
+
+  constructor(private http: HttpClient, private dialog: MatDialog) {}
+
+  uploadFile(
+    file: File,
+    entityId: string,
+    property: string,
+    blob = new Blob([file])
```
sleidig, line 24: when would file and blob be independent from each other?

src/app/core/file/file.service.ts
```diff
@@ -0,0 +1,74 @@
...
+  ): Observable<any> {
+    const attachmentId = `${this.attachmentsUrl}/${entityId}`;
+    return this.http.get<{ _id: string; _rev: string }>(attachmentId).pipe(
+      catchError(() =>
+        this.http
+          .put<{ rev: string }>(attachmentId, {})
+          .pipe(map((res) => ({ _rev: res.rev })))
+      ),
```
sleidig, line 32: this catch should probably only run on 404 errors?

src/app/core/file/edit-file/edit-file.component.ts
```diff
@@ -0,0 +1,86 @@
...
+  uploadProgress: number;
+  mode: ProgressSpinnerMode = "determinate";
+  done = false;
+
+  constructor(
+    private fileService: FileService,
+    private entityMapper: EntityMapperService,
+    private alertService: AlertService,
+    private logger: LoggingService,
+    private confirmationServer: ConfirmationDialogService
```
sleidig, line 28:
```suggestion
    private confirmationDialog: ConfirmationDialogService
```

src/app/core/file/edit-file/edit-file.component.ts
```diff
@@ -0,0 +1,86 @@
...
+  private handleError(err) {
+    this.logger.error("Failed uploading file: " + err);
+    this.alertService.addDanger("Could not upload file, please try again.");
+    this.done = true;
+  }
+
+  private async updateEntity(filename: string) {
+    this.formControl.setValue(filename);
+    this.entity[this.formControlName] = filename;
+    await this.entityMapper.save(this.entity);
```
sleidig, line 68: can this cause trouble / conflicts if the form is in edit mode for other fields in parallel, and the entity gets saved here automatically in between?

sleidig (commented):
Reads and works quite well already! :smiley: 

from functional testing:
- progress bar not showing yet in production deployment
- is the download progress possible to show inline, so that user can continue working / viewing the entity while downloading?
- Is there any way to detect whether browser allows popups (to display downloaded file)? then we could warn users why nothing is happening
- add a "remove file" button to allow delete the file without replacing it?
- to discuss: should we deactivate the upload function while the form is not in "Edit" mode? This would make it more consistent with behavior of other fields.

In terms of visuals I sometimes accidentally downloaded/uploaded when trying to do the other. Maybe because I was thinking of the import ui file upload component.
Also, if you like I can take over the tinkering to align the margins and look&feel a bit more with the other form fields.

src/app/core/file/edit-file/edit-file.component.ts
TheSlimvReal: As long as the `_id` and `_rev` are not part of the form (which should anyways not be the case) this still works because the entity mapper automatically overwrites the `_rev` of the saved entity to the latest value.

src/app/child-dev-project/children/demo-data-generators/demo-child-generator.service.ts
```diff
@@ -72,6 +77,49 @@ export class DemoChildGenerator extends DemoDataGenerator<Child> {
...
+                `file${j.toString()}`,
+                file
+              )
+            )
+          );
+        console.log("uploaded file", fileNum, fileName);
+        fileNum = (fileNum + 1) % 1064;
+      }
+    }
+    console.log("done");
```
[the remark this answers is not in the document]
TheSlimvReal: Ahem, this will be removed 😅 I used this code for creating my demo data. At the moment there is no concept on how this would work without a real database in the backend (which is the case in the demo mode)

src/app/core/file/file.service.ts
TheSlimvReal: I used it in the demo mode when I didn't use the http module for requesting the files. This can be removed later.

src/app/core/file/couchdb-file.service.ts
```diff
@@ -0,0 +1,129 @@
...
+  ): Alert {
+    const progress = obs.pipe(
+      filter(
+        (e) =>
+          e.type === HttpEventType.DownloadProgress ||
+          e.type === HttpEventType.UploadProgress
+      ),
+      map((e: HttpProgressEvent) => Math.round(100 * (e.loaded / e.total)))
+    );
+    return this.alerts.addProgress(message, progress);
```
TheSlimvReal, line 127: There might be an option how the alert can also be closed in this function but I didn't it. The difficulty is to not subscribe to the same observable again, because this will also re-trigger the request.

src/app/child-dev-project/children/model/child.ts
```diff
@@ -118,6 +118,18 @@ export class Child extends Entity {
...
+  @DatabaseField({ label: "File 3", dataType: "file" }) file3: string;
+  @DatabaseField({ label: "File 4", dataType: "file" }) file4: string;
+  @DatabaseField({ label: "File 5", dataType: "file" }) file5: string;
+  @DatabaseField({ label: "File 6", dataType: "file" }) file6: string;
+  @DatabaseField({ label: "File 7", dataType: "file" }) file7: string;
+  @DatabaseField({ label: "File 8", dataType: "file" }) file8: string;
+  @DatabaseField({ label: "File 9", dataType: "file" }) file9: string;
+  @DatabaseField({ label: "File 10", dataType: "file" }) file10: string;
+  @DatabaseField({ label: "File 11", dataType: "file" }) file11: string;
+
```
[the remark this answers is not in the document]
TheSlimvReal: Done

src/app/child-dev-project/children/demo-data-generators/demo-child-generator.service.ts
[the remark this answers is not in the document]
TheSlimvReal: Demo functionality is now here through the `MockFileService`

sleidig (changes requested):
- [ ] filename text can be clicked, but the hand cursor only shows over the label -> over text cursor should also change (and maybe clicking the label shouldn't download the file? but that's not a real problem) Also, the hand cursor shows hovering over the file field dotted line even for empty file fields.
- [ ] after clicking "No" in the "replace file" dialog, when selecting a new file again nothing happens (if it's the same file? for different file it does show the dialog again)
- [ ] file upload fails with conflict 409 from attachments couchdb when overwriting existing file (and in other instances - on a new entity before saving? when uploading the same file for multiple fields?)

sleidig (approved):
The code loooks really nice now and so does the UI.

Functional testing:
- when trying to delete a file that is not found in the backend (404 in console), the UI keeps the file name and for user nothing seems to happen. However, cancel/save updates the entity and the file name dissapears.
- In that case (broken files), I also saw "MaxListenersExceededWarning" in the console. Did we double check it's just because we have so many listeners - or could there be some bad registration? (https://stackoverflow.com/questions/50709059/maxlistenersexceededwarning-possible-eventemitter-memory-leak-detected-11-mess)

src/app/core/file/file.service.ts
```diff
@@ -1,18 +1,66 @@
...
  * This service allow handles the logic for files/attachments.
  * Files can be uploaded, shown and removed.
  */
 export abstract class FileService {
+  protected constructor(
+    protected entityMapper: EntityMapperService,
+    protected entities: EntityRegistry
+  ) {
+    // TODO maybe registration is to late (only when component is rendered)
+    this.deleteFilesOfDeletedEntities();
```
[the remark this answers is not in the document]
TheSlimvReal: Unfortunately no, it has to happen after the entity schemas are adjusted, which only happens in the subscription to the config service in the app component. So also doing it on config updates doesn't work, because the subscriptions are executed on the wrong order (first subscription is executed first).

PR #1503, opened 2022-10-26
fix: using initial username as id and preventing later edits

TheSlimvReal:
When creating new user they currently get a UUID (like all other entities). The problem is we need the user's id to match the username of the Keycloak user. This is important in order to assign users to entities and then later match them with actually logged in users.

This approach initially overwrites the entity id when the first time a username is saved but afterwards does not allow to change the username anymore. This way we ensure that user entities can be matched with the usernames.

### Visible/Frontend Changes
--

- [x] When creating a `User` entity, the name is used as `entityId`
- [x] When trying to change the name of a existing user, an error is thrown

### DIFF
--- a/src/app/core/user/user.ts
+++ b/src/app/core/user/user.ts
@@ -34,7 +34,23 @@ export class User extends Entity {
     label: $localize`:Label of username:Username`,
     validators: { required: true },
   })
-  name: string;
+  set name(value: string) {
+    if (this._name) {
+      // Throwing error if trying to change existing username
+      const label = User.schema.get("name").label;
+      throw new Error(
+        $localize`:Error message when trying to change the username|e.g. username cannot be changed after initialization:${label} cannot be changed after initialization`
+      );
+    }
+    this.entityId = value;
+    this._name = value;
+  }
+
+  get name(): string {
+    return this._name;
+  }
+
+  private _name: string;
 
   /**
    * settings for the mat-paginator for tables.
--- a/src/app/core/entity/model/entity.ts
+++ b/src/app/core/entity/model/entity.ts
@@ -139,12 +139,13 @@ export class Entity {
     let indices = [];
 
     // default indices generated from "name" property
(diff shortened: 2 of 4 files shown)


src/app/core/user/user.ts
```diff
@@ -34,7 +34,23 @@ export class User extends Entity {
     label: $localize`:Label of username:Username`,
     validators: { required: true },
   })
-  name: string;
+  set name(value: string) {
+    if (this._name) {
+      // Throwing error if trying to change existing username
+      const label = User.schema.get("name").label;
+      throw new Error(
+        $localize`:Error message when trying to change the username|e.g. username cannot be changed after initialization:${label} cannot be changed after initialization`
```
sleidig, line 42: best practices would suggest to not mix logic (the Entity model here) with UI (translating an error message), I think. The clean but more complex approach would be to define error codes or classes and only put human-readable and translated error messages when displaying it to the user. But I have my doubts that complexity is a worthy trade off here. :+1:

src/app/core/user/user.ts
```diff
@@ -34,7 +34,23 @@ export class User extends Entity {
     label: $localize`:Label of username:Username`,
     validators: { required: true },
   })
-  name: string;
+  set name(value: string) {
```
sleidig, line 37: let's maybe change the property to `username` (or will this break dozens of things?). For some clients we introduced additional firstname and lastname properties to be used in the UI. "username" would make the purpose of this one more explicit

TheSlimvReal: At the moment it would break a dozens of things (all the current user entities) So lets get this out and if we ever want to migrate this, we can do this independently.

src/app/core/user/user.ts
TheSlimvReal: Definitely a good thought. At the moment we are missing a general *display error* approach where we could translate these errors to a user friendly name.

PR #1511, opened 2022-10-27
fix(ui): properly position help icon for tooltips next to form fields

sleidig:
see Child Details --> tooltip for "mothertongue"


TheSlimvReal (approved):
Finally how it always should have been :smile:

Issue #1501, opened 2022-10-26
Link notes with any Entity type (e.g. also with ChildSchoolRelation)


sleidig:
*As a project coordinator I want to multiple observations related to a student's school year in order to have a better overview of my notes. I want to relate notes not only to the student overall but to some specific other records, which could include learning progress records or other.*

**Objective**
We have recently extended our `Note`s so that they can also be displayed in the school details view for the schools they are linked to. To cater for a specific use case of one client we need to also link and display notes at the `ChildSchoolRelation` details view.
In the future, it could become important to link notes to other records as well.

**Proposed Solutions & Alternatives**
As we are moving towards more generic data models (also see #1056 ) the best approach probably is a completely generic referencing from Notes to any other Entity.

Considerations & Questions
- Notes will be "manually" linked to the entity (no hidden automatic logic of inferring a link indirectly like linking a ChildSchoolRelation based on the Child that is linked. Only the `generateNewRecordFactory` could provide an initial starting point of what is linked for a new note)
  - [x] simple: add a entity-select field to link only ChildSchoolRelations
  - [ ] make note details view configurable?
  - [ ] allow multiple entity types to be selected from one entity-select field
- TODO: some entities do not have a human-readable name (toString) (e.g. what title should we display for a ChildSchoolRelation? Maybe need to generate something in a simple way) And how usable is a autocomplete select for such records, if there are a lot of them and they might not have an obvious name?
  - [x] (workaround: fixed "title" property to set manually)
  - [ ] ? generate a title when saving the entity (and save the generated title to db)
  - [ ] ? dynamically generate toString() with synchronuous access to "primary" entities like Child or School (so that we can use their toString internally)
- TO DISCUSS: Redundancy with existing child and school references (do we keep the system as is for these? Or do we try to migrate everything to a generic approach?)

sleidig:
![image](https://user-images.githubusercontent.com/1682541/199252629-ecc44fc6-ce40-4096-8e94-4e6a7205fb5d.png)
possible UI layout for linking related records (multi-type entity select)

sleidig:
Digging into the code to analyze the status quo and rethink which indices we need. Indexing on Notes is our most expensive operation, as there are a lot of notes in each system ...

Existing indices:
- `notes_index/by_<children|schools|authors>` -> only used for `NotesOfChildComponent`
  - do we need other indices for more effectively showing/filtering the Notes list (NotesManagerComponent)?
[...]

PR #1521, opened 2022-11-06
add Note.relatedEntities as property and index

sleidig:
see #1501

making notes linkable to ChildSchoolRelations (and other records in the future)

- should we take out some of the generic fixes/refactors (e.g. rename of component, avoid constantly generating new entities to get to the constructor) into separate PRs and commits?


src/app/core/entity/model/entity.ts
```diff
@@ -183,8 +183,9 @@ export class Entity {
    *
    * @returns {string} the unique id of this entity
    */
-  public getId(): string {
-    return this.entityId;
+  public getId(withPrefix: boolean = false): string {
```
TheSlimvReal, line 186: General question that I had during reading this PR. Is there any scenario in which we do **not** want to have the prefix stored? I cannot think of any at the moment so we should probably remove that option.

However, as this might have some implications on existing data we can also move this to another issue.

src/app/child-dev-project/children/children.service.ts
```diff
@@ -119,10 +120,15 @@ export class ChildrenService {
...
+  /**
+   * Query all notes that have been linked to the given other entity.
+   * @param entityId ID (with prefix!) of the related record
+   */
+  getNotesRelatedTo(entityId: string): Promise<Note[]> {
+    return this.dbIndexing.queryIndexDocsRange(
       Note,
-      `notes_index/by_${noteProperty}`,
+      `notes_related_index/note_by_relatedEntities`,
+      entityId + "", // in future, we can add a date limit here (see AttendanceService.getEventsForActivity)
```
TheSlimvReal, line 131:
```suggestion
      entityId + ":", // in future, we can add a date limit here (see AttendanceService.getEventsForActivity)
```
Same for `endkey` so we don't class with same prefixes, e.g. `Child` and `ChildSchoolRelation` both match the prefix `Child`

src/app/child-dev-project/children/children.service.ts
```diff
@@ -197,13 +203,38 @@ export class ChildrenService {
...
+            if (!Array.isArray(doc.relatedEntities)) return;
+
+            var d = new Date(doc.date || null);
+            var dateString = d.getFullYear() + "-" + String(d.getMonth()+1).padStart(2, "0") + "-" + String(d.getDate()).padStart(2, "0")
+
+            doc.relatedEntities.forEach((relatedEntity) => {
+              emit(relatedEntity + "_" + dateString);
+            });
+          }`,
+          // TODO: alternatively, we could build an index that combines all existing indexed properties for backward compatibility
```
TheSlimvReal, line 231: This, or we need to have a migration of existing data (which should not be too difficult though).

src/app/child-dev-project/children/children.service.ts
```diff
@@ -197,13 +203,38 @@ export class ChildrenService {
...
+          //  [...doc.relatedEntities, ...doc.children, ...doc.schools, ...doc.authors] (after adding relevant prefixes)
```
TheSlimvReal, line 232: Probably, we would also need to remove duplicates.

TheSlimvReal (commented):
I like the approach, the changes look quite simple and it seems like we get a nice level of abstraction/flexibility. Good that you also assessed a little bit on the status quo of the notes indices.

I think what we should decide on quite quickly is whether we want to change anything of the existing approach yet or whether we just add this on top for the moment. Right now the code seems like a little bit of both. Of course it would be nice to get the first thing done, but it might come with quite some migration effort. So I am also finde with for now, just adding a new property on top.

src/app/core/entity/model/entity.ts
sleidig: --> #1526

src/app/child-dev-project/children/children.service.ts
sleidig: discussed and postponed (see https://github.com/Aam-Digital/ndb-core/issues/1501#issuecomment-1308549111 )

sleidig:
Regarding the question on whether to refactor the overlapping existing logic, see here: https://github.com/Aam-Digital/ndb-core/issues/1501#issuecomment-1308549111
[tldr: we'll leave that as is for now]

sleidig:
deployed with activated config at qm.aam-digital.com

src/app/child-dev-project/notes/model/note.ts
```diff
@@ -166,6 +183,25 @@ export class Note extends Entity {
     }
 
     this.children = this.children.concat(childId);
+    this.relatedEntities.push(
```
TheSlimvReal, line 186: What is the benefit of this partial migration of children and schools (but not users and recurring activities)? Woudln't it be easiert to make a full migration at one point so we don't have to worry about what has already been migrated and what not?

src/app/child-dev-project/notes/model/note.ts
```diff
@@ -239,6 +275,8 @@ export class Note extends Entity {
   copy(): Note {
     const note: Note = super.copy() as Note;
     note.children = [...this.children];
+    note.schools = [...this.schools];
+    note.relatedEntities = [...this.relatedEntities];
```
TheSlimvReal, line 279:
```suggestion
    note.relatedEntities = [...this.relatedEntities];
    note.authors = [...this.authors];
```

src/app/child-dev-project/children/children.service.ts
```diff
@@ -119,14 +123,46 @@ export class ChildrenService {
...
   }
 
-  getNotesOf(entityId: string, noteProperty: string): Promise<Note[]> {
-    return this.dbIndexing.queryIndexDocs(
+  /**
+   * Query all notes that have been linked to the given other entity.
+   * @param entityId ID (with prefix!) of the related record
+   */
+  async getNotesRelatedTo(entityId: string): Promise<Note[]> {
+    let indirectlyLinkedNotes = [];
```
TheSlimvReal, line 131:
```suggestion
    let legacyLinkedNotes = [];
```
I would not call these *indirectly* because they are linked jus the same way as the `relatedEntities`.

src/app/child-dev-project/children/children.service.ts
```diff
@@ -197,13 +233,35 @@ export class ChildrenService {
...
 
-    return this.dbIndexing.createIndex(designDoc);
+    await this.dbIndexing.createIndex(designDoc);
+
+    const newDesignDoc = {
+      _id: "_design/notes_related_index",
+      views: {
+        note_by_relatedEntities: {
+          map: `(doc) => {
+            if (!doc._id.startsWith("${Note.ENTITY_TYPE}") && !doc._id.startsWith("${EventNote.ENTITY_TYPE}")) return;
```
TheSlimvReal, line 251: This is the only view where Notes and EventNotes are emitted. What is the reason for that?

I am a bit worried that this inconsistency will cause unexpected behaviour when using the views.

TheSlimvReal (changes requested):
So far only code review

src/app/child-dev-project/notes/note-details/note-details.component.html
```diff
@@ -193,5 +193,17 @@ <h1>{{ entity.date | date }}: {{ entity.subject }}</h1>
     >
     </app-entity-select>
 
+    <app-entity-select *ngIf="relatedEntitiesSchema?.additional"
+      class="form-section"
+      entityType="ChildSchoolRelation"
```
TheSlimvReal, line 198:
```suggestion
      [entityType]="relatedEntitiesSchema?.additional"
```
I don't see an advantage of the hard-coding.

TheSlimvReal (commented):
General look and feel is fine. I like the flexible approach with the potential to make this even more flexible.

src/app/child-dev-project/children/children.service.ts
sleidig: true. They are not exactly legacy yet either, if we stick with the current architecture for now ;-)  But yours is more clear :+1:

src/app/child-dev-project/children/children.service.ts
sleidig: I think this is left over from the initial draft, where we considered merging a lot of indices. Removing EventNotes here for now ...

sleidig: there is some risk that a user edits and event (EventNote) through our note details view, adds a relatedEntity and then this will not show up. But this has been a gap for the other references of EventNote already anyway

TheSlimvReal (approved):

PR #1522, opened 2022-11-07
Retry auto login

### LINKED ISSUE #1514: Do not report failed sync due to expired token as exception
**Describe the bug**
When the access token of a user expires, an error is thrown by PouchDB (and reported to the `"error"` callback of the live sync). These errors should not be reported to sentry as it is not really an exception.

**Expected behavior**
Either the system should be in a state, where errors like these cannot occur by e.g. automatically requesting a new access token via the refresh token and re-starting the sync. Otherwise we could also drop this error silently.

TheSlimvReal:
closes #1514 

Currently the access token is refresh a minute before it expires. In sentry, many errors popped up indicating, that is not working properly and users are left with expired access tokens. The reason for this is unclear, it could be due to unsteady internet connections or the timeout getting out of sync.

This PR changes the approach similar to what has also been implemented on the backend (after the same issue was faced):
When `PouchDB` or the `HttpClient` returns an error with status code `401` (unauthorised), then a `autoLogin` is tried and on success, the initial request is repeated with the new token. 

This has the advantage, that less request are needed to be done (token is only refreshed on demand) and we don't rely on timers any more. Unfortunately, more or less the same logic has to be implemented for `PouchDB` as well as the `HttpClient` as they don't work with the same interceptor logic. 

### Visible/Frontend Changes
--

- [x] access token is not anymore automatically refresh before it expires
- [x] if a request fails due to being unauthorised, the app retries the initial request after receiving a new token.


src/app/core/session/session-service/remote-session.spec.ts
```diff
@@ -48,5 +62,33 @@ describe("RemoteSessionService", () => {
...
+    spyOn(PouchDB, "fetch").and.returnValues(
+      Promise.resolve({
+        status: HttpStatusCode.Unauthorized,
+        ok: false,
+      } as Response),
+      Promise.resolve({ status: HttpStatusCode.Ok, ok: true } as Response)
+    );
+    let calls = 0;
+    mockAuthService.addAuthHeader.and.callFake((headers) => {
+      headers.Authorization = calls++ === 1 ? "active" : "inactive";
```
sleidig, line 76: I trust you, but I have no clue what this means - what's "active" of "inactive" as a header value for "Authorization"?

sleidig (approved):
Looks good to me, although some of this is very specific and I'm a bit out of my depth ;-) 

Don't know whether/where I could test this - but I trust you did, so good to go from my side.

Issue #1055, opened 2021-12-02
Notification Feature: Push Notifications to alert users


sleidig:
_As a user I should get actively pulled into the app with notifications in order to not miss new developments or required actions._

An overall notification feature should help users stay updated without the need to actively open the app and check data. Instead, both within the app (with some notification indicator) as well as through push notifications, the user should be automatically informed about new details.

-----
On a technical level, this feature includes:
- [x] Service to trigger push notifications in a PWA (if needed, including backend and external services)
- [x] UI to show details of notifications within the app and an "unread notifications" indicator
- [x] Configuration logic to allow users to enable/disable different types of notifications for their account individually
- Triggers for specific notifications for certain events:
  - [ ] notification when config is changed that affects certain behavior for users (e.g. a different pre-filter for the matching view)
  - [x] notification when a new/changed entity record matches a criteria (e.g. the current user has been added as the "assigned person" on a task record or a new participant with status "signed up through public form" has been added)
- in a future version:
  - [ ] notification when a planned date is reached (e.g. on the deadline of a task entity)
  - [ ] notification when an app update is available (to highlight new features)
  - [ ] emails in addition to push notifications

-----
Broken into user stories as implementation steps:

As a user I can
- [x] Aam-Digital/ndb-core#2662
- [x] Aam-Digital/ndb-core#2661
- [x] Aam-Digital/ndb-core#2660
- [x] Aam-Digital/ndb-core#2690
- [x] Aam-Digital/ndb-core#2696
- [x] Aam-Digital/ndb-core#2723
- [x] Aam-Digital/ndb-core#2724
- [x] Receive notification when an entity is updated
  - [x] Aam-Digital/ndb-core#2775
  - ~? support placeholders (e.g. "current user") in "entity update notification" rules~
- [x] Aam-Digital/ndb-core#2746
- [x] Aam-Digital/ndb-core#2754
- [x] Aam-Digital/ndb-core#2804

![Notifications drawio](https://github.com/user-attachments/assets/33fc4e90-dcef-4325-b822-9ff0eab641b8)

-----
# Remaining tasks:

### Open Issues (Frontend):
- [x] Aam-Digital/ndb-core#2847
- [x] Aam-Digital/ndb-core#2849
- [x] Aam-Digital/ndb-core#2850
- [ ] polish display of system notification message + events in frontend

### Open Issues (Backend):
- [x] filter operator for "in list" ("$elemMatch")
[...]

PR #1529, opened 2022-11-09
SwPush

TheSlimvReal:
see issue: #1055


sleidig:
@TheSlimvReal , is this still relevant? (And what problem was this targeting, actually?)

TheSlimvReal:
@sleidig this was to support #1055. I tried to implement the feature to allow our system to send push notifications to users phones and desktops. It god de-prioritised last time. I dont think we currently have the requirement.

sleidig:
Let's close this for now and use the approach from here in a new PR once we re-start work on the topic.

Issue #1525, opened 2022-11-09
entity select field does not show form label


sleidig:
The EditSingleEntity and DisplayEntity components do not show the form label like other fields. Only the empty component without an entity selected shows the label. After selecting, there is no chance for users to see and understand what the selected entity related to:

![image](https://user-images.githubusercontent.com/1682541/200783662-b6f8d372-de91-4754-aa0d-7ecdb29f774a.png)

sleidig:
also, margins and sizes should be improved:
![image](https://user-images.githubusercontent.com/1682541/200864321-825c42ef-8344-4826-b5d6-8087719464d7.png)

PR #1530, opened 2022-11-10
refactorings: schema dataTypes, getId use

sleidig:
code consolidations and ui fix for #1525


TheSlimvReal (approved):
Nice UI, looks much more in-line with the other components.

PR #1531, opened 2022-11-10
refactor(config): define entity names and icons in entity config inst…

### LINKED ISSUE #886: Add labels to the entities (for different language versions)
Right now, there's no separate label for entities like "Child" or "School". E.g. when trying to delete a child from the child details view, the message "Are you sure you want to delete this Child ?" is shown, where "Child" is taken directly from the entity type.  It is not read from the configuration and at the current state it is not possible to add different language versions for this label.

One problem might be that the entity name is also relevant for the URL, so care should be taken when adding the label feature.

_**also see #1056 **_

sleidig:
define entity names and icons in entity config instead individually in views

closes #886


src/app/core/config/config-fix.ts
```diff
@@ -940,6 +931,8 @@ export const defaultJsonConfig = {
   },
 
   "entity:Child": {
+    "label": "Participant",
```
TheSlimvReal, line 934: We either have to remove this or use `$localize` here otherwise the translation in the demo setup doesn't work.

TheSlimvReal (approved):
I really like it. The change took so little code and it really improves the whole look and feel a lot!

Issue #1512, opened 2022-10-27
Small addition to edit text with autocomplete (get back to adding new entity)


christophscheuing:
Up to now the edit-text-with-autocomplete is only used for adding recurring activities directly from the school's details view.

After once having selected an entry from the autocomplete dropdown list, I can still modify the title field and thereby edit the title of the selected entity or select another entity offered in the autocomplete dropdown list. However there is no possibility – apart from discarding the popped up dialogue and reopening it – to get back to adding a new entity. I think that as a user I'd expect that by completely deleting the text in the title field I can discard the selected entity and get back to adding a new entity (or selecting one via the autocomplete dropdown list), i.e. the status of the dialogue when it was opened at the beginning.

Implementing this would take away the possibility to first select an entity via the autocomplete dropdown list and then completely change its title. However the edit-text-with-autocomplete seems to be the wrong place for doing so anyway. If I want to change the title of an existing entity, I should rather find it in the displayed list of existing entities and edit it from there.

sleidig:
Good point. I think it might be good to make the title read-only there when it is an existing record being edited? That would also be a strong signal to the user that the current action is different from creating a new one. We could also add a small "x" button to remove the selected activity and start fresh.

PR #1532, opened 2022-11-10
Implemented readonly after having selected an entry

christophscheuing:
see issue: #1512

The additional "x" in the input field now allows to unselect a previously chosen entry from the autocomplete list. The requested user confirmation has been adapted to the two possible cases (having edited a new entity and then selecting one from the autocomplet list or having changed the values of an entity selected from the autocomplete list and then discarding this entry).


sleidig (approved):
Good job! This makes the UI more fool proof and clear :+1:

src/app/core/entity-components/entity-utils/dynamic-form-components/edit-text-with-autocomplete/edit-text-with-autocomplete.component.html
```diff
@@ -33,12 +34,21 @@
...
   class="tooltip-suffix"
 ></fa-icon>
 
-
+<fa-icon
+  *ngIf="!autocompleteDisabled && !!selectedEntity"
+  (click)="resetForm()"
+  #tooltipElement="matTooltip"
+  icon="circle-xmark"
+  [matTooltip]="'Remove link to this existing record.'"
```
sleidig, line 52:
```suggestion
  matTooltip="Unload existing record and reset form."
```

TheSlimvReal:
The reset does not work correctly:

1. Click the plus button
2. Select a autocomplete activity
3. Remove the autocomplete entity
4. Set a **new** name (don't select a another existing activity)
5. Save it
6. Go to the recurring activities and look for the previously "created" one
7. Open the details page
8. There you see that other students are already linked to this activity

Even after the `reset()` code is executed, the component still overwrites the previously selected entity.
This is because the hidden form fields like `_id` that are added in the `setAllFormValues` are **not** reset. In the reset function, these should even be removed from the form again.

TheSlimvReal (changes requested):
Fix error described in the above comment (with a (improved) test, that ensures that this error is actually fixed)

christophscheuing:
I've tried and fixed the described error. Maybe there's a more elegant way than using an extra class variable for the added controls.

TheSlimvReal (approved):
Nice works well now. But the component is getting a bit overly complicated. Lets hope we wont have to touch this too often

Issue #1527, opened 2022-11-09
UX touch ups after rework


sleidig:
A collection of minor UI issues as observed

![Screenshot from 2022-11-09 15-54-42](https://user-images.githubusercontent.com/1682541/200865412-4daad558-32d1-4c7a-8e50-6b33d9a17756.png)

- [x] Clearer separation of which version number belongs to which list of changes
- [ ] ~remove heading (in screenshot "Features" at 3.9.0) if all changes for the heading are hidden~  can be done in GitHub Releases by adding "." to the heading

![Screenshot from 2022-11-09 15-59-57](https://user-images.githubusercontent.com/1682541/200865614-599db6f9-db57-4910-88a7-dd7a31cfb2f6.png)
- [x] improve contrast (light orange/white on dark orange is hard to read for users with less than perfect vision); also: why do we introduce another shade of orange - this feels a bit clashing with our base orange color
- [x] in the concrete case of form errors: I wouldn't display an alert at all, the form fields already clearly indicate what's wrong

PR #1533, opened 2022-11-11
Ux tweaks

sleidig:
see #1527


src/app/child-dev-project/notes/dashboard-widgets/important-notes-dashboard/important-notes-dashboard.component.ts
```diff
@@ -12,14 +12,14 @@ import { concat, Observable } from "rxjs";
 import { first, map } from "rxjs/operators";
 import { UntilDestroy, untilDestroyed } from "@ngneat/until-destroy";
 
-@DynamicComponent("ImportantNotesComponent")
+@DynamicComponent("ImportantNotesDashboard")
```
TheSlimvReal, line 15:
```suggestion
@DynamicComponent("ImportantNotesDashboard")
@DynamicComponent("ImportantNotesComponent")
```
Or we have to migrate

src/app/core/entity-components/entity-form/invalid-form-field.error.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Custom error indicating details of a failed form validation.
+ */
+export class InvalidFormFieldError extends Error {
+  constructor(message: string) {
+    super(message);
+    Object.setPrototypeOf(this, InvalidFormFieldError.prototype);
```
TheSlimvReal, line 7: why is this necessary?

src/app/core/entity-components/entity-form/entity-form/entity-form.component.ts
```diff
@@ -97,7 +98,9 @@ export class EntityFormComponent<T extends Entity = Entity> implements OnInit {
       this.save.emit(this.entity);
       this.switchEdit();
     } catch (err) {
-      this.alertService.addWarning(err.message);
+      if (!(err instanceof InvalidFormFieldError)) {
```
TheSlimvReal, line 101: Should we make this consistent then? The same code is used in the `RowDetails` and the `EntitySubrecord` component.

src/app/core/entity-components/entity-form/entity-form.service.ts
```diff
@@ -129,7 +130,9 @@ export class EntityFormService {
     form.markAllAsTouched();
     if (form.invalid) {
       const invalidFields = this.getInvalidFields(form, schema);
-      throw new Error($localize`Fields: "${invalidFields}" are invalid`);
+      throw new InvalidFormFieldError(
+        $localize`Fields: "${invalidFields}" are invalid`
```
TheSlimvReal, line 134: If we don't want to display the form fields, we might as well remove this code.

TheSlimvReal (approved):
Thanks for taking the time to create the storybooks. This helped a lot for the review. 
Generally looks good, just the question whether we then generalise the changes with the invalid form fields directly.

src/app/child-dev-project/notes/dashboard-widgets/important-notes-dashboard/important-notes-dashboard.component.ts
sleidig: :+1: good point. Adding this into the existing commit as well as into the detectLegacyConfig logging

sleidig: hmm, actually automatic migration would be much more feasible then ... :see_no_evil:

src/app/core/entity-components/entity-form/invalid-form-field.error.ts
sleidig: because stackoverflow :wink: 

actually not necessary here but would be needed if we add methods to the custom class (https://stackoverflow.com/a/41429145/1473411)

src/app/core/entity-components/entity-form/entity-form.service.ts
sleidig: true. Removing the translation and adding a more dev centered error implementation

src/app/core/entity-components/entity-form/entity-form/entity-form.component.ts
sleidig: thanks, good suggestion! (And we should really try to merge some of that code)

implemented this now

TheSlimvReal (approved):

PR #1545, opened 2022-11-17
Remove flex layout

Schottkyc137:
Remove the last FxLayout dependency and remove it from the package.json


TheSlimvReal (approved):
Nice, lets hope this approach will be sufficient for a while

Issue #1539, opened 2022-11-15
Make dashboard widgets generic to work for any entity type


sleidig:
Our dashboard widgets are hard-coded to display details around Child entities. As we move more and more to a generic model, these should be configurable to also work for School or other entities.

- [x] ChildrenCountDashboard
- [x] NotesDashboard (cases with /without recent notes)
- [x] BirthdayDashboard

Issue #1432, opened 2022-09-08
provide "age" for any date property


sleidig:
e.g. also on school entities or others it would be useful to be able to use the "age" in the entity list

TheSlimvReal:
Just displaying this with a component (e.g. `DisplayAgeComponent`) will make it not sortable/filtereable.

sleidig:
maybe on the entity class we could use a custom "DateAndAge" object as property instead of a `date`. Then we can put the function to calculate age into that property instead of keeping it on the class itself.

TheSlimvReal:
another solution would be together with the #1056.
We would have a generic `Person` class in the application which has the date of birth logic which can then be subtyped by any entity.
As this required a bigger chance first it could still be worth to go with the `DateAndAge` object first.

Issue #1190, opened 2022-04-09
Generalize inactive status


TheSlimvReal:
Currently children can be marked as inactive and will not be shown in the default list views or dropdown selection.
This feature should be generalized so it can also be used for Users, School, etc.

A similar flag is also used for the `ChildSchoolRelation` where  a "active" status is calculated based on the start and end date.
Maybe this can also be included but as the logic is a bit different there this is not super necessary.
What can be done is to have the generic `isActive` function on the `Entity` class which can then be overwritten by subclasses.

sleidig:
This would also help block users from selecting a school/group that is missing some validation steps (e.g. signing an agreement with the organization).

sleidig:
As "Recurring Activities" should not be deleted, here also an option to archive them and hide them from the common screens is important

PR #1546, opened 2022-11-18
Flexible entity dashboards

TheSlimvReal:
see issue: #1539 #1432 #1190

### Visible/Frontend Changes
--

- [x] Any entity has `active` getter
- [x] Added `DateOfBirth` property, which has a `age` getter (and is a normal `Date` besides that)
- [x] Dashboard widgets support other/multiple entities


src/app/child-dev-project/children/dashboard-widgets/birthday-dashboard/birthday-dashboard.component.ts
```diff
@@ -17,40 +18,47 @@ export class BirthdayDashboardComponent
...
+    const data: EntityWithBirthday[] = [];
+    for (const [entityType, property] of Object.entries(this.entities)) {
+      const entities = await this.entityMapper.loadType(entityType);
+      data.push(
+        ...entities
+          .filter((entity) => entity.isActive && entity[property])
+          .map((entity) => ({
+            entity: entity,
+            birthday: this.getNextBirthday(entity[property]),
+            age: entity[property]?.age + 1,
```
sleidig, line 46: suggestions `age` -> `newAge`? (I was wondering why we don't add +1 in the template anymore)

src/app/child-dev-project/children/dashboard-widgets/entity-count-dashboard/entity-count-dashboard.component.ts
```diff
@@ -1,46 +1,48 @@
...
   implements OnInitDynamicComponent, OnInit
 {
+  private entity = Child.ENTITY_TYPE;
   /**
    * The property of the Child entities to group counts by.
    *
    * Default is "center".
    */
-  @Input() groupBy: string = "center";
+  private groupBy = "center";
```
sleidig, line 26: any particular reason to remove this as `@Input `?

I still like the idea that generally components can be configured equivalently through input or config.

src/app/child-dev-project/children/dashboard-widgets/entity-count-dashboard/entity-count-dashboard.component.ts
```diff
@@ -1,46 +1,48 @@
...
     private entityMapper: EntityMapperService,
     private router: Router
   ) {}
 
   onInitFromDynamicConfig(config: any) {
-    if (config?.groupBy) {
-      this.groupBy = config.groupBy;
-    }
+    this.groupBy = config?.groupBy ?? this.groupBy;
+    this.entity = config?.entity ?? this.entity;
```
sleidig, line 40: this.label also should be updated to the entity type's labelPlural

src/app/child-dev-project/children/model/child.ts
```diff
@@ -57,12 +57,12 @@ export class Child extends Entity {
   })
   projectNumber: string;
   @DatabaseField({
-    dataType: "date-only",
+    dataType: "date-of-birth",
```
sleidig, line 60: is this even necessary if the TypeScript type is `DateOfBirth` --> could infer the schema dataType?

src/app/child-dev-project/children/model/dateOfBirth.ts
```diff
@@ -0,0 +1,11 @@
+import { calculateAge } from "../../../utils/utils";
+
+/**
+ * Subclass of Date which provides the getter `age`.
+ * This age is calculated based on the date that this object represents.
+ */
+export class DateOfBirth extends Date {
```
sleidig, line 7: Maybe better named "DateWithAge" or similar? I can see a use for this also for admissionDate or similar properties.

src/app/child-dev-project/notes/model/note.ts
```diff
@@ -55,6 +55,23 @@ export class Note extends Entity {
     return instance;
   }
 
+  /**
+   * Returns the name of the Note property where entities of the given entity type are stored
+   * @param entityType
+   */
+  static getPropertyFor(entityType: string) {
```
sleidig, line 62: nice :-)

src/app/core/config/config-fix.ts
```diff
@@ -470,9 +470,10 @@ export const defaultJsonConfig = {
           "id": "name"
         },
         {
-          "view": "DisplayText",
+          "view": "DisplayAge",
           "label": $localize`:Column label for age of child:Age`,
-          "id": "age"
+          "id": "age",
```
sleidig, line 475: how about switching this to accept `id: "dateOfBirth"`?

TheSlimvReal: The problem is, we cannot have to times the same `"id"`. And `dateOfBirth` is already reserved for displaying the actual date, while this one should show the age. We could also put them both in one display component.

src/app/child-dev-project/children/model/child.ts
TheSlimvReal: Generally, nice idea. But it has one issue: Classes are renamed in the production build (we had the same issue with the `@DynamicComponent` annotation). Therefore, a static member like the `ENTITY_TYPE` would need to be set on the class to have a fixed name. But this can of course be a option to make this cleaner.

src/app/child-dev-project/children/dashboard-widgets/entity-count-dashboard/entity-count-dashboard.component.ts
TheSlimvReal: I remove it because we are not using it and I am also not sure what the support is (whether it still works). Other dashboards also don't allow it (newer ones have never been built this way). So consistency wise I find it a bit difficult to keep it here (why is it in one dashboard and not in others?). But of course not critical, just me trying to reduce things.

sleidig: my longer-term goal would still be that we do away with the onInitDynamic and let the system set config properties automatically to any input property and afterwards trigger the normal ngOnInit. I had gotten rather far with this for most components (like the dashboards, I think) and only got stuck on a few interlinked ones.

But we can re-introduce the @Input annotations also when we implement that.

src/app/child-dev-project/children/model/child.ts
sleidig: oh, right! not sure it's worth setting an ENTITY_TYPE - seems a bit counter-intuitive, so we could also stick with the current way?

sleidig: I saw you had already solved it. Actually nice! :100:

src/app/core/config/config-fix.ts
sleidig: This makes the config a bit cumbersome ... but I get the problem and have no idea for an alternative :+1:

sleidig (approved):
made a few small cleanups and corrections and tested functionally. Looks really nice! good to go

Issue #1543, opened 2022-11-15
Fields included in multiple tabs of entity details do not update


sleidig:
If the same property of an entity is displayed in multiple tabs (Form components), the field is not updated in the other tabs after changing it in one place.

PR #1549, opened 2022-11-22
fix: forms are automatically kept up-to-date

TheSlimvReal:
see issue: #1543

### Visible/Frontend Changes
- [x] Forms are automatically updated if a entity has been changed
- [x] Popup asks user, if existing changes should be overwritten

--


src/app/core/entity-components/entity-form/entity-form/entity-form.component.ts
```diff
@@ -72,16 +75,40 @@ export class EntityFormComponent<T extends Entity = Entity> implements OnInit {
...
+  private async applyChanges(entity) {
+    if (this.saveInProgress) {
+      // this is the component that currently saves the values -> no need to apply changes.
+      return;
+    }
+    if (
+      this.form.pristine ||
+      (await this.confirmationDialog.getConfirmation(
+        $localize`Load changes?`,
+        $localize`Local changes are in conflict with updated values. Do you want the local changes to be overwritten with the latest values?`
```
sleidig, line 107:
```suggestion
        $localize`Local changes are in conflict with updated values synced from the server. Do you want the local changes to be overwritten with the latest values?`
```

src/app/core/entity/model/entity.ts
```diff
@@ -191,8 +195,8 @@ export class Entity {
   /**
    * Get the class (Entity or the actual subclass of the instance) to call static methods on the correct class considering inheritance
    */
-  getConstructor(): EntityConstructor {
-    return <typeof Entity>this.constructor;
+  getConstructor<T extends Entity>(this: T): EntityConstructor<T> {
```
sleidig, line 198: what does this change mean?

sleidig (approved):
Works well :+1: 

Super fancy optimization would be to only reset those fields that were changed on the server. But that's probably not at all trivial, so let's keep this here.

src/app/core/entity/model/entity.ts
TheSlimvReal: Prior to that calling `new Child().getConstructor()` would return a constructor of the type `EntityConstructor<Entity>`, so calling `new (new Child().getConstructor())()` would have the type `Entity` instead of `Child`. This change correctly returns the constructor for the subclass, in this case `EntityConstructor<Child>`.

src/app/core/entity-components/entity-form/entity-form/entity-form.component.ts
TheSlimvReal: Not necessarily. the same issue happens, if a person was editing two forms at the same time but only saves one of them.

sleidig: true. But that case currently gives a very confusing UX anyway: the popup for the second tab will show immediately after saving the first tab (while you still see the first tab). So nobody apart from us two will understand which changes the question exactly relates to :see_no_evil:

Issue #1541, opened 2022-11-15
Allow filters on subrecord list (EntitySubrecordComponent)


sleidig:
In some cases embedded lists of subrecords (e.g. notes of a student) should be filtered - either enforced through the config in the background (only show health-related notes on the health tab) or giving the user a filter UI to filter manually.

- [ ] allow filtering in NotesRelatedToEntity by defining a filter value in config (and maybe which columns
- [ ] allow display of filter UI for EntitySubrecordComponent (enabled/disabled through @Input / config)

PR #1551, opened 2022-11-22
Configurable filters for EntitySubrecord

TheSlimvReal:
see issue: #1541 

### Visible/Frontend Changes
--

- [x] [MongoDB queries](https://www.mongodb.com/docs/manual/reference/operator/query/) can be used to filter records in the EntitySubrecord (thanks to [ucast](https://github.com/stalniy/ucast/tree/master/packages/mongo2js), which we also use for the permissions)
- [x] (?) merge with existing filter logic (EntityListComponent)
- [x] Adjust factory generator service (to automatically use values from filter)


src/app/core/entity-components/entity-subrecord/entity-subrecord/entity-subrecord.component.ts
```diff
@@ -172,6 +193,7 @@ export class EntitySubrecordComponent<T extends Entity>
             type === "update" &&
             !this._records.find((rec) => rec.getId() === entity.getId())
           ) {
+            // TODO maybe doc does not pass filter anymore -> remove
```
sleidig, line 196: do you still want to cover this as part of this PR? Would be good to have :-)

src/app/core/filter/filter.service.ts
```diff
@@ -0,0 +1,107 @@
...
+  createFactory,
+  allParsingInstructions,
+  allInterpreters,
+  Filter,
+  compare,
+} from "@ucast/mongo2js";
+import moment from "moment";
+
+/**
+ * This server handles the filter objects.
```
sleidig, line 16:
```suggestion
 * Utility service to help handling and aligning filters with entities.
```

src/app/core/filter/filter.service.ts
```diff
@@ -0,0 +1,107 @@
...
+  ) as Filter;
+
+  constructor(private configService: ConfigService) {}
+
+  /**
+   * Builds a predicate for a given filter object.
+   * This predicate can be used to filter arrays of objects.
+   * ```javascript
+   * const predicate = this.filterService.getFilterPredicate(filterObj);
+   * const filtered = this.data.filter(predicate)
```
sleidig, line 35: If we are always using it like this, we could also pack it into the function here. If we sometimes use the predicate separately, this implementation looks good.

`const filteredData = this.filterService.applyFilter(data, filterObj);`

src/app/core/filter/filter.service.ts
```diff
@@ -0,0 +1,107 @@
...
+        this.assignValueToEntity(key, value, schema, entity);
+      }
+    });
+  }
+
+  private assignValueToEntity<T extends Entity>(
+    key: string,
+    value,
+    schema: Map<string, EntitySchemaField>,
+    newNote: T
```
sleidig, line 63: rename --> newEntity?

src/app/core/filter/filter.service.ts
```diff
@@ -0,0 +1,107 @@
...
+  ) {
+    if (key.includes(".")) {
+      // TODO only one level deep nesting is supported (also by ucast https://github.com/stalniy/ucast/issues/32)
+      [key, value] = this.transformNestedKey(key, value);
+    }
+    const property = schema.get(key);
+    if (property.dataType === "configurable-enum") {
+      value = this.parseConfigurableEnumValue(property, value);
+    }
+    // TODO fail for unsupported data types
```
sleidig, line 73: let's add the check? Or is this sufficiently guarded through the condition above (`if (typeof value !== "object")`)?

src/app/core/filter/filter/filter.component.html
```diff
@@ -0,0 +1,12 @@
+<app-list-filter
+  *ngFor="let filter of filterSelections"
+  [filterConfig]="filter.filterSettings"
+  [(selectedOption)]="filter.selectedOption"
+  [displayAsToggle]="filter.display !== 'dropdown'"
```
sleidig, line 5: I think we recently discovered that `displayAsToggle` is not implemented anymore. Let's remove it completely as part of this refactor?

src/app/core/filter/filter/filter.component.ts
```diff
@@ -0,0 +1,144 @@
...
+  @Input() entityType: EntityConstructor<T>;
+  /**
+   * The list of entities. This is used to detect which options should be available
+   */
+  @Input() entities: T[];
+  /**
+   * If true, the filter state will be stored in the url and automatically applied on reload or navigation.
+   * default `false`.
+   */
+  @Input() withUrl = false;
```
sleidig, line 43: maybe more clear if named `useUrlQueryParams` or something similar?

src/app/core/filter/filter/filter.component.ts
```diff
@@ -0,0 +1,144 @@
...
+  /**
+   * If true, only filter options are shown, for which some entities pass the filter.
+   * default `false`
+   */
+  @Input() onlyUsed = false;
```
sleidig, line 48: rename for clarity? --> e.g. `onlyShowRelevantFilterOptions`

sleidig (commented):
I really like the refactor and functionality seems to run well. Just a few minor remarks on your open TODOs.

src/app/core/filter/filter.service.ts
TheSlimvReal: I think it is cheaper, to not re-create filters every time and I also just started using the predicate on a single entity 😅 So I suggest to keep it like this for now.

src/app/core/filter/filter.service.ts
TheSlimvReal: I added support for dates, but I cannot think of other conversions that are not caught by the object check.

src/app/core/filter/filter.service.ts
```diff
@@ -0,0 +1,109 @@
...
+   * This patch happens in-place.
+   * @param entity the entity to be patched
+   * @param filter the filter which the entity should pass afterwards
+   */
+  alignEntityWithFilter<T extends Entity>(entity: T, filter: DataFilter<T>) {
+    const schema = entity.getSchema();
+    Object.entries(filter ?? {}).forEach(([key, value]) => {
+      // TODO support arrays through recursion
+      if (typeof value !== "object") {
+        this.assignValueToEntity(key, value, schema, entity);
```
sleidig, line 54:
```suggestion
      if (typeof value !== "object") {
        // only simple equality filters are automatically applied to new entities, complex conditions (e.g. $lt / $gt) are ignored)
        this.assignValueToEntity(key, value, schema, entity);
```

src/app/core/filter/filter.service.ts
sleidig: sounds good :+1:   I was a bit confused when reviewing now because the !== object is quite far away from the function but with my additional code comment I am confident this is good

sleidig (approved):
works great :100:

PR #1553, opened 2022-11-24
Fix: Header stays on top when scrolling through entity details

### LINKED ISSUE #1542: Regression: Keep details tab titels fixed when scrolling
Entity Details used to hold the title and tabs fixed at the top of screen, visible regardless of scrolling. This is no the case anymore in recent versions.

Schottkyc137:
Closes #1542


TheSlimvReal (changes requested):
The header correctly stays on top. 

What is still missing from the current behaviour (and also in the issue, sorry for that)
-> The `Edit` button should also stay at the top when the form is scrolled (see current behaviour in the [demo](https://demo.aam-digital.com)

src/app/core/entity-components/entity-form/entity-form/entity-form.component.scss
```diff
@@ -37,3 +38,9 @@ app-entity-form {
 .tooltip-suffix {
   margin: 4px;
 }
+
+.form-buttons-wrapper {
+  position: sticky;
+  top: sizes.$regular;
+  z-index: 100;
```
TheSlimvReal, line 45:
```suggestion
  z-index: 100;
  margin-right: 2px;
```

TheSlimvReal (approved):
Looks and works nice now. Thanks for fixing

PR #1558, opened 2022-11-29
Fix demo note categories

sleidig:


TheSlimvReal (approved):
That was laying around for a while 😅

PR #1560, opened 2022-11-29
fix: incoming changes popup is not shown every time save is clicked

TheSlimvReal:
Currently, the popup that informs about updates of the entity is shown whenever a form is saved.
This should only happen if multiple forms are edited at the same time or if new changes are synced.

### Visible/Frontend Changes
- [x] 
- [ ] 

- [x] A short delay marks a form as the one that is currently saving


src/app/core/entity-components/entity-form/entity-form/entity-form.component.ts
```diff
@@ -130,7 +130,8 @@ export class EntityFormComponent<T extends Entity = Entity> implements OnInit {
         this.alertService.addDanger(err.message);
       }
     }
-    this.saveInProgress = false;
+    // Reset state after a short delay
+    setTimeout(() => (this.saveInProgress = false), 500);
```
sleidig, line 134:
```suggestion
    setTimeout(() => (this.saveInProgress = false), 1000);
```

sleidig (approved):

sleidig (changes requested):
Actually this still doesn't work well:

1. edit and save in tab A
2. tab B also shows the updated value correctly
3. edit and save in tab B now
4. popup appears (but we edited starting from the updated version - so where is the conflict?)

sleidig (approved):

Issue #139, opened 2018-06-28
integrate GoogleMaps/OSM to display Child and School locations


sleidig:
*As a social worker I want to see the exact location of a family's house on a map in order to easily find them for home visits and to plan visits of nearby families together.*

**Objective**
Social workers are often keeping contact with dozens of families in a chaotic neighbourhood. Especially new staff members or those usually working with other families may not know all the houses. Finding an address by street and house number can be extremely difficult and in some slums no official address exists. A visual map with the locations will be very useful for planning and finding houses.

**Proposed Solutions & Alternatives**
* A map component display in a separate section of the child details that displays a pin for the home of that child. Also displaying other children's homes on that map can be helpful for orientation.
* The map component can also be displayed as a stand-alone global view to display all children's home locations and schools to get a global overview of the neighbourhood.
* To save a new location, putting a pin on the map probably is most feasible. Later it would be interesting to allow users to add a GPS location when using the application on a smartphone.
* Similar to Child locations, a map section can also be added to school details.

**Notes for Implementation:**
* Location probably should be saved as long/lat coordinates with each Child entity

-----

**User Stories**
- [x] add/edit geo-location to an entity through drag & drop functionality on a map
- [x] delete geo-location of an entity
- [x] visualize the geo-location of an entity on a map within its details view
- [x] visualize multiple entities with their location on a single map

sleidig:
we should check out this: https://blog.angularindepth.com/google-maps-is-now-an-angular-component-821ec61d2a0

TheSlimvReal:
What kind of map integration do you have in mind? Would this be just another section in the child/school details view where the map is centered on this child/school or do you think of a separate component that displays all children & schools?

sleidig:
I updated the description with more details.

TheSlimvReal:
I see the following tasks and potential free (and not GoogleMaps) solutions:

- [ ] Have a form field that allows to enter a normal address. Upon confirmation, the geo-location of this address is selected.
- - Free&public solution would be [nominatim](https://nominatim.org/). It comes with limited auto-complete support though.
- - A more professional but potentially costly option would be [locationIQ](https://locationiq.com/demo#autocomplete)
- [ ] Calculate distances between two geo-locations 
- - Use Law of Haversines or Spherical Law of Cosines to calculate distance in frontend ([examples](https://henry-rossiter.medium.com/calculating-distance-between-geographic-coordinates-with-javascript-5f3097b61898))
[...]

PR #1563, opened 2022-12-01
Map integration

TheSlimvReal:
see issue: #139
The new feature can be tested on the 'address' field of a school or a child.

### Visible/Frontend Changes
- [x] form field to input text and select a existing location
- [x] select location on a map
- [x] calculate distances in matching view (see #1555)

--


src/app/features/location/geo.service.ts
```diff
@@ -0,0 +1,51 @@
...
+})
+export class GeoService {
+  readonly remoteUrl = "https://nominatim.openstreetmap.org";
+
+  constructor(private http: HttpClient) {}
+
+  /**
+   * Returns locations that match the search term
+   * @param searchTerm e.g. `Rollbergstraße Berlin`
+   * TODO countrycodes should come from the datatype?
```
sleidig, line 25: does this support multiple countrycodes? Probably should put the default into the config centrally (applying to all fields, possibly with option to overwrite)

src/app/features/location/map/map.component.ts
```diff
@@ -0,0 +1,142 @@
...
+      this.markers.forEach((marker) => marker.removeFrom(this.map));
+    }
+    this.markers = entities
+      .filter(({ entity, property }) => !!entity?.[property])
+      .map(({ entity, property }) => {
+        const marker = L.marker([entity[property].lat, entity[property].lon]);
+        const tooltip = marker.bindTooltip(entity.toString());
+        marker.on("click", () => {
+          this.entityClick.emit(entity);
+          tooltip.openTooltip();
```
sleidig, line 63: let's show tooltip already on hover?

src/app/features/location/map/map.component.ts
```diff
@@ -0,0 +1,142 @@
...
+  private setMultipleMarkers(coordinates: Coordinates[]) {
+    if (this.markers && this.map) {
+      this.markers.forEach((marker) => marker.removeFrom(this.map));
+    }
+    this.markers = coordinates
+      .filter((coord) => !!coord)
+      .map((coord) => L.marker([coord.lat, coord.lon]));
+  }
+
+  private setMarker(coordinates: Coordinates) {
```
sleidig, line 122: is it important to handle the single marker differently? Or could we just transform that to an array with just one element?

sleidig: the flyTo could be done if markers.length === 1, for example

sleidig (changes requested):
code looks really well designed :100: 

Functional testing observations:
- [x] disable buttons/actions on map popup if opened in disabled form (displaying map is good but no "Apply")
- [x] bigger map as popup (let's make it square or ~90vh?)
- [ ] bonus extension: use device GPS to set the current location (maybe a button next to apply/cancel in the map popup?)
- [x] didn't you say we are not allowed to do auto-complete with the free api?

notes for myself: not yet implemented/reviewed are
- [ ] viewcomponent
- [x] matching view (highlighting, color-coding of entities; displaying all available entities)

src/app/features/location/coordinates.ts
```diff
@@ -0,0 +1,4 @@
+export interface Coordinates {
+  lat: number;
+  lon: number;
```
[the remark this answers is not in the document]
TheSlimvReal: This is the structure [nominatim](https://nominatim.org/release-docs/develop/api/Output/#json) is using. I would rather not parse this every time.

src/app/features/location/geo.service.ts
TheSlimvReal: yes, multiple can be selected [docs](https://nominatim.org/release-docs/develop/api/Search/#result-limitation)

TheSlimvReal: set in app-wide config

src/app/features/location/map-utils.ts
```diff
@@ -0,0 +1,85 @@
...
+  const d = max - min;
+  switch (max) {
+    case r:
+      return ((g - b) / d + (g < b ? 6 : 0)) / 6;
+    case g:
+      return ((b - r) / d + 2) / 6;
+    case b:
+      return ((r - g) / d + 4) / 6;
+  }
+}
```
sleidig, line 62: shall we move this one to a style-utils.ts file at a more central folder (and take the `addAlphaToHexColor` from utils.ts into the new file as well)?

src/app/features/location/map/map.component.ts
```diff
@@ -1,93 +1,81 @@
...
-      this.setMultipleMarkers(coordinates);
-    } else {
-      this.setMarker(coordinates);
-    }
-    if (this.map) {
-      this.showMarkersOnMap();
-    }
+  @ViewChild("map") private mapElement: ElementRef<HTMLDivElement>;
+
+  @Input() height = "200px";
```
sleidig, line 29: I guess getting this just from css properties assigned to the parent container didn't work?

src/app/features/matching-entities/matching-entities/matching-entities.component.ts
```diff
@@ -186,4 +209,43 @@ export class MatchingEntitiesComponent
...
+
+  private getDistanceString(from: Entity, to: Entity, index: number): string {
+    const a = from[this.showMap[index]] as Coordinates;
+    const b = to[this.showMap[(index + 1) % 2]] as Coordinates;
+    if (!a || !b) {
+      return "-";
+    }
+    const res = getKmDistance(a, b).toFixed(2);
+    return $localize`:distance with unit|e.g. 5 km:${res} km`;
+  }
```
sleidig, line 250: Is there any feasible way to move the whole distance column logic to its own component? Maybe as a special ViewComponent? This would help keep the MatchingComponent more manageable, as it would be only about the immediate matching logic.

sleidig (commented):
Looks and feels amazing now. I love the map with multiple colored markers for the matching.

How are we storing the location in the database? It seems like we only keep the address - which ends up "undefined" if I select coordinates in the middle of the ocean. (Admittedly, not a realistic use case - but I can imagine use cases tracking locations in the countryside that do not have an address.

src/app/features/location/edit-location/edit-location.component.ts
```diff
@@ -0,0 +1,96 @@
...
+      },
+    });
+    ref
+      .afterClosed()
+      .pipe(
+        filter((res) => !!res),
+        concatMap((res) => this.location.reverseLookup(res))
+      )
+      // TODO maybe remove name of building (e.g. CRCLR House)
+      .subscribe((res) => this.formControl.setValue(res));
```
TheSlimvReal, line 94: Stores coordinates on form control if nothing can be found

src/app/features/matching-entities/matching-entities/matching-entities.component.ts
TheSlimvReal: I don't know if it was worth. The `MatchingComponent` only got a few lines shorter but at least the distance is always correctly up-to-date.

src/app/features/matching-entities/matching-entities/matching-entities.component.ts
```diff
@@ -220,32 +214,33 @@ export class MatchingEntitiesComponent
...
-      return this.getDistanceString(this.leftSide.selected, e, 0);
-    } else if (this.rightSide.selected && this.rightSide.selected !== e) {
-      return this.getDistanceString(this.rightSide.selected, e, 1);
-    } else {
-      return "-";
-    }
+  private getDistanceColumnConfig(index: number) {
+    return {
+      id: "distance",
+      label: "Distance",
```
sleidig, line 228:
```suggestion
      label: $localize`:Matching View column name:Distance`,
```

sleidig (approved):
looks good!

Can we add a "expand map" button for the normal map in matching view (maybe open in popup full screen?). For the overview I intentionally gave it very limited height but its slightly hard to use for search then.

PR #1564, opened 2022-12-01
Allow navigation instead of popup in subrecord lists

sleidig:
see commits: 
- refactored to generalize similar components
- added option to configure subrecord table to navigate instead of showing popup (suggestions for a name for this flag welcome! `clickMode` doesn't seem very expressive ...)


src/app/core/entity-components/entity-subrecord/entity-subrecord/entity-subrecord.component.ts
```diff
@@ -379,16 +384,35 @@ export class EntitySubrecordComponent<T extends Entity>
    * Show one record's details in a modal dialog (if configured).
    * @param row The entity whose details should be displayed.
    */
-  rowClick(row: TableRow<T>) {
+  onRowClick(row: TableRow<T>) {
+    //if (row.formGroup && !row.formGroup.disabled) {
+
```
TheSlimvReal, line 389:
```suggestion
```

src/app/core/entity/entity-config.service.ts
```diff
@@ -123,4 +124,9 @@ export interface EntityConfig {
    * icon used to visualize the entity type
    */
   icon?: string;
+
+  /**
+   * base route of views for this entity type
+   */
+  route?: string;
```
TheSlimvReal, line 131: Great idea. Could we use the chance to also use this for all entities? At the moment we have some places where we still create the route through `ENTITY_TYPE + ':' + id` (e.g. in the `EntityBlockComponent`) which is rather dangerous as it requires the route to be equal to the entity type.
With this approach we have much more flexibility and can make the routes however we want.

src/app/core/config/config-fix.ts
```diff
@@ -644,9 +644,10 @@ export const defaultJsonConfig = {
           "components": [
             {
               "title": $localize`:Title inside a panel:School History`,
-              "component": "PreviousSchools",
+              "component": "ChildSchoolOverview",
               "config": {
                 "single": true,
+                "clickMode": "navigate",
```
TheSlimvReal, line 650:
```suggestion
```
This is not working in the demo.

TheSlimvReal (changes requested):
Seems to be working well. Thanks for taking this up. 

Playing with the app I discovered the following styling issue:
- [ ] Somehow the tables in the details views are now missing some top-margin (as compared to the currently deployed demo) (might be an individual change though)

src/app/child-dev-project/schools/child-school-overview/child-school-overview.component.ts
```diff
@@ -0,0 +1,149 @@
...
+    switch (entity?.getConstructor()?.ENTITY_TYPE) {
+      case Child.ENTITY_TYPE:
+        return "child";
+      case School.ENTITY_TYPE:
+        return "school";
+    }
+  }
+
+  async loadData() {
+    if (!this.entity.getId() || this.entity.getId() === "" || !this.mode) {
```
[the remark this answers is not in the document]
sleidig:
```suggestion
    if (!this.mode) {
```
to be honest - I have no idea. These have been in one of the previous component implementations but I think we can assume a valid entity is set here

src/app/core/config/config-fix.ts
sleidig: :+1: thanks. forgot this here from testing, apparently

src/app/core/entity/entity-config.service.ts
sleidig: good point actually. Done.

TheSlimvReal (approved):
Great, works fine now

PR #1566, opened 2022-12-02
fix(ui): add padding to entity details tabs again

sleidig:
add padding to entity details tabs again that went missing with previous ui fix


TheSlimvReal (approved):
thanks!

PR #1568, opened 2022-12-03
fix: styling of disabled create button

TheSlimvReal:
When the create button in entity subrecord is disbled, it still looks like it can be used.

### Visible/Frontend Changes
- [x] 'plus' button looks disabled if permissions are missing

--

### DIFF
--- a/src/app/core/entity-components/entity-subrecord/entity-subrecord/entity-subrecord.component.scss
+++ b/src/app/core/entity-components/entity-subrecord/entity-subrecord/entity-subrecord.component.scss
@@ -3,26 +3,9 @@
 @use "@angular/material" as mat;
 
 .table-action-button {
-  font-size: 115%;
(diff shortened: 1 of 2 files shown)


src/app/core/entity-components/entity-subrecord/entity-subrecord/entity-subrecord.component.scss
```diff
@@ -3,26 +3,9 @@
 @use "@angular/material" as mat;
 
 .table-action-button {
-  font-size: 115%;
-  color: mat.get-color-from-palette(theme.$accent);
```
sleidig, line 7: We should stick to the color theme and display these actions in "accent" (i.e. blue). Now the button feels disabled even when it's not ;-)

sleidig (changes requested):

PR #1573, opened 2022-12-06
feat: make any primary entity type searchable

sleidig:
make any primary entity type searchable (e.g. also recurring activities)

if the entity type has a "human-readable" (i.e. overridden) toString method


src/app/core/entity/model/entity.ts
```diff
@@ -185,19 +185,21 @@ export class Entity {
...
    * <b>Overwrite this method in subtypes if you want an entity type to be searchable by other properties.</b>
    */
   @DatabaseField() get searchIndices(): string[] {
-    let indices = [];
-
-    // default indices generated from "name" property
-    if (typeof this["name"] === "string") {
-      indices = this["name"].split(" ");
+    if (
+      this.getConstructor().toStringAttributes === Entity.toStringAttributes &&
```
TheSlimvReal, line 194: When is this case necessary? Somebody could put `entityId` in the `toStringAttributes` but I don't think that case really needs to be covered.

TheSlimvReal (approved):
Works very well and thanks for the cleanup

src/app/core/entity/model/entity.ts
sleidig: I needed this because for `User` the entityId is the same as the name ... and then user entities didn't show up in search.

If you have a more solid idea to circumvent this, I'll be happy! :see_no_evil:

PR #1577, opened 2022-12-07
Multi entity select

sleidig:
Entity select field supports loading multiple different entity types in a single field now.


src/app/core/entity-components/entity-select/entity-select/entity-select.component.ts
```diff
@@ -38,16 +38,14 @@ export class EntitySelectComponent<E extends Entity> implements OnChanges {
...
-  @Input() set entityType(type: string) {
-    this.loading.next(true);
-    this.entityMapperService.loadType<E>(type).then((entities) => {
-      this.allEntities = entities;
-      this.loading.next(false);
-      this.formControl.setValue(null);
-    });
+  @Input() set entityType(type: string | string[]) {
+    if (typeof type === "string") {
+      type = [type];
```
TheSlimvReal, line 46: Lets make `withPrefix` true when `type` is an array. I cannot imagine a scenario where this would not be wanted.

sleidig: good point, I am pulling this from the EditEntityArrayComponent down into the general EntitySelectComponent :+1:

Issue #1570, opened 2022-12-05
Filters in "Record Attendance" not showing all options in dropdown


sleidig:
**Describe the bug**
If there are more than 5 activities available, the "Record Attendance" view is displaying dropdown filters to the user to make it easier to find the relevant activity. The options to select in the filters are not including all the values in the data, however. Some client has this issue both for category and group filter.

for example, check https://dev.aam-digital.net/attendance/add-day --> category "Guardian's Meeting" not included.

Issue #1478, opened 2022-10-15
Filter by group/team in Participants list not reliable


sleidig:
**Describe the bug**
In Participants list the filter for group/team is sometimes excluding records that are in the selected group.

**To Reproduce**
1. Go to 'Participants'
2. (very quickly!) select a group (i.e. team/school/...) from the filter dropdown.
3. The list now contains X participants in that group
4. Select "All" again in the filter for group and then re-select the same group to filter for it again.
5. The list now contains Y participants, a higher number.

**Expected behavior**
The filtered list must contain every participant in that group, regardless of how quickly the filter was used.

(probably group details are loaded asynchronously and is not available for all records yet at the time of first filtering here?)

sleidig:
fixed with #1754

PR #1584, opened 2022-12-12
fix: change detection is correctly triggered to build new filters

TheSlimvReal:
see issue: #1570 
As the additional events were only pushed to the existing array, no new change detection was triggered. Instead, we now correctly create a new array once all the additional events were created.

### Visible/Frontend Changes
- [x] Filters in roll call setup show all available options

--


sleidig (approved):
:100: 

This is not related to #1478, is it?

PR #1590, opened 2022-12-13
Fix entitysubrecord

sleidig:
see commits


sleidig:
Wait ... this actually breaks the columnsToDisplay. I wanted to be able to display a column added later programmatically - but now the EntityList tabs are messed up

src/app/core/entity-components/entity-subrecord/entity-subrecord/entity-subrecord.component.ts
```diff
@@ -224,9 +228,7 @@ export class EntitySubrecordComponent<T extends Entity>
   ngOnChanges(changes: SimpleChanges) {
     if (changes.hasOwnProperty("columns")) {
       this.initFormGroups();
-      if (this.columnsToDisplay.length < 2) {
```
sleidig, line 227: as unintuitive as this check looks, it is necessary - otherwise displayColumns provided by config are lost ...

TheSlimvReal (approved):

PR #1605, opened 2022-12-20
fix(matching): do not remember selected entities across pages

### LINKED ISSUE #1592: Prevent config mutation in `MatchingEntitiesComponent`
The matching component alters the config object which is received by the router. This leads to problems because the changes will be persisted throughout site changes. 

The component should be changed in order to not alter the config but only read from it.

sleidig:
closes #1592


sleidig:
see PR #1606

PR #1612, opened 2022-12-22
fix: more sophisticate applying of remote changes in entity-form

sleidig:
if external changes only affect untouched form fields or properties that are not even included in the form, these could directly be applied without interrupting the user with a popup.


src/app/core/entity-components/entity-form/entity-form/entity-form.component.ts
```diff
@@ -96,22 +97,44 @@ export class EntityFormComponent<T extends Entity = Entity> implements OnInit {
       .subscribe(({ entity }) => this.applyChanges(entity));
```
sleidig, line 97: If I remember correctly this would need an `untilDestroyed` here?

TheSlimvReal (approved):

Issue #1559, opened 2022-11-29
More e2e tests for regressions we recently fixed


sleidig:
With major refactorings we have (re-)introduced some problems. These would be good cases to write e2e tests for, so that we prevent future regressions.

- [ ] filter components above entity list should dissapear on smaller screeens
- [ ] tab headers and edit button should stay fixed on screen even when scrolling in entity details
- [ ] buttons disabled by a permission should look disabled + really not trigger any kind of action

sleidig:
closing this in favor of new e2e issue #2612

PR #1614, opened 2022-12-22
Add an E2E test that checks whether filter components are visible on small screens

Schottkyc137:
Towards: #1559


TheSlimvReal (approved):
Thanks for the nice looking test

Issue #1500, opened 2022-10-24
Add missing participants to events when opening details from the RecurringActivity Details view


sleidig:
**Describe the bug**

**To Reproduce**
1. Create an event for a recurring activity. 
2. Add a participant with a start date before that event to the group of the recurring activity (probably also works directly adding participant to Recurring Activity?)
3. View the Events & Attendance tab of the Activity Details
4. The event is now marked red in the calendar because there is attendance data missing (i.e. for the participant added in (2.))
5. Click on the date and "Details" to show the event details. The list of participants does not include the new participant and all existing participants in the event actually have attendance data. This is confusing the users, as they have now guidance on how to resolve the red warning of missing attendance data.

**Expected behavior**
When opening the event details, the system should automatically add the missing participants to the event with an empty attendance status. If the user cancels, the event remains unchanged (new participant not saved to database). If the user saves (with or without filling the attendance status for the new participant), the event is updated to include all expected participants.

also see related: #2621

PR #1616, opened 2022-12-22
feat(attendance): participants can be excluded from activities

sleidig:
participants can be excluded from activities if only a few group members should not show up in attendance workflows
(based on client request)


TheSlimvReal (changes requested):
There is a problem with the attendance calendar component. When a child has been previously excluded from a activity, but then is part again, the attendance calendar component throws errors when clicking on dates, where the child has been excluded, but attendance was reported.

This is because the `event.getAttendance(childId)` call returns undefined. Just returning an empty attendance here does not solve the problem, because changes on this will not be saved on the event. What would need to happen, is this child needs to be added to the event and then the new attendance saved. @sleidig what do you think should happen in this case?

sleidig:
Oh, interesting ... good find! This seems somewhat related to #1500 also.

The system could maybe add an new "NullAttendanceStatusType" attendance (which is counted as IGNORE) on the fly for that participant in the event and _not_ save it unless the user actually actively saves it? That way we would not unnecessarily edit existing events but offer users a straightforward way to add a status if necessary.

TheSlimvReal (approved):

PR #1617, opened 2022-12-27
Form refactor

TheSlimvReal:
This refactors the `EntityFormComponent` to only display the form and not handle any save/cancel logic.

### Visible/Frontend Changes
--

### Architectural/Backend Changes
- [x] `FormComponent` handles saving form
- [x] `JSON.stringfy` is used to compare complex properties (dates, objects, ...) for equality to better detect, when a form entity has been changed.
- [x] `RowDetailsComponent` uses `EntityFormComponent` to display form fields

Issue #1607, opened 2022-12-20
Embedabble form


TheSlimvReal:
*As a Project Manager I want to embed a form on my public website in order to allow people to quickly register to my program.*
It should be possible to create forms inside the application and then embed them on third party websites or send them around via a link.

Requirements are:
- [ ] From within Aam Digital a user can create a form and decide, which fields should be in the form
- [ ] After generating the form, the user has the option to either embed the form in a website (through `iframe`/`script` tags) or use a link to open the form on a single page
- [ ] Form submissions are directly saved to the database. This access to the databases is highly restricted and only allows to save certain kinds of documents (no reading or updating).
- [ ] Some properties should be automatically set on the form submission, e.g. a status that indicates that this comes from a external registration

## **Objective**
Project managers want to allow people to register to their program independently. This should make their work easier and also lower the threshold for a registration.

## **Proposed Solutions & Alternatives**
I see two approaches to this:

### 1. Use an external form builder
Requirements for this are

- Forms can be generated through API (from JSON)
- Forms are embedabble with plain HTML/JavaScript as well as shown in own pages
- It can connect to external database/send submissions automatically
- It is free (?)

[form.io](https://www.form.io) would provide all the required functionality but is rather expensive. I have not found a free alternative yet.

### 2. Integrate this in our app
Pros:
 - it is flexible, we can build it the way we need it
 - we can reuse the existing eco system with the schemas and form components we already have
 - its free und independent of external services

Cons:
 - our app has to be refactored in order to use more lazy loading, otherwise the app is too big for showing it in an iframe. However, this would generally be a great thing to have (e.g. for #229).
 - More development work as we have to implement things, other provider might have already solved (in a better way as they are specialised on that)

TheSlimvReal:
As we are going for approach 2 there are the following steps:

- [ ] Create a form on a empty page that can be accessed without logging in
- [ ] Allow submitting/saving of this from without logging in (and without offline mode)
- [ ] Add `public` rules to permission system which defines what can be done by not logged-in users in order to allow doc to be saved
- [ ] Make as much as possible lazy loaded in order to reduce initial loading time of app
- [ ] Implement UI that allows to defines new forms
- [ ] Create a code snippet that allows users to directly integrate the form into any website

PR #1618, opened 2022-12-27
Feature public/embeddable form 

TheSlimvReal:
see issue: #1607 

### Visible/Frontend Changes
- [x] Public form can be reacher without login under `/public-form/:id`
- [x] Improve UI of public form (requires #1617)
- [x] Login happens through redirects instead of `*ngIf` on same route

- [x] Created online-only mode through the remote session
- [x] Added `AuthGuard` for all routes that require login
- [x] Added support for `public` permissions in the `EntityAbility`, also see the [backend changes](https://github.com/Aam-Digital/replication-backend/pull/90)


src/app/features/public-form/public-form.component.html
```diff
@@ -0,0 +1,17 @@
...
+>
+  Submit Form
+</button>
+<button
+  mat-stroked-button
+  (click)="reset()"
+  i18n="Reset form button"
+>
+  Reset
+</button>
```
TheSlimvReal, line 17: The UI is still very raw. any suggestions for this? @sleidig

src/app/features/public-form/public-form.component.spec.ts
```diff
@@ -0,0 +1,109 @@
...
+    expect(initIndexedDBSpy).toHaveBeenCalled();
+  });
+
+  it("should initialize component with values from PublicFormConfig once config is ready", fakeAsync(() => {
+    expect(component.entity).toBeUndefined();
+
+    initConfig();
+    tick();
+
+    expect(component.entity.getConstructor()).toBe(Child);
```
sleidig, line 71: would be a bit easier to follow if you set the relevant formConfig property in this test here

src/app/features/public-form/public-form.component.ts
```diff
@@ -0,0 +1,107 @@
...
+    private route: ActivatedRoute,
+    private entities: EntityRegistry,
+    private entityMapper: EntityMapperService,
+    private entityFormService: EntityFormService,
+    private configService: ConfigService,
+    private entitySchemaService: EntitySchemaService,
+    private snackbar: MatSnackBar
+  ) {
+    // TODO the component should probably not handle this and it is very similar to the RemoteSession
+    this.database.initIndexedDB(
```
sleidig, line 49: what about that plan to have a RemoteSession instead of directly working with the db here? :blush:

src/app/utils/database-testing.module.ts
```diff
@@ -26,7 +27,8 @@ import { AppModule } from "../app.module";
   ],
 })
 export class DatabaseTestingModule {
-  constructor(pouchDatabase: PouchDatabase) {
+  constructor(pouchDatabase: PouchDatabase, components: ComponentRegistry) {
+    components.allowDuplicates();
```
sleidig, line 31: is that really linked with the Database in any way - i.e. should it be here or rather in a different test setup or module?

sleidig (commented):
code and UI looks really good! :smiley: 

For embedding the form somewhere (rather than just linking to it), I think we would still need an option to only display the component without the toolbar and other structure.

sleidig:
Further observations from testing:
- console still shows some 401 errors. I guess it doesn't break anything but would these bubble up into sentry? (and generally it's great if we can catch errors explicitly)
- permission to enforce the `status: "new"` and prevent overwriting or reading entities seems to work well. My quick attempt to break security using Postman didn't allow me to any sensitive actions

TheSlimvReal:
For the moment I don't see any why how the errors can be caught. They occur because the app is listening to updates of the config object which is done through the changes feed. This endpoint however requires a login. The endpoint should probably not be publicly available but it's also very difficult to prevent this request from happening.

Issue #1600, opened 2022-12-19
Make all components lazy loaded


TheSlimvReal:
Currently, all our components are referenced in the main bundle which makes the size for the initial loading very big.
Using Angular's lazy loading strategy this can be improved and we already use this for some hard coded components like the `ConflictResolutionListComponent`.
The difficulty is to get lazy loaded running together with the dynamic routes system that is introduce through the config and the `RouterService`.

I have described a potential solution [here](https://stackoverflow.com/a/74854987/10713841)

What needs to be done is transform components into smaller modules (maybe using Angulars latest [standalone components](https://angular.io/guide/standalone-components)) that can be lazy loaded and add these modules under a static path name as a lazy loaded route. Then add routes dynamically by looking for the route that lazy loads this component and re-use the `loadChildren` function for the new route.

PR #1624, opened 2023-01-03
Lazy loading

TheSlimvReal:
see issue: #1600 

### Visible/Frontend Changes
--

- [x] All dynamic components are lazy loaded
- [x] `DemoDataModule` is lazy loaded
- [x] Most services are `providedIn: "root"` in order to support lazy loading
- [x] All components are standalone in order to support lazy loading
- [ ] Fix storybook stories


src/app/core/coming-soon/coming-soon/coming-soon.component.ts
```diff
@@ -9,17 +9,31 @@ import {
...
+import { FontAwesomeModule } from "@fortawesome/angular-fontawesome";
+import { NgIf } from "@angular/common";
+import { RouteTarget } from "../../../app.routing";
 
 /**
  * Placeholder page to announce that a feature is not available yet.
  *
  * This integrates with analytics and allows user to explicitly request the feature.
  */
+@RouteTarget("ComingSoon")
```
TheSlimvReal, line 24: Should we replace the `RouteTarget` decorator with the `DynamicComponent` decorator as it is has no special meaning any more. In fact, both the registries are not really needed anymore as the lookup can happen through the global `dynamicComponents` map, which holds the lazy import functions. The `DynamicComponent` decorator has the benefit, that after the first loading, the component can be used without another lazy loading. What performance impact this has - I am not sure.

src/app/dynamic-components.ts
```diff
@@ -0,0 +1,488 @@
+import { Type } from "@angular/core";
+
+// TODO make a build script that looks for annotations and creates this
+export const dynamicComponents = new Map<string, () => Promise<Type<any>>>([
+  [
+    "Dashboard",
+    () =>
+      import("./core/dashboard/dashboard/dashboard.component").then(
```
TheSlimvReal, line 8: I think these functions are quite bad for the code coverage as none of them are executed in the tests.

src/app/core/view/dynamic-components/dynamic-component.directive.ts
```diff
@@ -22,16 +25,23 @@ export class DynamicComponentDirective implements OnChanges {
...
 
-    const component = this.registry.get(this.appDynamicComponent.component);
+    let component: ComponentType<OnInitDynamicComponent>;
+    if (this.registry.has(this.appDynamicComponent.component)) {
+      component = this.registry.get(this.appDynamicComponent.component);
+    } else {
+      component = await dynamicComponents.get(
+        this.appDynamicComponent.component
+      )();
+    }
```
sleidig, line 44: if (almost) everything is lazy-loaded, do we stil have use for the registry at all?

sleidig (approved):
Tons of code but actually pretty clear and good :slightly_smiling_face:

src/app/child-dev-project/attendance/attendance.module.ts
```diff
@@ -83,87 +33,11 @@ export class HorizontalHammerConfig extends HammerGestureConfig {
...
-    AddDayAttendanceComponent,
-    AttendanceManagerComponent,
-    ActivityAttendanceSectionComponent,
-    AttendanceWeekDashboardComponent,
-    GroupedChildAttendanceComponent,
-    AttendanceBlockComponent,
-    RecentAttendanceBlocksComponent,
-  ];
-}
+export class AttendanceModule {}
```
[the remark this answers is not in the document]
sleidig: same for entityRegistry, ChildrenModule, PwaInstallModule - if they only provide one or two services, wouldn't it be better to do that in a different way than through an empty module?

sleidig:
We should mark the final commit as QualitätsMENTOR also.

Issue #1626, opened 2023-01-04
invitation of new user fails if there are spaces in email field


sleidig:
if by pasting an email accidentally there is a leading space in the email field, inviting the new user fails and there is no very clear guidance to the user how to fix the problem. Maybe we should trim whitespaces. And add error messages to the form field like "email address not valid"?

PR #1631, opened 2023-01-05
Notify on incorrect email, automatically trim whitespaces

Schottkyc137:
see issue: #1626 

### Visible/Frontend Changes
- [x] An error will be shown when the entered email is not valid or when the field is empty
- [x] The email field will automatically be trimmed

The current implementation is quite 'aggressive' in that it doesn't even allow a user to enter leading or trailing whitespaces. I think that this is fine since there is no valid email that starts or ends with a whitespace.


TheSlimvReal (approved):
Nice, thanks for the quick fix

Issue #1634, opened 2023-01-07
Lazy loaded entities are not configured correctly


TheSlimvReal:
With the changes from #1624 also entity subclasses are loaded lazy. This creates the issue, that the schema extension only runs once at the beginning on the entities that have been registered so far. E.g. the `HistoricalEntityData` entity is only referenced inside the `HistoricalDataComponent` and therefore only registered, once this component is rendere. The config for this entity will never be applied because this happens much earlier.

The problem could be solved two ways:

1. (easy) create a static map with all entities
2. (better) run `EntityConfigService.addConfigAttributes()` only when the entity has been registered (through a callback)

PR #1635, opened 2023-01-08
Registry callback

TheSlimvReal:
see issue: #1634 

### Visible/Frontend Changes
--

- [x] `DynamicRegistry` allows to register callbacks
- [x] entity schemas are applied using a callback


TheSlimvReal:
Unfortunately, this does not fully solve the problem as entity can be used without a static reference, e.g. showing `HistoricalEntityData` in `EntityDetails` will still throw an error

Issue #1150, opened 2022-03-21
Allow groupBy for exports through the Query Service


sleidig:
We allow complex exports (resolving linked entities and relationships) through our "Reporting" feature and its QueryService. Aggregated statistics allow a "groupBy" field that splits the data and adds multiple outputs, one for each group-by value. This is not available for the "export" type queries at the moment.

In the example of attendance report for each activity type we work around this by copying the same query and filtering it by each interaction type in the config. That is really not feasible anymore for things like a group by team.

---
Create a Query Service that allows more flexibility with such exports and can duplicate/split exports based on group-by fields. This has some technical challenges and may even be easier with a complete re-design of the QueryService?

PR #1636, opened 2023-01-08
Export groupBy

TheSlimvReal:
see issue: #1150 

### Visible/Frontend Changes
--

- [x] A `groupBy` option is allowed on the top level of export configs


src/app/core/export/export-service/export.service.ts
```diff
@@ -91,14 +95,32 @@ export class ExportService {
...
     if (config) {
-      for (const dataRow of data) {
-        const extendedExportableRows = await this.generateExportRows(
-          dataRow,
-          config,
-          from,
-          to
-        );
-        flattenedExportRows.push(...extendedExportableRows);
+      if (groupByProperty) {
```
TheSlimvReal, line 98: So it is technically possible to extend the service with this new feature, however it just becomes more and more complex.

src/app/core/export/export-service/export-column-config.ts
```diff
@@ -24,4 +24,6 @@ export interface ExportColumnConfig {
    * => overall result: two export rows: [{ name: "child1", phone: "123"}, {name: "child2", phone: "567"}]
    */
   subQueries?: ExportColumnConfig[];
+
+  groupBy?: { label: string; property: string };
```
TheSlimvReal, line 28: So this would allow one `groupBy` on each level

sleidig (approved):
as discussed and refactored in pair programming.

Looks good to me know but please glance over it again with these changes!

PR #1641, opened 2023-01-11
fix: handle invalid dropdown (configurable-enum) options gracefully

### LINKED ISSUE #1628: handle loading of enum datatype values that are undefined in the options of the enum
If the value an entity holds for a enum-multi property (array - innerDataType configurable-enum) does not exist as an option in that configurable enum, an error gets logged:

> Error: Transformation for myCategory failed: TypeError: Cannot read properties of undefined (reading 'id')

This also leads to the form buttons dissapearing after pressing "Cancel". The problem does not happen for normal (non-array) enum values that do not exist.

Further thoughts in the longer run:
- maybe even give users a feedback, which (invalid) value is currently stored in this field so that they have full information to resolve the problem manually?

sleidig:
and display them to users for manual correction

closes #1628


sleidig:
UI now shows:
image

TheSlimvReal (approved):
Good handling of the error cases in the UI

Issue #1601, opened 2022-12-20
Allow multiple location properties to be displayed on overview map


sleidig:
*As a coordinator I want to match mentors and mentees based on their home as well as work address.*

The MatchingEntities view displays a map of all available entities' locations. For more advanced use cases this should allow multiple locations per entity as well:

- [ ] - display multiple location properties of each entity (e.g. workAddress, homeAddress)
- [ ] - allow to switch on/off which location properties are visualized and which ones are hidden
- [ ] - allow to switch on/off whether to only display locations for the two selected/matched entities or for all available ones
- [ ] - only display the filtered available entities for matchings

PR #1662, opened 2023-01-12
Multiple locations in Map

TheSlimvReal:
see issue: #1601 

the storybook showcases it.

### Visible/Frontend Changes
- [x] Added selection in matching view which allows to select one or more location properties of an entity
- [x] Only filtered entities are shown in the map

- [x] Matching component should work as standalone (currently failing) (already merged -> #1667)


src/app/core/config/config-fix.ts
```diff
@@ -1115,7 +1118,6 @@ export const defaultJsonConfig = {
       ["address", "address"],
       ["distance", "privateSchool"],
     ],
-    "showMap": ["address", "address"],
```
sleidig, line 1118: any chance we can keep the config for map properties in this central place? The usual setup integrates the matching in 3 views, each with (almost) same config so it is really tedious to copy and paste.

sleidig (commented):
Code looks okay (although the component is - and already has been before - extremely complex).

Visually I think the relation of the dropdowns (and what it means to select from them - i.e. that this is a filter) is not very clear yet. I could imagine that it is more clear if we can integrate the filter into the map view as a filter icon button next to the fullscreen button, which then opens a list of selectable properties. That could be a list of both entity types, I think ("Child: address", "Mentor: work", "Mentor: home", ...). And it might also allow people to filter while viewing the map fullscreen.

src/app/core/config/config-fix.ts
TheSlimvReal: Architecture wise it is very annoying to have this separate from the sides. Logically, I definitely belongs to the sides (_which properties of this sides entity should be displayed in the app_). So not having it in the side means a lot of back and forth with indices and searching. I would rather not having to do this. Of course we could just assign it at runtime but that just adds unnecessary complexity. I see the issue with the config but it just seems much more logical this way.

sleidig: we could look into setting default configs for the sides in `"appConfig:matching-entities"` (that is currently the other format but sides are also fixed through that).
But at the moment, maybe let's just keep it as you did and come back to this some time later if we have any ideas to simplify code and config.

sleidig (approved):

Issue #1610, opened 2022-12-21
Import array of values (for enum-multi)


sleidig:
For properties of type configurable-enum array (multi-select) an import for values from CSV is not yet supported. This would be useful to have.

One approach could be to allow a comma-separated string that is parsed into an array during import.

TheSlimvReal:
With the latest changes to the import feature, this functionality has been removed/not migrated. 
If the requirement comes up again it would be good to re-implement it.

sleidig:
To achieve this:

1. make sure you have a multi-select dropdown field in the entity type (e.g. change the "risk status" field to allow multiple values in the education demo's Child entity type)
2. Import a file with some multi-valued data, e.g.
```
name,risk status
John,"a,b"
Doe,a;b;c
```
3. In the import UI, step 2, you can expand and select a "multi-value separator"
4. in the value mapping dialog (button for the field in step 3 of the import UI), you should see not the exact cells to be mapped, but the individual values that are split to be mapped individually (that should make a lot less different values here also). Map the single values accordingly
5. in the preview you should see that multiple enum values (multiple chips) are imported for a single record

sleidig:
Thanks @george-neha . Polished a bit and added to the user guide now: https://chatwoot.help/hc/aam-digital/articles/1730371253-importing-data-from-a-spreadsheet-csv-file#importing-multiple-values-from-a-single-column-for-a-multi-select-dropdown-field

PR #1665, opened 2023-01-17
feat(import): array properties can be imported

TheSlimvReal:
Array properties can be imported with CSV. Have a look at the tests to see the supported formats

see #1610


src/app/features/data-import/data-import.service.ts
```diff
@@ -134,11 +140,29 @@ export class DataImportService {
       return Entity.createPrefixedId(importMeta.entityType, value);
     } else if (importMeta.dateFormat && this.dateDataTypes.includes(dataType)) {
       return this.transform2Date(value, importMeta.dateFormat);
+    } else if (this.arrayDataTyps.includes(dataType)) {
```
sleidig, line 143: do we need to stack this recursively in a way so that array items might also be parsed with transform2Date?

sleidig (approved):

src/app/features/data-import/data-import.service.ts
TheSlimvReal: Generally yes. I would however postpone this to when there is actually a use-case for this. At the moment our app is not supporting arrays of dates anyways and there are no other data types that are transformed during import.

Issue #1056, opened 2021-12-02
Define new Entity types completely through config


sleidig:
do not rely on pre-defined entity classes in Typescript
use specific interface/abstract classes the hold the high-level concept used in components
→ allow a flexible number of sub-entities and sections (e.g. HELGO: list of previous health checkups)

PR #1670, opened 2023-01-19
Dynamic entities

TheSlimvReal:
see issue: #1056 

### Visible/Frontend Changes
- [x] 
- [ ] 

- [x] New entities can be defined in config

### DIFF
--- a/src/app/core/entity/entity-config.service.spec.ts
+++ b/src/app/core/entity/entity-config.service.spec.ts
@@ -128,7 +128,7 @@ describe("EntityConfigService", () => {
     mockConfigService.getAllConfigs.and.returnValue([
       {
         _id: "entity:DynamicTest",
-        label: "DynamicTest",
+        label: "Dynamic Test Entity",
         extends: "Test",
         attributes: [{ name: "dynamicProperty", schema }],
       },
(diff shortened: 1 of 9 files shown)


sleidig:
reports do not support the custom entities yet?

src/app/core/entity/entity-config.service.spec.ts
```diff
@@ -134,6 +122,49 @@ describe("EntityConfigService", () => {
...
+    service.setupEntitiesFromConfig();
+
+    const dynamicEntity = entityRegistry.get("DynamicTest");
+    expect(dynamicEntity.ENTITY_TYPE).toBe("DynamicTest");
+    expect([...dynamicEntity.schema.entries()]).toEqual(
+      jasmine.arrayContaining([...Test.schema.entries()])
+    );
+    expect(dynamicEntity.schema.get("dynamicProperty")).toBe(schema);
+    const dynamicInstance = new dynamicEntity("someId");
+    expect(dynamicInstance instanceof Test).toBeTrue();
```
sleidig, line 146: but Test should still be usable without the extension in parallel. Does this work if the extended is still also instanceof Test?
Let's add an expectation that Test does not have the dynamicProperty here to be safe?

PR #1673, opened 2023-01-20
fix(demo): added public form in demo mode

TheSlimvReal:
A simple public form is available in the demo mode.

### Visible/Frontend Changes
- [x] 
- [ ] 

- [x] 
- [ ]

### DIFF
--- a/src/app/core/config/config-fix.ts
+++ b/src/app/core/config/config-fix.ts
@@ -168,6 +168,11 @@ export const defaultJsonConfig = {
                 "label": $localize`:Dashboard shortcut widget|record attendance shortcut:Add Child`,
                 "icon": "plus",
                 "link": "/child/new",
+              },
+              {
+                "label": $localize`:Dashboard shortcut widget|open public form:Registration Form`,
+                "icon": "file-circle-check",
+                "link": "/public-form/test",
               }
             ]
           }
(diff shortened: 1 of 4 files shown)


src/app/features/public-form/demo-public-form-generator.service.ts
```diff
@@ -0,0 +1,25 @@
...
+        provide: DemoPublicFormGeneratorService,
+        useClass: DemoPublicFormGeneratorService,
+      },
+    ];
+  }
+
+  protected generateEntities(): PublicFormConfig[] {
+    const form = new PublicFormConfig("test");
+    form.title = $localize`Example form`;
+    form.description = $localize`This is a form that can be sent around or embedded in a website.`;
```
sleidig, line 19:
```suggestion
    form.description = $localize`This is a form that can be shared as a link or embedded in a website. It can be filled by users without having an account. For example you can let participants self-register their details and just review the records within Aam Digital.`;
```

src/app/core/config/config-fix.ts
```diff
@@ -168,6 +168,11 @@ export const defaultJsonConfig = {
                 "label": $localize`:Dashboard shortcut widget|record attendance shortcut:Add Child`,
                 "icon": "plus",
                 "link": "/child/new",
+              },
+              {
+                "label": $localize`:Dashboard shortcut widget|open public form:Registration Form`,
```
sleidig, line 173:
```suggestion
                "label": $localize`:Dashboard shortcut widget|open public form:Public Registration Form`,
```

PR #1678, opened 2023-01-23
fix: updated manifest file

TheSlimvReal:
Fixed app install problem (again).

### Visible/Frontend Changes
- [x] 
- [ ] 

- [x] Fixed broken urls for manifest icons
- [x] Updated some namings


sleidig (approved):
seems to (still) work on my laptop and phone also :+1:

Issue #1683, opened 2023-01-26
UI has scrollbars


sleidig:
![image](https://user-images.githubusercontent.com/1682541/214812256-13214f6c-162f-4cd3-acb8-1734496d2f83.png)

sleidig:
![image](https://user-images.githubusercontent.com/1682541/216077452-34ec087b-ae7d-4b3e-a04c-920075934b67.png)

sleidig:
found another one :see_no_evil: 
![image](https://user-images.githubusercontent.com/1682541/217260728-3fb9ebcf-9d81-4376-a060-f21e0b1f8f62.png)

PR #1701, opened 2023-02-03
Scrollbar fixes

TheSlimvReal:
see issue: #1683
See commits

### Visible/Frontend Changes
- [x] 
- [ ] 

- [x] 
- [ ]

### DIFF
--- a/src/app/core/entity-components/entity-list/entity-list.component.ts
+++ b/src/app/core/entity-components/entity-list/entity-list.component.ts
@@ -179,8 +179,8 @@ export class EntityListComponent<T extends Entity>
     }
 
     this.listName =
-      this.listConfig.title ??
-      this.listName ??
+      this.listConfig.title ||
+      this.listName ||
       this.entityConstructor?.labelPlural;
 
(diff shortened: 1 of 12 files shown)


src/app/core/entity-components/entity-list/entity-list.component.ts
```diff
@@ -179,8 +179,8 @@ export class EntityListComponent<T extends Entity>
     }
 
     this.listName =
-      this.listConfig.title ??
-      this.listName ??
+      this.listConfig.title ||
```
sleidig, line 182: Any reason why you prefer || over ?? here? From my understanding, I would see ?? as the more logical choice here https://stackoverflow.com/a/61481631/1473411

sleidig (approved):
Great! :+1:

src/app/core/entity-components/entity-list/entity-list.component.ts
TheSlimvReal: Yes, but all our list titles are empty because the component has a default of `listName = ""` which is then always used instead of the actual plural title of the entity.

PR #1702, opened 2023-02-04
fix: memory leaks across app due to unsubscribed subscriptions

Arooba-git:
Hello 👋 
As part of our project, we are using Facebook's new Memlab tool to detect memory leaks in SPA applications and their libraries. While running the tool and analyzing the code of ndb-core, we saw that your project does a very good job of ensuring that all async operations are cancelled when components destroy. However, as per Memlab execution results, we found some subscriptions causing the memory to leak (screenshots below).

[before]
Screen Shot 2023-02-04 at 10 39 38 PM
<br />
Hence we added the fix by unsubscribing the subscriptions in destructors, and you can see the heap size and # of leaks reducing noticeably:
<br />
Screen Shot-after-subscription-fix

Note: Host listener is used to ensure that ngOnDestroy is called not only when the class destroy, but also when the page refreshes, tab or browser closes or the user navigates away from the page[1]

We ran the test suite before and after the patches and it executed 1031 of 1038 (skipped 7) successfully.
We also ensured all files pass linting after the fixes.

You can analyze this and other potential leak sources, if you like, by running [Memlab](https://facebook.github.io/memlab/) with a scenario file covering the maximum # of use cases.

Following is a sample of the scenario file we used (it needs to be a .js file but attaching here in .txt form):
[ndb-core-scenario-memlab.txt](https://github.com/Aam-Digital/ndb-core/files/10609105/ndb-core-scenario-memlab.txt)

Note that some other reported leaks (in Memlab) originated from Angular's internal objects, and hence were ignored.

References:
[1] https://wesleygrimes.com/angular/2019/03/29/making-upgrades-to-angular-ngondestroy.html

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -20,6 +20,8 @@ import {
   Injector,
   ViewContainerRef,
   ɵcreateInjector as createInjector,
+  OnDestroy,
+  HostListener,
 } from "@angular/core";
 import { AnalyticsService } from "./core/analytics/analytics.service";
 import { ConfigService } from "./core/config/config.service";
@@ -34,6 +36,9 @@ import { LoginState } from "./core/session/session-states/login-state.enum";
 import { LoggingService } from "./core/logging/logging.service";
 import { EntityRegistry } from "./core/entity/database-entity.decorator";
 
+import { takeUntil } from "rxjs/operators";
+import { Subject } from "rxjs";
+
 /**
  * Component as the main entry point for the app.
  * Actual logic and UI structure is defined in other modules.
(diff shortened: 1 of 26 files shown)


sleidig:
Thanks a lot for these pointers @Arooba-git !

TheSlimvReal:
@Arooba-git also thank you from my side for your analysis and contribution. This was very helpful. I have marked you in the PR #1704 where I used your suggestions but with the way how we generally unsubscribe from observables (see [until-destroy](https://www.npmjs.com/package/@ngneat/until-destroy)).

PR #1704, opened 2023-02-06
perf: temporary subscriptions are properly unsubscribed

TheSlimvReal:
I went through the suggestions in #1702 and implemented the ones that are actually problematic.
The following were left untouched:

- HTTP observables are closed once completed
- Route params subscriptions are cleaned up once user routes away (see [this post](https://stackoverflow.com/a/41177163/10713841))
- Services that are singletons

### Visible/Frontend Changes
--

- cleaning up subscriptions that are not needed any more

### DIFF
--- a/src/app/core/form-dialog/form-dialog.service.ts
+++ b/src/app/core/form-dialog/form-dialog.service.ts
@@ -1,4 +1,4 @@
-import { Injectable } from "@angular/core";
+import { Injectable, OnDestroy, HostListener } from "@angular/core";
 import { MatDialog, MatDialogRef } from "@angular/material/dialog";
 import { ComponentType } from "@angular/cdk/overlay";
 import { ConfirmationDialogService } from "../confirmation-dialog/confirmation-dialog.service";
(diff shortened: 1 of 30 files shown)


src/app/core/form-dialog/form-dialog.service.ts
```diff
@@ -61,7 +60,9 @@ export class FormDialogService {
     const dialogWrapper = dialogRef.componentInstance.formDialogWrapper;
     dialogWrapper.readonly = this.ability.cannot("update", entity);
 
-    dialogWrapper.close.subscribe((res) => dialogRef.close(res));
+    dialogWrapper.close
+      .pipe(untilDestroyed(dialogWrapper))
```
sleidig, line 64: does this work without some `@UntilDestroy()` annotation at a component? Or do the passed components all have that? (probably difficult to ensure via typing?)

sleidig (approved):
Looks good! Hoping this will help general performance already :blush: 

Can we manage to rerun the tool of @Arooba-git and quantify the improvements?

And let's add her `Co-authored-by:` in this PR.

src/app/core/form-dialog/form-dialog.service.ts
TheSlimvReal: The passed component is the `FormDialogWrapperComponent` which has the annotation. The tests needed to be adjusted because they use a mock component.

TheSlimvReal:
I ran the tool again, the results are vary significantly though. I does find fewer memory leaks but the actual memory usage does not change drastically/constantly.

Issue #1707, opened 2023-02-06
Record attendance list should exclude "inactive" activities


sleidig:
**Describe the bug**
Recurring Activities can be marked archived through a profile property (e.g. `inactive`), which will make sure the `isActive` flag is false and the record can be hidden from various lists. The "Record Attendance" roll call setup screen should also respect this and exclude any isActive === false activities.

TheSlimvReal:
Maybe with tooltip on inactive checkbox in entity class

PR #1709, opened 2023-02-07
Hide inactive activities from roll call

TheSlimvReal:
see issue: #1707 

### Visible/Frontend Changes
- [x] Inactive participants are hidden from the roll call

- [x] The `inactive` property has a default configuration in the entity class


sleidig (approved):
thanks! :-)

Issue #1697, opened 2023-02-01
Problems filtering dates in reports


sleidig:
- [x] It seems dates filtered for reports have an overlap issue, records are also included if the period ends on the day before

- [x] if a time period that's in the future is selected, the report fails 
error "query_parse_error: No rows can match your key range, reverse your start_key and end_key or set {descending : true}"
interestingly, this does not happen if previously a report period in the past was selected since the last reload of the app

sleidig:
time range is a config issue: use `<` instead of `<=` as filter condition

PR #1710, opened 2023-02-07
fix: report dates in the future don't throw errors

TheSlimvReal:
see issue: #1697

### Visible/Frontend Changes
--

- [x] Data is fetched in the actually provided date range, not just from the `from` date till today


sleidig (approved):
:100:

Issue #1682, opened 2023-01-25
support any (dynamic) entity type in reports


sleidig:
Currently, not all entity types can be used in queries for reports. Especially the newly introduced dynamic entities (#1670) are not supported yet.

PR #1711, opened 2023-02-07
Support all entity types in reports

TheSlimvReal:
see issue: #1682 

### Visible/Frontend Changes
--

- [x] query string is checked for referenced entities which are then loaded individually


src/app/core/export/query.service.ts
```diff
@@ -82,33 +99,51 @@ export class QueryService {
...
+      .filter((entity) => {
+        const loadingInfo = this.dataLoaded[entity.ENTITY_TYPE];
+        return (
+          loadingInfo === undefined ||
+          !(
+            loadingInfo === true ||
+            (loadingInfo.from <= from && loadingInfo.to >= to)
+          )
+        );
+      });
```
sleidig, line 146: this is a slightly complex pyramid of maps for keys and cache status. Still reasonably easy to understand but possibly there could be a way to have a single data construct to keep it all? (e.g. query regex, loading function, cached status, ... as part of one object?) Not sure whether this would make things better though, just a random thought.

sleidig (approved):
Code looks good and basic reports tested successfully for a custom extended entity type

feel free to have a look at PR #1216 also and close, amend or finish that one, if it seems feasible after having picked up this topic in general :blush:

TheSlimvReal:
@sleidig updates of the cache are now also implemented. Please have a look at the new changes

src/app/core/export/query.service.ts
```diff
@@ -99,25 +119,41 @@ export class QueryService {
...
 
-    return Promise.all(dataPromises);
+  private applyEntityUpdates(uncachedEntities: EntityConstructor[]) {
+    uncachedEntities
+      .filter(({ ENTITY_TYPE }) => !this.entityInfo[ENTITY_TYPE].updating)
+      .forEach(({ ENTITY_TYPE }) => {
+        this.entityInfo[ENTITY_TYPE].updating = true;
+        this.entityMapper
+          .receiveUpdates(ENTITY_TYPE)
+          .subscribe(({ entity, type }) => {
```
sleidig, line 149: will this not re-subscribe to the same entity type updates after a broader date range of that type is loaded?

And in general, when are we cleaning up these subscriptions? (or if we don't yet, shouldn't we clear that cache at some time rather than keeping all that stuff in memory forever after a user once used a report?)

sleidig (commented):
please have a last look at those subscriptions and potential memory implications. Otherwise reports seem to work and code looks good :+1:

src/app/core/export/query.service.ts
TheSlimvReal: No, the `updating` flag is only set to `true` once and after that this entity type will always be filtered out here.

Generally, there is no concept of cleaning up yet as this service lives outside of the component. 
A option would be to just clear the cache once the user leaves the site but then this whole caching concept was pointless (as you can make changes without going away from the reporting page).

Issue #1663, opened 2023-01-12
Service worker should only be enabled once a user is logged in


TheSlimvReal:
Currently, the service worker is registered at the app start. As we are now also offering the app publicly as a form, this is unnecessarily caching a lot of files that will never be used.

The service worker module allows to add a [registrationStrategy](https://angular.io/api/service-worker/SwRegistrationOptions#registrationStrategy) there a observable can be provided and the service worker will be registered once the observable emits a value. There we can use the `LoginState` to only register this once the user has logged in.

PR #1712, opened 2023-02-08
fix: service worker is only registered once user logs in

TheSlimvReal:
see issue: #1663 

### Visible/Frontend Changes
--

- [x] service worker registration is only started after user is logged in


sleidig (changes requested):
deployed PR / demo version does not allow to log in with demo:pass anymore (and does not do it automatically).

How can we test the install + offline etc. cases for this thoroughly? Does it need to be deployed specially?

sleidig (approved):

PR #1715, opened 2023-02-09
fix: created custom form input for boolean values

TheSlimvReal:
Once again see issue: #1683
It seemed like the issue occurred, because `EditBooleanComponent` was not yet using a `mat-form-field`. 
Using the new `CustomFormControlDirective` I created a `BooleanInput` component that can be used inside a `mat-form-field`.
Some stylings needed to be adjusted.

### Visible/Frontend Changes
- [x] Editing booleans looks the same as other form components

--


sleidig (approved):
Looks and works good :+1: 
I actually like how this has a similar mat-form-field boxing!

I would tend towards making the checkbox label font-size same as the actual values of other fields. While logically "inactive" is a label not a value here, from the look and feel it appears a bit off to me:
image
what do you think?

Issue #1717, opened 2023-02-13
Cancel in configurable enum does not work correctly


TheSlimvReal:
When pressing cancel in a from, the configurable enum component does not properly reset the value.

PR #1720, opened 2023-02-13
fix: cancel works correctly in autocomplete component

TheSlimvReal:
see issue: #1717


sleidig (approved):
I don't really understand the code, but functionally this works ;-)

PR #1721, opened 2023-02-13
fix: failed remote login correctly logs-out user

TheSlimvReal:
Currently, the case where the password is changed remotely and the user logs in with the old password, does **not** result in the user being logged out. This has been fixed. The user is now correctly logged out and the old password does not work anymore to log-in offline.

### Visible/Frontend Changes
- [x] 
- [ ] 

- [x] Auth interceptor returns initial error if authentication fails
- [x] Synced session uses logout flow if remote login fails

### DIFF
--- a/src/app/core/session/session-service/synced-session.service.spec.ts
+++ b/src/app/core/session/session-service/synced-session.service.spec.ts
@@ -171,7 +171,7 @@ describe("SyncedSessionService", () => {
     // Initially the user is logged in
     expectAsync(result).toBeResolvedTo(LoginState.LOGGED_IN);
     // After remote session fails the user is logged out again
-    expect(sessionService.loginState.value).toBe(LoginState.LOGIN_FAILED);
+    expect(sessionService.loginState.value).toBe(LoginState.LOGGED_OUT);
     flush();
   }));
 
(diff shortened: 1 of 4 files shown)


src/app/core/session/session-service/synced-session.service.spec.ts
```diff
@@ -171,7 +171,7 @@ describe("SyncedSessionService", () => {
     // Initially the user is logged in
     expectAsync(result).toBeResolvedTo(LoginState.LOGGED_IN);
     // After remote session fails the user is logged out again
-    expect(sessionService.loginState.value).toBe(LoginState.LOGIN_FAILED);
+    expect(sessionService.loginState.value).toBe(LoginState.LOGGED_OUT);
```
sleidig, line 174: why should this not be expected to be state "failed"?

sleidig (approved):

src/app/core/session/session-service/synced-session.service.spec.ts
TheSlimvReal: The failed state does not really make sense in this context anymore as we are reloading the app on logout (due to various services that only work once per session and the different databases per users) therefore the session state here is not relevant and we are now just using the default logout state.

PR #1736, opened 2023-02-22
fix: forms have better visibility on iPhone

TheSlimvReal:
On Safari and especially on iPhone, the disabled forms are a little transparent and not fully black.

### Visible/Frontend Changes
- [x] Disabled forms are better readable on iPhone and Safari
- [ ] 

- [x] 
- [ ]


sleidig (approved):
I think I had observed grey text in form/profile also at some device (not iPhone/Apple) recently, but couldn't replicate it again. Hope this helps :-)

Issue #1692, opened 2023-01-30
Shortcuts widget (dashboard) does not have pagination


sleidig:
If there are too many shortcuts configured into the widget, later ones are clipped and not visible. Instead, the widget should use our generic dashboard template and show pagination then.

PR #1738, opened 2023-02-22
fix: added pagination to shortcut widget

TheSlimvReal:
see issue: #1692 

### Visible/Frontend Changes
- [x] 
- [ ] 

- [x] 
- [ ]


sleidig (commented):
Should we hide pagination if there are less then one page lenght elements? Here this adds unnecessary complexity: 
image

sleidig (approved):
:smiley:

PR #1744, opened 2023-02-27
fix: missing queries don't throw errors

TheSlimvReal:
It's possible to make a row without actually defining a query on it (e.g. to group further queries) this however currently throws error.

### Visible/Frontend Changes
- [x] 
- [ ] 

- [x] 
- [ ]

### DIFF
--- a/src/app/core/export/query.service.spec.ts
+++ b/src/app/core/export/query.service.spec.ts
@@ -621,7 +621,7 @@ describe("QueryService", () => {
   it("does not throw an error if no query is provided", () => {
     return expectAsync(
       service.queryData(undefined, new Date(), new Date())
-    ).toBeResolved();
+    ).toBeResolvedTo({});
   });
 
   async function createChild(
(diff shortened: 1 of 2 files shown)


src/app/core/export/query.service.spec.ts
```diff
@@ -618,6 +618,12 @@ describe("QueryService", () => {
     expect(result).toEqual([maleChild.gender]);
   });
 
+  it("does not throw an error if no query is provided", () => {
+    return expectAsync(
+      service.queryData(undefined, new Date(), new Date())
+    ).toBeResolved();
```
sleidig, line 624: what is the expected result here actually? Should this be null/undefined/empty or something or passing through the values of the previous level? In the UI, will this just show a label and nothing in the count/value spot?

sleidig (approved):

src/app/core/export/query.service.spec.ts
TheSlimvReal: The value is a empty object (I added this to the test) what is displayed is just the label, without a value in this line.

PR #1750, opened 2023-03-01
Add option safari (and firefox?)

TheSlimvReal:
### Visible/Frontend Changes
- [x] There were some problems about adding a new option in safari which this solves.
- [x] Multi enum select is correctly reset on cancel

--

### DIFF
--- a/src/app/core/configurable-enum/basic-autocomplete/basic-autocomplete.component.ts
+++ b/src/app/core/configurable-enum/basic-autocomplete/basic-autocomplete.component.ts
@@ -42,6 +42,7 @@ interface SelectableOption<O, V> {
 @Component({
   selector: "app-basic-autocomplete",
   templateUrl: "basic-autocomplete.component.html",
+  styleUrls: ["./basic-autocomplete.component.scss"],
   providers: [
     { provide: MatFormFieldControl, useExisting: BasicAutocompleteComponent },
   ],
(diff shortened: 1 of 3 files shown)


src/app/core/configurable-enum/basic-autocomplete/basic-autocomplete.component.ts
```diff
@@ -142,6 +143,7 @@ export class BasicAutocompleteComponent<O, V = O>
     let filteredOptions = this._options;
     this.showAddOption = false;
     clearTimeout(this.addOptionTimeout);
+    setTimeout(() => (this.showAddOption = false), 100);
```
sleidig, line 146: observing actual users add options, I am tending towards removing the delay for new options again. It was complex enough for a non-savvy user to figure out they can do this without things moving and changing. Maybe instead highlight the "Add new" option with different color a bit more than already?

Sorry, somewhat off-topic here actually. Just popped into my mind.

sleidig (approved):
Couldn't test for the target devices but code looks good :+1:

src/app/core/configurable-enum/basic-autocomplete/basic-autocomplete.component.ts
TheSlimvReal: Got point. I removed the timeout and instead visually highlighted the add option text. Please have a look

sleidig: Nice. I like it.

In terms of UI logic, maybe we should use the blue accent color though (similar to add buttons above lists, etc.)?

TheSlimvReal: I get the point but it does not look good with the blue color.

Issue #1569, opened 2022-12-05
Allow upload of children photos


TheSlimvReal:
After #1136 has been implemented it is now possible to upload files and view them. 

The next extension should be to also allow user to upload the children photos. The catch here is that we need to have a strong caching capability as there will be hundreds of requests for pictures whenever a list view or similar is loaded. 

Requirements should be:

- [ ] Photos can not be loaded without being logged in (currently this is possible)
- [ ] Photos are requested from the server at maximum once per session
- [ ] Photos don't have to be requested again from the server if in the last session the same photo has already been loaded (e.g. through the service worker)
- [ ] Photos should update (after some time) if a new version has been uploaded by another user

Generally, this should be a datatype that can be added to all entity types.

PR #1755, opened 2023-03-07
Photo upload

TheSlimvReal:
see issue: #1569 

### Visible/Frontend Changes
- [x] Component allows to display and upload photos

- [x] Added cache for photo files to only be requested once per session

### DIFF
--- a/src/app/child-dev-project/children/child-block/child-block.component.ts
+++ b/src/app/child-dev-project/children/child-block/child-block.component.ts
@@ -13,6 +13,8 @@ import { DynamicComponent } from "../../../core/view/dynamic-components/dynamic-
 import { NgIf } from "@angular/common";
 import { TemplateTooltipDirective } from "../../../core/common-components/template-tooltip/template-tooltip.directive";
 import { ChildBlockTooltipComponent } from "./child-block-tooltip/child-block-tooltip.component";
+import { CouchdbFileService } from "../../../features/file/couchdb-file.service";
+import { SafeUrl } from "@angular/platform-browser";
 
 @DynamicComponent("ChildBlock")
 @Component({
@@ -32,7 +34,12 @@ export class ChildBlockComponent implements OnInitDynamicComponent, OnChanges {
   /** prevent additional details to be displayed in a tooltip on mouse over */
   @Input() tooltipDisabled: boolean;
 
-  constructor(@Optional() private childrenService: ChildrenService) {}
+  imgPath: SafeUrl;
+
+  constructor(
+    private fileService: CouchdbFileService,
+    @Optional() private childrenService: ChildrenService
+  ) {}
 
   async ngOnChanges(changes: SimpleChanges) {
     if (changes.hasOwnProperty("entityId")) {
@@ -50,5 +57,10 @@ export class ChildBlockComponent implements OnInitDynamicComponent, OnChanges {
     }
     this.linkDisabled = config.linkDisabled;
     this.tooltipDisabled = config.tooltipDisabled;
+    if (this.entity.photo2) {
+      this.fileService
+        .loadFile(this.entity, "photo2")
+        .subscribe((res) => (this.imgPath = res));
+    }
   }
 }
--- a/src/app/child-dev-project/children/model/child.ts
+++ b/src/app/child-dev-project/children/model/child.ts
@@ -124,6 +124,13 @@ export class Child extends Entity {
   };
 
   @DatabaseField({
+    dataType: "file",
+    label: $localize`:Label for the filename of a photo of a child:Photo Filename`,
+    editComponent: "NewPhoto",
+  })
+  photo2: string;
+
+  @DatabaseField({
     label: $localize`:Label for the phone number of a child:Phone Number`,
   })
   phone: string;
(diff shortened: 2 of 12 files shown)


TheSlimvReal:
TODO:

- [ ] Check offline functionality and maybe not request pictures every time (when they are cached)
- [ ] Restrict file types
- [ ] Clicking on photo should increase it

src/app/child-dev-project/children/child-block/child-block.component.ts
```diff
@@ -32,14 +43,28 @@ export class ChildBlockComponent implements OnInitDynamicComponent, OnChanges {
   /** prevent additional details to be displayed in a tooltip on mouse over */
   @Input() tooltipDisabled: boolean;
 
-  constructor(@Optional() private childrenService: ChildrenService) {}
+  imgPath: SafeUrl;
+  icon = Child.icon;
```
sleidig, line 47: this is rather repetitive to the child-block-tooltip implementation. But on the other hand I am no sure it's worth factoring out into its own component ... (might pay off thinking ahead toward giving any entity type the option of displaying a photo instead of icon)

src/app/child-dev-project/children/model/child.ts
```diff
@@ -112,19 +108,16 @@ export class Child extends Entity {
...
-  };
+  photo: string;
 
   @DatabaseField({
-    label: $localize`:Label for the phone number of a child:Phone Number`,
+    dataType: "file",
+    label: $localize`:Label for the filename of a photo of a child:Photo Filename`,
+    editComponent: "NewPhoto",
   })
   phone: string;
```
sleidig, line 122: revert?

src/app/features/file/couchdb-file.service.ts
```diff
@@ -49,24 +57,54 @@ export class CouchdbFileService extends FileService {
...
-          observe: "events",
-        })
+  private runFileUpload(
+    file: File,
+    entity: Entity,
+    property: string,
+    compression: number
+  ) {
+    let blobObs: Observable<Blob>;
+    if (compression) {
```
sleidig, line 85: what happens here if the file is not an image?

For a clean "separation of concerns" we may want to keep resizing out of the general file service and just call that separately from the photo component?

sleidig (changes requested):
UX looks good to me :+1: 

- in offline mode photos are currently not cached/displayed

src/app/features/file/couchdb-file.service.ts
TheSlimvReal: Good point, that actually made things muuuuch simpler 😅

src/app/child-dev-project/children/child-block/child-block.component.ts
TheSlimvReal: Definitely true, but seems to be out of scope for this PR as it would mostly require refactoring the tooltip component

ngsw-config.json
```diff
@@ -42,9 +42,9 @@
...
     },
     {
-      "name": "childPhotos",
+      "name": "attachments",
       "urls": [
-        "**/child-photos/**"
+        "**/app-attachments/**/**"
       ],
       "cacheConfig": {
         "maxSize": 500,
```
TheSlimvReal, line 50: Do these values (size 500 max age 10 days) still make sense? This means users will have an old image for up to 10 days.

src/app/features/file/couchdb-file.service.ts
```diff
@@ -145,7 +186,27 @@ export class CouchdbFileService extends FileService {
       });
   }
 
-  private reportProgress(message: string, obs: Observable<HttpEvent<any>>) {
+  loadFile(entity: Entity, property: string): Observable<SafeUrl> {
+    const path = `${entity.getId(true)}/${property}`;
+    if (!this.cache[path]) {
```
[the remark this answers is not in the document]
TheSlimvReal: For new files (no previous value) this works as expected. For the other cases its a general caching question (also across reloads) when and how we get updates. At the moment the limit is the `ngsw` config not so much this cache here.

ngsw-config.json
sleidig: if I understand the docs correctly, after remaining 10 days offline the image will not be displayed at all anymore ("considered invalid and eviced": https://angular.io/guide/service-worker-config#maxage)? In that sense, a big maxAge makes sense for offline support.

Maybe we should change this to the mentioned "staleWhileRevalidate" strategy though? In combination with our in-memory cache that could be a better fit?

sleidig (approved):
Offline photos also work now and things look good :100: 

Only in offline mode a ton of errors and failed requests show in the console. I guess they don't end up in sentry (because they can't be sent without internet?) but for debugging this might pose some pain. Not sure whether we can do anything about it?

src/app/features/file/couchdb-file.service.ts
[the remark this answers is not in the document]
TheSlimvReal: final approach ->[staleWhileRevalidate](https://developers.google.com/web/fundamentals/instant-and-offline/offline-cookbook/#stale-while-revalidate) (with no special in-memory cache handling

PR #1764, opened 2023-03-10
fix: entity count dashboards routes to the correct entity list

TheSlimvReal:
Currently, the entity count dashboard always routes to the `Child` route. This has been changed to use the route of the entity which is also used in the splitup.

### Visible/Frontend Changes
- [x] 
- [ ] 

- [x] 
- [ ]


src/app/child-dev-project/children/dashboard-widgets/entity-count-dashboard/entity-count-dashboard.component.ts
```diff
@@ -53,9 +56,9 @@ export class EntityCountDashboardComponent
   onInitFromDynamicConfig(config: any) {
     this.groupBy = config?.groupBy ?? this.groupBy;
     if (config?.entity) {
-      this.entity = config?.entity ?? this.entity;
-      this.label = this.entities.get(this.entity).labelPlural;
-      this.entityIcon = this.entities.get(this.entity).icon;
+      this.entity = this.entities.get(config.entity);
+      this.label = this.entity.labelPlural;
+      this.entityIcon = this.entity.icon;
```
sleidig, line 61: don't we need this also for the fallback case without config.entity defined, when it remains the default of this.entity?

sleidig (approved):

src/app/child-dev-project/children/dashboard-widgets/entity-count-dashboard/entity-count-dashboard.component.ts
TheSlimvReal: This is covered in the default values

PR #1767, opened 2023-03-14
Sync restart on errors

TheSlimvReal:
Trying to solve the `Failed to execute 'transaction' on 'IDBDatabase': The database connection is closing.` which sometimes causes the sync to stop and some data not being synchronised. At the moment this can only be solved by calling the `/db/admin/clear_local/app` endpoint (which triggers a check of all revs at source and target).

### Visible/Frontend Changes
- [x] 
- [ ] 

- [x] Sync restarts after complete or error callbacks are called
- [x] `complete` events are logged


sleidig (approved):
code looks okay, functionality seems hard to test (even going offline doesn't seem to trigger any of these sync events for me) but the general functionality seems to continue working :-)

Issue #1672, opened 2023-01-19
EditDateComponent with current date preselected


TheSlimvReal:
Either the current `EditDateComponent` should be extended or a new one created which automatically sets the current date as value on new entities.

Issue #1671, opened 2023-01-19
EditUserComponent with current user preselected


TheSlimvReal:
A component should extends the `EditSingleEntity` and/or `EditEntityArray` component with the only addition that it only works on the `User` entity and that the current user is preselected.

PR #1796, opened 2023-03-23
Select current date and user when this is wanted in the property schema

Schottkyc137:
see issues: #1672 #1671

- [x] Current date is automatically selected when the default Value of a date of an entity in the property schema is set to `now` 
- [x] Current User is automatically selected using the same scheme as above


sleidig (commented):
in some sense, this overlaps with the EntitySchemaField "defaultValue" (which, I think, we never use):
https://github.com/Aam-Digital/ndb-core/blob/0dae11599f79f85418e71ea449a90dfbc7411c09/src/app/core/entity/schema/entity-schema-field.ts#L48-L52

Might make sense to bring those two aspects together?

src/app/core/database/pouch-database.ts
```diff
@@ -272,7 +272,7 @@ export class PouchDatabase extends Database {
           pouchDB
             .changes({
               live: true,
-              since: "now",
+              since: "$now",
```
sleidig, line 275:
```suggestion
              since: "now",
```

TheSlimvReal (changes requested):
There are some problems with the `defaultValue` assignment in the `EntitySchemaService`. We somehow need to distinguish whether the actual value should be assigned or not. 
Generally, it would be nice if we could overwrite the constructor in a way that the default values are automatically assigned instead of having to take care of this in the edit components every time.

sleidig (commented):
as I stumbled into a few issues when fixing the tests, a different possible approach just came to my mind:
What if we move defaultValue initialization to the Entity constructor, away from the EntityService/SchemaService loading. Otherwise we have to handle the special placeholders also in the transformToEntityFormat to avoid errors ...

TheSlimvReal:
I made some changes and think this is ready for a final review:

- default values are assigned centrally in the form service (and only persisted once this form is saved)
- changes in edit components were reverted
- default values are also assigned if the entity is **not** new (based on recent requirements)
- added the default value to further schemas where appropriate

@sleidig any further thoughts?

src/app/child-dev-project/notes/model/note.ts
```diff
@@ -107,6 +109,7 @@ export class Note extends Entity {
     label: $localize`:Label for the social worker(s) who created the note:SW`,
     dataType: "entity-array",
     additional: User.ENTITY_TYPE,
+    defaultValue: PLACEHOLDERS.CURRENT_USER,
```
sleidig, line 112: would this need special handling to generically work for any innerDataType within an array?

src/app/core/entity/schema/entity-schema-field.ts
```diff
@@ -56,7 +56,7 @@ export interface EntitySchemaField {
    * Whether the field should be initialized with a default value if undefined
    * (which is then run through dataType transformation);
```
sleidig, line 57:
```suggestion
   * Default values are applied to form fields before they are displayed to users
```

sleidig (changes requested):
User Stories / Test cases:
- [x] new entity -> form shows default
- [x] existing entity with field defined -> form keeps existing value
- [x] existing entity with field undefined -> form keeps undefined value 
- [x] form shows default -> user deletes value -> entity gets saved with empty/undefined property (if not marked required)

:question: Is this the expected behavior? what happens with existing entities that have an empty field is maybe debatable. But I think setting defaults on existing entities creates some rather unexpected, slightly unpredictable user experience.

Once we are clear: Let's add these as additional unit tests also.

EDIT: behavior adapted now and all passed.

PR #1800, opened 2023-03-30
migration script for assets pictures

TheSlimvReal:
See #1755 

This is a simple helper to migrate all pictures of children from the `assets/child-photos` folder compressed to the CouchDB attachments database.

## Usage

1. Go to CouchDB admin page of instance and create the database `app-attachments` and give member permissions to `user_app` role
2. Checkout this branch locally
3. Copy photos to local directory (run at code root)
  > scp -r <SSH_HOST>:/var/docker/ndb-\<INSTANCE\>/child-photos src/assets/
4. Edit `/db` target in `proxy.conf.json` to point to `<INSTANCE>`
5. Start app
6. Login
7. Ensure everything looks normal (child pictures are not shown and error messages indicate the files cannot be found)
8. Comment in code in `app.module.ts` constructor and reload page
9. Upload bars should be shown and pictures of children should slowly become visible
10. Wait for `migration complete` to be shown in console
11. Child pictures can now be removed from remote assets folder. **Careful, there might also be the logo file inside which shouldn't be removed**


sleidig (approved):
Looks okay, although you will know better than me whether anything apart from "uploadFile" is required for the photos.

entity config of photo property is automatically using the new EditPhoto via the entity class, I guess?

In general, I feel it's more comfortable to integrate this into the actually deployed admin section behind a button and avoid the hassle of running a local clone. Then mark it for removal in the next release. But here I think we only have two (?) instances with photos so this sounds good

PR #1809, opened 2023-04-05
fix: migration from single to multi value select works automatically

TheSlimvReal:
When changing a schema value from single to multi, this currently creates problems.
With this change the migration happens automatically in the background and is persisted once the entity is saved again.

### Visible/Frontend Changes
--

- [x] Auto migration of single to multi schema


src/app/core/entity/schema-datatypes/datatype-array.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+    const db = arrayEntitySchemaDatatype.transformToDatabaseFormat(
+      obj,
+      schema,
+      schemaService
+    );
+
+    expect(db).toEqual(value);
+  });
+
+  it("should automatically transform values that are not entities", () => {
```
sleidig, line 43:
```suggestion
  it("should automatically wrap value into array (and transform to inner type) if not an array yet", () => {
```

sleidig (approved):

PR #1810, opened 2023-04-06
Further subrecord improvements

sleidig:
- avoid loading all entities in entity list if custom data is passed
- avoid re-triggering filter changes if the filter value actually doesn't change
- avoid multiple calls to init functions


src/app/core/entity-components/entity-list/entity-list.component.ts
```diff
@@ -84,7 +84,7 @@ import { DisableEntityOperationDirective } from "../../permissions/permission-di
 export class EntityListComponent<T extends Entity>
   implements OnChanges, AfterViewInit
 {
-  @Input() allEntities: T[] = [];
+  @Input() allEntities: T[] = undefined;
```
TheSlimvReal, line 87:
```suggestion
  @Input() allEntities: T[];
```
let's not make this too confusing (when do we use `undefined` when do we just not initialise it)

TheSlimvReal (approved):

PR #1819, opened 2023-04-12
Report performance

TheSlimvReal:
Improving performance of reports

### Visible/Frontend Changes
- [x] Added paginator to export table

- [x] Removed/refactored asynchrony from query service and using services

### DIFF
--- a/src/app/core/config/config-fix.ts
+++ b/src/app/core/config/config-fix.ts
@@ -885,77 +885,7 @@ export const defaultJsonConfig = {
               ]
             },
           ],
-        }, {
-          "title": "Participant Details Report (pseudonymized)",
-          "mode": "exporting",
-          "aggregationDefinitions": [
-            {
-              "query": "EventNote:toArray[* date >= ? & date <= ?]",
-              "subQueries": [
-                {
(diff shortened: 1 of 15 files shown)


src/app/core/config/config-fix.ts
```diff
@@ -885,7 +885,77 @@ export const defaultJsonConfig = {
               ]
             },
           ],
+        }, {
+          "title": "Participant Details Report (pseudonymized)",
+          "mode": "exporting",
+          "aggregationDefinitions": [
```
TheSlimvReal, line 891: Just for testing reasons, will be removed before merge

sleidig (approved):
Looks smart and seems to work well :+1:

sleidig:
Definitely learned something new about JavaScript from this :smile: 

a bit further reading on this: "adding an async modifier to the method can almost double [~81%] the time it takes to execute"
https://madelinemiller.dev/blog/javascript-promise-overhead/

two possible solutions to mitigate:
- perform data fetching or other asynchronous operations closer to the application's root and pass the resulting data down
- only use the async keyword for the inner-most method and in the next higher level work with the promise

Issue #1807, opened 2023-04-05
timezone influences reporting stats of events


sleidig:
**Describe the bug**
As events are internally stored including a time of day, running a report in a different time zone results in different events being included/excluded. More critically, for users in countries with summer/winter time shift the same person is naturally in two different time zones and probably receives wrong reports!

sleidig:
I think it makes sense that Note/Event entities continue holding a datetime (instead of a `date-only` which we store as "YYYY-MM-dd" by cutting time details). While we don't have use cases for it yet, time makes sense here and can be an important information.

This means we need to handle timezones in a smart way for reporting. I think a user would expect the following:
- wherever & whenever the report is run, the same numbers are reported
  - e.g. also on a holiday at the other end of the world, by a colleague in a different country, during winter and summer (with daylight savings time)
- an event recorded on 01.01.2023 according to local time of that user should be reported for that day. Even when the event took place at a time when it was already the next day in a different country from which you are running the report now.

sleidig:
Possible solution:
- store Date in entities including the timezone (i.e. "2023-01-01T00:00:00.000-01:00" instead of "2023-01-01T01:00:00.000Z")
  - (!) what if someone edits the entity and saves it again?
  - possible implementation: https://stackoverflow.com/a/17415677/1473411
- for reports, only extract the local date part and ignore timezone offsets
  - this might be a bit challenging, as we usually work with the entity (where this is just a date object), not with the database format?

Alternative solution:
- switch Note + EventNote entities' date to "date-only", discard time information.

sleidig:
after discussion: For the time being we will migrate Note + EventNote to use date-only (--> don't implement hypothetical use cases for date+time). An extended approach that supports date with time and multiple timezones while still generating consistent reports could be implemented following the thoughts above but we'll not go for this right now.

Issue #978, opened 2021-09-08
Wrong timezones in exports


TheSlimvReal:
**Describe the bug**
When selecting a date in the date picker (e.g. in the `NoteDetailsComponent`) it is done based on the current timezeone. E.g. if I select the date `01.01.2021` in Germany it is understood as `2021-01-01 00:00:00.000 CEST/UTC+2` which in turn is saved according to the [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Times) format with the UTC timezone (thats what the 'Z' stands for) as `2020-12-31T22:00:00.000Z`. This happens because the datepicker selects midnight of the selected date but according to the locale of the computer. However, the date is saved according to the ISO format and midnight in Germany is equivalent to 10pm UTC/Greenwich time. When displaying the date this is not a problem as long as a the `DatePipe` or `new Date().toLocaleDateString()` is used. However when exporting the data, the date is exported in raw (ISO) format where `2021-01-01` could become `2020-31-12` if the timezone of the user is not UTC.

**To Reproduce**
1. Go to 'Notes'
2. Click on 'Add Note'
3. Change the date
4. Add a child to the note
5. Click on 'save'
6. Click on 'Download CSV'
7. Compare the date of the newly created note in the CSV file with the app.
8. Dates should be different if local timezone is not UTC

**Expected behavior**
The date in the export and the table in the app should be the same

**Additional context**
Another problem that can come up here is if the users of the same app live in different timezones. E.g. I create a note for the `2021-01-01` in Germany. If a person from the USA now looks at the app, it will see the note I created but with the date `2020-12-31`. What should be the expected behavior here?

PR #1826, opened 2023-04-13
migrating date to date-only (to ensure consistent report results)

sleidig:
see issue: #1807 (maybe related #978 )

- [x] date values are migrated to the (timezone offset adjusted) date-only values during loading
- [ ] make date-only the default dataType for Date objects?


src/app/core/entity/schema-datatypes/datatype-date-only.spec.ts
```diff
@@ -35,4 +35,28 @@ describe("Schema data type:Date", () => {
...
+      { input: "2022-01-01", expected: "2022-01-01" },
+      // clean midnight dates are transformed to the inferred day
+      { input: "2023-01-20T00:00:00.000Z", expected: "2023-01-20" },
+      { input: "2022-12-31T22:00:00.000Z", expected: "2023-01-01" },
+      { input: "2022-12-31T19:30:00.000Z", expected: "2023-01-01" },
+      { input: "2023-01-01T05:00:00.000Z", expected: "2023-01-01" },
+      // exact dates do not allow to safely guess offset
+      // TODO: we could store these as "createdAt" of the entity (once that is implemented ...)
+      { input: "2023-01-06T10:03:35.726Z", expected: "2023-01-06" },
+      { input: "2023-01-06T23:59:39.726Z", expected: "2023-01-06" }, // TODO: what to do here? these are actually not deterministic ...
```
TheSlimvReal, line 50: I would say the moment we have an accurate time (minutes, seconds, milliseconds not 0) we should just use the date.
If minutes, seconds and milliseconds are 0 we know it has to be a flat date and we can use the closest one.

TheSlimvReal, line 50: What do you mean by not deterministic? I a sense that we cannot guess the timezone?

sleidig: yes, this currently results in a different date depending on the user's time zone.

Will cut off time details now to always have the same result.

TheSlimvReal (approved):

PR #1831, opened 2023-04-17
fix: demo data initialisation

TheSlimvReal:
Publishing the demo config last so the app only initialises after all other entities have been saved.
Currently, some dashboard components stay empty on first load and a refresh on school details sometimes results in a not found error.

### Visible/Frontend Changes
- [x] demo data is better initialised

--


src/app/core/demo-data/demo-data.module.ts
```diff
@@ -67,6 +66,7 @@ const demoDataGeneratorProviders = [
     maxCountAttributes: 5,
   }),
   ...DemoTodoGeneratorService.provider(),
+  ...DemoConfigGeneratorService.provider(),
```
sleidig, line 69:
```suggestion
  // keep Demo service last to ensure all entities are already initialized
  ...DemoConfigGeneratorService.provider(),
```

sleidig (approved):
good find!

PR #1835, opened 2023-04-19
fix: clicking version only opens popup once

TheSlimvReal:
see issue: #XX

### Visible/Frontend Changes
- [x] 
- [ ] 

- [x] 
- [ ]


sleidig (approved):
I would've prefered to keep the click on this component rather than in the UI component. But achieving that is not trivial with the style + sizes, so let's go with the simple solution :blush:

PR #1841, opened 2023-04-20
deps: updated dependencies

TheSlimvReal:
Updated all dependencies (but Storybook 🤮). Just can't get Storybook version 7 running. Somehow the dependency resolution is not working there.

### Visible/Frontend Changes
- [x] 
- [ ] 

- [x] 
- [ ]


sleidig (approved):
looks fine when clicking around the app and as these are minor/patch version changes things shouldn't break anyway :+1: 
Thanks for getting through this!

PR #1855, opened 2023-04-25
fix: photos in child block component are correctly updated

TheSlimvReal:
At the moment the child block component only loads the linked child and photo once. In the roll call it is required to update it multiple times.

### Visible/Frontend Changes
- [x] 
- [ ] 

- [x] 
- [ ]


sleidig (changes requested):
This properly loads new existing pics but the previous photo still remains visible for the next child if that child does not have a photo.

Issue #1424, opened 2022-09-06
Unsaved details forms not warning before navigating away


sleidig:
**Describe the bug**
Users can lose data they entered in a form by simply clicking on a different menu item. Instead we should warn them and ask whether to abort or discard changes, like we do when closing the note details popup form.

**To Reproduce**
1. Go to a Child details view and click Edit
2. Change the name
3. Without clicking save, navigate to "Schools" using the left menu
4. --> your changes are silently discarded

**Expected behavior**
Popup like for Note details for all entity details views in edit mode.

sleidig:
duplicate of #914

PR #1859, opened 2023-04-26
Save changes popup

TheSlimvReal:
see issue: #1424 

### Visible/Frontend Changes
- [x] User confirmation for closing edited popups
- [x] User confirmation when navigation from edited from details

--


sleidig (changes requested):
From a usability perspective:
- [x] confirmation shows while in the background navigation is already done. This seems uncommon for such a workflow, feels slightly off (dialog refers to previous view) and it may be useful for a user to still see the changes to save/discard.
- ~Navigating to Children list and confirming to save changes, the list does not get updated with the newly saved change (e.g. name is still the old one until reloading that view)~ (not applicable anymore, as navigation only when discarding)
- [ ] Should we ask for confirmation also when clicking "Cancel" instead of "Save" in the Form? (We do that on "Cancel" in the popup forms already)
- [x] Popup form "Save" button still asks whether I want to save changes. That shouldn't be necessary.
- [ ] Cancel (in addition to saving yes/no) would be a useful further option to go back to editing. Half the time I get such a message in other software I end up doing some further editing. (Could be moved to a separate PR/issue if we want this out quickly)
- [ ] some websites prevent you to navigate away even to a completely different domain in the browser if you have unsaved changes. This could be useful here as well. Currently I can navigate or reload to lose changes and the confirmation only happens for internal app navigation

TheSlimvReal:
@sleidig I have now implemented the logic in a way that it shows the form popup before navigation is done.

With this approach there are still some open questions:

- What do we do with different panels (we could deactivate the other panels while one is in edit mode)?
- When do we want to show the popup? At the moment it only occurs on navigation and when clicking the backdrop of a popup. Further options would also include "Cancel" and the "X" button in a popup.

src/app/core/form-dialog/dialog-buttons/dialog-buttons.component.ts
```diff
@@ -45,8 +48,20 @@ export class DialogButtonsComponent implements OnInit {
...
-  ) {}
+    private ability: EntityAbility,
+    private confirmation: ConfirmationDialogService
+  ) {
+    this.dialog.backdropClick().subscribe(() => {
+      if (this.form.dirty) {
+        this.confirmation
+          .getConfirmation(
+            $localize`:Save changes header:Save Changes?`,
+            $localize`:Save changes message:Do you want to save the changes you made?`
```
TheSlimvReal, line 59: This could also follow the same style as the popup for navigation where its about discarding instead of saving.

sleidig: good point, would be great to have the same pattern everywhere! (Maybe even define a default text in a method in confirmation service already to have it 100% consistent?)

src/app/core/entity-components/entity-details/form/unsaved-changes.service.ts
```diff
@@ -0,0 +1,30 @@
...
+  constructor(private confirmation: ConfirmationDialogService) {
+    // prevent navigation if changes are pending
+    window.onbeforeunload = () => (this.pending ? "" : undefined);
+  }
+
+  async checkUnsavedChanges() {
+    if (this.pending) {
+      const confirmed = await this.confirmation.getConfirmation(
+        "Discard changes?",
+        "All unsaved changes will be lost"
```
sleidig, line 19:
```suggestion
        "You have unsaved changes. Do you really want to leave this page? All unsaved changes will be lost."
```

sleidig (commented):
good improvements! :-)

> What do we do with different panels (we could deactivate the other panels while one is in edit mode)?

More comfortable would be if we can consider switching tabs the same as switching pages and show the dialog instead of preventing the switch per se. Technically difficult?

> When do we want to show the popup? At the moment it only occurs on navigation and when clicking the backdrop of a popup. Further options would also include "Cancel" and the "X" button in a popup.

Slightly unsure here. What do you think? Tending a bit towards showing it on cancel, too.

---
Recently read a design book that highlighted how the safeguard ("Do you want to save changes?") can become the most easy, default way to save a document for some users. In that regard, showing the dialog (and maybe considering to extend it with even more explicit buttons) could be worth considering.
Screenshot from 2023-05-09 09-48-52

TheSlimvReal:
The features that have not been implemented yet are very difficult to achieve:

### Restrict panel navigation with unsaved changes
It is not possible to keep the panel activated but still disable the navigation to another panel. We can either disable the panel completely or only show a popup after the navigation has happened. In the later case it's still difficult to implement the "leave/cancel" functionality as this would need to be communicated back to the form component (at the moment it would just stay in edit mode).

### Having a save button in the popup
This has the same problem as mentioned before that the "save" trigger needs to be forwarded to the underlying component which is quite difficult to achieve with our setup.

src/app/core/confirmation-dialog/confirmation-dialog.service.ts
```diff
@@ -55,4 +55,11 @@ export class ConfirmationDialogService {
         .afterClosed()
     );
   }
+
+  getSaveConfirmation() {
+    return this.getConfirmation(
+      $localize`:Save changes header:Save Changes?`,
```
sleidig, line 61: :warning: 
This is very misleading! Actually "Yes" is discarding changes, not saving them.

```suggestion
      $localize`:Save changes header:Discard Changes?`,
```

sleidig (changes requested):
Tested:

details view (in edit mode with changes):
- [x] navigate -> confirmation | yes -> navigating and discarding changes
- [x] navigate -> confirmation | no -> remain on view in edit mode
- [x] [edit mode without changes] -> navigate -> no confirmation dialog
- [x] close browser / leave app -> browser confirmation | cancel -> remain on view in edit mode
- [x] ~switch tab -> :question: no confirmation -> navigate -> :warn: confirmation (about unfocused tab)~ 
-> I agree with your suggestion and thinking about would now vote for disabling tab switches while in edit mode. This would also be consistent with our UX when creating a new entity (and we can maybe use the same code).

popup form (with changes):
- [x] backdrop click -> confirmation | no -> remain on popup with changes
- [x] backdrop click -> confirmation | yes -> closing popup, discarding changes (:rotating_light: changes are currently saved anyway)
- [x] cancel click -> no confirmation 
:warning: I would've expected that cancel and backdrop have the same behavior but it also makes sense to not additionally ask on cancel (and that is consistent with the details view behavior) :heavy_check_mark: 
- [x] [without changes] -> cancel/backdrop -> no confirmation dialog
- [x] close browser / leave app -> browser confirmation | cancel -> remain on view in edit mode
:warning: not sure whether technically not possible but for UX + consistency it would be great to prevent browser close in this case as well

src/app/core/session/auth.guard.ts
```diff
@@ -1,28 +1,23 @@
...
-      return true;
-    } else {
-      this.router.navigate(["/login"], {
-        queryParams: {
-          redirect_uri: state.url,
-        },
-      });
-      return false;
-    }
+export const AuthGuard: CanActivateFn = (
```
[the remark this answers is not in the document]
TheSlimvReal: I just saw that it was marked as deprecated when I came across it.

src/app/core/entity-components/entity-form/entity-form.service.ts
```diff
@@ -133,7 +155,10 @@ export class EntityFormService {
 
     return this.entityMapper
       .save(updatedEntity)
-      .then(() => Object.assign(entity, updatedEntity))
+      .then(() => {
+        this.unsavedChanges.pending = false;
```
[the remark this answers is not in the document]
TheSlimvReal: Good find -> #1911

Issue #1087, opened 2022-01-24
Remove function calls for assigning variables in templates


TheSlimvReal:
prevent usages like

```
<some-comp [title]="getTitle()></some-comp>
```
and instead use plain variables that are regularly updated

```
<some-comp [title]="title"></some-comp>
```

When using a function, this function is executed thousands of times and therefore has impact on the performance.

sleidig:
will be convered as part of the overall refactoring to use Angular signals

PR #1860, opened 2023-04-27
More robust autocomplete

TheSlimvReal:
These changes should make the basic autocomplete component more robust on different/slower browsers.

### Visible/Frontend Changes
- [x] 
- [ ] 

- [x] 
- [ ]


src/app/core/configurable-enum/basic-autocomplete/basic-autocomplete.component.ts
```diff
@@ -82,7 +82,17 @@ export class BasicAutocompleteComponent<O, V = O>
     startWith([] as SelectableOption<O, V>[])
   );
   showAddOption = false;
-  private delayedBlur: any;
+
+  get displayText() {
```
sleidig, line 86: isn't this going against #1087 and we should rather set a variable instead of a getter?

src/app/core/configurable-enum/basic-autocomplete/basic-autocomplete.component.ts
```diff
@@ -230,48 +239,46 @@ export class BasicAutocompleteComponent<O, V = O>
   }
 
   onFocusIn() {
```
sleidig, line 241: can we rename this function to describe what it does rather than when it is triggered (especially as this is not just an event handler but also explicitly called programmatically). Maybe activeateAutocompleteMode or resetAutocompleteMode or something?

src/app/core/configurable-enum/basic-autocomplete/basic-autocomplete.component.ts
```diff
@@ -230,48 +239,46 @@ export class BasicAutocompleteComponent<O, V = O>
...
-      if (this.multi) {
-        this.autocompleteForm.setValue("");
-      } else {
-        this.showAutocomplete();
-      }
-      this.focus();
+    if (this.multi) {
+      this.autocompleteForm.setValue("");
+    } else {
+      this.showAutocomplete();
```
sleidig, line 245: can we inline / simplify here? Why only showAutocomplete for single and for multi resetting the value is enough?

src/app/core/configurable-enum/basic-autocomplete/basic-autocomplete.component.ts
```diff
@@ -230,48 +239,46 @@ export class BasicAutocompleteComponent<O, V = O>
...
     }
+    setTimeout(() => this.inputElement.focus());
```
sleidig, line 247: let's comment extensively why we need a timeout here and what are the tradeoffs (as we might face some related bugs again in the future :wink: )

src/app/core/configurable-enum/basic-autocomplete/basic-autocomplete.component.ts
```diff
@@ -230,48 +239,46 @@ export class BasicAutocompleteComponent<O, V = O>
...
+    this.focus();
   }
 
   onFocusOut(event: FocusEvent) {
     if (
       !this.elementRef.nativeElement.contains(event.relatedTarget as Element)
     ) {
-      // use short timeout in order for creating an option to work
-      this.delayedBlur = setTimeout(() => this.notifyFocusOut(), 200);
+      if (!this.multi) {
```
sleidig, line 255: actually, why don't we select the exact match (as one new value) in multi mode as well?

sleidig (changes requested):
looks good and seems to work well (although I didn't experience the bug previously either).

Tested:

**single select:**
- [x] existing value displayed initially
- [x] able to use autocomplete, see all options dropdown, type to narrow options
- [x] click outside to reset (discarding anything typed; showing selected options as string again)
- [x] select a new/additional option (value shown in string, dropdown closed, able to open dropdown again || multi: )
- [ ] add a new option :warning: 

clicking to add new option currently does not show the yes/no confirmation and instead shows the dropdown at the top left corner of the screen. No option is added.

**multi select:**
- [x] existing value displayed initially
- [x] able to use autocomplete, see all options dropdown, type to narrow options
- [x] click outside to reset (discarding anything typed; showing selected options as string again)
- [x] select a new/additional option (dropdown remains open, able to select further values, value shown in string on focus out)
- [x] add a new option (confirmation dialog, nothing on "No", created and selected on "Yes")

If we can, we should hide that checkmark on the right of the option in multi mode. This is somewhat inconsistent with the checkboxes.
image

src/app/core/configurable-enum/basic-autocomplete/basic-autocomplete.component.ts
sleidig: reworked the implementation a bit and removed the exact match select logic (except for deleting the selected option on empty field).

sleidig (approved):
succesfully tested after the further changes

PR #1869, opened 2023-04-28
fix: editing entity array correctly marks a form as dirty

TheSlimvReal:
Save button is disabled in note details if only an entity-array field is changed.

### Visible/Frontend Changes
- [x] 
- [ ] 

- [x] 
- [ ]


sleidig (approved):
We may want to review the EditEntityArrayComponent and EntitySelectComponent and see whether we want implement them similar to EditSingleEntity using the BasicAutocompleteComponent. Not sure this would work but my first instinct was to wonder why entity array is different from entity single

PR #1884, opened 2023-05-25
fix: allowing spaces in enum IDs

TheSlimvReal:
see issue: #XX

### Visible/Frontend Changes
- [x] 
- [ ] 

- [x] 
- [ ]

### DIFF
--- a/src/app/core/export/query.service.ts
+++ b/src/app/core/export/query.service.ts
@@ -307,7 +307,7 @@ export class QueryService {
     key: string,
     value: string
   ): any[] {
-    const values = value.replace(new RegExp(" ", "g"), "").split("|");
+    const values = value.trim().split(/\s*\|\s*/);
     return objs.filter((obj) => {
       if (obj?.hasOwnProperty(attr)) {
         return values.includes(obj[attr][key]?.toString());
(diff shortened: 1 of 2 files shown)


src/app/core/export/query.service.ts
```diff
@@ -307,7 +307,7 @@ export class QueryService {
     key: string,
     value: string
   ): any[] {
-    const values = value.replace(new RegExp(" ", "g"), "").split("|");
+    const values = value.trim().split(/\s*\|\s*/);
```
sleidig, line 310: why is the simple `.split("|")` not enough here anymore? (maybe add comment to document this even?)

sleidig (approved):

src/app/core/export/query.service.ts
sleidig: your comment explains the regex (I got that) but not why we need a regex and can't just split at any "|"  ;-)

TheSlimvReal: It's just safeguard so both `male|female` as well as `male | female` works. Spaces around the `|` are ignored allowing people to define the query either way.

PR #1885, opened 2023-05-25
fix: improved logging for failed sync

TheSlimvReal:
Sync completions are not send to Sentry directly and instead just logged as a information in case other errors occur.

Errors are still send to sentry but hold information about the last successful authentication which should help to understand, why these errors occur. The last authentication timestamp has been improved to be updated whenever a new access token has been received (in the Keycloak case).

### DIFF
--- a/src/app/core/session/auth/couchdb/couchdb-auth.service.ts
+++ b/src/app/core/session/auth/couchdb/couchdb-auth.service.ts
@@ -9,6 +9,7 @@ import {
 import { firstValueFrom } from "rxjs";
 import { AppSettings } from "../../../app-config/app-settings";
 import { AuthUser } from "../../session-service/auth-user";
+import { tap } from "rxjs/operators";
 
 @Injectable()
 export class CouchdbAuthService extends AuthService {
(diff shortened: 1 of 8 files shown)


src/app/core/session/session-service/synced-session.service.ts
```diff
@@ -226,16 +226,19 @@ export class SyncedSessionService extends SessionService {
         // replication was resumed: either because new things to sync or because connection is available again. info contains the direction
         this.syncState.next(SyncState.STARTED);
       })
-      .on("error", this.handleStoppedSync("error"))
-      .on("complete", this.handleStoppedSync("complete"));
+      .on("error", this.handleStoppedSync())
+      .on("complete", (info) =>
+        this.loggingService.info(`Life sync completed: ${JSON.stringify(info)}`)
```
sleidig, line 231:
```suggestion
        this.loggingService.info(`Live sync completed: ${JSON.stringify(info)}`)
```

src/app/core/session/session-service/synced-session.service.ts
```diff
@@ -226,16 +226,19 @@ export class SyncedSessionService extends SessionService {
...
+      );
   }
 
-  private handleStoppedSync(reason: string) {
+  private handleStoppedSync() {
```
sleidig, line 235: rename -> `handleFailedSync`?

src/app/core/session/auth/couchdb/couchdb-auth.service.ts
```diff
@@ -25,11 +26,13 @@ export class CouchdbAuthService extends AuthService {
...
-        { name: username, password: password },
-        { withCredentials: true }
-      )
+      this.http
+        .post<AuthUser>(
+          `${AppSettings.DB_PROXY_PREFIX}/_session`,
+          { name: username, password: password },
+          { withCredentials: true }
+        )
+        .pipe(tap(() => this.logSuccessfulAuth()))
```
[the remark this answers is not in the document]
TheSlimvReal: Yes, before the authorisation was only logged once on the initial login. What it did not cover is further authentications with the refresh token, so it was not possible to actually see if the user is still authenticated or if the token already expired due to inactivity.

PR #1890, opened 2023-05-30
Improved app initialisation

TheSlimvReal:
At the moment we have the problem that properties defined in the config are not yet available in the demo data generators.
This is because the config is saved last. If the config is saved first we have another problem where the app is loaded but on the initial screen some views might be empty because the app builds once the config is ready but the demo data is only created and saved afterwards.

The new solution is to save the config first but postpone the creation of the app. 
This is done by moving the initialisation logic from the app component to the `APP_INITIALIZER` provider.
This way all the initialisation is done and only afterwards the `AppComponent` gets build.

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -116,6 +116,12 @@
               "src/assets",
               "src/favicon.ico",
               "src/manifest.webmanifest"
+            ],
+            "fileReplacements": [
+              {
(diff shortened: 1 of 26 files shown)


sleidig (approved):
Code looks good.

Tested in both demo+mock and local (persistent indexeddb) modes.

Issue #1135, opened 2022-03-11
Feature: Import UI


sleidig:
_As a user I want to import data from other sources through a CSV file without requiring any technical support, even if I am not familiar with concepts like data formats, etc._

https://docs.google.com/document/d/1K8eNgKXDkvVN8OuRG87YB07izPdtFExLTG1Mc9vBVJY/edit

---

The base feature with a UI is developed in PR #1059 . This issue is a summary thread to discuss further improvements and user stories, which we can split out into their own issues to be implemented.

see Milestone [Import UI](https://github.com/Aam-Digital/ndb-core/milestone/23)
- [x] #974
- [x] #975
- [x] allow undo of import
- [x] #1771
- [x] #1772
- [x] map string values from import to configurable-enum values in app
- [x] map string values from import to boolean values in app
- [x] automatically infer date format
- [x] test date format to be able to parse dates in that sheet/column
- [ ] #1768
- [ ] split column by space (e.g. from name to first + last name)

Smaller bugs:

- [x] #1799
- [ ] import config and auto assignment sometimes uses same column multiple times
- [ ] trim spaces on values for a better mapping experience (probably also ignore case)

TheSlimvReal:
Inspiration by an accounting software to import transactions that also uses a column mapping approach.

![image](https://user-images.githubusercontent.com/33730997/158783172-4355aa17-7d42-439b-a097-c470f1279cc0.png)

TheSlimvReal:
Improve usability of undoing/viewing previous imports
- [ ] Replace usage of `transactionID` with a (new) optional property `importedAt` in the `Entity` class that is set with the timestamp `YYYY-MM-DD HH:MM:SS` of the import
- [ ] Create a new entity (some name like `ImportMetaData`) that stores for each import the timestamp of the import, the amount and type of entities imported and the configuration of the import (entity type, date-formatting, column mapping, ...)
- [ ] Show `ImportMetaData` entities in import widget and allow to delete a recent import or use the configuration of a recent import

PR #1916, opened 2023-06-30
basic import ui layout

sleidig:
see issue: #1135 
all changes that implement a new, easy-to-use import UI module.

TODO:
- [x] routing / import of module (currently a workaround ... to get demo running; should we keep a hard-coded path, though?)
- [x] execute additionalActions (e.g. add to existing activity, etc.)
- more details/warnings for summary popup (unmapped columns, etc.) #1943

BUGS:
- [x] assigned values (enum mapping) not displayed in popup when reused from previous import
- [x] additional actions not applied from previous import
- [x] (?) ~dates still get parsed even if invalid mapping format defined~ (could not reproduce this problem currently)
- [x] console errors + continue button from mapping
- [x] make "continue" button more prominent and move to top right float (like edit entity details)


src/app/features/import/import-review-data/import-review-data.component.ts
```diff
@@ -0,0 +1,74 @@
...
+    } else {
+      return this.schemaService
+        .getDatatypeOrDefault(mapping.property.schema.dataType)
+        .transformToObjectFormat(
+          val,
+          mapping.property.schema,
+          this.schemaService,
+          entity
+        );
+    }
```
sleidig, line 72: @TheSlimvReal here we also need some kind of access to the mapping function for a specific property (or propertyType) in the reviewComponent

src/app/features/import/import/import.component.ts
```diff
@@ -0,0 +1,108 @@
...
+  onDataLoaded(data: ParsedData<any>) {
+    this.rawData = data.data;
+
+    if (this.columnMapping) {
+      this.alertService.addInfo(
+        $localize`:alert info after file load:Column Mappings have been reset`
+      );
+    }
+    this.onColumnMappingUpdate(
+      data.fields.map((field) => ({ column: field, propertyName: undefined }))
```
TheSlimvReal, line 102: Why settings `propertyName` to `undefined` instead of omitting it?

src/app/features/import/import/import.component.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    // apply columns individually to ensure only valid mappings are merging on top of existing mapping
+    const applicableMapping: ColumnMapping[] = [];
+    for (const existingCol of this.columnMapping) {
+      let applicableCol = importMetadata.config.columnMapping.find(
+        (c) => existingCol.column === c.column
+      );
+      if (!applicableCol) {
+        // reset any existing mapping - the loading should also apply which columns are ignored
+        applicableCol = { column: existingCol.column, propertyName: undefined };
```
TheSlimvReal, line 124: Same as above: Why not omit the `propertyName` completely instead of setting it to undefined?

src/app/features/import/import/import.component.ts
```diff
@@ -0,0 +1,108 @@
...
+      }
+      applicableMapping.push(Object.assign({}, existingCol, applicableCol));
```
TheSlimvReal, line 126: Whats the point of assigning this to the initial column? This only seems relevant when `additional` is set but this seems rather dangerous as the `propertyName` has been overwritten here and we don't know whether the additional even makes sense for this new property.

TheSlimvReal (changes requested):

src/app/features/import/import-column-mapping/import-column-mapping.component.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+      { column: "y", propertyName: "projectNumber" }, // property does not exist on new entityType
+    ];
+
+    component.entityType = "School";
+    component.ngOnChanges({
+      entityType: new SimpleChange("Child", "School", false),
+    });
+
+    const expectedNewMapping: ColumnMapping[] = [
+      { column: "x", propertyName: "name" },
```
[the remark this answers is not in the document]
sleidig: the problem is that the mapping assumes a very specific input rawData (the imported column names). Applying a previous mapping will only work if the new imported spreadsheet has 100% same columns as the historic one. Otherwise new columns will be inaccessible to re-map. And I am not sure what would happen to mappings that do not have an actual column anymore.

We could of course also just disallow to load a previous mapping for any case where imported columns do not match exactly.

src/app/features/import/import-column-mapping/import-column-mapping.component.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+  });
+
+  it("should emit on updateMapping from UI", () => {
+    component.updateMapping();
+
+    expect(component.columnMappingChange.emit).toHaveBeenCalled();
+  });
+
+  it("should emit on changes from popup form with additional details", () => {
+    // TODO: test additional mapping details from popup form to be applied and emitted
```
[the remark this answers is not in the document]
sleidig: I think the latest implementation requires a `columnMappingChange` event beyond editing in place for the updates in related components (e.g. re-parsing in preview) to work correctly.

src/app/features/import/import-column-mapping/import-column-mapping.component.ts
```diff
@@ -0,0 +1,125 @@
...
+        disableClose: true,
+      }
+    );
+  }
+
+  /**
+   * Emit an updated columnMapping array and emit change event, to ensure smooth change detection.
+   */
+  updateMapping() {
+    this.columnMappingChange.emit([...this.columnMapping]);
```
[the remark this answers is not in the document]
sleidig: for change detection to seamlessly work (update preview component, etc.), in place editing of elements is not sufficient. Therefore I wanted to avoid any side effect on the input value and instead emit a new array on each change. 

Admittedly, I only did that half-way because the data-binding in the template already edits the array ... If there is a better way to get change detection across components, we would consider that as well. Or clean up and on init create a copy, rather then here on emit?

src/app/features/import/import-history/import-history.component.ts
```diff
@@ -0,0 +1,67 @@
...
+import { applyUpdate } from "../../../core/entity/model/entity-update";
+
+@UntilDestroy()
+@Component({
+  selector: "app-import-history",
+  templateUrl: "./import-history.component.html",
+  styleUrls: ["./import-history.component.scss"],
+})
+export class ImportHistoryComponent implements OnInit {
+  @Input() highlightedPreviousImport: ImportMetadata;
```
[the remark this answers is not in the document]
sleidig: the idea was to allow the parent view to highlight a specific item, e.g. the newly completed one - but maybe also a specific other import for different reasons. Currently not used though.

src/app/features/import/import/import.component.ts
sleidig: Here this is required because _the loading should also apply which columns are ignored_ (e.g. the user in previous import had consciously decided not to map the email column, so apply that mapping should also reset in case it had been mapped.

src/app/features/import/import/import.component.ts
sleidig: only added for consistent data structure as it is needed below (see other comment)

src/app/features/import/import-column-mapping/import-column-mapping.component.spec.ts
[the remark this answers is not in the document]
sleidig: yes, you are probably right. I am not really sure anymore why I thought this was needed

TheSlimvReal (approved):

PR #1924, opened 2023-07-07
Release 3.22.0

TheSlimvReal:
### Configurable date range filter

Replace 
```
{
          "id": "date",
          "type": "prebuilt"
}
```
with 
```
{
          "id": "date"
}
```

### Dynamic default values

If one of the following properties is overwritten in the config, then a the appropriate default value needs to be set (`$now` or `$current_user`):

- `Aser.date`
- `EducationalMaterial.materialAmount` (default value `1`)
- `Note.date` and `Note.authors`
- `HistoricalEntityData.date` (and any entity extending it)
- `Todo.assignedTo`


sleidig (approved):
testeed on dev.aam-digital.com

PR #1944, opened 2023-07-24
Dependency updates

TheSlimvReal:
A hell lot of breaking things...

- Mostly the storybook update to version 7 required **a lot** of adjustments how stories are written as well as how dependencies are provided. It seems like the support for standalone components is now even better (requires less boilerplate) but the existing approach did not work anymore. While adjusting these I also fixed a few stories so there will be some conflicts on percy once we get some new quota.
- I cleaned up the `tsconfig` files to fix the wrong types we sometimes had in the test files (it used Chai/Mocha instead of Jasmine types).
- Faker had a major update with a lot of deprecated/changed a few APIs
- ESlint had an update which required lint changes and also included special rules for stories


src/polyfills.ts
```diff
@@ -29,11 +29,6 @@
  *
  * Learn more in https://angular.io/docs/ts/latest/guide/browser-support.html
  */
-/***************************************************************************************************
- * BROWSER POLYFILLS
- */
-/** Evergreen browsers require these. **/
-import "core-js/es/reflect";
```
TheSlimvReal, line 36: I did not find anything that indicates that this is actually still needed. New angular projects don't come with it.

sleidig (approved):
thank you! :heart:

PR #1952, opened 2023-08-01
Toplevel query

TheSlimvReal:
At the moment export queries are not allowed to have a result in the first hierarchy level but need subqueries.
This feature has been added here.


sleidig (approved):
Let's add a unit test for the new functionality also?

Took me a minute to understand what happens when between these cases. But I don't see any easy way to make this code much clearer and it looks correct to me :+1:

sleidig (approved):
Thanks :pray: . This unit test is great documentation also

PR #1960, opened 2023-08-07
fix(login): fix error upon login for keycloak user without exact_username

### LINKED ISSUE #1959: first login shows error (trim on undefined)
**Describe the bug**
At least for the latest instances set up, the first time trying to log in results in an error. Retrying the login button then works immediately.

> Unexpected login error: TypeError: Cannot read properties of undefined (reading 'trim')

sleidig:
throwing "Cannot read properties of undefined (reading 'trim')"

fixes #1959


sleidig:
@TheSlimvReal where do we actually map `exact_username` in our code?

TheSlimvReal:
I have adjusted it to log a better error message instead of silently failing.

sleidig (commented):
Thanks! :-)

PR #1962, opened 2023-08-08
fix(forms): clear dropdown value in field after deleting its option

### LINKED ISSUE #1804: deleting dropdown option re-shows invalid value in record
**Describe the bug**
The UI is inconsistent at the moment when deleting an option in the dropdown (configurable enum):

1. Click the tool icon of a dropdown field while editing an entity and delete the option that is currently selected value for this entity
2. The form shows an empty field for the dropdown now before and after saving the form.
3. Go back to the list and re-open the record: The old deleted value is not shown again (as expected with the tag "[invalid option]")

The field should either immediately after deleting the option show "[invalid option] OLD_VALUE" or (preferably) reset the field for that entity to be saved as undefined.

sleidig:
closes #1804


TheSlimvReal (approved):

TheSlimvReal:
I have adjusted it to log a better error message instead of silently failing.

PR #1965, opened 2023-08-09
doc(core): document extendability approach

sleidig:


TheSlimvReal (approved):

doc/compodoc_sources/concepts/extendability.md
```diff
@@ -0,0 +1,31 @@
...
+Aam Digital is designed to be an extendable platform.
+We try to define core interfaces that can be implemented in additional feature modules
+to implement further functionality in a modular way.
+
+The following aspects are specifically designed to be extended:
+- **DataTypes**
+  - transformation functions how data is stored in / read from database
+  - `editComponent` how data is displayed and edited in forms
+  - `viewComponent` how data is displayed in tables
+  - `importValueMapping` to support smart import into the data type
```
TheSlimvReal, line 11: Does this exist already?

doc/compodoc_sources/how-to-guides/create-new-entity-type.md
```diff
@@ -1,21 +1,67 @@
...
+
+There are two approaches to define entity types:
+- at runtime in the config file
+- at build time in the code
+
+Unless you have special requirements and write further code specifically for your entity type, 
+you usually should rely on the more flexible and easy approach defining them in the config. 
+
+
+## Defining a custom Entity type only through config
```
TheSlimvReal, line 18: I would rather not have this duplicate section here. Lets just have a link to the one in `configuration` and not having to maintain and update the info in two places.

doc/compodoc_sources/how-to-guides/create-new-entity-type.md
```diff
@@ -1,21 +1,67 @@
...
+    "labelPlural": "My Entities",
+    "toStringAttributes": ["firstname", "lastname"], // used when an entity is displayed as a reference/link
+    "attributes": [
+        {"name": "name", "schema": { "dataType": "string", "label": "name" } },
+        {"name": "remarks", "schema": { "dataType": "string", "label": "Further remarks", "editComponent": "EditLongText" } },
+    ]
+}
+```
+
+The [config-fix.ts](https://github.com/Aam-Digital/ndb-core/blob/master/src/app/core/config/config-fix.ts) in the repository is an example and demo configuration from which you can copy parts of configuration definitions also.
```
TheSlimvReal, line 44: Why would we have this information here? This is very general and also relevant in other cases. Lets keep all the information about the configuration system in one place.

TheSlimvReal (changes requested):
I don't like the fact that we will have to maintain information about the configuration system in two places. Lets just have one section for how entities are defined in the config and only link it here. Otherwise one of the two will always be outdated.

doc/compodoc_sources/how-to-guides/create-new-entity-type.md
sleidig: agreed. My tendency would be to use this "How To" guide to explain the details of entity config then and use the "Concepts" article about configuration only for the big picture approach of config + linking to this guide (+ linking to the interface .ts file specifying the available config props for entity configs - if we have that?).

What are your thoughts to disentangle this in the most readable way?

TheSlimvReal: As I understood it, this how to guide was about adding a new entity as actual code. Most of the content refers to code concepts. For people that only want to configure something I do not see this as a good starting point (they would have to read through a lot of stuff until we tell them that for doing it through the config they do not require any of that knowledge). 
So I would prefer having a very good documentation in the configuration section as this should hold all the required information that a administrator would need to adjust the config and keeping this guide solely for developers/contributors that want to extend the code. So I think a simple hint at the beginning that explain this (adding vs configuring entity) and refers to the config guide would make it easier for people to find the for them relevant place.

doc/compodoc_sources/how-to-guides/create-new-entity-type.md
TheSlimvReal: As mentioned above I would like to have all information regarding the config system in one place and not mix this in any way with implementation details. So the hint about the example config should only be required in the documentation of the config once.

doc/compodoc_sources/how-to-guides/create-new-entity-type.md
TheSlimvReal: We could also consider to move the part that explains the config file into a how to guide as it basically describes how to make adjustments to the app and split it from the implementation details that explain how the architecture behind loos like.

sleidig: I like that idea. Have added a stub for now and split the config part out of this guide :+1:

TheSlimvReal (approved):
Great, thanks for cleaning up some documentation

Issue #1844, opened 2023-04-20
Location input should allow to manually extends the address 


TheSlimvReal:
The location input only allows to select an existing address. 

Sometimes it might be necessary to extend this address with additional information (while still keeping the coordinates) e.g. to save street number 6b instead of just 6.

Problems / aspects related to this:
- import of address may loose information (if the address cannot be mapped, it doesn't get imported at all)
- additional address details (like floor, building number, etc.) cannot be saved
- the entered address is reformated (e.g. showing house number before the street name) and displayed differently from what the user typed - which can cause confusion.
- Address field should be a long text field to display the details fully (?)

---
_Als Projektkoordinator:in muss ich manchmal Zusatzinformationen zu einer Adresse angeben oder eine Haus-/Wohnungsnummer speichern, die nicht im offiziellen Adress-Register vorhanden ist (das für die Kartenfunktion abgefragt wird)._

**To implement:**
- [x] datatype: change `LocationDatatype`
  - [x] store "raw address" string in addition to the location lookup object `{ locationText: string; geoLocation: GeoResult }`
  - [x] migration function to provide backward compatibility
- [x] UI:
  - [x] `ViewLocationComponent`: display locationText (+ icon if geo available) 
  - [x] `EditLocationComponent`: textbox for custom location text + button to "locate on map" (which triggers search, allows to select geo-result); From lookup popup, can copy address back into custom address - or otherwise system will just display "[selected on map]"
- [x] Import:
  - [x] module saves the exact address string imported in the locationText, even if the address cannot be found and mapped on the location API. (currently invalid address details are not imported at all)

sleidig:
We would need to store (and display) an "extraAddress" text in addition to the lookup GeoResult. This leads to a number of variations to consider regarding UX. We might either display two form fields or smartly combine things into one (if we can figure out a good way for that).

User stories & edge cases (assuming two strings: _address_ and _extraAddress_):
- [x] _address_ = lookup location (existing cases to set / edit); no additional extra_address defined
- [ ] set _extraAddress_ to non-existing location
- [ ] edit _extraAddress_ from non-existing to existing location
- [ ] edit _address_ from existing location to non-existing location

---
The easiest implementation would be to just have an independent additional text field:

> Address: _..address.._ [map icon]
> Additional address details: _..extraAddress.._ [?]
> 
[...]

PR #1974, opened 2023-08-17
feat(import): import address by lookup of location data

### LINKED ISSUE #1957: Import address into "location" data field
During imports we may have address value as simple text that should be mapped into a property that has "location" dataType (through which we enable map integration). At the moment, this cannot be imported correctly.

sleidig:
closes #1957 

Implemented basic use case without a configuration UI or user feedback. Looking into #1844 to enable any address information to be imported without data loss.

### DIFF
--- a/src/app/features/location/location.datatype.spec.ts
+++ b/src/app/features/location/location.datatype.spec.ts
@@ -1,0 +1,78 @@
+import { TestBed } from "@angular/core/testing";
+import { of } from "rxjs";
+import { LocationDatatype } from "./location.datatype";
+import { GeoResult, GeoService } from "./geo.service";
+
+describe("Schema data type: location", () => {
+  let service: LocationDatatype;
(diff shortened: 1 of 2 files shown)


TheSlimvReal (approved):
Works really nice. Good job. Glad to see our refactoring pay off

src/app/features/location/location.datatype.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+  });
+
+  async function testImportMapping(
+    importedValue: string,
+    mockedLookup: GeoResult[],
+    expectedResult,
+  ) {
+    mockGeoService.lookup.and.returnValue(of(mockedLookup));
+
+    const actualResult = await service.importMapFunction(importedValue, null);
```
TheSlimvReal, line 30:
```suggestion
    const actualResult = await service.importMapFunction(importedValue);
```

PR #1969, opened 2023-08-15
Generalized ChildSchoolRelation to allow group relationships

sleidig:
Generalize the data structure and components around `ChildSchoolRelation` to allow group relations (e.g. array  of children linked to one school within one relation).

### Visible/Frontend Changes
- [x] A generic alternative to ChildSchoolOverview can work on any kind of entity type, having any kind of referenced entity types
- [x] ~NotesRelatedToEntity can reference any indirectly (2-hop) related entity type in new Notes~ --> PR #1970 

### Architectural/Backend Changes
- ~(maybe) generalize index / query of ChildrenService.createChildSchoolRelationIndex + others~
  - --> skip this and go for completely generated indices later
- ~(maybe) generate indices from config / schema without hard-coding pouchdb query~
  - --> later (first get basic functionality working), an issue for that already exists and this could replace a lot of custom services and components that are mostly boilerplate:
  - #262
- [x] (maybe) generalize ChildSchoolRelation to simply be a time-period-based entity relationship [???]

PR #1975, opened 2023-08-17
Refactor folder structure

sleidig:
... to align things with the logic that has become more and more established.

---
- Merge _after_ #1969 
- but please keep as a separate commit rather than merging into the other PR
- sorry in advance for the million file changes :blush:


src/app/core/config/config-fix.ts
```diff
@@ -618,6 +618,31 @@ export const defaultJsonConfig = {
...
+                    "visibleFrom": "sm",
+                  },
+                  "schoolId",
+                  "schoolClass",
+                  "result",
+                ],
+              }
+            },
+
+
```
sleidig, line 645:
```suggestion
```

src/app/core/config/config-fix.ts
```diff
@@ -86,7 +86,7 @@ export const defaultJsonConfig = {
     "config": {
       "widgets": [
         {
-          "component": "DashboardShortcutWidget",
+          "component": "ShortcutDashboard",
```
TheSlimvReal, line 89: This will require a (simple) migration.

src/app/core/entity-components/entity-properties/edit/edit-age/edit-age.component.html
TheSlimvReal, line 1: I am wondering on what would be the best place for these. Another approach would be to have everything regarding a single datatype in one folder (components + datatype) as they are tightly coupled. With the configurable enum we already have this.

src/app/core/entity-components/entity-select/display-entity-array/display-entity-array.component.ts
TheSlimvReal, line 1: I don't really see why the components should be in a different folder. Especially as only one of them uses the entity select (edit entity array) the others have nothing to do with it. I would rather see the entity select as part of the edit entity array as it is not use anywhere else (anymore).
So I would suggest to also move them into the. entity properties folder.

src/app/core/entity-components/related-entities/related-entities.component.ts
TheSlimvReal, line 1: Should we also have a folder with all the components that can be shown in a details view? On the other hand sometimes they can also be used for a full screen (e.g. `MatchingEntitiesComponent`)

TheSlimvReal: But I don't like that currently `EntityDetails` and `EntityList` are right next to `EntitySubrecord`, `EntityForm` or `RelatedEntities`

TheSlimvReal (changes requested):
I really like the cleanup. That definitely clarifies things and moves into the right directions.

I added some general thoughts I have regarding the folder structure.

src/app/core/config/config-fix.ts
sleidig: "DashboardShortcutWidget" is registered as an alias in the new module. Or do we prefer to run a migration?

src/app/core/entity-components/entity-properties/edit/edit-age/edit-age.component.html
sleidig: yes, I think I'd actually prefer that (comps + datatype). Keeps closely related things together and also reflects the same structure that extensions for new features follow.

src/app/core/entity-components/entity-select/display-entity-array/display-entity-array.component.ts
sleidig: good point. would create datatype-oriented folders as discussed above and give these their own, clearly sorted one then

src/app/core/entity-components/related-entities/related-entities.component.ts
sleidig: also fair ... 

how about
- core/basic-datatypes
- core/entity-list
- code/entity-details (including the very generic comps like RelatedEntities - or they get a (whatever named?) own core/x folder ...)
- feature/matching

src/app/core/config/config-fix.ts
TheSlimvReal: Ah right, but generally the migration is easy to achieve with search and replace so we might as well do it now.

TheSlimvReal: Maybe a point to also finally do this for the other aliases.

src/app/core/entity-components/related-entities/related-entities.component.ts
TheSlimvReal: Sounds good to me. At the moment I cannot think of anything where this doesn't make sense

src/app/core/config/config-fix.ts
sleidig: :+1:

src/app/core/entity-components/related-entities/related-entities.component.ts
sleidig: what about special display variants (e.g. display-percentage, edit-long-text) that currently don't have a datatype on their own? Add as additional comps in the related datatype?

TheSlimvReal (approved):

Issue #1980, opened 2023-08-21
build-storybook / Percy CI is failing with "JavaScript heap out of memory"


sleidig:
See CI: e.g. https://github.com/Aam-Digital/ndb-core/actions/runs/5923239290/job/16058498260

While `npm run storybook` seems to run okay on local machines, our CI job to do visual regression testing using Storybook is failing at the step of building storybook statically:
`npm run build-storybook` (which uses `ng run ndb-core:build-storybook`, i.e. `"builder": "@storybook/angular:build-storybook"`) exits with

> FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory

sleidig:
This is probably related to this: https://github.com/storybookjs/storybook/issues/6408

setting `storyStoreV7: false` didn't resolve it for me, however.

sleidig:
The build storybook seems to behave a bit different from the hot-reload "npm run storybook". Now I see an error with ɵɵinvalidFactoryDep

https://github.com/storybookjs/storybook/issues/23534
has a workaround (set `"enableProdMode": false`) that seems to help

PR #1982, opened 2023-08-21
Fix percy / storybook build

sleidig:
closes: #1980


TheSlimvReal:
Its possible to watch how percy works (https://docs.percy.io/docs/debugging-sdks#display-the-asset-discovery-browser) but this did not really lead me anywhere

src/app/core/basic-datatypes/configurable-enum/enum-dropdown/enum-dropdown.stories.ts
TheSlimvReal, line 1: The last problem comes from this story. I have now configured this to use the default form field story generator, this however does not work too great as the enum service is not properly set up.

With the previous story the problem is the injection of the `NgControl` in the `BasicAutocompleteComponent` (https://github.com/Aam-Digital/ndb-core/blob/fix-percy/src/app/core/common-components/basic-autocomplete/basic-autocomplete.component.ts#L123).
I don't know why this is a problem with the previous story. Also the problem does not occur when only running Percy on this single story (by restricting the stories in Storybooks `main.js` file). I was not able to mock this dependency.

sleidig: added a way to provide mockServices there, so that enumService is available

PR #1986, opened 2023-08-24
fix: removed migration for configurable enums

TheSlimvReal:
The migration was already done a few months ago and is not required anymore


src/app/core/config/config.service.ts
```diff
@@ -73,51 +64,4 @@ export class ConfigService {
...
-
-  private async detectLegacyConfig(config: Config): Promise<Config> {
-    // ugly but easy ... could use https://www.npmjs.com/package/jsonpath-plus in future
-    const configString = JSON.stringify(config);
-
-    if (configString.includes("ImportantNotesComponent")) {
-      this.logger.warn(
-        "Legacy Config: ImportantNotesComponent found - you should use 'ImportantNotesDashboard' instead",
-      );
-    }
```
sleidig, line 85: do we want to keep this functionality in general to detect "legacy" config pieces?

sleidig (approved):

src/app/core/config/config.service.ts
TheSlimvReal: Hmm good point. I was thinking that we rather use the admin service to always automatically migrate things without relying on the frontend.

sleidig: True, I think the admin API is the better approach. Maybe config scans are something to include in our release process or could even be monitored somehow on the server-side.

PR #1987, opened 2023-08-24
fix(forms): display label for photo fields

sleidig:
The photo field until now did not display any form-field label. This is becoming an issue as the field is more widely used and there are forms with multiple photo fields.

- photo field has a label like any other field now
- I failed to keep the image centered within form-field, but the left bound layout also looks fairly clean and consistent now in my opinion.


TheSlimvReal (changes requested):
I was wondering whether we could have the label next to the buttons below the image so we also don't have the issues of the component becoming bigger once the user switches into edit mode (because the buttons appear). The current way the label is seems quite hidden.

Issue #1342, opened 2022-07-05
2-Factor authentication


TheSlimvReal:
2-Faktor authentication (2FA) should be available in the application

- [ ] Organizations can decide whether they want to use 2FA
- [ ] 2FA with email and/or app works
- [ ] Configurable time period after which the second factor is required

[overview of login user stories / test cases (GoogleDoc)](https://docs.google.com/document/d/1VBx_Zsk0d7KdfcXyUhf77a-nWHArHvsKmST3Ss0jr_k/edit)

TheSlimvReal:
Possible approach:

1. Instance is marked with *OTP required* (potentially through `config.json`).
2. When creating a user, `CONFIGURE_OTP` is put into the `requiredActions` of the user.
3. When a user clicks on the registration mail, OTP setup will also be part of the registration process in Keycloak.
4. Aam Digital is extended with a login step for entering the OTP, potentially in same form as username and password.
5. By sending username, password and OTP to keycloak, a token can be requested ([example](https://ultimatesecurity.pro/post/2fa-api/#access-keycloak-apis-using-two-factor-authentication)).
6. If everything is correct the user is logged in.

Open questions/issues:

- How is the offline login affected? OTP check does not work offline.
- It is not possible to find out whether the username and password **OR** the OTP were incorrect. All three things have to be valid in order to get a successful response.

sleidig:
We have discussed the technical approach for 2FA again and came to the conclusion to re-work the login logic in the client/app in order to use Keycloak "natively", i.e. forward users to the Keycloak login page and on the app only use the token issued. This requires re-thinking how we handle offline authentication, however.

PR #1990, opened 2023-08-30
Login with Keycloak UI

TheSlimvReal:
closes: #1342 

Open questions
- [x] Should the offline login support multiple users or only the last one? 
- [x] What capabilities do we want for the demo mode? Do we need different users there or is one sufficient?
- [x] What are the restrictions for offline usage (e.g. only when there is no internet, or how long does it work without a remote login)? -> for now no time restrictions, but online login is required while online
- [x] Is it okay to completely remove CouchDB login? -> has been removed
- [x] What should happen if no email configured? -> Using a condition we can only ask for OTP if a email is configured


sleidig:
**_questions & observations from functional testing / UX:_**
- Overall this feels really smooth and straight forward :smiley: 
- The style + title ("DEV") of the login page will be shared across all instances using the same Keycloak, right?
- the "... is loading for the first time" after login is slightly disrupting (maybe we should at least rephrase this), although it is only showing very briefly on my computer
- no offline detection yet to disable the "Please login" button (I think?)
- To make it even more seamless, I am thinking whether there are possible options to remove the "Please login" button and get people directly to the login form. But then the switch between offline and online login is difficult? (although, if we can detect state reliably, we could only forward to keycloak if no active login + online).

with custom styling + reducing clicks in mind: embedding the whole keycloak login page with an iframe (rather than linking to it with the button) is discouraged, you said (I think). Do you know more about the downsides of that?

src/app/core/session/auth/local/local-auth.service.ts
```diff
@@ -0,0 +1,40 @@
...
+import { AuthUser } from "../auth-user";
+
+@Injectable({
+  providedIn: "root",
+})
+export class LocalAuthService {
+  private static LAST_LOGGED_IN_KEY = "LAST_USER";
+
+  login(): AuthUser {
+    return this.getStoredUser(LocalAuthService.LAST_LOGGED_IN_KEY);
```
sleidig, line 11: `getStoredUser()` doesn't seem to get reused and is itself only a short method. I think it might be more readable to inline it here.

src/app/core/session/auth/local/local-auth.service.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  canLoginOffline(): boolean {
+    return !!localStorage.getItem(LocalAuthService.LAST_LOGGED_IN_KEY);
+  }
+
+  private getStoredUser(username: string): AuthUser {
+    const stored = localStorage.getItem(username);
+    if (!stored) {
+      throw Error("OFFLINE LOGIN ERROR: No user found locally");
```
sleidig, line 21: is this behavior intuitive to throw an error if the login is not possible, rather than returning false / null / ...?

src/app/core/session/auth/local/local-auth.service.ts
```diff
@@ -0,0 +1,40 @@
+import { Injectable } from "@angular/core";
+import { AuthUser } from "../auth-user";
+
+@Injectable({
+  providedIn: "root",
+})
+export class LocalAuthService {
```
sleidig, line 7: could you add JsDoc to the class and public methods here? :blush:

src/app/core/session/session-service/local-session.ts
```diff
@@ -15,18 +15,12 @@
...
 import { PouchDatabase } from "../../database/pouch-database";
 import { AppSettings } from "../../app-settings";
 import { SessionType } from "../session-type";
 import { environment } from "../../../../environments/environment";
-import { AuthUser } from "./auth-user";
+import { AuthUser } from "../auth/auth-user";
+import { Database } from "../../database/database";
 
 /**
  * Responsibilities:
```
sleidig, line 26: the JsDoc might need an update.
Also, is "LocalSession" still the best name for this, as we are moving away from this core concept of a session, I think?

src/app/core/session/session-service/session-manager.service.ts
```diff
@@ -0,0 +1,103 @@
...
+import { LoginStateSubject, SessionType } from "../session-type";
+import { LoginState } from "../session-states/login-state.enum";
+import { Router } from "@angular/router";
+import { KeycloakAuthService } from "../auth/keycloak/keycloak-auth.service";
+import { LocalAuthService } from "../auth/local/local-auth.service";
+import { environment } from "../../../../environments/environment";
+
+/**
+ * A synced session creates and manages a LocalSession and a RemoteSession
+ * and handles the setup of synchronisation.
```
sleidig, line 33: This JsDoc might be left from the previous concept?

src/app/core/session/session-service/session-manager.service.ts
```diff
@@ -0,0 +1,103 @@
...
+  canLoginOffline(): boolean {
+    return this.localAuthService.canLoginOffline();
+  }
+
+  logout() {
+    if (this.remoteLoggedIn) {
+      this.remoteAuthService.logout();
+    } else {
+      this.userService.user = undefined;
+      this.loginStateSubject.next(LoginState.LOGGED_OUT);
```
sleidig, line 69: shouldn't loginState etc. also reset for remote logout? I guess this does happen somewhere else already but for cleaner code, is there any chance to do this here, in the same place?

src/app/core/session/session-service/session-manager.service.ts
```diff
@@ -0,0 +1,103 @@
...
+      this.router.navigate(["/login"], {
+        queryParams: { redirect_uri: this.router.routerState.snapshot.url },
+      });
+    }
+  }
+
+  /**
+   * Do log in automatically if there is still a valid CouchDB cookie from last login with username and password
```
sleidig, line 77:
```suggestion
   * Do log in automatically if there is still a valid auth cookie from previous remote login
```

src/app/core/session/session-service/session-manager.service.ts
```diff
@@ -0,0 +1,103 @@
...
+   */
+  checkForValidSession() {
+    // TODO maybe also in offline case?
+    if (environment.session_type === SessionType.synced) {
+      return this.remoteAuthService
+        .autoLogin()
+        .then((user) => this.initialiseUser(user))
+        .then(() => this.handleRemoteLogin())
+        .catch(() => undefined);
```
sleidig, line 86: log errors

src/app/core/session/session-service/session-manager.service.ts
sleidig: a similar point for `remoteLogin()`: from the service here, it's not really clear how/when `initializeUser()` is called.

src/app/core/session/session.module.ts
```diff
@@ -39,27 +31,17 @@ import { serviceProvider } from "../../utils/utils";
...
-        ? injector.get(KeycloakAuthService)
-        : injector.get(CouchdbAuthService);
-    }),
-    {
-      provide: HTTP_INTERCEPTORS,
-      useClass: AuthInterceptor,
-      multi: true,
-    },
+    LoginStateSubject,
+    SyncStateSubject,
```
sleidig, line 40: I really like the simplicity in tests and other places to access these states now! However, I do wonder if it wouldn't be easier to find the relevant state (for someone less familiar with the code base) if we do collect them in a single provided service (containing loginState + syncState + currentUser)?

src/app/core/user/user.service.ts
```diff
@@ -0,0 +1,13 @@
...
+@Injectable({
+  providedIn: "root",
+})
+export class UserService {
+  user: AuthUser;
+
+  getCurrentUser(): AuthUser {
+    return this.user;
+  }
+}
```
sleidig, line 13: any reason for this to be a service method while loginState is a directly exposed subject?

src/app/core/database/sync.service.spec.ts
```diff
@@ -0,0 +1,103 @@
...
+    // logout + error -> no restart
+    syncSpy.calls.reset();
+    loginState.next(LoginState.LOGGED_OUT);
+    tick();
+    errorCallback();
+    expect(syncSpy).not.toHaveBeenCalled();
+  }));
+
+  // TODO need to access initRemoteDB somehow
+  // it("should try auto-login if fetch fails and fetch again", async () => {
```
sleidig, line 76: is this still a relevant, open TODO?

src/app/core/database/sync.service.ts
```diff
@@ -0,0 +1,172 @@
...
+import { SyncState } from "../session/session-states/sync-state.enum";
+import { LoginStateSubject, SyncStateSubject } from "../session/session-type";
+import { LoginState } from "../session/session-states/login-state.enum";
+import { filter } from "rxjs/operators";
+import { KeycloakAuthService } from "../session/auth/keycloak/keycloak-auth.service";
+
+@Injectable({
+  providedIn: "root",
+})
+export class SyncService {
```
sleidig, line 17: brief JsDoc please

src/app/core/session/auth/keycloak/keycloak-auth.service.ts
```diff
@@ -1,20 +1,15 @@
...
 import { environment } from "../../../../../environments/environment";
-import { AuthUser } from "../../session-service/auth-user";
-import { catchError } from "rxjs/operators";
+import { AuthUser } from "../auth-user";
+import { KeycloakService } from "keycloak-angular";
+import { SessionType } from "../../session-type";
 
 @Injectable()
-export class KeycloakAuthService extends AuthService {
+export class KeycloakAuthService {
```
sleidig, line 11: short JsDoc please, to clarify the role of this within the overall system :blush:

src/app/core/session/auth/keycloak/keycloak-auth.service.ts
```diff
@@ -1,20 +1,15 @@
...
 import { parseJwt } from "../../../../utils/utils";
 import { environment } from "../../../../../environments/environment";
-import { AuthUser } from "../../session-service/auth-user";
-import { catchError } from "rxjs/operators";
+import { AuthUser } from "../auth-user";
+import { KeycloakService } from "keycloak-angular";
+import { SessionType } from "../../session-type";
 
 @Injectable()
-export class KeycloakAuthService extends AuthService {
```
sleidig, line 17: my heart is still bleeding a tiny bit for losing the generic AuthService interface. But I think there is not too much value in keeping that (with the exact same interface as this KeycloakAuthService) just for the sake of abstraction ...

src/assets/silent-check-sso.html
```diff
@@ -0,0 +1,7 @@
+<html>
+<body>
+<script>
+    parent.postMessage(location.href, location.origin);
```
sleidig, line 4: I know this is required by Keycloak, but can you explain to me what it actually does? :sweat_smile:

sleidig (changes requested):
This code looks incredibly clean and clear. Great job! :heart_eyes: 
(of course, I still had a number of minor comments but looking at the big picture this seems pretty much good to go)


- [ ] update developer docs
  - https://aam-digital.github.io/ndb-core/documentation/additional-documentation/concepts/session-and-authentication-system.html
  - https://aam-digital.github.io/ndb-core/documentation/additional-documentation/concepts/overall-architecture.html

src/app/core/session/session-service/local-session.ts
TheSlimvReal: Move to SessionManager

src/app/core/session/session-service/session-manager.service.ts
TheSlimvReal: Improve function names and add comments for explanation

src/app/core/user/user.service.ts
TheSlimvReal: Should become subject now with multi user offline login

sleidig (commented):
Code looks good :+1: 

Functional testing & UX:
- :slightly_smiling_face: overall the flow and forwarding feels quite good for me
- Upon logout a lot of stuff happens (the app reloads; the login screen says it's trying login; then the keycloak login page appears ...) This shouldn't be much of an issue but feels slightly weird if you watch it closely (and at slowed down CPU speed)
- :question: If the remote connection fails (for syncing or loading roles on the user account component) while the user is logged in: Are we handling that in any way? Can/Should we send the user to remote login again?
- [ ] When resetting remote login (after logout previously while offline) there also is first shown "login in progress", then briefly the "building application" and only then the login screen 
- [ ] tweak layout of login view a bit (message when remote login fails (+ tooltip?); margins)
- [ ] UI of "Use Offline": I am wondering if this would be more seamless to display the user accounts directly without the popup. I'll briefly experiment with that alternative.

sleidig (changes requested):
Code looks good to me :100: 

**Functional Testing including 2FA**
(on dev.aam-digital.net - some things may just be configuration issues):
- 2FA via email code works really smoothly and the "remember device" option is super useful :smiley: 
- [ ] no "forgot password" option on (keycloak) login form
- [ ] 2FA email with codes could have a slightly more elaborate email template (maybe a footer stating the system name in the body - where are these templates maintained?)
- [ ] make keycloak login light orange background also white? (considering our recent UI background switch, and the fact that the color scheme of the app might have been customized to other colors)
- [ ] trying login with "demo-admin" somehow puts me into a "We are sorry ... invalid username or password" screen without any chance to try with different password (console logs a 400 Bad Request)
- [x] on slow devices and connections the offline login section becomes available even before the "re-check" of login after an online login is complete. Maybe we need to increase the timeout a bit further?
- [ ] offline login section should be available immediately if the device is completely offline (the "we couldn't connect to server" text is showing immediately, but the offline login option only appears with the delay).
- [x] Also, I would maybe just disable the offline login section instead of not displaying it at all?
- [ ] Do we have documentation about how to set this up (and possible adjust the keycloak setup)?

sleidig (approved):
Good to merge :heavy_check_mark: 

I'd prefer to deploy this (as a pre-release?) to selected pilot instances first, however.

Issue #1939, opened 2023-07-24
dropdown field (edit-configurable-enum) shows [object Object] when selecting with keyboard


sleidig:
**Describe the bug**
When you confirm a dropdown option by pressing ENTER on the keyboard instead of clicking with the mouse, the field shows "[object Object]" as selected value. The correct label gets displayed after you click somewhere else (i.e. when the field loses focus) but in the meantime this confuses users.

**To Reproduce**
1. Go to 'Child Details' and activate Edit mode
2. Click on 'Center', navigate with keyboard arrows through the options, hit ENTER key to confirm the selection
3. See error

PR #1993, opened 2023-09-07
Added call of blur() in basic-autocomplete

christophscheuing:
closes: #1939

### Visible/Frontend Changes
- When hitting enter after selecting an option from the dropdown menu, the input field looses its focus. This also is the case when selecting an option with the mouse.


sleidig (approved):
seems to work well. Thanks a lot!

PR #1988, opened 2023-08-24
Daterangepicker fixes

christophscheuing:
closes: #1967

### Visible/Frontend Changes
In the date range panel component an "All" option has been included.

PR #2005, opened 2023-09-18
fix(*): correctly allow date ranges where start date is in future

sleidig:
allow date ranges where start date is in future and no end date is set

:warning: mark commit for @codo-mentoring

### DIFF
--- a/src/app/core/entity-details/related-time-period-entities/time-period.spec.ts
+++ b/src/app/core/entity-details/related-time-period-entities/time-period.spec.ts
@@ -73,10 +73,17 @@ describe("TimePeriod Entity", () => {
     expect(() => relation.assertValid()).toThrowError();
   });
 
-  it("does pass validation when the start date is before the end date", () => {
+  it("should pass validation when the start date is before the end date", () => {
     const relation = new TimePeriod();
     relation.start = moment().subtract(1, "day").toDate();
     relation.end = new Date();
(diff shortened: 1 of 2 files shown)


TheSlimvReal (commented):
I am not sure that I can reproduce the bug. When I use the latest master, I can also enter or select a date range in the future.

sleidig:
hmm, interesting. In the demo I can't reproduce either @TheSlimvReal . But here and in a client system it is happening: https://qm.aam-digital.com/matching  (maybe due to synced mode or because the "end" field is not part of the form?)

TheSlimvReal:
@sleidig I think you can reproduce it in the demo. The actual fix to the problem has not yet been deployed (#1988) With the "All" filter option it is also possible to only have a start or end date.

sleidig:
@TheSlimvReal , indeed this bug only occurs if the "end" field is not part of the confirmation popup of the matching view.

TheSlimvReal:
It seems like moment has a different behaviour depending on whether `undefined` or `null` is passed as end date.

sleidig:
> It seems like moment has a different behaviour depending on whether `undefined` or `null` is passed as end date.

ah, that makes sense, I guess.
I'd suggest to go ahead with this PR though, as the fix makes the behavior more explicit and reliable instead of using a not very well known default of moment.js

PR #2007, opened 2023-09-19
fix(filters): "false" option of boolean filters includes undefined

sleidig:
so that entities for which the user has not touched a checkbox field (leaving it unchecked) are correctly included

:warning: mark commit for @codo-mentoring


src/app/core/filter/filter/filter.component.spec.ts
```diff
@@ -58,38 +62,57 @@ describe("FilterComponent", () => {
   });
 
   it("should support a boolean filter", async () => {
```
TheSlimvReal, line 64: I think the correct place for testing the actual filter functionality would be in `filters.spec.ts`

TheSlimvReal (approved):
Makes a lot of sense this addition

PR #2008, opened 2023-09-19
Design tuning

sleidig:
### Visible/Frontend Changes
- [x] switch sidenav and main content backgrounds

and a bit of code cleanup.

### DIFF
--- a/src/app/core/pwa-install/pwa-install.component.scss
+++ b/src/app/core/pwa-install/pwa-install.component.scss
@@ -4,6 +4,6 @@
   border-top: solid 1px rgba(0, 0, 0, 0.12);
   border-radius: 0;
   overflow: hidden;
-  background-color: colors.$primary;
   width: 100%;
+  color: white;
 }
--- a/src/app/core/ui/navigation/navigation/navigation.component.html
+++ b/src/app/core/ui/navigation/navigation/navigation.component.html
@@ -24,6 +24,7 @@
       [angularticsLabel]="item.label"
       [routerLink]="[item.link]"
       [class.matched-background]="item.link === activeLink"
+      onclick="this.blur();"
     >
       <a class="flex-row gap-small">
         <app-fa-dynamic-icon
--- a/src/styles/variables/_colors.scss
+++ b/src/styles/variables/_colors.scss
@@ -43,7 +43,11 @@ $disabled: rgba(0, 0, 0, 0.38);
 $disabled-transparent: $grey-transparent;
 $border-color: $grey-darker;
 
-$background: mat.get-color-from-palette(theme.$primary, 50);
+/** background for the main view */
+$background: white;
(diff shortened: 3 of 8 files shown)


src/app/core/pwa-install/pwa-install.component.scss
```diff
@@ -4,6 +4,6 @@
   border-top: solid 1px rgba(0, 0, 0, 0.12);
   border-radius: 0;
   overflow: hidden;
-  background-color: colors.$primary;
   width: 100%;
+  color: white;
```
TheSlimvReal, line 8: is there a specific reason for making this white?

src/app/core/ui/navigation/navigation/navigation.component.html
```diff
@@ -24,6 +24,7 @@
       [angularticsLabel]="item.label"
       [routerLink]="[item.link]"
       [class.matched-background]="item.link === activeLink"
+      onclick="this.blur();"
```
TheSlimvReal, line 27: A bit funky but I guess this removes the focus indicator from the button?

src/styles/variables/_colors.scss
```diff
@@ -43,7 +43,11 @@ $disabled: rgba(0, 0, 0, 0.38);
 $disabled-transparent: $grey-transparent;
 $border-color: $grey-darker;
 
-$background: mat.get-color-from-palette(theme.$primary, 50);
+/** background for the main view */
+$background: white;
```
TheSlimvReal, line 47: At the moment the completely white screen feels quite intense on the eyes but I guess this is something to get used to. What I find a bit problematic is that with the white background the tabs for different columns groups on the entity lists are quite difficult to find. At least for me it feels much less clear what they are there for when they have no distinction from the table below. Similarly with the tabs in the entity details.

TheSlimvReal (changes requested):
Nice to see how little code was required to change this UI layout. I think I will need quite some time to get used to this though.

src/styles/variables/_colors.scss
sleidig: :sweat_smile: I heard the exact opposite from codo team ("the colored background is quite overwhelming")

But your point about contrasts and UX relations is super important - thanks for looking at these details. I reviewed that aspect carefully now. To me it seems like this may be an issue but previously this was similarly hidden. I have raised the elevation level a bit to give more distinction between table and background, however.

src/app/core/pwa-install/pwa-install.component.scss
sleidig: the top toolbar has the same background and white text, so this seems more consistent (although I don't feel good setting it explicitly). At different color schemes the contrast with black text was even worse, also.

TheSlimvReal (approved):

PR #2009, opened 2023-09-20
fix: range background color in date picker

TheSlimvReal:
The date range picker does not highlight the selected range anymore.


sleidig (approved):
good find! Thanks

Issue #1674, opened 2023-01-20
Deleting / anonymizing data


TheSlimvReal:
## Goal
It must be possible to completely delete personal data from the system and the database.
It should be possible to only delete personal data while keeping statistically relevant data in the system.

## Status quo
When deleting an entity the personal data is not directly visible in the app anymore. However, the old revisions of this document still exist. This makes is possible to receive the already deleted data. By calling the `_compact` endpoint of the CouchDB the non-leave revisions are fully deleted. 

## Requirements

- [x] anonymization method where predefined set of values are set to undefined; make this configurable?
- [x] UI button that allows "Anonymization" of a record (as an alternative next to deleting it completely), make the difference intuitively clear to users
- [ ] permissions (distinct between delete and anonymize)
- [x] delete attachments (if property is cleared)
- [ ] `_compact` endpoint needs to be called on a regular basis to properly clear the deleted data
- do a cascading delete / anonymization; see #220

sleidig:
UX Drafts:
![image](https://github.com/Aam-Digital/ndb-core/assets/1682541/9f1ff3d4-55f2-4801-b61c-51211975e349)

sleidig:
## Config format:
Entities need to be configurable regarding what data is removed / retained when "anonymizing" a record.

additional flag in entity config `anonymize`
with possible options:
- `remove`
- `retain`
- `retain-anonymized` (e.g. keep year of birth but remove day + month to remove personal identifiable details)

### What should be the default (i.e. if this is not defined for a property in the config)?
Consider all properties by default to be removed? This would be along the lines of data protection by default. It may be easy to miss / forget to flag a personal data field as such.

Would we normally have more properties to remove or to retain?
Looking at the actual use cases and setups, there often are a lot of fields for contact details that have to be deleted. Also, often only some selected properties are at all relevant for reports, so other details could be removed without loosing much.

--> remove by default, flag explicitly to retain after anonymization?

### example config:
```
"entity:Person": {
  "attributes": [
    {
      "name": "lastname",
      "schema": {
        "label": "Lastname",
        "dataType": "string",
        "anonymize": "delete" // flag a property as "personal data" to be removed during anonymization (could also be omitted as default)
      }
    },
    {
      "name": "dateOfBirth",
      "schema": {
        "label": "Date of Birth",
        "dataType": "date-only",
        "anonymize": "retain-anonymized" // do not delete completely but anonymize by keeping only the year
      }
    },
    {
      "name": "city",
[...]

Issue #899, opened 2021-07-20
clear, consistent behavior of "inactive" (make dropout more visible)


TheSlimvReal:
*As a social worker I want to know about the dropout logic in order to mark children as inactive instead of deleting them.*
The delete button should only be used if e.g. a child was created by accident. In case a child drops out of the program, the dropout logic should be used instead. This will automatically mark the child as inactive, once a dropout date has been set.
Currently, most users don't know about this feature and therefore delete the children instead. In the delete popup, the user should be asked if instead the child should be marked as dropout.

**Objective**
This should inform the user about the dropout logic in order to not delete children but keep them in the system marked as inactive. That allows users to read up on old children later on.

**Proposed Solutions & Alternatives**
In the popup where the user is asked if the child should be deleted, there should be an option asking if the child should be marked as dropout and potentially show a datepicker which allows to select the dropout date etc.

**Additional context**
Currently all entities (school, child, recurring activity) use the same delete logic and in some cases, the dropout logic does not make sense. Only in the case for a child, this additional info should be shown.

sleidig:
Very good points. I will think this through and write some user stories to highlight the dropout/inactive feature from all angles. This is a key aspect, we should maybe work through this thoroughly and add a couple of workflows and pointers.

considering also:
- how on technical level entities are marked as "inactive": #619
- how to count active children externally (for analytics and in the future also for billing of customers)
- if a child is actually deleted, what to do with the linked other documents? #874
- where to add a guide or documentation on the concept of "inactive" status, to explain that in detail?

sleidig:
Users [YGSP] have accidentally marked records as "inactive" (through the checkbox property, because they were not fully aware what this implies). The project manager asked to have permissions for this state to get better control.

sleidig:
a solution to turn the "inactive" status into a core feature, more deeply integrated into the UI, is discussed here currently: https://github.com/Aam-Digital/ndb-core/issues/1674#issuecomment-1731712668

PR #2012, opened 2023-09-25
Anonymization & Archive functionality

sleidig:
see #1674 (and #899)

## UX Changes
- Entity Details context (three-dot menu) items to "anonymize" and "archive" in addition to "delete" (+ extensive confirmation dialogs
- inactive records always hidden by default + toggle to include always available below table
- info box in Entity Details for archived + anonymized records
- anonymization logic

## Tasks
- after #2011 
- [x] create a unit testing setup
- [x] define all test cases
- [x] trigger anonymization logic
- [x] implement special "retain-anonymized" logic in datatype implementations
- [x] entity actions menu with "anonymize" option
- [x] display info box if archived/anonymized (both popup + entity details component, ...)
- [x] update Entity.isActive to ensure it always prioritizes the "archive" functionality (do we need a migration?)
  -  :question: what about logic for Todo and TimePeriod?
- [x] default filter for isActive on all lists? (or a way to very easily activate it in config)
- [x] ~somehow need to trigger an update to the entity details view after anonymize~ improve applyChanges of forms for entityUpdates
- [x] delete file attachments
- ~create a new permissions "anonymize"~
- [x] documentation
- _cascading anonymization of related records (--> will be done in separate PR)_


src/app/core/entity-details/entity-actions-menu/entity-actions-menu.component.ts
```diff
@@ -0,0 +1,115 @@
...
+  ],
+})
+export class EntityActionsMenuComponent implements OnChanges {
+  @Input() entity: Entity;
+
+  /**
+   * whether the "delete" action will trigger a navigation back to the parent list.
+   * This is useful when the entity is deleted from a fullscreen detail view but not for an overlay.
+   */
+  @Input() navigateOnDelete: boolean = false;
```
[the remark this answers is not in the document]
TheSlimvReal: I guess this doesn't work with the undo behaviour...

src/app/core/entity-details/entity-actions-menu/entity-actions-menu.component.ts
```diff
@@ -0,0 +1,115 @@
...
+      }
+      if (!this.entity?.isActive) {
+        return action.action !== "archive";
+      }
+      return true;
+    });
+  }
+
+  async executeAction(action: EntityMenuAction) {
+    const result = await this.entityRemoveService[action](
```
TheSlimvReal, line 107: I see the how this is a handy syntax but I am very afraid that this will break easily once something is changed in the `entityRemoveService`. The IDEA does not detect usage of the functions in the remove service so an error will most likely be unnoticed.

src/app/core/entity/entity-remove.service.ts
```diff
@@ -29,8 +33,36 @@ export class EntityRemoveService {
...
+    private schemaService: EntitySchemaService,
+    private fileService: FileService,
   ) {}
 
+  private showSnackbarConfirmation(
+    entity: Entity,
+    action: string,
+    navigateBackToUrl?: string,
+  ) {
+    const snackBarTitle = $localize`:Entity action confirmation message:${action} ${
```
TheSlimvReal, line 45: I don't think this part needs localize.

src/app/core/entity/model/entity.ts
```diff
@@ -196,6 +198,12 @@ export class Entity {
   })
   inactive: boolean;
 
+  /**
+   * Whether this entity has been anonymized and therefore cannot be re-activated.
+   */
+  @DatabaseField()
+  anonymized: boolean;
```
TheSlimvReal, line 205: Anonymize will often break the `toString` method. Should we add a fallback like `Anonymized child`? I think confluence does something like this when a user is deleted but still referenced in some docs.

src/app/core/entity/entity-remove.service.ts
```diff
@@ -65,45 +103,120 @@ export class EntityRemoveService {
...
+      originalEntity,
+      $localize`:Entity action confirmation message verb:Archived`,
+    );
+    return true;
+  }
+  /**
+   * Undo the archive action on the given entity.
+   * @param entity
+   */
+  async archiveUndo<E extends Entity>(entity: E) {
```
TheSlimvReal, line 200:
```suggestion
  async undoArchive<E extends Entity>(entity: E) {
```

src/app/core/entity-details/related-time-period-entities/time-period.ts
```diff
@@ -26,6 +26,7 @@ export class TimePeriod extends Entity {
    * Returns true if this relation is currently active
    */
   get isActive(): boolean {
+    // TODO: isActive time period vs. "archived" record?
```
TheSlimvReal, line 29: I have the feeling we might need to split these two concepts

src/app/core/entity/entity-remove.service.spec.ts
```diff
@@ -47,22 +66,27 @@ describe("EntityRemoveService", () => {
...
   });
 
   it("should delete entity, show snackbar confirmation and navigate back", async () => {
     // onAction is never called
     mockSnackBarRef.onAction.and.returnValues(NEVER);
+    // mock that dialog is dismissed immediately
+    const afterDismissed = new Observable<MatSnackBarDismiss>((subscriber) =>
+      subscriber.next({} as MatSnackBarDismiss),
+    );
+    mockSnackBarRef.afterDismissed.and.returnValue(afterDismissed);
```
TheSlimvReal, line 83:
```suggestion
    mockSnackBarRef.afterDismissed.and.returnValue(of(undefined));
```
Or is there any reason for this more complex approach?

src/app/core/entity/entity-remove.service.spec.ts
```diff
@@ -72,19 +96,209 @@ describe("EntityRemoveService", () => {
...
+      innerDataType: "date-only",
+    })
+    retainAnonymizedDates: Date[];
+
+    @DatabaseField({ dataType: "file" }) file: string;
+
+    @DatabaseField({ anonymize: "retain-anonymized", dataType: "entity-array" })
+    referencesToRetainAnonymized: string[];
+
+    static create(properties: Object) {
```
TheSlimvReal, line 160:
```suggestion
    static create(properties: Partial<AnonymizableEntity>) {
```

src/app/core/entity/entity-remove.service.spec.ts
```diff
@@ -72,19 +96,209 @@ describe("EntityRemoveService", () => {
...
+    entity.retainedField = "test";
+
+    await testAnonymization(
+      entity,
+      [entity],
+      [AnonymizableEntity.create({ retainedField: "test" })],
+    );
+  });
+
+  it("should anonymize and keep even record without any fields", async () => {
```
TheSlimvReal, line 202:
```suggestion
  it("should anonymize and keep empty record without any fields", async () => {
```

src/app/core/entity/schema/entity-schema-field.ts
```diff
@@ -104,6 +104,15 @@ export interface EntitySchemaField {
...
   showInDetailsView?: boolean;
+
+  /**
+   * whether the field will be retained when the entity is "anonymized".
+   *
+   * By default, fields are removed (data minimization by default).
+   *
+   * "retain-anonymized" triggers a special dataType action to retain the data partially in a special, anonymized form.
+   */
+  anonymize?: "retain" | "retain-anonymized" | undefined;
```
TheSlimvReal, line 115:
```suggestion
  anonymize?: "retain" | "retain-anonymized";
```
`undefined` is implicit when the property is optional.

src/app/core/site-settings/site-settings.service.spec.ts
```diff
@@ -44,36 +42,42 @@ describe("SiteSettingsService", () => {
     expect(service).toBeTruthy();
   });
 
-  it("should only publish changes if property has changed", () => {
+  it("should only publish changes if property has changed", fakeAsync(() => {
```
TheSlimvReal, line 45: Where does this change come from?

TheSlimvReal (changes requested):
I really like how most of the anonymization/archiving logic is found in very few places. Really good approach to it. The entity remove service gets quite big, especially all the tests around anonymization. But I also see the shared logic between the functions.

src/app/core/basic-datatypes/entity/entity.datatype.ts
```diff
@@ -57,4 +61,29 @@ export class EntityDatatype extends StringDatatype {
...
+      schemaField.additional,
+      value,
+    );
+
+    if (!referencedEntity) {
+      // TODO: remove broken references?
+      return value;
+    }
+
+    await this.removeService.anonymize(referencedEntity);
```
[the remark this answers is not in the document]
sleidig: Yes, the logic for direct references (indirect for cascading action are not implemented here yet) currently would be:
- "retain" -> keep unchanged
- "remove" (i.e. no explicit anonymize action configured) -> reference removed, related entity kept unchanged (:question: maybe not always what we want)
- "retain-anonymized" -> trigger "cascading" anonymization (and keep reference)

This is probably not always what is expected ... I would postpone this to the overall implementation of complete cascading anonymization, however. (tagged this comment explicitly in #220 now)

src/app/core/basic-datatypes/entity/entity.datatype.spec.ts
```diff
@@ -42,4 +43,28 @@ describe("Schema data type: entity", () => {
...
+
+    const anonymizedValue = await dataType.anonymize(
+      testValue,
+      testSchemaField,
+      null,
+    );
+
+    expect(anonymizedValue).toEqual(testValue);
+    expect(mockRemoveService.anonymize).toHaveBeenCalledWith(referencedEntity);
+    // expect(entityMapper.save).toHaveBeenCalledWith(referencedEntity); --> is called by EntityRemoveService (mocked here)
```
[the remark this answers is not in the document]
sleidig:
```suggestion
```

src/app/core/entity-details/entity-actions-menu/entity-actions-menu.component.ts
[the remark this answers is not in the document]
sleidig: yes, that's the issue - I had been flip-flopping a bit around where to put this myself ...

src/app/core/entity/entity-remove.service.spec.ts
sleidig: have kept this from previous code - yours looks nice!

src/app/core/site-settings/site-settings.service.spec.ts
sleidig: I think this was an attempt to fix this unstable test. Has nothing to do with this PR however, will remove.

src/app/core/entity-details/entity-actions-menu/entity-actions-menu.component.ts
sleidig: that's a really good alternative, thanks!

src/app/core/entity/entity-remove.service.ts
sleidig: may need to switch the order of action and type in different languages (in fact for most it needs to be the other way round, probably)

src/app/core/entity/model/entity.ts
sleidig: great suggestion! Will implement a fallback

src/app/core/config/config-fix.ts
```diff
@@ -562,15 +562,6 @@ export const defaultJsonConfig = {
         ]
       },
       "filters": [
-        {
```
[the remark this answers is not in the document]
sleidig: yes, probably important - the filter will be broken with a default of showInactive = false automatically applied as well.

src/app/core/entity-details/related-time-period-entities/time-period.ts
sleidig: _as discussed just now:_

we will keep the overlapping logic with special isActive functions of Todo and TimePeriodEntity and consider these as not active similar to the manual archiving.

TheSlimvReal (approved):
Thanks for the adjustments. Looks and feels very solid now.

src/app/child-dev-project/attendance/model/recurring-activity.ts
TheSlimvReal, line 1: It feels a little bit weird to talk give the annonymize option for recurring activities but then marking all fields as retain. Should we rather mark a whole entity as `not-personal` so that also the UI can hide this option and we don't have to worry about forgetting a property?

sleidig: yeah, actually had the exact same thought ... and for the cascading scenarios (https://github.com/Aam-Digital/ndb-core/issues/220#issuecomment-1740656623) we may need that also. Configuration options are growing explosively a bit, but I don't see good alternatives.

However, I think, I'd leave this as a TODO for the new PR.

Issue #2013, opened 2023-09-27
DisplayNumber in table columns is showing empty instead of displaying "0" as a value


sleidig:
**Describe the bug**
A value of "0" is not displayed in table columns (even if the user explicitly entered "0"). Instead the column cell remains empty.

**To Reproduce**
1. see in demo: https://demo.aam-digital.com/child/1?tabIndex=5
2. Edit an EducationalMaterial record, changing "amount" to "0"
3. --> Now that value is not displayed anymore in the column

PR #2015, opened 2023-09-28
fix: 0 value is ignored

TheSlimvReal:
closes: #2013 

### Visible/Frontend Changes
- [x] Number 0 is shown in view string component


sleidig (changes requested):
In the test deployment I still see empty instead of "0" columns in the Health Check sub-table when saving "0".

sleidig (approved):
Oh, my complaint is related but actually due to explicit checks in the other *DisplayComponents. Added there now.

PR #2026, opened 2023-10-13
fix: support avg queries

TheSlimvReal:
Actually, I am not sure if something like sum of all actual divided by sum of all totals is more required. Will check.


src/app/core/export/query.service.ts
```diff
@@ -248,6 +249,25 @@ export class QueryService {
...
+   * @private
+   */
+  private avg(data: any[], decimals = 0): string {
+    const numbers = data
+      .map((d) => Number.parseInt(d))
+      .filter((i) => !Number.isNaN(i));
+    const result =
+      numbers.length === 0
+        ? 0
+        : numbers.reduce((i, sum) => sum + i, 0) / numbers.length;
```
TheSlimvReal, line 267: Should the avg be calculated over all values or just the valid/existing ones?

src/app/core/export/query.service.spec.ts
```diff
@@ -635,6 +635,30 @@ describe("QueryService", () => {
...
+    // Returns valid number if no values are available
+    res = service.queryData("a:avg", undefined, undefined, [{ b: 1 }]);
+    expect(res).toBe("0");
+
+    // Numbers are fixed to provided decimals
+    res = service.queryData("a:avg(2)", undefined, undefined, [
+      { a: 3 },
+      { a: 2 },
+    ]);
+    expect(res).toBe("2.50");
```
sleidig, line 659: let's add a test with one entry `= 0` (I agree that missing and invalid should be ignored, but 0 should be included in calculation)

src/app/core/export/query.service.ts
sleidig: I would ignore missing / invalid. (that might help if we have a mix of entity types or badly maintained data)

sleidig (approved):
Looks good :-)

Issue #1918, opened 2023-07-05
Improve default CSV export


TheSlimvReal:
The default CSV export in the list pages (accessed through the three-dot menu at the top right of list views) is currently not very user friendly as it just shows all data with the property names. This then also includes fields like `_id` and `_rev` that users normally shouldn't see.

To make the default better I suggest the following:

- Only export properties that also have a `.label` in the given `Entity.schema`
- Show the labels as column headers
- Allow columns to have same header (e.g. multiple fields with the same label)
- (to be discussed) use the order in which the fields are present in the details page forms, or at least take order in which the property schemas are defined in the config

PR #2033, opened 2023-10-16
Improve default csv export

brajesh-lab:
closes: #1918 

### Visible/Frontend Changes

- [x] The column label is shown in the exported CSV file headers.

- [x] Added a new function  to add column labels in the CSV file.


src/app/core/entity-list/entity-list/entity-list.component.ts
```diff
@@ -178,6 +182,7 @@ export class EntityListComponent<T extends Entity>
     if (!this.allEntities) {
       // if no entities are passed as input, by default load all entities of the type
       await this.loadEntities();
+      this.exportFile()
```
sleidig, line 185: to avoid having to call this in multiple places - and for following the separation of concerns principle - I would suggest to move your `exportFile` method to the `ExportDataDirective` - or maybe even better to the DownloadService.

We have extracted the entityConstructur from the data objects in another place also, using something like `entityConstructor = records[0]?.getConstructor();`, so you could also get to the schema there.

sleidig (changes requested):
Functionality works well :100: 

A few remarks on fitting this into our software architecture in my comments.

src/app/core/export/download-service/download.service.spec.ts
```diff
@@ -54,7 +55,7 @@ describe("DownloadService", () => {
       { _id: "Test:2", other: 2 },
     ];
 
-    const csvExport = await service.createCsv(docs);
+    const csvExport = await service.createCsv(docs,exportConfig);
```
sleidig, line 58: @rudresh do you know why our eslint is not complaining about this without a space after the comma?

rudresh: @sleidig There wasn't a rule to check for space after comma. I've added a rule to check this in [this draft PR](https://github.com/Aam-Digital/ndb-core/pull/2039)

src/app/core/export/download-service/download.service.spec.ts
sleidig, line 1: looks like a test is still failing here:

> DownloadService should transform object properties to their label for export FAILED
#17 256.4 	Expected [ '' ] to have size 2.
#17 256.4 	    at <Jasmine>
#17 256.4 	    at UserContext.<anonymous> (src/app/core/export/download-service/download.service.spec.ts:113:18)

src/app/core/export/download-service/download.service.ts
```diff
@@ -90,17 +90,59 @@ export class DownloadService {
...
+
+    const result = this.exportFile(data, entityConstructor);
+    return result;
+  }
+
+  exportFile(data: any[], entityConstructor: { schema: any; }) {
+    const entitySchema = [entityConstructor.schema];
+    const columnLabel = {};
+
+    entitySchema[0].forEach((value: { value: string, label: string }, key: string) => {
```
sleidig, line 120:
```suggestion
    const entitySchema = entityConstructor.schema;
    const columnLabel = {};

    entitySchema.forEach((value: { value: string, label: string }, key: string) => {
```

A bit simpler - or was there a special reason for this?

src/app/core/export/download-service/download.service.spec.ts
```diff
@@ -99,15 +100,19 @@ describe("DownloadService", () => {
     testEntity.dateProperty = new Date(testDate);
     testEntity.boolProperty = true;
 
+    const objectToArray = [
+      testEntity.enumProperty.label,
+      testEntity.dateProperty.toISOString(),
+      testEntity.boolProperty.toString(),
+      ];
+      
     const csvExport = await service.createCsv([testEntity]);
```
sleidig, line 109: `TestEntity`'s fields do not have any labels here, so this is not really testing the new functionality.

Let's update above:
```suggestion
        @DatabaseEntity("TestEntity")
    class TestEntity extends Entity {
      @DatabaseField({ "label": "test enum" }) enumProperty: ConfigurableEnumValue;
      @DatabaseField({ "label": "test date" }) dateProperty: Date;
      @DatabaseField({ "label": "test boolean" }) boolProperty: boolean;
    }
```

Can you also add expectations (or a separate unit test) to check that the header indeed exports the field labels?

src/app/core/export/download-service/download.service.spec.ts
```diff
@@ -99,15 +100,19 @@ describe("DownloadService", () => {
...
     const rows = csvExport.split(DownloadService.SEPARATOR_ROW);
     expect(rows).toHaveSize(1 + 1); // includes 1 header line
-    const columnValues = rows[1].split(DownloadService.SEPARATOR_COL);
-    expect(columnValues).toHaveSize(3 + 1); // Properties + _id
-    expect(columnValues).toContain('"' + testEnumValue.label + '"');
-    expect(columnValues).toContain('"' + testDate + '"');
-    expect(columnValues).toContain('"true"');
+    expect(objectToArray).toContain(testEnumValue.label);
+    expect(objectToArray).toContain(new Date(testDate).toISOString());
+    expect(objectToArray).toContain('true');
```
sleidig, line 115: You are not testing the result of the `createCsv` method anymore now. objectToArray seems to only be something manually set up for the test?

src/app/core/export/download-service/download.service.ts
sleidig, line 120: I don't think this typing is correct. entitySchema is a `Map<string, EntitySchemaField>` and EntitySchemaField does not have a "value" property

sleidig (changes requested):

src/app/core/export/download-service/download.service.spec.ts
```diff
@@ -77,11 +77,12 @@ describe("DownloadService", () => {
       '"_id","_rev","propOne","propTwo"' +
       DownloadService.SEPARATOR_ROW +
       '"TestForCsvEntity:1","2","first","second"';
+    spyOn(service, "exportFile").and.returnValue(expected);
     const result = await service.createCsv([test]);
     expect(result).toEqual(expected);
   });
 
-  it("should transform object properties to their label for export", async () => {
+  it("should transform schema label to their label for export when label is defined in entity schema", async () => {
```
sleidig, line 85: although the test name says it, I do not see and "expect" check regarding the column headers of the exported csv (which would be the values in `row[0]`)

This test originally checked whether the value of a `ConfigurableEnumValue` (which is a dropdown options field) gets exported by its human-readable label rather than the option id. That is still an important test, and is still tested here.
But your new functionality is missing a real test case. Maybe it would be better to create another, new test for your part.

src/app/core/export/download-service/download.service.spec.ts
```diff
@@ -105,10 +106,27 @@ describe("DownloadService", () => {
...
-    expect(columnValues).toContain('"' + testEnumValue.label + '"');
-    expect(columnValues).toContain('"' + testDate + '"');
-    expect(columnValues).toContain('"true"');
+    expect(columnValues).toHaveSize[3 + 1]; // Properties + _id
+    expect(columnValues).toContain['"' + testEnumValue.label + '"'];
+    expect(columnValues).toContain['"' + testDate + '"'];
+    expect(columnValues).toContain['"true"'];
+  });
+
+  it("should transform object properties to their label for export when entity schema is not defined", async () => {
```
sleidig, line 115:
```suggestion
  it("should export property keys as column headers for export when entity schema is not defined", async () => {
```

src/app/core/export/download-service/download.service.ts
```diff
@@ -90,17 +91,59 @@ export class DownloadService {
...
+    const entitySchema = entityConstructor.schema;
+    const columnLabel = new Map<string, EntitySchemaField>();
+    
+    entitySchema.forEach((value: { label: string }, key: string) => {
+      if (value.label) columnLabel[key] = value.label;
+    });
+  
+    const exportEntities = data.map((item) => {
+      for (const key in item) {
+        if (!columnLabel.hasOwnProperty(key)) delete item[key];
```
sleidig, line 127: I think the `delete item[key]` creates the side effect of changing the original data array elements. It's probably better to work on a copy of the item instead.
this could otherwise affect the data that we work with within the app's list as well.

sleidig, line 127: don't you mean `columnLabel.has()` of the [Map](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/has) class instead?

src/app/core/export/download-service/download.service.ts
```diff
@@ -90,17 +91,59 @@ export class DownloadService {
...
+      columnKeys.forEach((key) => {
+        orderedItem.push(item[key]);
+      });
+      orderedData.push(orderedItem);
+    });
 
-    return this.papa.unparse(data, {
-      quotes: true,
-      header: true,
-      newline: DownloadService.SEPARATOR_ROW,
```
sleidig, line 102: why no config for `newline` and `quotes` anymore?

src/app/core/export/download-service/download.service.ts
sleidig: also I would rename the variable to `columnLabels` (plural) as it stores all labels.

sleidig (changes requested):

src/app/core/export/download-service/download.service.ts
brajesh-lab: I have updated it.

src/app/core/export/download-service/download.service.spec.ts
brajesh-lab: @sleidig I have added a new test case for check labels in CSV export

src/app/core/export/download-service/download.service.ts
brajesh-lab: @sleidig I am Specifying fields and data explicitly, So I am not using for `newline` and `quotes`

src/app/core/export/download-service/download.service.spec.ts
sleidig: thanks! This covers all the logic well now :+1: 

Just a few glitches in your test code - I am pushing a small commit to fix that.

src/app/core/export/download-service/download.service.ts
sleidig: ah, I see your point. However, quotes and newline are config to specify how the output / result looks and we should still add them additionally.

I am pushing a small fix for this.

src/app/core/export/download-service/download.service.ts
sleidig: thanks :slightly_smiling_face:

PR #2034, opened 2023-10-17
Timezone index fix

TheSlimvReal:
The events/notes on a specific date did not work with a timezone `UTC -0500` (negative offset) because `new Date("2022-01-01")` is then interpreted as `Fri Dec 31 2021 19:00:00 GMT-0500 (Eastern Standard Time)`. 
The moment adapter handles this better or `new Date(2022, 0, 1)` also works.

- [x] Tests are also executed in the `America/Detroit` (UTC -0500) timezone
- [x] Failing tests were fixed

### DIFF
--- a/src/app/core/basic-datatypes/date/date-range-filter/date-range-filter.component.spec.ts
+++ b/src/app/core/basic-datatypes/date/date-range-filter/date-range-filter.component.spec.ts
@@ -6,6 +6,7 @@ import { MatNativeDateModule } from "@angular/material/core";
 import { NoopAnimationsModule } from "@angular/platform-browser/animations";
 import { DateFilter } from "../../../filter/filters/filters";
 import { defaultDateFilters } from "./date-range-filter-panel/date-range-filter-panel.component";
+import moment from "moment";
 
 describe("DateRangeFilterComponent", () => {
   let component: DateRangeFilterComponent<any>;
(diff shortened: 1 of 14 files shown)


src/app/core/basic-datatypes/date/date-range-filter/date-range-filter.component.spec.ts
```diff
@@ -101,8 +102,8 @@ describe("DateRangeFilterComponent", () => {
 
   it("should set the correct date filter when changing the date range manually", () => {
     component.filterConfig = new DateFilter("test", "test", []);
-    component.fromDate = new Date("2021-10-28");
-    component.toDate = new Date("2024-02-12");
+    component.fromDate = moment("2021-10-28").toDate();
+    component.toDate = moment("2024-02-12").toDate();
```
sleidig, line 106: in practice , these are set using `new Date(selectedOptionDates[0])` only - will this potentially become an issue filtering things?

sleidig (approved):

src/app/core/basic-datatypes/date/date-range-filter/date-range-filter.component.spec.ts
sleidig: ... actual code is fixed as well, using moment

Issue #1872, opened 2023-05-04
add new option from enum popup window


sleidig:
As the dialog is offering to rename and delete options already, it would be helpful to also have the option to add completely new options not only from the dropdown field but also in its option edit dialog shown when clicking the tool icon.

-----
### Implementation details:
- use `npm run storybook` to test and improve the component "Core / Entities / Edit Properties / Enum Dropdown / Primary"

- Component / template to be extended: `ConfigureEnumPopupComponent` 
(add a button to "add new option" between the last existing field and the close button)
![image](https://github.com/Aam-Digital/ndb-core/assets/1682541/16a369e6-40fa-4d19-adef-3ea56d41a867)

- current logic to add a new value to the options in `EnumDropdownComponent`: move this to `ConfigurableEnumService` and reuse
https://github.com/Aam-Digital/ndb-core/blob/8db14d8c0ca4a9412f853700eebfc1bf814dfa44/src/app/core/configurable-enum/enum-dropdown/enum-dropdown.component.ts#L78-L83

TheSlimvReal:
Seems to be more necessary than I expected. A lot of people only find the rename option but don't realise that you have to enter something in order to add a new option.

PR #2040, opened 2023-10-18
create new dropdown option from popup config window

sadaf895:
closes: #1872

### Visible/Frontend Changes
- [x] 
- [ ] 

- [x] 
- [ ]


src/app/core/basic-datatypes/configurable-enum/configurable-enum.service.ts
```diff
@@ -47,4 +49,10 @@ export class ConfigurableEnumService {
     }
     return this.enums.get(entityId);
   }
+  private addNewOption(name: string) {
+    const option = { id: name, label: name };
+    this.enumEntity.values.push(option);
+    this.entityMapper.save(this.enumEntity);
+    return option;
+  }
```
TheSlimvReal, line 57: Its not necessary to add this function here. When the popup is closed, it automatically saves the whole entity enum entity. All that needs to be done is to add the new option to the enum.

TheSlimvReal:
I would like to push this feature as it comes up more often.

PR #2041, opened 2023-10-18
cascading delete

sleidig:
closes: #220 

- [x] unit tests for all scenarios
- [x] recursive delete logic for related entities
- [x] undo function for all affected entities
- [x] schema config for entity-ref as aggregate / composite
- [x] mark entities as a whole as "no personal data"
- [x] :question: manual review in some cases (and/or warn user in confirmation dialog that related entities will also be affected)
- [x] :question: discuss whether code of anonymize and delete should be split to reduce complexity
- [x] progress indicators while processing action + undo

cascading-delete


src/app/core/entity/model/entity.ts
```diff
@@ -168,6 +168,15 @@ export class Entity {
...
   }
 
+  /**
+   * whether this entity type can contain "personally identifiable information" (PII)
+   * and therefore should follow strict data protection requirements
+   * and offer a function to anonymize records.
+   *
+   * Default, we always take a "privacy-first" approach - assuming that the entity type can contain PII.
+   */
+  static hasPII: boolean = true;
```
sleidig, line 178: or should we invert this to `hasNoPII`?

I feel like hasPII is much more natural to read and understand. But as the default assumes hasPII = true ...

src/app/core/entity/schema/entity-schema-field.ts
```diff
@@ -66,6 +67,14 @@ export interface EntitySchemaField {
    */
   additional?: any;
 
+  /**
+   * (Optional) Assign any custom "extension" configuration you need for a specific datatype extension.
+   *
+   * You can pass any kind of value here to allow complex custom datytypes' transformations
+   * that are not part of the core datatypes and therefore not included in this core interface.
+   */
+  entityReferenceRole?: EntityReferenceRole;
```
TheSlimvReal, line 76: I am very confused by the code comment. Doesn't this just allow to set `composite` or `aggregate`? What are these transformations/extensions?

TheSlimvReal (changes requested):
Really nice job. I like how little places needed to be modified and also how little core code is necessary in general. Also a very extensive test suite makes pretty clear what happens. 
My comments are mostly around making thins a bit more precise in terms of naming. Happy to discuss this a bit more.

src/app/core/entity/model/entity.ts
sleidig: yeah, considering the range of different entity types, including more technical one, that sounds smart :+1: 

In the setup UI we can still be default make this checked - or force the user to actively set it to either checked or unchecked when setting up a new entity type.

src/app/core/entity/entity-remove.service.spec.ts
```diff
@@ -263,38 +276,292 @@ describe("EntityRemoveService", () => {
...
+    refComposite: string[];
+
+    static create(name: string, properties?: Partial<RelatedEntity>) {
+      return Object.assign(new RelatedEntity(), { name: name, ...properties });
+    }
+  }
 
   // for direct references (e.g. x.referencesToRetainAnonymized --> recursively calls anonymize on referenced entities)
   //    see EntityDatatype & EntityArrayDatatype for unit tests
+  // TODO: should we allow an additional option to delete a direct referenced entity completely during anonymization?
```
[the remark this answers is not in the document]
sleidig:
```suggestion
```

src/app/core/entity/entity-remove.service.spec.ts
```diff
@@ -263,38 +276,292 @@ describe("EntityRemoveService", () => {
...
+    onSnackbarAction.complete();
+    tick();
+
+    const actualEntitiesAfterUndo = // @ts-ignore
+      (service.entityMapper as MockEntityMapperService).getAllData();
+    expectEntitiesToMatch(actualEntitiesAfterUndo, [entity, ref1, ref2], true);
+  }));
+
+  // TODO: test Note with attendance (?)
+  // TODO: test for entity with two properties referencing the same entity type
```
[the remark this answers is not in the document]
sleidig: the way the search for affected entities is implemented, it might possibly be detected twice and then be edited twice - if badly implemented, this could lead to things being overwritten or conflicts to occur.

src/app/core/entity/entity-remove.service.ts
```diff
@@ -104,13 +104,135 @@ export class EntityRemoveService {
...
+        entity,
+        (e) => this.deleteEntity(e),
+        (e, refField, entity) =>
+          this.removeReferenceFromEntity(e, refField, entity),
+      );
+
+    const originalEntity = entity.copy();
+    await this.entityMapper.remove(entity);
+
+    return [originalEntity, ...affectedEntitiesBeforeAction];
```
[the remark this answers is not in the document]
sleidig: what I wanted to say with that is, "affected entities in their original state" (i.e. before any id refs have been removed, etc.), as this is for the undo functionality here.

src/app/core/entity/entity-remove.service.ts
```diff
@@ -104,13 +104,135 @@ export class EntityRemoveService {
...
+   * @param compositeAction
+   * @param aggregateAction
+   * @private
+   */
+  private async cascadeActionToRelatedEntities(
+    entity: Entity,
+    compositeAction: (
+      relatedEntity: Entity,
+      refField?: string,
+      entity?: Entity,
```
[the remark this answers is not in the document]
sleidig: yes, not yet used ...
But I want to support the same interface / signature for compositeAction and aggregateAction

src/app/core/entity/entity-remove.service.ts
```diff
@@ -134,18 +256,30 @@ export class EntityRemoveService {
...
 
+  /**
+   * The actual anonymize action without user interactions.
+   * @param entity
+   * @private
+   */
   private async anonymizeEntity(entity: Entity) {
+    if (!entity.getConstructor().hasPII) {
+      // entity types that are generally without PII by default retain all fields
+      // this should only be called through a cascade action anyway
```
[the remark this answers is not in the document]
sleidig: yes, I think that's why I implemented it here :sweat_smile:

src/app/core/entity/schema/entity-schema-field.ts
sleidig: thanks! :see_no_evil:   this seems to be a copy of the code comment for `additional` only

sleidig:
```suggestion
  /**
   * (Optional) If the dataType of this field references another entity,
   *    define the role of this relationship for the entity containing this field.
   * 
   * i.e. how "important" is the entity this field is referencing?
   * Does this the entity containing this field (not the referenced entity) still have meaning after the referenced entity has been deleted?
   * 
   * see options of the `EntityReferenceRole` type
   */
  entityReferenceRole?: EntityReferenceRole;
```

src/app/core/entity/entity-config.service.ts
```diff
@@ -78,13 +78,15 @@ export class EntityConfigService {
         ),
       );
     }
+    // TODO: shall we just assign all properties that are present in the config object?
```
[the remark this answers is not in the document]
sleidig: I'll keep this aside for now but once the next static attribute gets added here, maybe it's finally time. For now I do see some advantage of having a clear whitelist of available/allowed properties.

src/app/core/entity/entity-remove.service.spec.ts
```diff
@@ -263,38 +276,292 @@ describe("EntityRemoveService", () => {
...
+    await testAction(
+      // @ts-ignore calling private method to simulate this being called through cascade action
+      "anonymizeEntity",
+      entity,
+      [entity],
+      [AnonymizableEntity.create({ defaultField: "test" })],
+    );
+
+    // reset actual state
+    AnonymizableEntity.hasPII = false;
```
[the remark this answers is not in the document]
sleidig: you are right - actually this was supposed to revert back to the (previously default) hasPII = true.
Fixed now

src/app/core/entity/entity-remove.service.spec.ts
```diff
@@ -169,7 +173,7 @@ describe("EntityRemoveService", () => {
     // @ts-ignore
     service.entityMapper = entityMapper;
 
-    await service.anonymize(entity);
+    await service[action](entity);
```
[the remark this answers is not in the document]
sleidig: I had left this in the tests when we removed in from the actual service because I can reuse the whole test setup/function easily. I'll see if I can remove this here as well when I refactor the tests as you suggested.

src/app/core/entity/entity-remove.service.ts
```diff
@@ -104,13 +104,135 @@ export class EntityRemoveService {
...
+        for (const e of affectedEntities) {
+          if (
+            refType.entityType.schema.get(refField).entityReferenceRole ===
+              "composite" &&
+            asArray(e[refField]).length === 1
+          ) {
+            // is only composite
+            const furtherAffectedEntities = await compositeAction(e);
+            furtherAffectedEntities.forEach((e) =>
+              originalAffectedEntitiesForUndo.push(e.copy()),
```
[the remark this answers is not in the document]
sleidig: You are right. Nice, in-depth discovery! :monocle_face:

src/app/core/entity/entity-remove.service.spec.ts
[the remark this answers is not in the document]
sleidig: Thanks for the suggestion!

I split up the service and also the tests, doing a rather big refactoring to simplify code. Please have a look

src/app/core/entity/entity-actions/entity-anonymize.service.spec.ts
```diff
@@ -0,0 +1,293 @@
...
+import moment from "moment";
+import {
+  allEntities,
+  ENTITIES,
+  EntityWithRelations,
+  expectAllUnchangedExcept,
+} from "./entity-actions-test-utils";
+import { EntityAnonymizeService } from "./entity-anonymize.service";
+
+fdescribe("EntityAnonymizeService", () => {
```
sleidig, line 26: These tests get weird errors with a HttpResponse 404 although there shouldn't be any services making http requests ... 

I am a bit lost (and maybe becoming blind to my own code's errors). Can you help me out here, @TheSlimvReal :blush:

src/app/core/entity/entity-actions/entity-actions.service.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+  });
+
+  it("should re-save all affected entities and navigate back to entity on undo", fakeAsync(() => {
+    const anotherAffectedEntity = new Entity();
+    mockedEntityDeleteService.deleteEntity.and.resolveTo([
+      primaryEntity,
+      anotherAffectedEntity,
+    ]);
+
+    // Mock a snackbar where 'undo' is immediately pressed
```
TheSlimvReal, line 92: Is 'immediate' relevant here?

src/app/core/entity/entity-actions/entity-actions-test-utils.ts
TheSlimvReal, line 1: Its a bit confusing that the file is called `entity-actions-test-utils` but the test file for the `EntityActionsService` is not even using it.

src/app/core/entity/entity-actions/entity-anonymize.service.spec.ts
```diff
@@ -0,0 +1,293 @@
...
+  it("should cascade anonymize the 'composite'-type entity that references the entity user acts on", async () => {
+    await service.anonymizeEntity(ENTITIES.ReferencedAsComposite);
+
+    expectAnonymized(
+      [ENTITIES.ReferencedAsComposite, ENTITIES.ReferencingSingleComposite],
+      entityMapper,
+    );
+  });
+
+  it("should not cascade anonymize the 'composite'-type entity that still references additional other entities but ask user", async () => {
```
TheSlimvReal, line 254: I guess the ask user part is not yet implemented?

src/app/core/entity/entity-actions/entity-anonymize.service.spec.ts
TheSlimvReal: The 404 came from the `MockEntityMapper` if the `.get()` call fails. It happened because a entity was never saved and the testing function tries to load it.

I also fixed some other tests that did not correctly check for the `inactive` and `anonymized` flag but for the cascading ones it got a bit difficult and I was not sure whether you had a different idea in mind to get these running.

src/app/core/entity/entity-actions/entity-anonymize.service.spec.ts
```diff
@@ -0,0 +1,299 @@
...
+
+    @DatabaseField({ anonymize: "retain-anonymized", dataType: "entity-array" })
+    referencesToRetainAnonymized: string[];
+
+    static create(properties: Partial<AnonymizableEntity>) {
+      return Object.assign(new AnonymizableEntity(), properties);
+    }
+
+    static expectAnonymized(
+      entityId: string,
```
TheSlimvReal, line 79: Is there a reason why you don't just pass the entity over?

src/app/core/entity/entity-actions/entity-anonymize.service.spec.ts
sleidig: Thanks! :heart:

src/app/core/entity/entity-actions/entity-anonymize.service.spec.ts
sleidig: I want an implicit check that the actual entity is not just a return value but really stored in the database

src/app/core/entity/entity-actions/entity-actions-test-utils.ts
sleidig: then maybe `cascading-entity-test-utils`?

src/app/child-dev-project/schools/model/school.ts
```diff
@@ -10,6 +10,7 @@ export class School extends Entity {
   static label = $localize`:label for entity:School`;
   static labelPlural = $localize`:label (plural) for entity:Schools`;
   static color = "#9E9D24";
+  static override hasPII = true;
```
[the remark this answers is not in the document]
sleidig: fair point. I guess I was thinking of notes on contact persons or other details - but mostly that wouldn't be PII either.
```suggestion
  static override hasPII = false;
```

src/app/core/entity/entity-actions/entity-actions.service.spec.ts
sleidig: this is from the tests that had been present originally. Indeed, I would understand this also that it is not immediate but once "next" is fired.

Simplified/Corrected now.

sleidig:
Almost done now from my perspective.

One problem: After anonymization, the sub-record lists somehow don't update automatically (e.g. anonymized related notes are shown with full details until you re-open that anonymized child's details again).

In EntitySubrecord we ignore simple updates and only act if an update leads to the entity being newly added. Shouldn't this also trigger a replacement of the changed entity, @TheSlimvReal ?
```
 else if (
            type === "update" &&
            !this.records.find((rec) => rec.getId() === entity.getId())
          ) {
            this.addToTable(entity);
          }
```

TheSlimvReal:
> Almost done now from my perspective.
> 
> One problem: After anonymization, the sub-record lists somehow don't update automatically (e.g. anonymized related notes are shown with full details until you re-open that anonymized child's details again).
> 
> In EntitySubrecord we ignore simple updates and only act if an update leads to the entity being newly added. Shouldn't this also trigger a replacement of the changed entity, @TheSlimvReal ?
> 
> ```
>  else if (
>             type === "update" &&
>             !this.records.find((rec) => rec.getId() === entity.getId())
>           ) {
>             this.addToTable(entity);
>           }
> ```

@sleidig I thinks this was initially excluded, because most of the parent components (e.g. `NotesManager`, `ActivitiesOverview`...) also had some kinds of listening to entity updates. As it becomes very confusing how exactly the order of these executions are it would be great to get this all bundled into one function (optimally in the `EntitySubrecord`). That might be a bigger task though and I am not sure how easy it is.

src/app/core/entity/entity-actions/entity-actions.service.ts
```diff
@@ -56,7 +46,7 @@ export class EntityRemoveService {
 
     // Undo Action
     snackBarRef.onAction().subscribe(async () => {
-      await this.entityMapper.save(entity, true);
+      await this.entityMapper.saveAll(previousEntitiesForUndo, true);
```
TheSlimvReal, line 49: This doesn't seem to be working anymore

TheSlimvReal (changes requested):

TheSlimvReal (approved):
Seems to all work well now. Nice fixes.

src/app/core/basic-datatypes/date/display-date/display-date.component.html
```diff
@@ -0,0 +1,14 @@
+<ng-container
+  *ngIf="!(displayAsAnonymized || isPartiallyAnonymized); else anonymized"
```
TheSlimvReal, line 2: A bit annoying, that both of these properties are necessary. Every component that is also used in templates will have to implement both which is a bit against the idea that they can be used everywhere.

src/app/core/entity/entity-actions/entity-actions.service.ts
```diff
@@ -149,13 +147,10 @@ export class EntityActionsService {
...
-        $localize`:post-anonymize related PII warning dialog:
-        Some related records (e.g. notes) may still contain personal data in their text.
-        If such records have been linked only to this ${
+        $localize`:post-anonymize related PII warning dialog:Some related records (e.g. notes) may still contain personal data in their text. We have automatically anonymized all records that are linked to ONLY this ${
           entity.getConstructor().label
-        }, we have automatically also anonymized them for you.
-        However, there have been some records that are related to multiple other records.
-        These are still unchanged, so that you will not lose relevant data. Please review the linked records manually to ensure all sensitive information has been removed (e.g. by editing a note's text).`,
+        }.
+        However, there are some records that are linked to multiple records. We have not anonymized these, so that you will not lose relevant data. Please review the them manually to ensure all sensitive information is removed (e.g. by looking through the linked notes and editing a note's text).`,
```
TheSlimvReal, line 153: Why do these two cases have such a different warning text? As far as I understood, the scenario described here is also true for deleting entities where the linked ones are deleted if this entity was the only link.

src/app/core/entity-details/entity-details/entity-details.component.ts
```diff
@@ -91,12 +96,25 @@ export class EntityDetailsComponent implements EntityDetailsConfig, OnChanges {
     }
     if (changes.id) {
       this.loadEntity(this.id);
-      // `initPanels()` is already called inside `loadEntity()`
+      this.subscribeToEntityChanges();
     } else if (changes.panels) {
+      // `initPanels()` is already called inside `loadEntity()`
```
TheSlimvReal, line 101: Why did you move this? I think at this place, the comment is a bit out of place: `initPanels` is called in this if statement, so what is this comment actually saying?

TheSlimvReal (approved):
Good UI tweaks. Works very well.

I was wondering whether we should also write to the URL if inactive records are currently shown in the entity list. This makes it a little bit easier to navigate around.

src/app/core/basic-datatypes/date/display-date/display-date.component.html
sleidig: yes, agreed. I wanted to reuse date-component's template for how month-component is displaying. Unfortunately isPartiallyAnonymized doubling as an `@Input` just gets overwritten ...

src/app/core/entity-details/entity-details/entity-details.component.ts
sleidig: sorry, I messed this up during merge conflicts. Thanks for pointing it out!

sleidig, line 101:
```suggestion
      this.subscribeToEntityChanges();
      // `initPanels()` is already called inside `loadEntity()`
    } else if (changes.panels) {
```

src/app/core/entity/entity-actions/entity-actions.service.ts
sleidig: fair point. I pushed an updated to make the delete warning (which has been shorter) similar to the anonymize warning. There still is some flaw here - in case of delete, the user doesn't have the entity anymore to go through linked related records. But I don't have a solution for that ...

Issue #1362, opened 2022-07-12
Duplicate a record


sleidig:
*As a coordinator I want to duplicate a school or activity in order to easily set up records that are similar to each other.*

There may be a number of similar records - the same type of activities for different groups, multiple groups with almost/exactly the same list of participants, etc.
To make things easy, some of our actual users requested a way to copy/duplicate an entry to create a new one.

**Proposed Solutions & Alternatives**
- check/select a row in the entity list (how to do this in the UI?)
- show a context menu button (or use the three-dot-menu?) with an option to "Duplicate record"
- on clicking this: clone the record(s), add "Copy of" to their names, save them and show them in the list
and/or
- in the details screen of an entity in the three-dot-menu show an option to "Duplicate record"
- on clicking this: clone the record, add "Copy of" to the name, forward to the details view of the copy

PR #2042, opened 2023-10-19
Duplicate a record

brajesh-lab:
see #1362


src/app/core/duplicate-records/duplicate-records.service.ts
```diff
@@ -0,0 +1,44 @@
...
+    let data = []
+    const final =[]
+    const entityConstructor = this.entityTypes.get(schema);
+    const keys =[...entityConstructor.schema.keys()]
+    const filteredKeys = keys.filter(key => key !== '_id' && key !== '_rev');
+    originalData.map((item: { record: { [x: string]: any; }; })=> {
+      delete item.record["_rev"];
+      delete item.record["_id"];
+      const strData = JSON.stringify(item.record);
+      const jsonData = JSON.parse(strData)
```
sleidig, line 29: As briefly discussed during standup:
This might be the issue why the age is not showing (I think JSON.parse just keeps a date string as a string and can't create a Date object - let alone our patched version `DateWithAge`

`Entity` class has a "copy" method (although I am not sure it will handle every potential complex property correctly):
https://github.com/Aam-Digital/ndb-core/blob/e7147b01a2cf900bfe885c5ed9bb17e11f8387c1/src/app/core/entity/model/entity.ts#L342-L352

Another possible approach (I am not sure whether this is much more than necessary, however) could be to transform the entity into "database format" and back into "entity format". This transformation using the `EntitySchemaService` makes sure that all properties are initialized as the correct object instances etc., if necessary.
For example, we use that here to store SiteSettings in localStorage and then load and transform it back:
https://github.com/Aam-Digital/ndb-core/blob/e7147b01a2cf900bfe885c5ed9bb17e11f8387c1/src/app/core/site-settings/site-settings.service.ts#L101-L102

src/app/core/duplicate-records/duplicate-records.service.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+  let entitySchemaService: EntitySchemaService;
+  let database: Database; 
+  let sessionService: SessionService;
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      providers: [
+        DuplicateRecordsService,
+        EntityMapperService,
+        EntityRegistry,
```
sleidig, line 24:
```suggestion
  @DatabaseEntity("your_schema_name")
  class YourTestClass extends Entity {}

  beforeEach(() => {
    TestBed.configureTestingModule({
      imports: [CoreModule],
      providers: [
        DuplicateRecordsService,
        EntityMapperService,
        { provide: EntityRegistry, useValue: entityRegistry },
        ComponentRegistry,
```

This will make the test work:
- the automatic registration for any `@DatabaseEntity()` annotation happens in the global variable `entityRegistry`, so this has to be provided specifically ...
- to have the data types for DatabaseFields, the CoreModule registers a lot of the common ones. And that in turn requires ComponentRegistry also ...

An alternative would be to simply import our `MockedTestingModule` here, which brings all of those related things with it. But it also imports a lot of other things, making it more of an integration than a unit test then.

brajesh-lab:
@sleidig, Please take a look.

src/app/core/duplicate-records/duplicate-records.service.ts
```diff
@@ -0,0 +1,68 @@
...
+      duration: 3000, 
+      horizontalPosition: horizontalPosition,
+      verticalPosition: verticalPosition,
+    });
+  }
+
+  async duplicateRecord(sourceData: any, schemaName: string) {
+    if (!sourceData || sourceData.length === 0) {
+      const flashMessage = this.config.getConfig("flashMessage") as { message: string };
+      this.showFlashMessage(flashMessage.message);
```
sleidig, line 36: thanks for trying to make this configurable!

However, I think this message can be consistent across our clients and should always say the same thing. Translations are handled using `$localize` (see https://aam-digital.github.io/ndb-core/documentation/additional-documentation/how-to-guides/build-localizable-(translatable)-ui.html), so let's have the message hard-coded here directly instead of using config.

src/app/core/duplicate-records/duplicate-records.service.ts
```diff
@@ -0,0 +1,68 @@
...
+    private config: ConfigService
+  ) {}
+
+  showFlashMessage(message: string) {
+    const horizontalPosition: MatSnackBarHorizontalPosition = 'center';
+    const verticalPosition: MatSnackBarVerticalPosition = 'top';
+    this.snackBar.open(message, 'Close', {
+      duration: 3000, 
+      horizontalPosition: horizontalPosition,
+      verticalPosition: verticalPosition,
```
sleidig, line 29:
```suggestion
```

Let's use the default positioning, so that such messages are displayed consistently everywhere in the app.

src/app/core/duplicate-records/duplicate-records.service.ts
```diff
@@ -0,0 +1,68 @@
...
+    });
+  }
+
+  async duplicateRecord(sourceData: any, schemaName: string) {
+    if (!sourceData || sourceData.length === 0) {
```
sleidig, line 34: Should we hide the "duplicate" action from the menu when no items are selected?

src/app/core/duplicate-records/duplicate-records.service.ts
```diff
@@ -0,0 +1,68 @@
...
+    const horizontalPosition: MatSnackBarHorizontalPosition = 'center';
+    const verticalPosition: MatSnackBarVerticalPosition = 'top';
+    this.snackBar.open(message, 'Close', {
+      duration: 3000, 
+      horizontalPosition: horizontalPosition,
+      verticalPosition: verticalPosition,
+    });
+  }
+
+  async duplicateRecord(sourceData: any, schemaName: string) {
```
sleidig, line 33: can you find the exact type instead of using `any`?

src/app/core/duplicate-records/duplicate-records.service.ts
```diff
@@ -0,0 +1,68 @@
...
+    formattedData.forEach((item)=> {
+      const entity = new entityConstructor();
+      for (const key of keys) {
+        if (key === 'name' || key === 'title' || key === 'subject') {
+          item[key] = `Copy of ${item[key]}`;
+        }
+       entity[key] = item[key];
+      }
+      duplicatedData.push(entity);
+    })
```
sleidig, line 64: Any particular reason why you use two loops here, instead of doing this all in one?

src/app/core/duplicate-records/duplicate-records.service.spec.ts
```diff
@@ -0,0 +1,92 @@
...
+  it('should transform data correctly', () => { 
+    const duplicateTest = new DuplicateTestEntity();
+    duplicateTest.name = "TestName"
+    duplicateTest.boolProperty = true;
+    
+    const originalData = [
+      {
+        record: duplicateTest
+      }
+    ];
```
sleidig, line 64: wouldn't things be easier if we implement the service to just work with an array of entities (`[duplicateTest]`) instead of this object?

src/app/core/duplicate-records/duplicate-records.service.ts
```diff
@@ -0,0 +1,68 @@
...
+    if (!sourceData || sourceData.length === 0) {
+      const flashMessage = this.config.getConfig("flashMessage") as { message: string };
+      this.showFlashMessage(flashMessage.message);
+      return;
+    }
+    const duplicateData = this.transformData(sourceData, schemaName);
+    return await this.entitymapperservice.saveAll(duplicateData);
+  }
+
+  transformData(sourceData: any, schemaName: string): any {
```
sleidig, line 43: more descriptive name would be "duplicate" or "clone"?

sleidig (changes requested):
Looks pretty good already! :slightly_smiling_face: 
I left a few remarks on the code

src/app/core/duplicate-records/duplicate-records.service.ts
rudresh: @sleidig I think we should show the "duplicate" action all the time. That way the user knows that there is an option to duplicate the records. The duplicate action itself looks a bit hidden under the 3 dots. Conditionally showing them would make it a bit difficult for the users to figure out themselves. Let us know your thougts.

sleidig: That's a valid point! In fact, I was also thinking how to make the action more visible at least whenever something is selected ... an animation or some element that is visible without opening the menu, possibly.

For now, I think keeping the option always visible, as you suggest, is good. I would then disable the button when nothing is selected and when disabled display a matTooltip explaining to first select some rows.

rudresh: Yes sounds good. @brajesh-lab let's update it accordingly.

src/app/core/duplicate-records/duplicate-records.service.ts
brajesh-lab: @sleidig I have updated it with `clone`

src/app/core/duplicate-records/duplicate-records.service.spec.ts
brajesh-lab: @sleidig, Updated it.

src/app/core/duplicate-records/duplicate-records.service.ts
brajesh-lab: @sleidig, I have merged two loops into a single one

src/app/core/duplicate-records/duplicate-records.service.ts
brajesh-lab: @sleidig, I have changed with `Entity[]`

src/app/core/duplicate-records/duplicate-records.service.ts
brajesh-lab: @rudresh @sleidig  I have updated it.

src/app/core/duplicate-records/duplicate-records.service.ts
brajesh-lab: @sleidig, I have removed it.

src/app/core/duplicate-records/duplicates-data-directive/duplicate-records.directive.ts
```diff
@@ -0,0 +1,19 @@
+import { Directive, HostListener, Input } from "@angular/core";
+import { DuplicateRecordService } from '../duplicate-records.service';
+
+@Directive({
+  selector: '[appDuplicateRecords]',
+  standalone: true,
+})
+
+export class DuplicateRecordDirective {
```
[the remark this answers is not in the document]
brajesh-lab: @sleidig, Updated it.

src/app/core/common-components/entity-subrecord/entity-subrecord/entity-subrecord.component.ts
```diff
@@ -96,6 +98,8 @@ export class EntitySubrecordComponent<T extends Entity> implements OnChanges {
   @Input() isLoading: boolean;
 
   @Input() clickMode: "popup" | "navigate" | "none" = "popup";
+  
+  @Output() selectedData: EventEmitter<any> = new EventEmitter<any>
```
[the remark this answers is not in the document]
brajesh-lab: @sleidig, Updated it.

src/app/core/entity-list/entity-list/entity-list.component.html
```diff
@@ -201,6 +202,20 @@ <h2>{{ listName }}</h2>
     ></fa-icon>
     <span i18n> Import from file </span>
   </button>
+  
+  <button
+    mat-menu-item
+    (click)="duplicateRecords()"
+    [disabled]="selectedRows.length === 0"
+    [matTooltip]="selectedRows.length === 0 ? 'Please select rows' : ''" i18n-matTooltip
```
sleidig, line 210:
```suggestion
    matTooltip="Please select rows" [matTooltipDisabled]="selectedRows.length > 0" i18n-matTooltip
```

src/app/core/duplicate-records/duplicate-records.service.ts
sleidig: Looks great :+1:

src/app/core/duplicate-records/duplicate-records.service.ts
```diff
@@ -0,0 +1,43 @@
...
+  ) {}
+
+  async duplicateRecord(sourceData: Entity[], schemaName: string)  {
+    const duplicateData = this.clone(sourceData, schemaName);
+    return await this.entitymapperservice.saveAll(duplicateData);
+  }
+
+  clone(sourceData: Entity[], schemaName: string): any {
+    const duplicateData = [];
+    const entityConstructor = this.entityTypes.get(schemaName);
```
sleidig, line 26: You can get the entity constructor easily even without passing the schemaName as parameter, from the entity instances:

`sourceData[i].getConstructor()` returns the constructor with its schema of the current item. As we only have a few items being cloned usually, I think it won't hurt to move this inside the loop and simply use `item.getConstructor()`. We can even remove the schemaName parameter from the duplicateRecord method then.

src/app/core/common-components/entity-subrecord/entity-subrecord/entity-subrecord.component.ts
```diff
@@ -96,6 +98,8 @@ export class EntitySubrecordComponent<T extends Entity> implements OnChanges {
   @Input() isLoading: boolean;
 
   @Input() clickMode: "popup" | "navigate" | "none" = "popup";
+  
+  @Output() selectedRecords: EventEmitter<{ event: any, row: T[] }> = new EventEmitter<{ event: any, row: T[] }>
```
sleidig, line 102:
```suggestion
  /** outputs an event containing an array of currently selected records (checkmarked by the user) */
  @Output() selectedRecords: EventEmitter<T[]> = new EventEmitter<T[]>
```

I would like to simplify this interface: Let's simply output the array of all currently selected records, whenever the selection changes. I don't think any component outside of this will need access to individual select and unselect events.

That'll let you simplify the code in EntityList, but requires you to adapt selectRow() here

src/app/core/duplicate-records/duplicate-records.service.ts
```diff
@@ -0,0 +1,43 @@
...
+    const keys = [...entityConstructor.schema.keys()].filter(key => key !== '_id' && key !== '_rev');
+
+    sourceData.map((item: Entity)=> {
+      const dbEntity = this.entityService.transformEntityToDatabaseFormat(item);
+      const entityformat = this.entityService.transformDatabaseToEntityFormat(dbEntity, entityConstructor.schema);
+      const entity = new entityConstructor();
+      for (const key of keys) {
+        if (key === 'name' || key === 'title' || key === 'subject') {
+          entityformat[key] = `Copy of ${entityformat[key]}`;
+        }
```
sleidig, line 36: Nice and important to update the name!

To make this even more generic: Can we use entityConstructor.toStringAttributes to decide which property needs to be prefixed with "Copy of"? Maybe we can add it to the first item in the "toStringAttributes" array (except when that item is "entityId"). That's the properties that are used to display the entity's name

src/app/core/duplicate-records/duplicate-records.service.ts
brajesh-lab: @sleidig Updated it.

src/app/core/duplicate-records/duplicate-records.service.ts
brajesh-lab: @sleidig, Thanks 

updated it.

src/app/core/common-components/entity-subrecord/entity-subrecord/entity-subrecord.component.ts
brajesh-lab: Changed it.

src/app/core/duplicate-records/duplicate-records.service.ts
```diff
@@ -0,0 +1,44 @@
...
+    return await this.entitymapperservice.saveAll(duplicateData);
+  }
+
+  clone(sourceData: Entity[]): any {
+    const duplicateData = [];
+    const entityConstructor = sourceData[0].getConstructor();
+    const keys = [...entityConstructor.schema.keys()].filter(key => key !== '_id' && key !== '_rev');
+
+    sourceData.map((item: Entity)=> {
+      const dbEntity = this.entityService.transformEntityToDatabaseFormat(item);
```
sleidig, line 30:
```suggestion

    sourceData.map((item: Entity)=> {
      const entityConstructor = item.getConstructor();
      const keys = [...entityConstructor.schema.keys()].filter(key => key !== '_id' && key !== '_rev');
      const dbEntity = this.entityService.transformEntityToDatabaseFormat(item);
```

src/app/core/common-components/entity-subrecord/entity-subrecord/entity-subrecord.component.ts
```diff
@@ -96,6 +98,8 @@ export class EntitySubrecordComponent<T extends Entity> implements OnChanges {
   @Input() isLoading: boolean;
 
   @Input() clickMode: "popup" | "navigate" | "none" = "popup";
+  
+  @Output() selectedRecords: EventEmitter<T[] > = new EventEmitter<T[]>
```
sleidig, line 102:
```suggestion
  /** outputs an event containing an array of currently selected records (checkmarked by the user) */
  @Output() selectedRecords: EventEmitter<T[] > = new EventEmitter<T[]>
```

sleidig (approved):
I found and fixed one more functional problem: 

If you select some rows and then switch the "tab" (i.e. the section above the table, which changes the columns that the table displays), then the selection in the UI did not keep the checkboxes selected (although internally it remembers the state).

I think this is good to go now :+1:

PR #2050, opened 2023-10-25
refactor: remove redundant, duplicate entity-config constant

sleidig:


TheSlimvReal (approved):
Good find. Thanks for fixing.

PR #2056, opened 2023-10-26
Inactive records in entity select

### LINKED ISSUE #1983: Entity Selection should by default hide inactive records
**Describe the bug**
We used to have a filter for the autocomplete dropdown "entity-select", so that the records available to select only show if active (e.g. inactive archived children are not offered in the list when linking people to a new note). There was then a checkbox to extend the search to all, including inactive records.

This functionality seems to have disappeared or broken.

christophscheuing:
Closes #1983

Functionality already there, but still working on the tests.


christophscheuing:
Unfortunately making the message "None, but XX inactive found" itself clickable hasn't worked out. But UI-wise it should also be OK if this message is only a non-clickable hint and if the user has to click on the option with the checkbox in the entry below.

src/app/core/common-components/entity-select/entity-select.component.html
```diff
@@ -43,12 +43,23 @@
...
     </mat-option>
+    <mat-checkbox
+    *ngIf="getNumberOfInactive() > 0"
+    labelPosition="after"
+    (change)="toggleIncludeInactive()"
+    [checked]="includeInactive"
+    i18n="Label for checkbox|e.g. include inactive children"
+  >
+    Also show inactive {{ label }}
+  </mat-checkbox>
```
TheSlimvReal, line 63: Can we merge this and the `None, but ... inactive ...` option in to one? E.g. `Also show ... inactive`?

TheSlimvReal (changes requested):
The functionality already looks good. But I would have a little wish-list to make the component a little clearer

- [ ] I don't see a reason why the `filter` function every time applies the `additionalFilter` and the `isSelected` check. This rarely changes. Can we instead create a `relevantEntities` property that only holds the entities which pass these two checks, so the `filter` function only does the filtering based on the search term.
- [x] I don't like the multiple calls of the `getNumberOfInactive` function and also the complicated way to get this number. Let's rather have two arrays `filteredEntities` which holds the actually displayed entities (as it is currently) and `inactiveFilteredEntities` which only holds the inactive ones that match the filter. Both are calculated based on the `relevantEntities`, as described before, inside the `filter` function. This way we drastically remove the amount of times this has to be calculated and it also makes the code a little more straight forward.

Let me know if you need more details.

src/app/core/common-components/entity-select/entity-select.component.spec.ts
```diff
@@ -65,6 +72,10 @@ describe("EntitySelectComponent", () => {
     component.entityType = User.ENTITY_TYPE;
     tick();
     fixture.detectChanges();
+    console.log(
+      "should suggest ... filteredEntities",
+      component.filteredEntities,
+    );
```
sleidig, line 78:
```suggestion
```

sleidig: @christophscheuing can you remove the other debugging console outputs in the tests (or turn them into "expect()" statements?).

src/app/core/common-components/entity-select/entity-select.component.ts
```diff
@@ -89,8 +94,10 @@ export class EntitySelectComponent<E extends Entity> implements OnChanges {
         filter((isLoading) => !isLoading),
       )
       .subscribe((_) => {
-        this.selectedEntities = this.allEntities.filter((e) =>
-          sel.find((s) => s === e.getId(true) || s === e.getId()),
+        this.selectedEntities = sel.map((id) =>
+          this.allEntities.find(
+            (s) => id === s.getId(true) || id === s.getId(),
+          ),
```
sleidig, line 100: what's supposed to happen here if an item in "sel" is not available in allEntities? Should this really put an undefined value in the selectedEntities array?

src/app/core/common-components/entity-select/entity-select.component.ts
```diff
@@ -156,7 +163,10 @@ export class EntitySelectComponent<E extends Entity> implements OnChanges {
   inputPlaceholder = this.loadingPlaceholder;
 
   allEntities: E[] = [];
+  relevantEntities: E[] = [];
```
sleidig, line 166: can you add a code comment or an even clearer variable name here? in what sense are they "relevant"? what's the difference to being "filtered"?

src/app/core/common-components/entity-select/entity-select.component.html
```diff
@@ -43,12 +43,29 @@
...
     </mat-option>
+    <mat-option
+      *ngIf="
+        filteredEntities.length < 1 &&
+        !includeInactive &&
+        inactiveFilteredEntities.length > 0
+      "
+      [disabled]="true"
+      style="display: none"
+    ></mat-option>
```
sleidig, line 60: This doesn't seem to ever be displayed?

sleidig (changes requested):
@christophscheuing thanks for implementing this! Looks pretty good to me now. I had a few questions and small remarks still, however.

src/app/core/common-components/entity-select/entity-select.component.spec.ts
christophscheuing: I removed those debugging logs.

christophscheuing: I removed those logs.

src/app/core/common-components/entity-select/entity-select.component.html
christophscheuing: True, but it has to be there in order for the mat-checkbox to be displayed. I've added a comment for that in the code.

sleidig (approved):
Great :+1:

PR #2059, opened 2023-10-31
Export only displayed data

brajesh-lab:
closes: #XX

### Visible/Frontend Changes
- [x] 
- [ ] 

- [x] 
- [ ]


brajesh-lab:
@sleidig, This is ready for your review. Please take a look and suggest if changes are needed.

src/app/core/common-components/entity-subrecord/entity-subrecord/entity-subrecord.component.ts
```diff
@@ -283,6 +285,10 @@ export class EntitySubrecordComponent<T extends Entity> implements OnChanges {
     }
   }
 
+  setFilteredData( filterData: any) {
+    this.filteredData.emit(filterData);
+  }
```
sleidig, line 290: as a one-liner, we can inline this instead (i.e. remove this method and directly call filteredData.emit )

sleidig: or maybe instead we use this to map the filteredData so that it outputs an array of entities instead of the weird, wrapped objects that the datasource uses?

src/app/core/common-components/entity-subrecord/entity-subrecord/entity-subrecord.component.ts
```diff
@@ -94,7 +94,7 @@ export interface TableRow<T extends Entity> {
 })
 export class EntitySubrecordComponent<T extends Entity> implements OnChanges {
   @Input() isLoading: boolean;
-
+  @Output() filteredData: EventEmitter<any[]> = new EventEmitter<any[]>()
```
sleidig, line 97: what's the exact type of this? Is it `{ record: T}`? Or can we make it `T[]`?

sleidig (changes requested):
Seems to work well :+1: 
And nice how little code was necessary!

Just very small code remarks

src/app/core/entity-list/entity-list/entity-list.component.html
```diff
@@ -170,7 +171,7 @@ <h2>{{ listName }}</h2>
 
   <button
```
sleidig, line 172: let's give users two options as actions, to either export all or only filtered data: "Download all data (.csv)" and "Download current (.csv)" (open to better wording, if anyone has suggestions).

brajesh-lab: @sleidig, I have added one more button to `Download Current (.csv)`

src/app/core/common-components/entity-subrecord/entity-subrecord/entity-subrecord.component.ts
brajesh-lab: @sleidig, The Type is  `{ record: T}`

src/app/core/common-components/entity-subrecord/entity-subrecord/entity-subrecord.component.ts
brajesh-lab: @sleidig, I have removed the function and calling directly `filteredData.emit`

src/app/core/entity-list/entity-list/entity-list.component.ts
```diff
@@ -301,4 +306,11 @@ export class EntityListComponent<T extends Entity>
     }
     this.addNewClick.emit();
   }
+
+  getFilteredData(filterData: any[]) {
+    this.filteredData = [];
+    filterData.map((item) => {
+      this.filteredData.push(item.record);
+    });
+  }
```
sleidig, line 315: I suggest we do this mapping within entity-subrecord-component. I don't see any value of that format having `{ record: Entity }` outside of that component. It will make the interfaces easier to understand and use.

brajesh-lab: @sleidig, Updated.

sleidig: I was talking about something a bit different.

Implemented it myself now. Do have a look at my last commit to understand what I was trying to say :blush: , @brajesh-lab

sleidig (approved):

Issue #1762, opened 2023-03-10
Click on Entity in disabled form should open details page


TheSlimvReal:
When having a reference to an entity in a form and the form is disabled (e.g. visiting child details) then the form element should be clickable and forward to the details page of the linked entity. 

This should be implemented for both `EditSingleEntity` and `EditEntityArray`.

sleidig:
Related to this: For better UX the EditSingleEntity should, in my opinion, have the same style as the multi-select, EditEntityArray (i.e. both showing the rounded block to indicate this is more than just free text).

PR #2061, opened 2023-11-04
#1762 Updated UI for singleEntities and entityArray

brajesh-lab:
closes: #1762 
### Visible/Frontend Changes
- [x] Clicking on entity in disabled form opens the details page


brajesh-lab:
@sleidig, This is ready for your review. Please take a look and suggest if changes are needed.

sleidig (changes requested):
Nice!

Some suggestions for fine-tuning the UX:

currently cursor suggests the whole chip is clickable, but navigation only happens when clicking the text:
image

The design (chips with background) also applies to our "configurable-enum" (simple dropdown options) fields now. That makes it hard for users to distinguish these (and understand why they can click some and not others). entity chips should be visually distinguishable. Maybe the other (the configurable-enum options) could be shown with only an outlined chip instead of a chip with grey background?

brajesh-lab:
@

> Nice!
> 
> Some suggestions for fine-tuning the UX:
> 
> currently cursor suggests the whole chip is clickable, but navigation only happens when clicking the text: image
> 
> The design (chips with background) also applies to our "configurable-enum" (simple dropdown options) fields now. That makes it hard for users to distinguish these (and understand why they can click some and not others). entity chips should be visually distinguishable. Maybe the other (the configurable-enum options) could be shown with only an outlined chip instead of a chip with grey background?





@sleidig, Is this looking better?
image

sleidig:
Nice, @brajesh-lab . I like the visual distinction now.

src/app/core/common-components/entity-select/entity-select.component.ts
```diff
@@ -138,13 +138,16 @@ export class EntitySelectComponent<E extends Entity> implements OnChanges {
...
-      this.formControl.disable();
-    } else {
-      this.formControl.enable();
-    }
+    this._disabled = disabled;
+  }
+  
+  get isDisabled(): boolean {
+    return this._disabled;
   }
```
sleidig, line 149: Did this cause any problems for you with using `formControl.disabled` in the previous implementation?

To be honest, I am not sure why we took the complex path of disabling the formControl there, but this may well have had a good reason.

sleidig:
The Note Details (if you click to blue bubble button in the bottom right) also still has issues:
image

src/app/core/common-components/entity-select/entity-select.component.ts
brajesh-lab: @sleidig, When `fromControl` was `disabled`, I could not make the mat-chip clickable. The `disabled` state prevented an action to be performed on the form. 

Now I am setting a boolean flag on the `Edit` button click event and checking that for enabling or disabling the input fields.

sleidig: We are including this component in complex forms, where the disabled state may indeed be set from the parent form. Therefore should not change this here to de-couple.

I remember a workaround for the clicks not firing on disabled components, that we have used for a similar problem in the EditFileComponent: setting `pointer-events: none` for the disabled input let's the clicks fall through to the parent component (see https://github.com/Aam-Digital/ndb-core/blob/2e39cec3dbc45ce8a34126080d453fe29d9a15ef/src/app/features/file/edit-file/edit-file.component.scss).
Maybe we can use this here instead.

sleidig (changes requested):
The UI looks pretty good now!

There is some bug with forms opening in popups, however. The values of dropdown fields are not displayed there anymore:
e.g. go to https://pr-2061.aam-digital.net/child/1?tabIndex=1 and click on one of the listed "School History" entries - the school field in the popup form remains empty.

brajesh-lab:
@sleidig, Summarising the work I have done so far on this issue:

On issue [#1762](https://github.com/Aam-Digital/ndb-core/issues/1762), the form control is disabled. When I removed the form control properties and hid the input field, it worked properly and navigated to the details page. However, this may impact other areas of the app.

When adding a CSS property and applying it to `mat-form-field`, the function worked on all form fields(input fields), and I couldn't click on a particular `chip `to navigate to the details page.

I also tried using `span` and `div,` but it only worked at the top of the `mat-form-field`, making it impossible to pass a particular entity to navigate to the details page.

PR #2062, opened 2023-11-06
refactor(core): implement RoutedViewComponent 

sleidig:
to generalize loading of config from route data

as a basis for the new admin / setup UI

:warning: mark as codo


src/app/child-dev-project/attendance/add-day-attendance/add-day-attendance.component.ts
```diff
@@ -38,8 +36,8 @@ export interface AddDayAttendanceConfig {
   ],
   standalone: true,
 })
-export class AddDayAttendanceComponent {
-  config?: AddDayAttendanceConfig;
+export class AddDayAttendanceComponent implements AddDayAttendanceConfig {
+  @Input() sortParticipantsBy: any;
```
TheSlimvReal, line 40:
```suggestion
  @Input() sortParticipantsBy: string;
```
Also the interface is not really useful as it only provides this field which is overwritten here and its only used once.

src/app/child-dev-project/children/children-list/children-list.component.ts
```diff
@@ -34,6 +34,8 @@ export class ChildrenListComponent implements OnInit {
 
   async ngOnInit() {
     this.route.data.subscribe(
+      // TODO replace this use of route and rely on the RoutedViewComponent instead
+      // see  that flattens the config option, assigning individual properties as inputs however, so we can't easily pass on
```
TheSlimvReal, line 38: I guess it will be quite some work to refactor this in the entity list component and then pass all the properties individually.

src/app/core/entity-details/entity-details/entity-details.component.ts
```diff
@@ -60,77 +59,79 @@ import { EntityArchivedInfoComponent } from "../entity-archived-info/entity-arch
...
+  }
+  @Input() entityType: string;
   entityConstructor: EntityConstructor;
 
+  @Input() id: string;
+  record: Entity;
+
+  @Input() panels: Panel[] = [];
+  /** the actual, fully resolved panel configs */
+  panelsComponents: Panel[] = [];
```
TheSlimvReal, line 78:
```suggestion
  panelComponents: Panel[] = [];
```

TheSlimvReal: And why does this now need to be in a different property? I think nobody will understand what the differences between these two is.

src/app/core/ui/routed-view/routed-view.component.html
TheSlimvReal, line 1: Delete?

src/app/core/ui/routed-view/routed-view.component.ts
```diff
@@ -0,0 +1,47 @@
...
+      // merge updated config properties from route params
+      route.paramMap.subscribe((params) => {
+        const config = this.config;
+        for (const key of params.keys) {
+          config[key] = params.get(key);
+        }
+        this.config = { ...config };
+      });
+
+      // TODO: should we also assigned queryParams from the url to input properties?
```
TheSlimvReal, line 44: I am not sure whether this could result in a problem at one point. There are no restrictions as to what is written into the params so it would be possible to overwrite other config options from there by using the same key. Don't you think the query params object on the route object is already easy enough to use?

TheSlimvReal (approved):
Really good idea and nice to see how easy it was to implement this. This makes the difference between route components and panel components even smaller. That might be useful at one point.

sleidig:
> Can you explain a little why this is needed in general? What will change with the admin UI?

In general, I've had the thought to make components usable in both dialog popups and main full-screen views. (Although here I realized that making the Dialog-based components generic is still some more work, so skipping this here for now)
In particular, with regards to the Admin UI, I wanted to keep the components to configure views and entities independent of router data because we might want to display these inside some kind of stepper for a guided setup process also.

src/app/child-dev-project/children/children-list/children-list.component.ts
sleidig: I did start refactoring it but then realized it is quite complex. Also, we might prefer to move away from this ChildrenListComponent and use the EntityListComponent only (once we have those last few special bits generalized), so I didn't want to invest too much time for this now.

src/app/core/ui/routed-view/routed-view.component.html
sleidig: good point! Thanks.

src/app/core/entity-details/entity-details/entity-details.component.ts
sleidig: I started writing a long comment here, explaining why this was the best I could come up with ... and then realized that apparently my other refactorings made this unnecessary now. :sweat_smile: 

Good find. Removed now.

src/app/core/ui/routed-view/routed-view.component.ts
sleidig: probably true. I also had some concerns about potential security issues.

PR #2063, opened 2023-11-06
test(e2e): disable regularly failing e2e test

sleidig:
the whole app bootstrap including generation of indices for attendance view may just be too slow and extensive to make this work reliably. For now, we need to get a reliable CI pipeline again, so I would disable this e2e test (we anyways don't have good e2e coverage)


TheSlimvReal:
The test is actually failing because the demo data for activities assigned to the demo user are different each day the test is run. Sometime you need the "show more" button, sometimes not.

sleidig:
Ah, thanks for clarifying the root cause!

We might want to check if we can provide a seed for the randomization to have that consistent across test runs. But for now, let's put the e2e optimizations aside.

TheSlimvReal (approved):

PR #2064, opened 2023-11-08
fix: entity list works with new input configs

TheSlimvReal:
closes: #XX

### Visible/Frontend Changes
- [x] 
- [ ] 

- [x] 
- [ ]

### DIFF
--- a/src/app/core/entity-list/entity-list/entity-list.component.ts
+++ b/src/app/core/entity-list/entity-list/entity-list.component.ts
@@ -21,7 +21,6 @@ import { EntitySubrecordComponent } from "../../common-components/entity-subreco
 import { entityFilterPredicate } from "../../filter/filter-generator/filter-predicate";
 import { AnalyticsService } from "../../analytics/analytics.service";
 import { RouteTarget } from "../../../app.routing";
-import { RouteData } from "../../config/dynamic-routing/view-config.interface";
 import { EntityMapperService } from "../../entity/entity-mapper/entity-mapper.service";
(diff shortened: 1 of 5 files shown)


src/app/core/entity-list/entity-list/entity-list.component.ts
```diff
@@ -108,12 +112,13 @@ export class EntityListComponent<T extends Entity>
 
   isDesktop: boolean;
 
-  listName = "";
-  columns: (FormFieldConfig | string)[] = [];
-  columnGroups: GroupConfig[] = [];
+  @Input() title = "";
+  @Input() columns: (FormFieldConfig | string)[] = [];
+  @Input() columnGroups: ColumnGroupsConfig;
+  groups: GroupConfig[] = [];
```
TheSlimvReal, line 118: I am not too happy with this naming change but I did not want to require a migration

sleidig: maybe we can collect (or then do a detailed review of) all config namings as part of the Admin UI work and then prepare one large, automatic migration to clean all of this up and make it fully consistent across different components.

sleidig (approved):
Seems to work well now.

Thanks a lot for fixing this so quickly!

src/app/core/entity-list/entity-list/entity-list.component.ts
TheSlimvReal: Lets collect them here https://github.com/Aam-Digital/ndb-core/issues/2066

PR #2065, opened 2023-11-08
fix: no more changed-after-checked error from filteredRecords binding

sleidig:


TheSlimvReal (approved):
Good find for the solution

PR #2068, opened 2023-11-08
fix(core): support "required" indicator for custom form-controls

sleidig:
until now, custom form-controls (like our basic-autocomplete) did not display a * as required indicator in the mat-label. This is now supported.

Code taken from https://material.angular.io/guide/creating-a-custom-form-field-control#required

-----
:warning: mark as codo


src/app/core/common-components/basic-autocomplete/custom-form-control.directive.ts
```diff
@@ -139,5 +149,15 @@ export abstract class CustomFormControlDirective<T>
...
     }
+
+    if (control) {
+      if (control.hasValidator(Validators.required)) {
+        this.required = true;
+        this.stateChanges.next();
+      } else if (this.required) {
+        this.required = false;
+        this.stateChanges.next();
+      }
```
TheSlimvReal, line 160:
```suggestion
      if (this.required !== coerceBooleanProperty(control.hasValidator(Validators.required))) {
        this.required = control.hasValidator(Validators.required);
      }
```
The state change notification already happens in the setter and this makes is a bit clearer, when it is set to required.  Currently the required is always updated even if the value was always true.

sleidig: nice!

PR #2071, opened 2023-11-10
fix: panels are only initialized once

TheSlimvReal:
Go to Children List and click add new. Console shows deprecation warning.

### DIFF
--- a/src/app/core/entity-details/entity-details/entity-details.component.ts
+++ b/src/app/core/entity-details/entity-details/entity-details.component.ts
@@ -91,8 +91,7 @@ export class EntityDetailsComponent implements EntityDetailsConfig, OnChanges {
     }
     if (changes.id) {
       this.loadEntity(this.id);
-    }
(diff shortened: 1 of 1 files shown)


src/app/core/entity-details/entity-details/entity-details.component.ts
```diff
@@ -91,8 +91,7 @@ export class EntityDetailsComponent implements EntityDetailsConfig, OnChanges {
     }
     if (changes.id) {
       this.loadEntity(this.id);
-    }
-    if (changes.panels) {
+    } else if (changes.panels) {
```
sleidig, line 94:
```suggestion
    }
    if (changes.panels && !changes.id) {
      // on "id" changes, panels are automatically re-initialized already
```
as we are mostly using independent if-clauses, this might be more consistent and make the whole interdependence clear in the single line?

sleidig (approved):
Thanks for discovering this!

PR #2076, opened 2023-11-14
Correctly marking edited state of controls

TheSlimvReal:
Some controls where not correctly picked up as edited. This resulted in the app not detecting, that a form should be saved before navigating away and therefore data could be lost.

I have went through our custom controls and fixed this edited state detections. Some general observations:

- `markAsDirty` needs to come **before** `setValue`
- both `markAsDirty` alone is not enough. `setValue` also needs to be called on the control. Even if the same value as before is set


sleidig (approved):
Great find and analysis. Thanks a lot! :smiley:

Issue #2081, opened 2023-11-17
Broken `EntityListComponent`


TheSlimvReal:
**Describe the bug**
When the config for `EntityListComponent` only holds properties in the `columnGroups` properties, but they are not mentioned in the `columns` list, then the component breaks. That is a regression and worked before. The `columns` should be automatically filled with missing properties from the `columnGroups`.

**Additional context**
This has probably been introduced with #2062

PR #2085, opened 2023-11-18
fix: new columns correctly trigger input update

TheSlimvReal:
closes: #2081


sleidig (approved):

sleidig (changes requested):
When running locally, this causes "duplicate column definition" errors on my machine. I'll look into it further ...

PR #2086, opened 2023-11-19
Bulk actions

sleidig:
better UX for bulk actions. Don't always show select checkboxes in lists


TheSlimvReal (approved):
Thats a good solution. Thanks

PR #2089, opened 2023-11-20
fix(dashboard): attendance weeks only start from Monday correctly

sleidig:
Somehow I am too stupid to get this right. This is the second try after initially tests had failed when running on Sundays ...
Please double-check, @TheSlimvReal

### DIFF
--- a/src/app/core/common-components/entity-form/entity-form.service.spec.ts
+++ b/src/app/core/common-components/entity-form/entity-form.service.spec.ts
@@ -173,7 +173,7 @@ describe("EntityFormService", () => {
 
     schema.defaultValue = PLACEHOLDERS.NOW;
     form = service.createFormGroup([{ id: "test" }], new Entity());
-    expect(form.get("test")).toHaveValue(new Date());
+    expect(form.get("test").value).toEqual(new Date());
 
     schema.defaultValue = PLACEHOLDERS.CURRENT_USER;
     form = service.createFormGroup([{ id: "test" }], new Entity());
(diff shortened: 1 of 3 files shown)


TheSlimvReal:
As far as I understand the problem, its just a different start of the week day for moment and native date: Moment uses Monday as first day so on Sunday the Monday this week is the previous Monday. In native this would be the following day so the next Monday. So the solution to just use the same calculation method for both works but it kind of makes the test unnecessary as its the same code on both sides.

sleidig:
True, I also saw the parallel of test and implementation. I just want Sunday to be considered the last day of the current week (as is common on almost every "normal" country :sweat_smile: ). For tests maybe a few fixed sample times would be more useful than calculating the expected relative date on the fly.

src/app/core/common-components/entity-form/entity-form.service.spec.ts
```diff
@@ -173,7 +173,7 @@ describe("EntityFormService", () => {
 
     schema.defaultValue = PLACEHOLDERS.NOW;
     form = service.createFormGroup([{ id: "test" }], new Entity());
-    expect(form.get("test")).toHaveValue(new Date());
+    expect(form.get("test").value).toEqual(new Date());
```
TheSlimvReal, line 176: Whats the difference here?

sleidig: the test sometimes failed although the date was exactly the same - I guess if MockDate was used before, it's not reference identical anymore?

TheSlimvReal (approved):

PR #2091, opened 2023-11-21
fix(core): correctly init columns in subrecord component

sleidig:
for special subrecord tables like AserComponent


TheSlimvReal:
Can you explain what this fixes?

sleidig:
> Can you explain what this fixes?

@TheSlimvReal see the current master's "Child Details > Education" for example: Aser and EducationHistory do not show any columns (because of multiple change cycles the system initially adds only the "actions" column to columnsToDisplay and the bad implementation did not re-update if some columnsToDisplay already exist - because at other times we give an explicit list as `@Input` for columnsToDisplay also.

TheSlimvReal (approved):
Thanks for fixing.

I think that would have actually been discovered by our e2e tests 😅

PR #2095, opened 2023-11-22
fix: deleted rules object does not throw error

TheSlimvReal:
closes: #XX

### Visible/Frontend Changes
- [x] 
- [ ] 

- [x] 
- [ ]

### DIFF
--- a/src/app/core/permissions/ability/ability.service.ts
+++ b/src/app/core/permissions/ability/ability.service.ts
@@ -42,7 +42,11 @@ export class AbilityService extends LatestEntityLoader<Config<DatabaseRules>> {
   }
 
   private updateAbilityWithUserRules(rules: DatabaseRules): Promise<any> {
-    const userRules = this.getRulesForUser(rules);
+    // If rules object is empty, everything is allowed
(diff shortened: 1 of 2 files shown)


src/app/core/permissions/ability/ability.service.ts
```diff
@@ -42,15 +42,19 @@ export class AbilityService extends LatestEntityLoader<Config<DatabaseRules>> {
   }
 
   private updateAbilityWithUserRules(rules: DatabaseRules): Promise<any> {
-    const userRules = this.getRulesForUser(rules);
+    // If rules object is empty, everything is allowed
+    const userRules: DatabaseRule[] = rules
+      ? this.getRulesForUser(rules)
+      : [{ action: "manage", subject: "all" }];
```
sleidig, line 48: do we still need the separate init with all permissions then? (line 37)

sleidig (approved):

src/app/core/permissions/ability/ability.service.ts
TheSlimvReal: Yes, in case no rules were found at all this is relevant

PR #2097, opened 2023-11-23
fix: clicking filename does not open file twice

TheSlimvReal:
When a file has been uploaded and the file name is clicked while in edit mode, then the file is opened twice. This is because the event is caught multiple times.


sleidig:
I actually can't reproduce this issue in v3.26.0, am I missing something? Or would this only happen in non-mock mode for some reason?

sleidig (approved):
I'd love not having to use so much special code (css "pointer-events: none", and now the ".stopPropagation()", ...). But can't think of a way to do this smarter. Let's go with this and maybe make it reusable later for issues like #2061

Issue #2049, opened 2023-10-25
Setup Wizard: Step-wise onboarding / setup process


sleidig:
Upon first login the app offers a guided setup assistant. The process leads the user through all required steps to do an initial configuration of a system and explains the context as well as tips and tricks to non-technical users.

- Welcome + Overview
- Tips to create a setup that is easy to use later
- Configuration of data fields
- Configuration of system settings
- Tips for next steps towards usage

Improvements & TODOs:
- [x] save progress
  - [x] re-open at the last used stage
  - [x] display for each user (?) until dismissed by user
- [x] configurable step descriptions w markdown?

PR #2101, opened 2023-11-24
refactor(core): change FormConfig structure to use .fieldGroups instead of .cols and .headers

sleidig:
To ease the development of the setup wizard (PR #2049 ) this refactors config structure including an on-the-fly migration.

TODO:
- [ ] adapt tests
- [ ] adapt config-fix to new format
- [ ] refactor EntityFormComponent to also use new format

-----
:warning: mark as codo

### DIFF
--- a/src/app/child-dev-project/notes/note-details/note-details.component.ts
+++ b/src/app/child-dev-project/notes/note-details/note-details.component.ts
@@ -27,6 +27,7 @@ import { MAT_DIALOG_DATA, MatDialogModule } from "@angular/material/dialog";
 import { DialogButtonsComponent } from "../../../core/form-dialog/dialog-buttons/dialog-buttons.component";
 import { DialogCloseComponent } from "../../../core/common-components/dialog-close/dialog-close.component";
 import { EntityArchivedInfoComponent } from "../../../core/entity-details/entity-archived-info/entity-archived-info.component";
+import { FormFieldComponent } from "../../../core/common-components/entity-form/form-field/form-field.component";
 
 /**
  * Component responsible for displaying the Note creation/view window
@@ -48,6 +49,7 @@ import { EntityArchivedInfoComponent } from "../../../core/entity-details/entity
     MatMenuModule,
     DialogCloseComponent,
     EntityArchivedInfoComponent,
+    FormFieldComponent,
   ],
   standalone: true,
   encapsulation: ViewEncapsulation.None,
--- a/src/app/core/common-components/entity-form/entity-form.service.ts
+++ b/src/app/core/common-components/entity-form/entity-form.service.ts
@@ -18,6 +18,10 @@ import {
   PLACEHOLDERS,
 } from "../../entity/schema/entity-schema-field";
 import { isArrayDataType } from "../../basic-datatypes/datatype-utils";
+import {
+  ColumnConfig,
+  toFormFieldConfig,
+} from "../entity-subrecord/entity-subrecord/entity-subrecord-config";
 
 /**
  * These are utility types that allow to define the type of `FormGroup` the way it is returned by `EntityFormService.create`
@@ -60,27 +64,32 @@ export class EntityFormService {
    * @param forTable
    */
   public extendFormFieldConfig(
-    formFields: FormFieldConfig[],
+    formField: ColumnConfig,
     entityType: EntityConstructor,
     forTable = false,
-  ) {
-    formFields.forEach((formField) => {
-      try {
-        this.addFormFields(formField, entityType, forTable);
-      } catch (err) {
-        throw new Error(
-          `Could not create form config for ${formField.id}: ${err}`,
-        );
-      }
-    });
+  ): FormFieldConfig {
+    const fullField: FormFieldConfig = toFormFieldConfig(formField);
+
+    try {
+      this.addSchemaToFormField(
+        fullField,
+        entityType.schema.get(fullField.id),
+        forTable,
+      );
+    } catch (err) {
+      throw new Error(
+        `Could not create form config for ${fullField.id}: ${err}`,
+      );
+    }
+
+    return fullField;
   }
 
-  private addFormFields(
+  private addSchemaToFormField(
     formField: FormFieldConfig,
-    entityType: EntityConstructor,
+    propertySchema: EntitySchemaField,
     forTable: boolean,
   ) {
-    const propertySchema = entityType.schema.get(formField.id);
     formField.edit =
       formField.edit ||
       this.entitySchemaService.getComponent(propertySchema, "edit");
@@ -101,6 +110,8 @@ export class EntityFormService {
     if (propertySchema?.validators) {
       formField.validators = propertySchema?.validators;
     }
+
+    return formField;
   }
 
   /**
@@ -115,7 +126,9 @@ export class EntityFormService {
     entity: T,
     forTable = false,
   ): EntityForm<T> {
-    this.extendFormFieldConfig(formFields, entity.getConstructor(), forTable);
+    formFields = formFields.map((f) =>
+      this.extendFormFieldConfig(f, entity.getConstructor(), forTable),
+    );
     const formConfig = {};
     const entitySchema = entity.getSchema();
     const copy = entity.copy();
--- a/src/app/core/common-components/entity-form/form-field/form-field.component.html
+++ b/src/app/core/common-components/entity-form/form-field/form-field.component.html
@@ -1,0 +1,17 @@
+<div *ngIf="_field">
+  <div [class.field-with-tooltip-suffix]="_field.tooltip">
+    <ng-container
+      [appDynamicComponent]="{
+        component: _field.edit,
+        config: {
+          formFieldConfig: _field,
+          propertySchema: entity.getSchema().get(_field.id),
+          formControl: form.get(_field.id),
+          entity: entity
+        }
+      }"
+    >
+    </ng-container>
+  </div>
+  <app-help-button [text]="_field.tooltip"></app-help-button>
+</div>
--- a/src/app/core/common-components/entity-form/form-field/form-field.component.ts
+++ b/src/app/core/common-components/entity-form/form-field/form-field.component.ts
@@ -1,0 +1,50 @@
+import { Component, Input, OnChanges, SimpleChanges } from "@angular/core";
+import { DynamicComponentDirective } from "../../../config/dynamic-components/dynamic-component.directive";
+import { HelpButtonComponent } from "../../help-button/help-button.component";
+import { Entity } from "../../../entity/model/entity";
+import { EntityForm, EntityFormService } from "../entity-form.service";
+import { ColumnConfig } from "../../entity-subrecord/entity-subrecord/entity-subrecord-config";
+import { FormFieldConfig } from "../entity-form/FormConfig";
+import { NgIf } from "@angular/common";
+
+/**
+ * Generic component to display one form field (editComponent) of an entity.
+ *
+ * Dynamically extends field details from entity schema and loads the relevant, specific EditComponent implementation.
+ */
+@Component({
+  selector: "app-form-field",
+  templateUrl: "./form-field.component.html",
+  styleUrls: ["./form-field.component.scss"],
+  standalone: true,
+  imports: [DynamicComponentDirective, HelpButtonComponent, NgIf],
+})
+export class FormFieldComponent<T extends Entity> implements OnChanges {
+  /** field id or full config */
+  @Input() field: ColumnConfig;
+  /** full field config extended from schema (used internally and for template) */
+  _field: FormFieldConfig;
+
+  @Input() entity: T;
+  @Input() form: EntityForm<T>;
+
+  constructor(private entityFormService: EntityFormService) {}
+
+  ngOnChanges(changes: SimpleChanges): void {
+    if (changes.field || changes.entity) {
+      this.updateField();
+    }
+  }
+
+  private updateField() {
+    if (!this.entity?.getConstructor()) {
+      this._field = undefined;
+      return;
+    }
+
+    this._field = this.entityFormService.extendFormFieldConfig(
+      this.field,
+      this.entity.getConstructor(),
+    );
+  }
+}
--- a/src/app/core/config/config.service.spec.ts
+++ b/src/app/core/config/config.service.spec.ts
@@ -98,4 +98,69 @@ describe("ConfigService", () => {
     const result = service.exportConfig();
     expect(result).toEqual(expected);
   }));
+
+  it("should migrate Form field group headers into combined format", fakeAsync(() => {
+    const testConfigId = "view:test/:id";
+
+    const form1Config = {
+      cols: [["name"], ["other"]],
+      headers: [null, "2nd group"],
+    };
+    const form2Config = {
+      cols: [["name"], ["other"]],
+    };
+
+    const viewConfig = {
+      component: "EntityDetails",
+      config: {
+        entity: "Entity",
+        panels: [
+          {
+            title: "Tab 1",
+            components: [{ title: "", component: "Form", config: form1Config }],
+          },
+          {
+            title: "Tab 2",
+            components: [
+              { component: "ActivitiesOverview" },
+              { component: "Form", config: form2Config },
+            ],
+          },
+        ],
+      },
+    };
+
+    updateSubject.next({
+      entity: Object.assign(new Config(), {
+        data: { [testConfigId]: viewConfig },
+      }),
+      type: "update",
+    });
+    tick();
+
+    const result: any = service.getConfig(testConfigId);
+
+    const expectedForm1Config = {
+      fieldGroups: [
+        { fields: form1Config.cols[0] },
+        { fields: form1Config.cols[1], header: form1Config.headers[1] },
+      ],
+    };
+    const expectedForm2Config = {
+      fieldGroups: [
+        { fields: form2Config.cols[0] },
+        { fields: form2Config.cols[1] },
+      ],
+    };
+
+    expect(result.config.panels[0].components[0].config).toEqual(
+      expectedForm1Config,
+    );
+    expect(result.config.panels[1].components[1].config).toEqual(
+      expectedForm2Config,
+    );
+    expect(result.config.panels[1].components[0].config).toEqual(
+      viewConfig.config.panels[1].components[0].config,
+    );
+  }));
 });
--- a/src/app/core/config/config.service.ts
+++ b/src/app/core/config/config.service.ts
@@ -4,6 +4,8 @@ import { Config } from "./config";
 import { LoggingService } from "../logging/logging.service";
 import { LatestEntityLoader } from "../entity/latest-entity-loader";
 import { shareReplay } from "rxjs/operators";
+import { EntitySchemaField } from "../entity/schema/entity-schema-field";
+import { FieldGroup } from "../entity-details/form/form.component";
 
 /**
  * Access dynamic app configuration retrieved from the database
@@ -35,7 +37,7 @@ export class ConfigService extends LatestEntityLoader<Config> {
   }
 
   public getConfig<T>(id: string): T {
-    return this.currentConfig.data[id];
+    return this.applyMigrations(id, this.currentConfig.data[id]);
   }
 
   public getAllConfigs<T>(prefix: string): T[] {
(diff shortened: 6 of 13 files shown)


src/app/core/config/config-fix.ts
```diff
@@ -269,10 +269,10 @@ export const defaultJsonConfig = {
...
               "component": "Form",
               "config": {
-                "cols": [
-                  ["logo", "favicon"],
-                  ["siteName", "defaultLanguage", "displayLanguageSelect"],
-                  ["primary", "secondary", "error", "font"],
+                "fieldGroups": [
+                  { fields: ["logo", "favicon"] },
+                  { fields: ["siteName", "defaultLanguage", "displayLanguageSelect"] },
+                  { fields: ["primary", "secondary", "error", "font"] },
```
sleidig, line 275: TODO: update whole config-fix with new format

for testing the migration, not done yet

src/app/core/config/config.service.ts
```diff
@@ -46,6 +47,67 @@ export class ConfigService extends LatestEntityLoader<Config> {
...
+          group.header = formConfig.headers[i];
+        }
+      });
+    }
+
+    return newFormConfig;
+  }
+
+  const newConfig = JSON.parse(
+    JSON.stringify(configData),
```
TheSlimvReal, line 100:
```suggestion
    configString,
```

src/app/core/entity-details/form/form-config.ts
```diff
@@ -0,0 +1,16 @@
+import { ColumnConfig } from "../../common-components/entity-subrecord/entity-subrecord/entity-subrecord-config";
+
+/**
+ * The (possibly abbreviated) configuration for a "FormComponent", as it is stored in the config file.
+ */
+export interface FormConfig {
+  fieldGroups: FieldGroup[];
+}
```
TheSlimvReal, line 8: What is this interface there for?

src/app/core/common-components/entity-form/form-field/form-field.component.html
TheSlimvReal, line 1: Good idea with this component. Makes creating form components even easier.

src/app/core/config/config.service.ts
```diff
@@ -35,7 +36,7 @@ export class ConfigService extends LatestEntityLoader<Config> {
   }
 
   public getConfig<T>(id: string): T {
-    return this.currentConfig.data[id];
+    return this.applyMigrations(id, this.currentConfig.data[id]);
```
TheSlimvReal, line 39: Same as in the other PR. Only patching this ad-hoc does not allow to save the full config later.

src/app/child-dev-project/notes/note-details/note-details.component.ts
```diff
@@ -77,22 +81,21 @@ export class NoteDetailsComponent implements OnInit {
...
+    this.bottomForm = formConfig?.bottomForm ?? this.bottomForm;
   }
 
   ngOnInit() {
+    this.topFieldsGroups = this.topForm.map((f) => ({ fields: [f] }));
+    this.bottomFieldsGroups = [{ fields: this.bottomForm }];
+
     this.form = this.entityFormService.createFormGroup(
-      this.middleForm.concat(...this.topForm, this.bottomForm),
+      [...this.middleForm, ...this.topForm, ...this.bottomForm],
```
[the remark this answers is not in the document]
sleidig: oh, I wasn't aware of that. Sounds smart!

src/app/core/entity-details/form/form-config.ts
sleidig: This is slightly trivial at the moment but defines the format allowed in config doc (similar to `EntityDetailsConfig` and others). I think this is useful, even though we don't actively validate config formats usually.
Should we make the code comment clearer, somehow?

src/app/core/common-components/entity-form/entity-form.service.ts
```diff
@@ -111,15 +122,17 @@ export class EntityFormService {
...
   public createFormGroup<T extends Entity>(
-    formFields: FormFieldConfig[],
+    formFields: ColumnConfig[],
     entity: T,
     forTable = false,
   ): EntityForm<T> {
-    this.extendFormFieldConfig(formFields, entity.getConstructor(), forTable);
+    const fullFields = formFields.map((f) =>
+      this.extendFormFieldConfig(f, entity.getConstructor(), forTable),
+    );
```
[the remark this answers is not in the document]
sleidig: with the new `<app-form-field` NoteDetails (and others) do not need field details anymore, so that should be fine to not run this in those components. 

To create the form with its validators, we do need the field schema in that service. But currently it's actually a weird mix of loading the schema and using the extended field ...
Doing extendFormFieldConfig here also means it will happen twice - again in the FormFieldComponent. Although that shouldn't create much problems.

Crazy other idea: What if FormFieldComponents would self-register their formControls to the given (empty) form object passed as input?

src/app/core/config/config.service.ts
sleidig: to be honest, I am not sure why I opted for migrating during getConfig instead of loading of config from DB. I'll change this as you suggest

src/app/core/common-components/entity-form/entity-form.service.ts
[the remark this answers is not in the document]
TheSlimvReal: Then lets split it up. The validators are only used to create the form group and not in the components. At the moment I think the validators always come from the schema and never from the field definition which I think also makes sense. 
Regarding the self registration. I am not sure whether its a good idea to move all this logic to the components instead of having it in a central service. Or do you see any advantages?

src/app/core/entity-details/form/form-config.ts
sleidig: ok, removed as discussed

src/app/core/config/config-fix.ts
```diff
@@ -673,7 +673,7 @@ export const defaultJsonConfig = {
                 "cols": [
                   ["health_bloodGroup"],
                   [
-                    { "id": "_description_health", "edit": "EditDescriptionOnly", "label": $localize`:description section:Health checkups are to be done regularly, at least every 6 months according to the program guidelines.`},
+                    { "id": "_description_health", "editComponent": "EditDescriptionOnly", "label": $localize`:description section:Health checkups are to be done regularly, at least every 6 months according to the program guidelines.`},
```
TheSlimvReal, line 676: There are still quite some places where "view" instead of "viewComponent" is used in this file.

src/app/core/config/config.service.ts
```diff
@@ -48,4 +49,88 @@ export class ConfigService extends LatestEntityLoader<Config> {
...
+  private applyMigrations(config: Config): Config {
+    const migrations: ConfigMigration[] = [
+      migrateFormHeadersIntoFieldGroups,
+      migrateFormFieldConfigView2ViewComponent,
+    ];
+
+    const newConfig = JSON.parse(JSON.stringify(config), (_that, rawValue) => {
+      let configPart = rawValue;
+      for (const migration of migrations) {
+        if (migration.filter(_that, configPart)) {
```
TheSlimvReal, line 62: The word `filter` is confusing me here a little as its not really filtering anything. What about `shouldMigrate`?

src/app/core/config/config.service.spec.ts
```diff
@@ -98,4 +99,148 @@ describe("ConfigService", () => {
...
+      expectedForm2Config,
+    );
+    expect(result.config.panels[1].components[0].config).toEqual(
+      viewConfig.config.panels[1].components[0].config,
+    );
+    // keep new format unchanged:
+    expect(result.config.panels[2].components[0].config).toEqual(formNewConfig);
+  }));
+
+  it("should migrate FormFieldConfig view/edit to viewComponent/editComponent", fakeAsync(() => {
```
TheSlimvReal, line 180: Is "edit" to "editComponent" also checked here?

src/app/core/config/config.service.ts
```diff
@@ -48,4 +49,88 @@ export class ConfigService extends LatestEntityLoader<Config> {
...
+        }
+      });
+    }
+
+    configPart.config = newFormConfig;
+    return configPart;
+  },
+};
+
+const migrateFormFieldConfigView2ViewComponent: ConfigMigration = {
```
TheSlimvReal, line 115: I think for this kind of migration it would be much simpler to just use search and replace. I don't see "view" and "edit" used anywhere else in our config.

src/app/core/entity/default-datatype/edit-component.ts
```diff
@@ -62,9 +58,6 @@ export abstract class EditComponent<T> implements OnInit, OnChanges {
     if (!this.formFieldConfig?.forTable) {
       this.label = this.formFieldConfig?.label ?? this.propertySchema?.label;
     }
-    if (this.formFieldConfig?.forTable) {
-      this.tooltip = undefined;
-    }
```
TheSlimvReal, line 67: As far as I understand this now, the description is only not shown in the edit mode of the entity subrecord because it does not use the `FormFielComponent` and therefore the description is not rendered? If we would want to use the component there as well we would still need to remove the description. Maybe this should already happen in the `EntityFormService`?

src/app/core/entity/schema/entity-schema-field.ts
```diff
@@ -78,6 +78,8 @@ export interface EntitySchemaField {
    */
   entityReferenceRole?: EntityReferenceRole;
 
+  // TODO: possibly remove viewComponent + editComponent from schema and only infer via dataType (or overwrite in FormFieldConfig)
```
TheSlimvReal, line 81: For self-configurability it would definitely be helpful.

TheSlimvReal (changes requested):
This topic has become much more complex than I expected 😅 Let me know if I should help with some of the topics.

src/app/core/common-components/entity-subrecord/entity-subrecord/entity-subrecord.component.spec.ts
```diff
@@ -184,24 +183,6 @@ describe("EntitySubrecordComponent", () => {
     expect(sortedNames).toEqual(["A", "b", "C"]);
   });
 
-  it("should log a warning when the column definition can not be initialized", () => {
```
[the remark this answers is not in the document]
sleidig: This does not seem to have been handled in subrecord component anyway? Was this supposed to come from appDynamicComponent?

src/app/core/config/config-fix.ts
sleidig: :+1:  will do the config-fix update now that we have settled on the format

src/app/core/config/config.service.ts
sleidig: true - we could actually also leave this within the single migration function and let them just return the object unchanged if nothing should be migrated.

I would suggest to make `type ConfigMigration = (key: string, configPart: any) => any`, i.e. a "reviver" function implementation. And then we simply chain all migration revivers within the JSON.parse to apply migrations.

src/app/core/config/config.service.ts
sleidig: while that might work well, I would like to avoid a risky approach - there may well be some user defined pieces that accidentally match. Let's do this properly.

src/app/core/entity/schema/entity-schema-field.ts
sleidig: noted for a possible implementation within the setup wizard, when we rework the datatype vs. component selection question.

src/app/core/common-components/entity-form/form-field/form-field.component.ts
```diff
@@ -0,0 +1,54 @@
...
+import { EntityForm, EntityFormService } from "../entity-form.service";
+import { ColumnConfig } from "../../entity-subrecord/entity-subrecord/entity-subrecord-config";
+import { FormFieldConfig } from "../entity-form/FormConfig";
+import { NgIf } from "@angular/common";
+
+/**
+ * Generic component to display one form field (editComponent) of an entity.
+ *
+ * Dynamically extends field details from entity schema and loads the relevant, specific EditComponent implementation.
+ */
```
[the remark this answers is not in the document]
sleidig: Thanks for pointing this out! I knew we had something for the ViewComponent but just couldn't find it ...

Should we have a single component that supports view + edit modes via input switch - or keep the two completely separate components? (i.e. keeping this implementation and deleting EntityPropertyViewComponent; Or removing the view mode here and using EntityPropertyViewComponent for that part).
What do you think, @TheSlimvReal ?

src/app/core/entity/default-datatype/edit-component.ts
sleidig: I think this tooltip property was not used here anyway. Tooltip has always been displayed one or two parent levels higher. For table inline edit mode it should be omitted for layout reasons, I guess.

I am moving this logic to the FormFieldComponent, where the tooltip is actually displayed and the full formFieldConfig parsed. It was actually still missing there. Thanks!

src/app/core/config/dynamic-components/dynamic-component.directive.ts
```diff
@@ -40,9 +45,20 @@ export class DynamicComponentDirective implements OnChanges {
...
-    )();
+    let component;
+    try {
+      component = await this.components.get(
+        this.appDynamicComponent.component,
+      )();
+    } catch (e) {
+      this.logger.error(
+        `Failed to load dynamic component:\n${JSON.stringify(
+          this.appDynamicComponentContext,
```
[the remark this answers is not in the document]
sleidig: actually true. I was trying to get something more useful than "undefined" component not found. Somehow missed that we have a bunch of config available anyway :+1:

src/app/core/common-components/entity-form/entity-form.service.ts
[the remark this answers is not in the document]
sleidig: Thanks a lot for finding this!

In Child > Education History we had a "ChangesAfterChecked" (due to the description for table column header tooltip being extended in place later via EntityFormService). Changed with a medium sized refactoring that I will need in setup-wizard also (introducing a "FormFieldLabelComponent").

Working to fix the special fields in RowDetails now and incorporating that into FormFieldComponent as well, so that we could even use them seamlessly anywhere.

src/app/core/common-components/entity-form/form-field/form-field.component.ts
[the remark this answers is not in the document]
sleidig: Looking further into special cases in subrecord and rowdetails, where we use viewComponents as fallbacks or alternatives to editComponents, I am tending towards a single integrated component that handles view and edit mode.

src/app/core/common-components/entity-form/entity-form.service.ts
```diff
@@ -111,42 +120,63 @@ export class EntityFormService {
...
+    formFields: ColumnConfig[],
     entity: T,
     forTable = false,
   ): EntityForm<T> {
-    this.extendFormFieldConfig(formFields, entity.getConstructor(), forTable);
     const formConfig = {};
-    const entitySchema = entity.getSchema();
     const copy = entity.copy();
-    formFields
-      .filter((formField) => entitySchema.get(formField.id))
```
TheSlimvReal, line 123: Do you know why this filter is not needed anymore?

src/app/core/common-components/entity-form/form-field-label/form-field-label.component.html
```diff
@@ -0,0 +1,3 @@
+<span *ngIf="_field" [matTooltip]="_field.description">
+  {{ _field?.label ?? _field?.id }}
```
TheSlimvReal, line 2: Is this our current approach as well to fall back to the `.id`? I am not sure whether this should be supported or if we rather obviously show that there is something missing.

src/app/core/common-components/entity-form/form-field-label/form-field-label.component.ts
```diff
@@ -0,0 +1,47 @@
...
+    }
+  }
+
+  private updateField() {
+    if (!this.entityType) {
+      this._field = undefined;
+      return;
+    }
+
+    this._field = this.entityFormService.extendFormFieldConfig(
```
TheSlimvReal, line 42: Does this really have to be another component which extends the config itself? In case of the entity subrecord this now means that is created 3 times there (1x `createFormGroup`, 1x `extendFormFieldConfig`, 1x this component). And when clicking on the row it happens another 2 times (1x row details `createFormGroup` and 1x in form field component). Currently this is at least only 3 times in total.

src/app/core/common-components/entity-form/form-field/form-field.component.html
```diff
@@ -0,0 +1,37 @@
...
+    else displayComponent
+  "
+>
+  <div [class.field-with-tooltip-suffix]="_field.description">
+    <ng-container
+      [appDynamicComponent]="{
+        component: _field.editComponent,
+        config: {
+          formFieldConfig: _field,
+          propertySchema: entity.getSchema().get(_field.id),
```
TheSlimvReal, line 13: As we are passing the entity as well we can remove this here and only access it in the `EditComponent`.

src/app/core/common-components/entity-form/form-field/form-field.component.ts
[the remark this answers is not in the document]
TheSlimvReal: On the other hand we currently only have one place where edit and view components are switched (entity subrecord inline edit) which I don't think is a very useful mode. In all other cases there is quite a clear cut between edit components and view component. Therefore I would like to make the view component as simple as possible and keep to code a little more obvious. Otherwise its always quite difficult to understand why in a certain place you are supposed to use a form field component when you just want to display that value.

src/app/core/common-components/entity-form/entity-form.service.ts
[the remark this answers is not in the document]
TheSlimvReal: I don't think this was the error I was talking about. I still see it when opening the row details as described above. I think it has to do with the missing filter in `createFormGroup`. I was not able to reproduce the error you are describing in the earlier version.

src/app/core/common-components/entity-subrecord/entity-subrecord/entity-subrecord.component.spec.ts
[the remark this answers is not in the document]
sleidig: yes, there is a catch, which could catch the error thrown by EntityFormService.extendFormFieldConfig  But actually I do not see how any error could occur in the EntityFormService in the current implementation. Do you? Otherwise let's remove both try-catch blocks and only rely on the errors for broken field config in the individual FormFieldComponent?

TheSlimvReal (changes requested):

src/app/core/common-components/entity-form/form-field-label/form-field-label.component.ts
sleidig: the refactoring overall removes us calling getEntityConstructor as a method - that in itself has way way way more performance implication than (once on changes) doing an object.assign.

The alternative (which we decided against initially) could be to do the extending already in config-service or extremely early, centrally - so that everyone has the full details automatically. (but I do want to get back to "actual work" from all this refactoring here also :blush: )

src/app/core/common-components/entity-form/form-field/form-field.component.html
sleidig: :+1: good point. In fact we don't need it in EditComponent anymore at all, because it's just used to (again ... :roll_eyes: ) re-apply some fallback configs from schema to the field config

src/app/core/common-components/entity-form/entity-form.service.ts
sleidig: I think I had assumed because we don't rely on the schema here anymore, we do not need to explicitly remove them. But actually there may have been more reason behind filtering them, like the form not being relevant for those that are not related to entity properties ...
I am putting this back in, thanks :+1:

src/app/core/common-components/entity-form/form-field/form-field.component.ts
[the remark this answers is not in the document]
sleidig: That is also a valid point. implementation itself is probably cleaner and simpler if we keep this separated ...

Currently Subrecord and RowDetails use the view-mode switch. But they could both instead use the other component, passing fieldConfig to that to render the viewComponent.

Only one limitation: subrecord in inline edit mode is doing a fallback, showing viewComponents for those columns that do not have a editComponent. And for that again would need the full fieldConfig :roll_eyes: 
Alternatively: let FormFieldComponent use the "ViewFieldComponent" (for view mode) as a fallback but have that as a separated component?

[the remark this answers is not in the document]
TheSlimvReal: Ah I see. Then I think go ahead with the approach that you currently feel most comfortable with.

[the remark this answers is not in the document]
sleidig: I will clean this up now:
- FormFieldComponent will only handle edit mode
- ViewFieldComponent will follow the same pattern and handle view mode (renaming + moving EntityPropertyViewComponent for this)

TheSlimvReal (approved):
I see an end. Thanks a lot for the motivation to get this PR to this level. I really like it now.

The names for the label, view and edit fields makes a lot of sense for me. I am also happy to see that in many places we are now able to pass less arguments around, especially that we dont even need the `EntitySchemaField` anymore at a lot of places. Good job there. 
Lets hope we get the feeling that this works is paying off soon.

src/app/core/config/config-fix.ts
```diff
@@ -734,10 +710,10 @@ export const defaultJsonConfig = {
               "title": "",
               "component": "Form",
               "config": {
-                "cols": [
-                  ["dropoutDate"],
-                  ["dropoutType"],
-                  ["dropoutRemarks"]
+                "fieldGroups": [
+                  { "fields": ["dropoutDate"] },
```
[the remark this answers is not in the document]
sleidig: interesting thought. Let's see where this all leads when implementing the Admin UI further.

src/app/core/common-components/entity-field-view/entity-field-view.component.ts
```diff
@@ -0,0 +1,63 @@
...
+ */
+@Component({
+  selector: "app-entity-field-view",
+  templateUrl: "./entity-field-view.component.html",
+  styleUrls: ["./entity-field-view.component.scss"],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  imports: [NgIf, DynamicComponentDirective, PillComponent],
+  standalone: true,
+})
+export class EntityFieldViewComponent<E extends Entity = Entity>
```
sleidig, line 32: This `standalone: true` might be the reason for #2167

PR #2102, opened 2023-11-24
refactor(core): change EntityConfig to flattened structure with EntitySchemaField including .id

sleidig:
To ease the development of the setup wizard (PR https://github.com/Aam-Digital/ndb-core/issues/2049 ) this refactors config structure including an on-the-fly migration.

---
:warning: mark as codo


src/app/core/entity/entity-config.ts
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Dynamic configuration for a entity.
+ * This allows to change entity metadata based on the configuration.
+ */
+export interface EntityConfig {
+  /**
+   * A list of attributes that will be dynamically added/overwritten to the entity.
+   */
+  attributes?: EntitySchemaField[];
```
TheSlimvReal, line 11: Thinking again about this structure I am a bit sad that it will still be an array. For the property names it is important that they are unique and using a object structure we could actually enforce this (at the moment I don't even know what happens a name exists twice).
What was again the use case where you need direct access to the name of the property?

src/app/core/entity/schema/entity-schema-field.ts
```diff
@@ -26,6 +26,13 @@ import { EntityReferenceRole } from "../../basic-datatypes/entity/entity-referen
  * `@DatabaseField(fieldConfig: EntitySchemaField)`
  */
 export interface EntitySchemaField {
+  /**
+   * The key (property name) on the entity object for this field.
+   *
+   * Do not change after initial field creation!
+   */
+  readonly id: string;
```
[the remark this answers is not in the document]
sleidig: "name" to me sounds like something that might be a "human-readable" name as well instead of a technical, internal name. "key" or "id" (which we also use for filters, for example) makes that more intuitively clear, I think.

src/app/core/entity/entity-config.ts
sleidig: that is a valid point. (I think the later attribute definition would simply overwrite the former one).
We need the id within the individual schema-field because in a few places we pass around only a single field's schema - and then often need pass two separate params (schemaField + id) instead of the single object.

However, we could also make this an object here and take care of the transformation, even writing the id to the schema-field, in the entity-config-service. Only downside would be that again config and runtime structure are not identical then and we need to juggle with two interfaces (and potentially confusing two models in our head).

TheSlimvReal: Looking at how you are currently "saving" the attributes I already see a big advantage of having a key-value object there. Instead of just having a list of changes you would just have a single object with the changed schemas and these can be written to the config.

So far I only found the `ConfigFieldComponent` which uses both the schema and the attribute ID at the same time. Lets discuss whether we can find a more suitable structure for the attribute config.

sleidig: there are only slight differences for using the Map instead of an object also (https://stackoverflow.com/a/18541990). Maybe refactoring the Entity.schema to a simple object would also help (no need for function calls to access schema in templates, etc.)

TheSlimvReal: I am not talking about the map in entity schema. Although we can also adjust that. I mean the config structure for attributes which is just an array and does therefore enforce the unique keys even when adjusting in fauxton.

Issue #2104, opened 2023-11-27
Migrate report structure


TheSlimvReal:
The reporting structure needs to be changed to we can define permissions on a per-report basis.

- [x] Make report entity
- [x] Use report entity instead of `config-fix` in `ReportingComponent`
- [ ] Migrate existing reports to new entity

PR #2106, opened 2023-11-27
fix: reports are own entities

TheSlimvReal:
closes: #2104 

MIGRATION REQUIRED -> See https://github.com/Aam-Digital/ndb-admin/pull/15


src/app/features/reporting/report-config.ts
```diff
@@ -0,0 +1,25 @@
...
+import { DatabaseEntity } from "../../core/entity/database-entity.decorator";
+import { Aggregation } from "./data-aggregation.service";
+import { ExportColumnConfig } from "../../core/export/data-transformation-service/export-column-config";
+import { DatabaseField } from "../../core/entity/database-field.decorator";
+
+@DatabaseEntity("ReportConfig")
+export class ReportConfig extends Entity {
+  static create(data: Partial<ReportConfig>) {
+    return Object.assign(new ReportConfig(), data);
+  }
```
sleidig, line 11:
```suggestion
/**
 * A report can be accessed by users to generate aggregated statistics or customized exports calculated from available data.
 * "read" permission for a ReportConfig entity is also used to control which users can generate the report's results.
 */
@DatabaseEntity("ReportConfig")
export class ReportConfig extends Entity {
  static create(data: Partial<ReportConfig>) {
    return Object.assign(new ReportConfig(), data);
  }
  
```

sleidig (approved):
Very clear and simple. :+1: 
Functionality seems unaffected (only tested in mock mode).

Do we want/need an automated migration via admin backend?

PR #2107, opened 2023-11-28
Dependency updates

TheSlimvReal:
All app dependencies updated to latest version.
Also the used Node version has been updated to 18.

There were some issues with the style paths. Maybe this fixes the sometimes broken styles 🤞


sleidig (approved):
Seems to build and work well. Awesome!

Only the icon color for the primary action somehow switched from white to black, although I prefer the look and contrast of white icon. I switched that manually now
image

Any new features or migrations from Angular v17 we should consider? Maybe the new build system? https://angular.io/guide/esbuild

PR #2109, opened 2023-11-28
Query test fix

TheSlimvReal:
Some tests in the query service sometimes failed.

This is because the property `religion` is only defined in the config and the config attributes were not applied in the test environment. So in some cases the entity was not set up properly and the queries using this attribute failed.


sleidig (approved):
Thanks a lot for looking into this! I had noticed the occasional failing tests also

Issue #2117, opened 2023-12-01
Automatically generate SQL schema from config


TheSlimvReal:
To get the SQL database working it needs the schema for the tables. This should be automatically created from the customer config. The difficulty there is that the "final" entity attributes are build by merging the built-in attributes with the ones defined in the config. This also needs to be reflected in the SQL schema. As this should probably run in the backend its not super straight forward to achieve this as the built-in attributes are only available in the frontend.

1) One approach could be to built a default schema for each release/commit and upload this somewhere. The service can then request this schema and extend it with the custom config. This approach does not require any changes to the app but it needs a lot of infrastructure to work together.

2) An alternative would be the much bigger refactoring where all the entity related code is extracted to a library that can be used in the front and backend (browser and NodeJS). This is quite a big project but might be something we anyways need at one point. It would also allow to build the SQL schema without having to version the default schemas (but we will have to version the library though so maybe this is actually quite similar to the before approach but with just more stuff uploaded somewhere).

3) We could also generate the schema in the frontend but this would have the issue that the schema cannot be rebuilt without accessing the app. 

The decision probably depends a little on the question where we see the default config go in the future. Is this a concept we want to stick with or do we rather want to go to a format where we plan on having "full" customer configs where they already include all information. This would mean the first or third option could be viable as they are simpler to achieve and allow to be removed once they are not required anymore.

If we however want to build a library for the entity system in the near future anyway, then of course option 2 would be the most straight forward. Until that has happened we could go with option 3 and then just copy/paste the code over.

Issue #2105, opened 2023-11-27
Backend service for reports


TheSlimvReal:
Add a backend service (or extends the permission backend) with an endpoint that allows to query data.
This endpoint receives the name of the organization and the ID of the report.
The report than uses the access token of the request to request the report entity with the given ID (user needs read access on the report entity).
The service then executes the query of the report on the database and returns the result as `JSON`.

Issue #1887, opened 2023-05-25
API endpoint for accessing reporting data


sleidig:
To integrate Aam Digital with other systems like TolaData's monitoring & evaluation platform, data should be accessible for external systems through a secure API.

Currently we achieve a "manual integration" by configuring custom reports, have users download the csv file of a report from Aam Digital and upload that file in the other system. The idea is to make the same data available not only as a csv download but also through a server API automatically.

also see https://docs.google.com/document/d/1kXhP01F8EvHBAPNHqQhWUqJYSJBEut3a02MIYUrGkF8/edit#

sleidig:
@TheSlimvReal does the current API already support access via some kind of API token?

TheSlimvReal:
@sleidig No, this still needs to be added

sleidig:
see https://github.com/Aam-Digital/aam-services/issues/11

PR #2123, opened 2023-12-07
SQS Integration

TheSlimvReal:
closes: #2117 #2105 #1887

- report config has been adjusted to also support reports with `"mode": "sql"`.
- SQL reports are handled by `SqlReportService`
- `SqlReportService` automatically creates and, if necessary, updates the SQS schema before sending the query request to the baceknd
- [Query Backend](https://github.com/Aam-Digital/query-backend) is available through an reverse proxy


src/app/features/reporting/reporting/reporting.component.spec.ts
```diff
@@ -213,4 +217,22 @@ describe("ReportingComponent", () => {
     expect(component.data).toEqual(data);
     expect(component.mode).toBe("exporting");
   });
+
+  it("should use the sql report service when report has mode 'sql'", async () => {
+    const report = new SqlReport();
+    report.mode = "sql";
```
sleidig, line 223: could init .mode within the constructor already?

src/app/features/reporting/reporting/reporting.component.ts
```diff
@@ -34,8 +31,8 @@ import { ReportConfig } from "../report-config";
   standalone: true,
 })
 export class ReportingComponent {
-  reports: ReportConfig[];
-  mode: "exporting" | "reporting" = "exporting";
+  reports: ReportType[];
+  mode: ReportType["mode"];
```
sleidig, line 35: I find `ReportType` a bit confusing, as this here is more than just the type. From the names I would always intuitively write ReportConfig instead of ReportType. Can't think of a better naming right now, however ...

Alternatively, maybe work with a generic on ReportConfig instead of the sup-classes? Then we wouldn't have 1 + 3 + 1 different types in this context

src/app/features/reporting/sql-report/sql-report.service.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+  it("should be created", () => {
+    expect(service).toBeTruthy();
+  });
+
+  it("should query the external service with the provided report data", async () => {
+    const result = [{ some: "data" }];
+    mockHttpClient.post.and.returnValue(of(result));
+    const report = new SqlReport();
+
+    await service.query(
```
sleidig, line 48: let's also check that the result is returned here?

sleidig (approved):
Impressive how simple this code is :smiley: 

From my quick testing in the PR deployment, the existing reports still seem to work fine.
Only the order in the dropdown of the reporting component had a designed logical order (most important/general first) which is now lost.

src/app/features/reporting/reporting/reporting.component.ts
TheSlimvReal: Lets see if can find a better approach here

TheSlimvReal:
@sleidig the issue with the unstable ordering was already introduced with #2106. Should we sort them by name?

src/app/features/reporting/report-config.ts
```diff
@@ -7,38 +7,48 @@ import { DatabaseField } from "../../core/entity/database-field.decorator";
 /**
  * A report can be accessed by users to generate aggregated statistics or customized exports calculated from available data.
  * "read" permission for a ReportConfig entity is also used to control which users can generate the report's results.
+ *
+ * This is the class which is saved to the database.
+ * However, when when using, use the {@link ReportEntity} instead which provides better type safety.
```
sleidig, line 12:
```suggestion
 * However, when using this in code, use the {@link ReportEntity} instead which provides better type safety.
```

sleidig (approved):
Really nice resolution for the ReportConfig thing (although it feels a bit like dark magic ...)

Regarding report ordering: I guess alphabetic order would at least allow us to consciously influence it, so that would be nice.

PR #2125, opened 2023-12-11
new EntityPermissionGuard to check for EntityAbility

sleidig:
- [x] PROBLEM: on new load (i.e. when typing the special URL in the browser), the EntityAbility does not seem to have the permissions loaded yet and is allowing any access.
- [x] make this apply on all routes by default
- [x] make Navigation menu check also for this guard, not only for userRoleGuard when hiding menu items


src/app/core/permissions/permission-guard/entity-permission.guard.ts
```diff
@@ -0,0 +1,42 @@
...
+      if (route instanceof ActivatedRouteSnapshot) {
+        // Route should only change if this is a "real" navigation check (not the check in the NavigationComponent)
+        this.router.navigate(["/404"]);
+      }
+      return false;
+    }
+  }
+
+  private canAccessRoute(routeData) {
+    const operation = routeData?.["requiredPermissionOperation"];
```
TheSlimvReal, line 31: Hmm what about making this behaviour default? If a "entity" or "config.entity" exists then the "read" permission for this entity is checked. Otherwise its ignored. This would make most route permissions automatic without anything required in the config.

TheSlimvReal: Ok but I see the issue that sometimes its not just about reading. I just wonder, whether we can get this a bit more automated because you will always need the same permissions for a certain config.

TheSlimvReal: Yeah I think the `RouterService` is the right place for that. In my mind I would have the following cases:

1. If entity can be read from config and a operation is defined check this
2. If entity can be read from config but no operation defined then check with "read"
3. If no entity can be found return true

Potentially we could also save on the component which permission is required (e.g. through a annotation) but thats not possible with the current architecture.

sleidig:
To test critical path:

(1) create a route with EntityPermissionGuard, e.g. in app.routing.ts:
```
  {
    path: "user-account",
    component: UserAccountComponent,
    canActivate: [AuthGuard, EntityPermissionGuard],
    data: {
      requiredPermissionOperation: "update",
      config: { entityType: "Config" },
    },
  },
```

(2) change DatatabaseRules in demo-permission-generator.service.ts for user_app, to have some missing entity permissions

(3) visit that path, and also reload page on the forbidden path as this has previously been an error
try this in both mock and local mode, as behavior may be different.

src/app/core/permissions/ability/ability.service.spec.ts
```diff
@@ -159,22 +198,29 @@ describe("AbilityService", () => {
       entity: new Config(Config.PERMISSION_KEY, rules),
       type: "update",
     });
+
+    //TODO: is the current implementation as a replaySubject rather than an explicit triggering after updating the ability really correct?
```
sleidig, line 202: Do you know the background of this functionality (and test), @TheSlimvReal ?

TheSlimvReal: Yes, the common use cases were: notify once it is there and then notify me whenever it gets updated. This is where the replay subject comes in handy. We do not need to check whether it already has been loaded. We just subscribe and in case it is already loaded we get notified instantly. Is there any problem with it?

sleidig: I think the update might trigger before the code in the service, actually updating the EntityAbility runs.
(And also, with the rework here this test somehow fails :wink: )

How about "manually" triggering updatedAbility after we have applied ability.update() ?

TheSlimvReal: Yes, it looks like the `async` call breaks this. As the ability updated is barely used I guess its fine to change that. I also just found `this.ability.on('update', () => {})` which allows to listen for updates that are published by casl itself. So maybe thats even easier.

TheSlimvReal (approved):

PR #2126, opened 2023-12-11
fix: prevent usage of superclass schema

TheSlimvReal:
If currently only the `@DatabaseEntity` property is used without any property being decorated with `@DatabaseField`, then the schema object from the superclass is used. This results in new properties also being added to the superclass's schema.


sleidig (approved):
Thanks so much for figuring this out! :1st_place_medal:

Issue #2128, opened 2023-12-13
Reloading page redirects to 404


sleidig:
Probably since PR #2125 the system sends users to 404 page when reloading on a route that has permissions (e.g. /school or /child). While we have changed the system to avoid giving access to any route (including ones without permission), for routes with permission the user should be able to reload.

PR #2129, opened 2023-12-14
Guards fix

TheSlimvReal:
closes: #2128 

- [x] `EntityPermissionGuard` waits for some rules to be initialised
- [x] `AuthGuards` needs to always be added before using another guard

### DIFF
--- a/src/app/core/ui/navigation/navigation/navigation.component.spec.ts
+++ b/src/app/core/ui/navigation/navigation/navigation.component.spec.ts
@@ -15,7 +15,13 @@
  *     along with ndb-core.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-import { ComponentFixture, TestBed, waitForAsync } from "@angular/core/testing";
+import {
+  ComponentFixture,
+  fakeAsync,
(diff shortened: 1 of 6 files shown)


src/app/core/ui/navigation/navigation/navigation.component.spec.ts
```diff
@@ -103,11 +111,39 @@ describe("NavigationComponent", () => {
...
     mockConfigUpdated.next(null);
+    tick();
 
     expect(component.menuItems).toEqual([
       new MenuItem("Children", "child", "/child"),
     ]);
-  });
+  }));
+
+  it("should add menu items where entity permissions are missing", fakeAsync(() => {
```
sleidig, line 121: The title of this tests sounds counter-intuitive - why should it add where permissions are missing? and the actual test case looks more like it is simply filtering out those that do not pass the guard

sleidig (commented):
after reload the menu item is not highlighted anymore (maybe that also needs to be async until items are available?)

sleidig (approved):

Issue #2122, opened 2023-12-05
Add permissions to dashboard


TheSlimvReal:
Currently all users see the whole dashboard even if they might not even have access to the entities which are required to the dashboard widgets to work or they might not be able to visit the routes which are mentioned in the shortcuts widget.

The app should automatically detect which widgets the user cannot use/view and disable them, based on the route permissions or entity permissions.

sleidig:
Shall we try to merge the two permissions logics (operations on entities + roles to access a route)? For the menu entry to the setup wizard I am struggling with that a bit

sleidig:
related / overlapping with https://github.com/Aam-Digital/ndb-core/issues/1552

PR #2131, opened 2023-12-14
Dashboard permissions

### LINKED ISSUE #1552: Custom Dashboard for individual user accounts
*As a project manager I want to see different details on the dashboard (higher level overview) than my field workers (who need more daily task management).*

**Objective**
Allow different users to have a custom dashboard, seeing different widgets either based on their user role or their individual settings for their personal user account.

TheSlimvReal:
closes: #2122, closes #1552

- [x] Permissions for shortcut widget
- [x] Permissions for entity widgets

### DIFF
--- a/src/app/core/dashboard/dashboard/dashboard.component.ts
+++ b/src/app/core/dashboard/dashboard/dashboard.component.ts
@@ -20,6 +20,7 @@ import { DynamicComponentConfig } from "../../config/dynamic-components/dynamic-
 import { NgFor } from "@angular/common";
 import { DynamicComponentDirective } from "../../config/dynamic-components/dynamic-component.directive";
 import { RouteTarget } from "../../../route-target";
+import { EntityAbility } from "../../permissions/ability/entity-ability";
 
 @RouteTarget("Dashboard")
 @Component({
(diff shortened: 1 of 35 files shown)


src/app/core/permissions/permission-guard/entity-permission.guard.ts
```diff
@@ -53,6 +53,7 @@ export class EntityPermissionGuard implements CanActivate {
     path = path.replace(/^\//, "");
 
     function isPathMatch(genericPath: string, path: string) {
+      // TODO this does not seem to work with children routes (admin module)
```
sleidig, line 56: yes, I was a bit lazy implementing this ... :sweat_smile: 

I think the logic could actually be shared between all guards, so we can make the code from UserRoleGuard reusable:
 https://github.com/Aam-Digital/ndb-core/blob/3b4168b92b9ef214b66e09e594fb77bec2a49a43/src/app/core/permissions/permission-guard/user-role.guard.ts#L75
 
(side note: router.config already contains all the data from config, so parsing the config there is actually unnecessary)

src/app/core/dashboard/dashboard/dashboard.component.ts
```diff
@@ -33,7 +34,58 @@ import { RouteTarget } from "../../../route-target";
...
+      }
+    }
+    // No entity relation -> show widget
+    return true;
+  }
+
+  /**
+   * Detect, which entity is required for which widget.
+   *
+   * TODO in the future the widget itself should expose this
```
sleidig, line 63: That would be much nicer, of course. Any idea how to get there with the dynamic components?

What if we implement this as part of `DashboardWidgetComponent` (maybe with an input the custom widgets can set, if appropriate)?

sleidig (commented):
Seems to work well already :+1: 

Shall we also add the option similar to views, to assign `"permittedUserRoles": ["custom_role"]` to any dashboard widget?

src/app/core/dashboard/dashboard/dashboard.component.ts
TheSlimvReal: The only way I see for this would be by loading the components here already so we can access any static members of it.

src/app/core/permissions/permission-guard/entity-permission.guard.ts
TheSlimvReal: I am not sure what you mean here. What is the shared logic? And also regarding the `router.config` I am not fully following. Do you want to extend it here if you have a improvement in mind?

sleidig: generalized the code. Please review my commit :-)

sleidig (approved):
Really nice generalizations and great cleanup of the interfaces! :smiley: 

@TheSlimvReal , please review my refactoring of the guards that is part of this PR as well, before merging this.

src/app/core/permissions/permission-guard/abstract-permission.guard.ts
```diff
@@ -0,0 +1,92 @@
...
+   *
+   * @param path
+   */
+  public checkRoutePermissions(path: string) {
+    let routeData = this.getRouteDataFromRouter(path, this.router.config);
+    return this.canAccessRoute(routeData?.data);
+  }
+
+  /**
+   * Extract the relevant route from Router, to get a merged route that contains the full trail of `permittedRoles`
```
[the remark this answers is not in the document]
sleidig: this includes the "permittedRoles" from parent routes of nested routing. Not perfect merge of route data, however - for more complex use cases this might potentially break, I think.

PR #2134, opened 2023-12-18
fix(core): new unique-id validator

### LINKED ISSUE #1557: Creating a user with a username that already exists produces a local db conflict
Trying to create a new user entity with a name that already exists fails - but not gracefully as a conflicting document is produced in the local database. We should prevent this more explicitly and with an even more clear error message for users.

sleidig:
to prevent creating duplicate user ids in form

fixes #1557

### Visible/Frontend Changes
- [x] form field is marked invalid if the username field value already exists

- [x] a general "uniqueId" validator is added, usable via entity schema config
- [x] the DynamicValidatorService has to run async now for this advance validator to be build

## TODO
- [x] detect original value and allow to revert to that


src/app/core/demo-data/demo-data-initializer.service.ts
```diff
@@ -62,7 +62,7 @@ export class DemoDataInitializerService {
 
     this.localAuthService.saveUser(this.normalUser);
     this.localAuthService.saveUser(this.adminUser);
-    await this.sessionManager.offlineLogin(this.normalUser);
+    await this.sessionManager.offlineLogin(this.adminUser);
```
TheSlimvReal, line 65:
```suggestion
    await this.sessionManager.offlineLogin(this.normalUser);
```

src/app/core/common-components/entity-form/unique-id-validator/unique-id.service.ts
```diff
@@ -0,0 +1,15 @@
+import { Injectable } from "@angular/core";
+import { EntityMapperService } from "../../../entity/entity-mapper/entity-mapper.service";
+import { uniqueIdValidator } from "./unique-id-validator";
+
+@Injectable({ providedIn: "root" })
+export class UniqueIdService {
```
TheSlimvReal, line 6: Did you have a look at [async validators](https://angular.io/api/forms/AsyncValidator)? As far as I understood it, they would allow to move the async part to the validation check instead of where the validator is built.

TheSlimvReal (commented):
I would definitely want to keep the validators already built in the form service. This is supposed to be a _complete_ service so that it returns you the **full** form with everything set up so it can also be used at other places. 

What exactly do you mean with the DI validator architecture? Like the refactoring we recently did for the data types?

src/app/core/common-components/entity-form/entity-form.service.ts
```diff
@@ -174,7 +174,7 @@ export class EntityFormService {
     formConfig[field.id] = [value];
 
     if (field.validators) {
-      const validators = this.dynamicValidator.buildValidators(
+      const validators = await this.dynamicValidator.buildValidators(
```
TheSlimvReal, line 177: I don't think this is `async` anymore so these changes could all be reset.

TheSlimvReal (changes requested):
If I directly press "save" I sometimes still get the conflict update message. It seems like it doesn't wait for the validator. Can we ensure this somehow?

TheSlimvReal (changes requested):
I discovered one more issue on existing records. E.g. when editing the name field of a existing user entity but changing the original name there, it still doesn't allow to save it.

sleidig:
> I discovered one more issue on existing records. E.g. when editing the name field of a existing user entity but changing the original name there, it still doesn't allow to save it.

This is consistent with current behavior, enforced by the `User` entity. I think that will become obsolete with your PR #2143 (are you removing the special username property there already?). Not super user-friendly but as this unique-id validator is intended for more general use, I would ignore this here. What do you think, @TheSlimvReal ?

PR #2135, opened 2023-12-19
fix: highlight active menu item again

sleidig:
fix regression from last PR


TheSlimvReal (commented):
How did that get lost again 😅

TheSlimvReal (approved):

Issue #2137, opened 2023-12-20
Remove usage of `AuthUser`


TheSlimvReal:
Currently we use the `name` property of `AuthUser` to check whether a given task or event is assigned to the currently logged in user.
This should be refactored so that the `name` property is **only** used to check permissions and to find the connected user entity. After that only the user entity and it's ID (which is currently identical to the `name` property of `AuthUser`) should be used in the application.
For this it might be possible to replace the `CurrentUserSubject` to a `CurrentlyLoggedInSubject` which holds the entity (e.g. `User`) that is logged in.

PR #2139, opened 2023-12-20
Reduce usage of `AuthUser`

TheSlimvReal:
closes: #2137 

- [x] add a `CurrentlyLoggedInSubject` (better names welcome) which hold the entity to which a user account is linked
- [x] use new subject whenever the user is used for business logic
- [x] rename `AuthUser` subject to `SessionInfo` to better distinguish difference from new subject


src/app/app-initializers.ts
```diff
@@ -35,18 +36,19 @@ export const appInitializers = {
...
-          LoggingService.setLoggingContextUser(undefined);
-          analyticsService.setUser(undefined);
-        }
-      });
+      combineLatest([loginState, currentlyLoggedIn]).subscribe(
+        ([newState, user]) => {
+          if (newState === LoginState.LOGGED_IN) {
+            const username = user?.getId(true);
+            LoggingService.setLoggingContextUser(username);
+            analyticsService.setUser(username);
```
sleidig, line 45: it might still make sense to have the username (auth/keycloak layer) here for logging, rather than the entity (business logic layer). But I think it doesn't hurt much to have it this way also.

src/app/core/entity/entity-mapper/entity-mapper.service.ts
```diff
@@ -195,7 +195,8 @@ export class EntityMapperService {
   }
 
   protected setEntityMetadata(entity: Entity) {
-    const newMetadata = new UpdateMetadata(this.currentUser.value?.name);
+    // TODO use Entity (CurrentlyLoggedIn) or keep AuthUser?
+    const newMetadata = new UpdateMetadata(this.sessionInfo.value?.name);
```
sleidig, line 199: really interesting question ...
AuthUser feels a bit more like secured accountability. But for display this in the UI in future, the UserEntity would be much easier to use. So maybe rather optimize for that?

src/app/core/session/currently-logged-in.ts
```diff
@@ -0,0 +1,10 @@
+import { BehaviorSubject } from "rxjs";
+import { Entity } from "../entity/model/entity";
+import { Injectable } from "@angular/core";
+
+@Injectable()
+export class CurrentlyLoggedInSubject extends BehaviorSubject<Entity> {
```
sleidig, line 6: this name is missing a bit that you are getting an Entity (not just a boolean state, for example). Why not keep the name "CurrentUserSubject"?

src/app/core/session/currently-logged-in.ts
```diff
@@ -0,0 +1,10 @@
+import { BehaviorSubject } from "rxjs";
+import { Entity } from "../entity/model/entity";
+import { Injectable } from "@angular/core";
+
+@Injectable()
```
sleidig, line 5:
```suggestion
/**
 * The Entity linked to the currently logged in user, which can be used to pre-fill forms or customize workflows.
 */
@Injectable()
```

src/app/core/session/session-service/session-manager.service.ts
```diff
@@ -84,20 +92,38 @@ export class SessionManagerService {
...
+  offlineLogin(user: SessionInfo) {
     return this.initializeUser(user);
   }
 
-  private async initializeUser(user: AuthUser) {
+  private async initializeUser(user: SessionInfo) {
     await this.initializeDatabaseForCurrentUser(user);
-    this.currentUser.next(user);
+    // TODO can we remove this?
+    this.sessionInfo.next(user);
```
sleidig, line 102: Do we use it via DI anywhere? Maybe if needed only rarely, we could instead make it part of SessionManagerService only?

sleidig (approved):
I really like how this separates the two concepts and having a SessionInfo for the auth layer seems intuitively clear.

src/app/core/session/session-service/session-manager.service.ts
TheSlimvReal: It's surprisingly still quite a few places where we need this. So I would keep it a separate subject for now.

Issue #1277, opened 2022-05-18
column: number of active participants of group


sleidig:
The Teams/Schools/Groups list should allow to show a column stating the total number of participants.

PR #2152, opened 2024-01-04
feat(schools): school list show a column stating the total number of participants

tomwwinter:
closes: #1277 

### Visible/Frontend Changes
- [ ] new column in school list

- [ ] none

### DIFF
--- a/src/app/child-dev-project/schools/display-participants-count/display-participants-count.component.html
+++ b/src/app/child-dev-project/schools/display-participants-count/display-participants-count.component.html
@@ -1,0 +1,3 @@
+<p>
+  {{ participantRelationsCount() }}
+</p>
--- a/src/app/child-dev-project/schools/display-participants-count/display-participants-count.component.ts
+++ b/src/app/child-dev-project/schools/display-participants-count/display-participants-count.component.ts
@@ -1,0 +1,44 @@
+import { Component, OnChanges, signal, WritableSignal } from "@angular/core";
+import { CommonModule } from "@angular/common";
+import { ChildrenService } from "../../children/children.service";
+import { ViewDirective } from "../../../core/entity/default-datatype/view.directive";
+import { DynamicComponent } from "../../../core/config/dynamic-components/dynamic-component.decorator";
+import { ChildSchoolRelation } from "../../children/model/childSchoolRelation";
+import { LoggingService } from "../../../core/logging/logging.service";
+
+@DynamicComponent("DisplayParticipantsCount")
+@Component({
+  selector: "app-display-participants-count",
+  standalone: true,
+  imports: [CommonModule],
+  templateUrl: "./display-participants-count.component.html",
+})
+export class DisplayParticipantsCountComponent
+  extends ViewDirective<any>
+  implements OnChanges
+{
+  participantRelationsCount: WritableSignal<number> = signal(null);
+
+  constructor(
+    private _childrenService: ChildrenService,
+    private _loggingService: LoggingService,
+  ) {
+    super();
+  }
+
+  override async ngOnChanges(): Promise<void> {
+    super.ngOnChanges();
+
+    return this._childrenService
+      .queryActiveRelationsOf("school", this.entity.getId())
+      .then((relations: ChildSchoolRelation[]) => {
+        this.participantRelationsCount.set(relations.length);
+      })
+      .catch((reason) => {
+        this._loggingService.error(
+          "Could not calculate participantRelationsCount, error response from ChildrenService." +
+            reason,
+        );
+      });
+  }
+}
--- a/src/app/core/config/config-fix.ts
+++ b/src/app/core/config/config-fix.ts
@@ -309,6 +309,7 @@ export const defaultJsonConfig = {
       "entity": "School",
       "columns": [
         "name",
+        { id: "DisplayParticipantsCount", viewComponent: "DisplayParticipantsCount", label: $localize`Children` },
         "privateSchool",
         "language"
       ],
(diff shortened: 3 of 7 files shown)


src/app/child-dev-project/schools/display-participants-count/display-participants-count.component.html
```diff
@@ -0,0 +1,3 @@
+<p>
+  {{ participantRelationsCount }}
```
TheSlimvReal, line 2: Do you have insights, how expensive it is to call a getter in a template? So far we have tried to get rid of function call in templates (#1087) as they might end up getting called hundreds of times. So far we mostly just made the private property (`_participantRelationsCount`) public to directly access it in the template.

src/app/child-dev-project/schools/display-participants-count/display-participants-count.component.ts
```diff
@@ -0,0 +1,53 @@
...
+  extends ViewDirective<any>
+  implements OnChanges
+{
+  get participantRelationsCount(): number {
+    return this._participantRelationsCount;
+  }
+
+  set participantRelationsCount(value: number) {
+    this._participantRelationsCount = value;
+    this._changeDetectorRef.markForCheck();
```
TheSlimvReal, line 25: Good idea to connect this with a setter.

src/app/child-dev-project/schools/display-participants-count/display-participants-count.component.ts
```diff
@@ -0,0 +1,53 @@
...
+      .queryActiveRelationsOf("school", this.entity.getId())
+      .then(
+        (relations: ChildSchoolRelation[]) => {
+          this.participantRelationsCount = relations.length;
+        },
+        (reason) => {
+          console.log(
+            "Could not calculate participantRelationsCount, error response from ChildrenService.",
+            reason,
+          );
```
TheSlimvReal, line 49: Is there a reason why you use this promise syntax? So far we always you the distinct `then` and `catch` functions. I think this makes it much easier to understand what each function belongs to. In rxjs subscribers they also have deprecated the syntax for multiple functions as arguments.

src/app/core/entity/default-datatype/view.directive.ts
```diff
@@ -12,6 +12,12 @@ export abstract class ViewDirective<T, C = any> implements OnChanges {
   /** indicating that the value is not in its original state, so that components can explain this to the user */
   isPartiallyAnonymized: boolean;
 
+  /**
+   * Attention:
+   * When content is loaded async in your child component, you need to manually trigger the change detection
+   * See: https://angularindepth.com/posts/1054/here-is-what-you-need-to-know-about-dynamic-components-in-angular#ngonchanges
+   *
+   */
```
TheSlimvReal, line 20: After reading through this blogpost  I had another idea how me might be able to get rid of this issue in our dynamic components:
We could also `await` the `ngOnChanges` call [where we create the dynamic component](https://github.com/Aam-Digital/ndb-core/blob/master/src/app/core/config/dynamic-components/dynamic-component.directive.ts#L94).
This would make the `markForCheck` call unnecessary and it also mirrors the "normal" behaviour when using the component in a template and passing the entity as a input parameter.

src/app/core/config/config-fix.ts
```diff
@@ -309,6 +309,7 @@ export const defaultJsonConfig = {
       "entity": "School",
       "columns": [
         "name",
+        { id: "DisplayParticipantsCount", viewComponent: "DisplayParticipantsCount", label: "Active Participants" },
```
TheSlimvReal, line 312:
```suggestion
        { id: "DisplayParticipantsCount", viewComponent: "DisplayParticipantsCount", label: $localize`Schools` },
```
I think this makes it a bit more easier to understand in our demo setup. Also adding `i18n`. The translations we normally add as part of our release process.

TheSlimvReal (changes requested):
Thanks for finishing this so quickly, especially with the thorough testing.

The functionality is great, most comments comments are around code style.

src/app/child-dev-project/schools/display-participants-count/display-participants-count.component.html
tomwwinter: Good point, i remember darkly... To avoid the getter and also to get rid of the ChangeDetection injection, i just use a Signal here. I just read an article some weeks ago, that they got rid of unnecessary calculations within the template when you use Signals.

src/app/child-dev-project/schools/display-participants-count/display-participants-count.component.ts
tomwwinter: No, just a little rusty with JS ;)

Done.

src/app/core/config/config-fix.ts
tomwwinter: Should be "$localize`Students`", right?

src/app/child-dev-project/schools/display-participants-count/display-participants-count.component.ts
```diff
@@ -0,0 +1,47 @@
...
+  override async ngOnChanges(): Promise<void> {
+    super.ngOnChanges();
+
+    return this._childrenService
+      .queryActiveRelationsOf("school", this.entity.getId())
+      .then((relations: ChildSchoolRelation[]) => {
+        this.participantRelationsCount.set(relations.length);
+      })
+      .catch((reason) => {
+        console.log(
```
TheSlimvReal, line 41: Let's use our logging service here

TheSlimvReal (approved):

Issue #2092, opened 2023-11-21
More bulk actions: bulk delete + bulk archive multiple records at once from list


sleidig:
*As a project manager I want to quickly delete or archive a selection of participants without open each ones details view.*

**Objective**
From the list it should be possible to execute actions like delete, archive (and anonymize, if applicable) for a selected set of records together. We have recently implemented a bulk action to duplicate multiple records. This functionality can be extended for additional actions on the selected records.

PR #2154, opened 2024-01-04
More bulk actions: bulk delete + bulk archive multiple records at once from list

christophscheuing:
relates to: #2092


src/app/core/entity/entity-actions/entity-actions.service.spec.ts
```diff
@@ -121,21 +156,113 @@ describe("EntityActionsService", () => {
...
-  it("should archive and save entity", async () => {
-    await service.archive(primaryEntity);
+  it("should anonymize and save a single entity", async () => {
+    // onAction is never called
+    mockSnackBarRef.onAction.and.returnValues(NEVER);
+    mockSnackBarRef.afterDismissed.and.returnValue(of(undefined));
 
-    expect(primaryEntity.isActive).toBeFalse();
-    expect(mockedEntityMapper.save).toHaveBeenCalledWith(primaryEntity);
+    // const testEntity = new Entity();
```
sleidig, line 164:
```suggestion
```

src/app/core/entity/entity-actions/entity-actions.service.ts
```diff
@@ -59,22 +60,39 @@ export class EntityActionsService {
    *
    * This also triggers a toast message, enabling the user to undo the action.
    *
-   * @param entity The entity to remove
+   * @param entityParam The entity to remove
```
sleidig, line 63:
```suggestion
   * @param entityParam The entity or entities to remove
```

Also, I think "entities" would be more clear than entityParam as a name

src/app/core/entity/entity-actions/entity-actions.service.ts
```diff
@@ -140,24 +186,28 @@ export class EntityActionsService {
...
+    }
     progressDialogRef.close();
 
     if (result.potentiallyRetainingPII.length > 0) {
       await this.confirmationDialog.getConfirmation(
         $localize`:post-anonymize related PII warning title:Related records may still contain personal data`,
-        $localize`:post-anonymize related PII warning dialog:Some related records (e.g. notes) may still contain personal data in their text. We have automatically anonymized all records that are linked to ONLY this ${
-          entity.getConstructor().label
-        }.
+        $localize`:post-anonymize related PII warning dialog:Some related records (e.g. notes) may still contain personal data in their text. We have automatically anonymized all records that are linked to ONLY ${textForAnonymizeEntity}.
```
sleidig, line 198: this doesn't work here, because the sentence structure is different (for delete it was the same structure).

src/app/core/entity/entity-actions/entity-actions.service.spec.ts
```diff
@@ -80,19 +92,42 @@ describe("EntityActionsService", () => {
...
     expect(result).toBe(true);
     expect(snackBarSpy.open).toHaveBeenCalled();
-    expect(mockedEntityDeleteService.deleteEntity).toHaveBeenCalled();
+    expect(mockedEntityDeleteService.deleteEntity).toHaveBeenCalledWith(
+      primaryEntity,
+    );
     expect(mockRouter.navigate).toHaveBeenCalled();
   });
 
+  it("should delete several entities and show snackbar confirmation", async () => {
```
sleidig, line 110: We should add a test to confirm that the undo function actually includes all the original entities if multiple were acted on

sleidig (changes requested):
Looks really good in code (although the message generation has some complex layers :sweat_smile: ) and works smoothly.

From Testing (almost all good):
- bulk actions can be activated on EntityList and special lists like ChildrenList and trigger as expected
- archive works
- anonymize works
  - [x] notification text problem: "[..] linked to ONLY these 2 Children have ." 
- delete works
- undo of bulk action through snackbar works

src/app/core/entity/entity-actions/entity-actions.service.ts
sleidig: but actually you use entityParam everywhere consistently, so let's leave that as is.

src/app/core/entity/entity-actions/entity-actions.service.spec.ts
christophscheuing: I added such tests for the deletion and anonymization of several entities.

src/app/core/entity/entity-actions/entity-actions.service.ts
christophscheuing: I fixed that.

sleidig (approved):
:100:

PR #2160, opened 2024-01-08
Delete docker pr tags

TheSlimvReal:
Currently, the PR and pre-release tags stay on [Docker Hub](https://hub.docker.com/repository/docker/aamdigital/ndb-server/tags?page=1&ordering=last_updated).

- [x] delete PR tags from Docker Hub when PR is closed
- [x] delete pre-release tag from Docker Hub when new release is created

I have already used the same scripts to delete all old PR and pre-release tags manually.


tomwwinter (approved):
LGTM

sleidig (approved):
Looks good (although of course I can't run the CI in my head and know it'll run :wink: )

Just as a general note (especially for @tomwwinter ): We do also automatically remove GitHub pre-releases with the CI, but that is its own GitHub Action yml.
But I think as they have different triggers, they will not fit into a joint file.

PR #2174, opened 2024-01-11
fix(core): related-entities can handle records where the mapped field…

### LINKED ISSUE #2172: RelatedEntities component loads with error if some records are missing the field
If a RelatedEntities table is displayed on the details screen and some of the records to be loaded there are missing the mapping "property" field, then that component loads indefinetely (throwing an error trying to read ".includes" on undefined)

sleidig:
… when it is an array property but undefined

fixes #2172


tomwwinter (approved):
LGTM

Issue #1912, opened 2023-06-28
Field level permissions


TheSlimvReal:
*As a project manager I want to make certain fields readonly for some users in order to not allowing them to edit it.*
This request has come up with some organisation. 

**Objective**
At the moment we can only provide permissions on a whole entity.
This should extend this functionality so that also certain fields can be enabled/disabled for some user roles.

**Proposed Solutions & Alternatives**
[CASL](https://casl.js.org) which we use do define our permissions already allows to define so called `fields` to which a certain rule applies.
This theoretically already covers what we need. 
To make this feature more usable, this should also result in a adaption of the UI.
Fields, which the user is not allowed to edit should stay in readonly mode when the edit button is clicked.
To detect which fields a user is not allowed to edit, CASL provides the function [permittedFieldsOf](https://casl.js.org/v6/en/api/casl-ability-extra#permitted-fields-of) to find out which fields a user can perform a certain action on.

In the backend this should also be properly implemented.
At the moment this case is only covered in the single document `POST` endpoint.

**Steps**
- [ ] Individual properties should stay disabled in forms if user has no "create" or "update" permissions
- [ ] Implement property based write in the [backend](https://github.com/Aam-Digital/replication-backend) for all endpoints receiving documents
- [ ] Individual properties should not be shown in forms if user has no "read" permissions
- [ ] Implement property based "read" in backend for all endpoints sending documents

PR #2180, opened 2024-01-15
feat(permissions): filter fields in forms based on read/write permissions

tomwwinter:
- filter fields in forms with no read permission
- disable fields in edit mode with read-only permissions

Can be tested with this permissions:

```js
const rules: DatabaseRules = {
  user_app: [
    { subject: "Child", action: "read", fields: ["name"] },
    { subject: "Child", action: "manage", fields: ["name", "dateOfBirth"] },
  ],
  admin_app: [{ subject: "all", action: "manage" }],
}
```

related to: #1912

### Visible/Frontend Changes
- [x] filter fields with now read permission in detail view 
- [x] disable fields with no write access in edit mode 

- [ ] none


src/app/core/entity-details/form/form.component.ts
```diff
@@ -32,28 +34,49 @@ import { FieldGroup } from "./field-group";
...
     private router: Router,
     private location: Location,
     private entityFormService: EntityFormService,
     private alertService: AlertService,
+    private abilityService: AbilityService,
   ) {}
 
+  ngOnDestroy(): void {
+    this._destroyed.next();
+    this._destroyed.complete();
```
TheSlimvReal, line 61: We use the [@ngneat/until-destroy](https://www.npmjs.com/package/@ngneat/until-destroy) package to make the unsubscribing a little easier (e.g. see the `NoteDetailsComponent` on how to use it)

src/app/core/entity-details/form/form.component.ts
```diff
@@ -82,6 +105,25 @@ export class FormComponent<E extends Entity> implements FormConfig, OnInit {
     this.entityFormService.resetForm(this.form, this.entity);
     this.form.disable();
   }
+
+  private filterFieldGroupsByPermissions() {
```
TheSlimvReal, line 109: Do you see any way on how we can move (most of) this logic into the `EntityFormService` or the `EntityFormComponent` so we can re-use the functionality in all forms across the map?

src/app/core/entity-details/form/form.component.ts
```diff
@@ -82,6 +105,25 @@ export class FormComponent<E extends Entity> implements FormConfig, OnInit {
...
+          this.abilityService.hasReadPermission(this.entity, field),
+        );
+        return group;
+      })
+      .filter((group) => group.fields.length !== 0);
+  }
+
+  private disableReadOnlyFormControls() {
+    Object.keys(this.form.controls).forEach((fieldId) => {
+      if (!this.abilityService.hasEditPermission(this.entity, fieldId)) {
```
TheSlimvReal, line 122:
```suggestion
      if (!this.ability.can('read', this.entity, fieldId)) {
```
This way you wouldn't event need the extra functions in the ability service.

TheSlimvReal (changes requested):
Thanks for the goods start already. The functionality looks good, lets try to refactor this so we can use the same approach in all forms.

src/app/core/entity-details/form/form.component.ts
tomwwinter: done

src/app/core/permissions/ability/ability.service.ts
```diff
@@ -27,6 +39,14 @@ export class AbilityService extends LatestEntityLoader<Config<DatabaseRules>> {
     super(Config, Config.PERMISSION_KEY, entityMapper, logger);
   }
 
+  hasReadPermission(entity: Entity, field: string | FormFieldConfig): boolean {
+    return this._hasFieldPermission(entity, field, this.READ_PERMISSION_GROUP);
```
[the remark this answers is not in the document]
tomwwinter: done.

src/app/core/entity-details/form/form.component.ts
tomwwinter: Logic moved to EntityFormComponent and EntityFormService

src/app/core/entity-details/form/form.component.ts
tomwwinter: done

src/app/core/common-components/entity-form/entity-form/entity-form.component.ts
```diff
@@ -78,6 +81,31 @@ export class EntityFormComponent<T extends Entity = Entity>
...
+  private applyFormFieldPermissions() {
+    if (this.fieldGroups) {
+      this.fieldGroups = this.entityFormService.filterFieldGroupsByPermissions(
+        this.fieldGroups,
+        this.entity,
+      );
+    }
+
+    if (this.form) {
+      this.form.statusChanges.pipe(untilDestroyed(this)).subscribe((_) => {
```
TheSlimvReal, line 102: I guess this does not need to be run for every status change (e.g. not when the status changes to `disabled`).

src/app/core/common-components/entity-select/entity-select.component.spec.ts
```diff
@@ -36,6 +38,8 @@ describe("EntitySelectComponent", () => {
         ]),
       ],
     }).compileComponents();
+    entityAbility = TestBed.inject(EntityAbility);
+    spyOn(entityAbility, "can").and.returnValue(true);
```
TheSlimvReal, line 42: Thats a bit annoying, maybe we should add the default "everything is allowed"-rule to the `MockedTestingModule`.

src/app/core/common-components/entity-form/entity-form.service.ts
```diff
@@ -78,6 +79,35 @@ export class EntityFormService {
     }
   }
 
+  filterFieldGroupsByPermissions<T extends Entity = Entity>(
+    fieldGroups: FieldGroup[],
```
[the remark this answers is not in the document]
tomwwinter: It's a lot of logic for be handled inside the component, so i would keep inside the service. But lets discuss this in a call.

src/app/core/common-components/entity-form/entity-form/entity-form.component.ts
tomwwinter: I will add a check for disabled state.

src/app/core/common-components/entity-select/entity-select.component.spec.ts
tomwwinter: Problem is also that our "unit tests" are more like integration tests with a lot of real dependencies... But yes, the default could be to allow everything.

src/app/core/common-components/entity-form/entity-form/entity-form.component.ts
```diff
@@ -78,6 +81,31 @@ export class EntityFormComponent<T extends Entity = Entity>
...
+        this.entityFormService.disableReadOnlyFormControls(
+          this.form,
+          this.entity,
+        );
+      });
+    }
```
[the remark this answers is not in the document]
tomwwinter: Implemented

src/app/core/common-components/entity-form/entity-form.service.spec.ts
```diff
@@ -54,6 +55,46 @@ describe("EntityFormService", () => {
...
+    TestBed.inject(EntityAbility).update([
+      { subject: "Entity", action: "update", fields: ["foo"] },
+    ]);
+
+    service.disableReadOnlyFormControls(formGroup, entity);
+
+    expect(formGroup.get("name").disabled).toBeTruthy();
+    expect(formGroup.get("foo").disabled).toBeFalse();
+    expect(formGroup.get("bar").disabled).toBeTruthy();
+  });
```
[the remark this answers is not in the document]
tomwwinter: Test created

src/app/core/common-components/entity-form/entity-form/entity-form.component.ts
```diff
@@ -78,6 +83,22 @@ export class EntityFormComponent<T extends Entity = Entity>
       this.initialFormValues = this.form.getRawValue();
       this.disableForLockedEntity();
     }
+
+    if (this.form && this.entity) {
+      this.entityFormService.disableReadOnlyFormControls(
```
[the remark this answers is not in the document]
tomwwinter: Good point, i forgot to remove it, after moved to the service ;)

src/app/core/common-components/entity-form/entity-form/entity-form.component.ts
```diff
@@ -78,6 +83,22 @@ export class EntityFormComponent<T extends Entity = Entity>
...
+        this.form,
+        this.entity,
+      );
+    }
+  }
+
+  private applyFormFieldPermissions() {
```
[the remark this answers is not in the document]
tomwwinter: The function was quite long and therefore easier to read. However, as it has now become smaller, I have dragged it back into the function.

TheSlimvReal (commented):
Code looks very good now. I only changed some small things. When playing with it I discovered a functionality issue though:

Currently the `disabled` state is always applied based on the `update` permissions. This means that the property is also disabled when a new entity (e.g. a new Note) is created even if the user has `create` permissions on the property. I think there is a valid use case where users are allowed to set a property initially, but once the entity already exists they are not allowed to change it anymore. Therefore I think for new entities the disabled state should be evaluated based on the `create` permission for a field.

TheSlimvReal (approved):
Works great now. Thanks.

Issue #2066, opened 2023-11-08
Migrate configs


TheSlimvReal:
This is a collection issue to collect potential places where it would make sense to migrate the config. Once we do a migration, all of these should be included.

- [ ] `EntityListComponent` rename property `columnGroups` to `columnGroupsConfig` and apply the same to the config property and the interface `EntityListConfig` (see [comment](https://github.com/Aam-Digital/ndb-core/pull/2064#discussion_r1386219013))
- [x] Align `NavigationMenuConfig` with `MenuItem` interface. Currently one uses `name` and the other uses `label`. The `DashboardShortcutWidgets` already uses the `MenuItem` interface.
- [x] Add subproperty for `HistoricalDataComponent` (see #2187)

### To discuss
- switch the config model a bit: one config doc per entity type? Would allow more useful typing of the Config entity also and group things in a useful way:
```
class Config extends Entity {
  @DatabaseField() entitySchema: EntitySchema;
  @DatabaseField() viewDetails: EntityDetailsConfig;
  @DatabaseField() viewList: EntityListConfig;
}
```

sleidig:
moved the last remaining config migration point to the related Admin UI ticket: #2067

Let's close this and open new individual issues where needed. With the ConfigService migration function it has become fairly easy to just implement an on-the-fly migration also.

PR #2187, opened 2024-01-22
fix: show new health an observation records immediately

### LINKED ISSUE #2186: [Regression] New health checks are not directly added to table
**To Reproduce**
1. Go to 'Child details'
2. Click on 'Health'
3. Click '+' button
4. Fill data
5. Click 'save'
6. New entry is not visible in table
7. After re-entering the view, the entry is visible

**Expected behavior**
Entry should instantly be visible.

**Additional context**
This is probably because of the changes in #2142. Now the changes listener is only applied in the `RelatedEntitiesComponent`, which is not (yet) used in the `HealthCheckComponent`, and not in the entities table as before.
Its probably easiest to also start using `RelatedEntitiesComponent` and check whether other components might also be affected by the same issue.

sleidig:
fixes #2186

### DIFF
--- a/src/app/core/config/config-fix.ts
+++ b/src/app/core/config/config-fix.ts
@@ -640,14 +640,16 @@ export const defaultJsonConfig = {
             {
               "title": "",
               "component": "HistoricalDataComponent",
-              "config": [
-                "date",
-                { "id": "isMotivatedDuringClass", "visibleFrom": "lg" },
-                { "id": "isParticipatingInClass", "visibleFrom": "lg" },
-                { "id": "isInteractingWithOthers", "visibleFrom": "lg" },
-                { "id": "doesHomework", "visibleFrom": "lg" },
-                { "id": "asksQuestions", "visibleFrom": "lg" },
-              ]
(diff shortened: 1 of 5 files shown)


src/app/core/config/config-fix.ts
```diff
@@ -640,14 +640,16 @@ export const defaultJsonConfig = {
...
-              "config": [
-                "date",
-                { "id": "isMotivatedDuringClass", "visibleFrom": "lg" },
-                { "id": "isParticipatingInClass", "visibleFrom": "lg" },
-                { "id": "isInteractingWithOthers", "visibleFrom": "lg" },
-                { "id": "doesHomework", "visibleFrom": "lg" },
-                { "id": "asksQuestions", "visibleFrom": "lg" },
-              ]
+              "config": {
+                "columns": [
```
TheSlimvReal, line 644: Seems to be backwards compatible but I added the migration to our list in #2066

TheSlimvReal (approved):
Thanks for the quick fix

PR #2190, opened 2024-01-23
fix: plus button in notes / tasks of child correctly opens popup again

sleidig:
closes #2188

PR #2192, opened 2024-01-23
[Simon] fix: plus button in notes / tasks of child correctly opens popup again 

### LINKED ISSUE #2188: '+' button not working in `TodosRelatedToEntity` and `NotesRelatedToEntity`
**To Reproduce**
1. Go to Child Details
2. Go to 'Notes and Tasks'
3. Click on '+' in any table
4. Nothing happens

**Expected behavior**
It should open the custom popup for the respective entity.

**Additional context**
It only seems to happen in tables where a custom popup component is used (`NoteDetails` / `TodoDetails`) instead of the generic `RowDetailsComponent`.

TheSlimvReal:
Closes #2188
Alternative to #2190

TODO
- [ ] Changes after Checked in `AttendanceDetailsComponent`

### DIFF
--- a/src/app/child-dev-project/attendance/activities-overview/activities-overview.component.ts
+++ b/src/app/child-dev-project/attendance/activities-overview/activities-overview.component.ts
@@ -20,7 +20,7 @@ export class ActivitiesOverviewComponent
   extends RelatedEntitiesComponent<RecurringActivity>
   implements OnInit
 {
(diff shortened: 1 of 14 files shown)


sleidig (approved):
I like your implementation. Seems to work well - I did not find a "Changed after checked" warning.

The updates of new notes and todos currently throws in everything - even unrelated entities. But I would hope to fix this with further index generalization at some point, so maybe acceptable right now.

PR #2193, opened 2024-01-23
refactor(core): migrate navigationMenu config format

sleidig:
to align with actually used MenuItem interface

see #2066

---
once this has been released, we can run the set of migrations in #2066 server-side and remove them from the client-side code


TheSlimvReal (approved):
Thanks for the cleanup. Looks good.

Issue #1999, opened 2023-09-14
CSP warnings in real operation


sleidig:
After deploying the CSP (report-only), some content is still triggering warnings and not properly whitelisted in our default CSP:

- Refused to load the image 'blob:URL'
- Fetch API cannot load https://o167951.ingest.sentry.io/api/1242399/security/. Request mode is "no-cors" but the redirect mode is not "follow".

TheSlimvReal:
One problem is that the `index.html` script hash has not been correctly updated. How can this be done?

sleidig:
I documented this here in den Developer Docs "Security" Concept (happy for suggestions about a better place): https://aam-digital.github.io/ndb-core/documentation/additional-documentation/concepts/security.html

TheSlimvReal:
The instructions dont seem to work on all browsers (this is from Chrome on Mac, Safari seems to show even less info)
<img width="619" alt="Screenshot 2024-01-22 at 09 20 49" src="https://github.com/Aam-Digital/ndb-core/assets/33730997/214cb69d-ca26-44a0-8952-610c2a18632b">

PR #2194, opened 2024-01-23
some CSP fixes

sleidig:
see #1999 

---
allow blob images and fix angular blocking CSP reports going to sentry

CSP configuration is a huge pain ***  I still somehow couldn't get a new hash for the index.html script. And setting up nginx with nonces also doesn't look trivial ...


TheSlimvReal:
The security endpoint seems to work now but as long as we don't have the correct hash, this still will send a bunch of reports to sentry and also mill the console. So is this really something we want to keep enabled?

TheSlimvReal:
The security endpoint seems to work now but as long as we don't have the correct hash, this still will send a bunch of reports to sentry and also mill the console. So is this really something we want to keep enabled?

TheSlimvReal (approved):

Issue #2202, opened 2024-01-29
New field-level permissions break public forms if only "create" permissions without "read" are given


sleidig:
Through public forms we allow users to create entities without being logged in (e.g. for feedback surveys, self registration, etc.). This usually implies permissions to "create" without permissions to "read" existing records.

With the recent changes in PR #2180, introducing field-level permissions, this is not possible anymore: Without "read" permissions, the client shows an empty form

https://github.com/Aam-Digital/ndb-core/blob/017216b1395e79ca2a9579dbf2bc37bfc5736eee/src/app/core/common-components/entity-form/entity-form/entity-form.component.ts#L142-L153

PR #2203, opened 2024-01-29
fix: use create permissions when read public form fields

tomwwinter:
closes: #2202 

### Visible/Frontend Changes
- [ ] public forms should be usable again


TheSlimvReal (approved):

sleidig (approved):
Works :+1: 

Thanks for the super quick fix!

Issue #2198, opened 2024-01-25
EntityCount Dashboard should display human-readable categories when disaggregating by a referenced entity


sleidig:
![image](https://github.com/Aam-Digital/ndb-core/assets/1682541/b936c5ed-6d21-47a5-97a1-c57a78d2b744)

The grouping by a property in the EntityCountDashboard currently displays either .label or the string of the groupBy property. This means grouping by an entity reference (which is a string uuid on the entities) is not displayed well.

--> use toString or EntityBlockComponent

sleidig:
duplicate of #2002

Issue #2002, opened 2023-09-14
EntityCountDashboard should use viewComponent for labels in list?


sleidig:
In some cases disaggregation for a referenced entity would be nice, but this requires displaying it better:
![image](https://github.com/Aam-Digital/ndb-core/assets/1682541/bfa3cf2e-4e73-4949-b1cc-cd3b41267455)

PR #2205, opened 2024-01-29
Count by entity in Dashboard

sleidig:
closes: #2198 #2002 

- [x] display entity block if grouped by entity field
- [ ] check whether we can generalize the displaying using viewComponent instead of condition to use entity display
- [x] groupBy array properties, disaggregating summed up for each individual item (e.g. per category, also supporting multiple categories in the counted entities)

### DIFF
--- a/src/app/core/config/config-fix.ts
+++ b/src/app/core/config/config-fix.ts
@@ -96,11 +96,7 @@ export const defaultJsonConfig = {
           }
         },
         {
-          "component": "EntityCountDashboard",
-          "config": {
-            "entity": "Note",
-            "groupBy": "children"
-          }
+          "component": "EntityCountDashboard"
         },
         {
           "component": "ImportantNotesDashboard",
--- a/src/app/core/config/config-fix.ts
+++ b/src/app/core/config/config-fix.ts
@@ -96,7 +96,11 @@ export const defaultJsonConfig = {
           }
         },
         {
-          "component": "EntityCountDashboard"
+          "component": "EntityCountDashboard",
+          "config": {
+            "entity": "Note",
+            "groupBy": "children"
+          }
         },
         {
           "component": "ImportantNotesDashboard",
--- a/src/app/core/config/config-fix.ts
+++ b/src/app/core/config/config-fix.ts
@@ -705,7 +705,7 @@ export const defaultJsonConfig = {
               "config": {
                 "fieldGroups": [
                   { "fields": ["title"] },
-                  { "fields": ["type", "inactive"] },
+                  { "fields": ["type"] },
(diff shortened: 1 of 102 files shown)


src/app/core/config/config-fix.ts
```diff
@@ -96,7 +96,11 @@ export const defaultJsonConfig = {
           }
         },
         {
-          "component": "EntityCountDashboard"
+          "component": "EntityCountDashboard",
+          "config": {
+            "entity": "Note",
+            "groupBy": "children"
+          }
```
sleidig, line 103:
```suggestion
          "component": "EntityCountDashboard"
```

src/app/features/dashboard-widgets/entity-count-dashboard-widget/entity-count-dashboard/entity-count-dashboard.component.ts
```diff
@@ -52,14 +58,21 @@ export class EntityCountDashboardComponent
     this._entity = this.entities.get(value);
   }
 
-  private _entity: EntityConstructor = Child;
+  protected _entity: EntityConstructor = Child;
```
TheSlimvReal, line 61:
```suggestion
  private _entity: EntityConstructor = Child;
```

src/app/utils/utils.ts
```diff
@@ -40,25 +40,36 @@ export function getParentUrl(router: Router): string {
...
- * @param propertyToGroupBy The key of the property in the elements by whose value the result is grouped.
+ * @param propertyToGroupBy The key of the property in the elements by whose value the result is grouped. If this is an array, grouped by individual elements.
  * @returns an array where the first entry is the value of this group and the second all entries that have this value.
  */
-export function groupBy<T, P extends keyof T>(
+export function groupBy<T, P extends keyof T, E>(
   array: T[],
   propertyToGroupBy: P,
-): [T[P], T[]][] {
+): [T[P] extends Array<E> ? E | undefined : T[P], T[]][] {
```
TheSlimvReal, line 49:
```suggestion
): [T[P] extends Array<infer E> ? E | undefined : T[P], T[]][] {
```
or without `E`
```suggestion
): [T[P] extends Array<any> ? T[P][number] | undefined : T[P], T[]][] {
```
Otherwise I get `unknown` as the type for the key. With this approach I correctly get string in the test (as long as I correctly type the empty array as string too). 
See https://stackoverflow.com/a/50693866/10713841

TheSlimvReal (approved):
Works very well. Also the filtering when clicking on the button at the side of each entry works, but only if the filter is available in the list component (in this case children filter on notes manager). But I think that problem might be fixed once we make the filters more self-configurable?

PR #2209, opened 2024-02-01
Fixed bug of navigating to entity details even in selectable mode

christophscheuing:
When in selectable mode in an entity table, clicking the checkbox navigated to the details of that entity. This is fixed here.

It is also added here that clicking on a row in selectable mode has the same effect as clicking the checkbox.


sleidig (approved):
:100:

PR #2221, opened 2024-02-06
fix: align functionality and UX of single and multi entity-select fields

### LINKED ISSUE #2175: EditSingleEntity should probably use EntitySelectComponent
Currently, the implementation for single entity reference fields (EditSingleEntity) is significantly different from the multi-select entity references (EditEntityArray). We do have a general component for entity selection (EntitySelectComponent) and it would be best to use this in both edit fields to have consistent UX and shared functionality (e.g. the hiding of inactive records until the user explicitly includes them).

Implementation TODOs:
- [x] Add an input for EntitySelectComponent to function as a single select (currently always multi-select).
  - [x] Do not show additional "add other ..." field if selected.
  - [x] Open dropdown on click into the field. unselect existing when selecting a new record.
  - [x] handle array or non-array input/output for value.
- [x] Refactoring of EditSingleEntity is required, using EntitySelectComponent instead of directly building it with app-basic-autocomplete

### LINKED ISSUE #2212: "entity" datatype (DisplayEntity, EditSingleEntity) does not support multiple entity types
**Describe the bug**
`entity-array` datatype fields can be configured to allow a selection of multiple entity types (e.g. select a mix of Child and School records in the same field). This is done by configuring and array for the "additional" schema field config: `additional: ["Child", "School"]`

When configuring the same for an `entity` datatype, the references are not found.

sleidig:
closes #2175, closes #2212

- [x] replace EditSingleEntity with a generalized implementation of EditEntityArray --> now EditEntity
- [x] replace the custom autocomplete implementation of entity-select by using basic-autocomplete
- [x] extend basic-autocomplete to offer some UI variant to display chips and templates
- [x] decide what tests still need to be moved from entity-select and which are obsolete and to be removed
- [x] improve the "include inactive" toggle back to its full UX, by implementing an Output in basic-autocomplete that shares the current autocomplete filter (?)


src/app/core/basic-datatypes/entity-array/edit-entity/edit-entity.component.ts
```diff
@@ -2,28 +2,40 @@ import { Component, Input, OnInit } from "@angular/core";
...
+  multi: boolean = false;
+
   ngOnInit() {
     super.ngOnInit();
 
     this.entityName = this.formFieldConfig.additional;
 
+    this.multi =
+      this.formFieldConfig.dataType === ArrayDatatype.dataType ||
+      this.formFieldConfig.dataType === EntityArrayDatatype.dataType;
```
TheSlimvReal, line 37:
```suggestion
    this.multi = isArrayDataType(this.formFieldConfig.dataType);
```

src/app/core/common-components/entity-select/entity-select.component.html
```diff
@@ -49,15 +52,21 @@
...
     <mat-option
       *ngIf="
         filteredEntities.length < 1 &&
         !includeInactive &&
         inactiveFilteredEntities.length > 0
       "
       style="display: none"
-    ></mat-option>
+    >
+      <!-- The following mat-option is never displayed ("display: none")
```
TheSlimvReal, line 67:
```suggestion
      <!-- This mat-option is never displayed ("display: none")
```

src/app/core/common-components/entity-select/entity-select.component.ts
```diff
@@ -68,61 +73,70 @@ export class EntitySelectComponent<E extends Entity> implements OnChanges {
...
-   * @param sel The initial selection
+   * @param sel The initial selection (single id string for multi=false, array of id strings for multi=true)
    */
-  @Input() set selection(sel: string[]) {
-    if (!Array.isArray(sel)) {
-      this.selectedEntities = [];
-      return;
-    }
+  @Input() set selection(sel: T) {
+    const selArray = Array.isArray(sel) ? sel : [sel];
```
TheSlimvReal, line 87: There might be an issue when passing `undefined` which will result in `[undefined]` which might not result in the correct behavior.

src/app/core/common-components/entity-select/entity-select.component.ts
```diff
@@ -297,10 +312,22 @@ export class EntitySelectComponent<E extends Entity> implements OnChanges {
   }
 
   private emitChange() {
-    this.selectionChange.emit(this.selectedEntities.map((e) => e.getId()));
+    const newSelection = this.selectedEntities.map((e) => e.getId());
+
+    if (isMulti(this)) {
+      this.selectionChange.emit(newSelection);
+    } else {
+      this.selectionChange.emit(newSelection[0] as T);
```
TheSlimvReal, line 320: A bit annoying that the correct type `T = string` is not picked up for the `false` case here.

TheSlimvReal (changes requested):
Nice how little code was required to merge these two cases.

- [ ] The component does not seem to display errors (Child > Education > add new school history and press save without selecting a school)

sleidig:
> * [ ]  The component does not seem to display errors (Child > Education > add new school history and press save without selecting a school)

Thanks for discovering this! This seemingly minor flaw turned out to be a major challenge ... After a lot of forth and back with different ideas, the most feasible and maintainable approach for me is the rewrite the entity-select to use basic-dropdown component. A major change but consolidating even more duplicate code and UX.

src/app/core/common-components/entity-select/entity-select.component.ts
```diff
@@ -45,16 +35,22 @@ import { LoggingService } from "../../logging/logging.service";
...
+    FormsModule,
   ],
   standalone: true,
 })
 @UntilDestroy()
-export class EntitySelectComponent<E extends Entity> implements OnChanges {
-  readonly separatorKeysCodes: number[] = [ENTER, COMMA];
+export class EntitySelectComponent<
+  E extends Entity,
+  T extends string[] | string = string[],
```
sleidig, line 49: This is a bit annoying for typing to make things work as a single and multi select, although the code and template works very well for both ...
If you have any ideas for improvement here, @tomwwinter ,  happy to hear them!

tomwwinter: Is there a way to pass always an array?

tomwwinter (approved):

src/app/core/common-components/entity-select/entity-select.component.ts
sleidig: This component is used in `EditEntityComponent`, which in turn already receives a `FormControl` from parent through which the value input and output is handled (and which maps to either a single (non-array) entity property or a multi entity-array property). We could maybe use value change subscriptions to transform the value a level higher or create a internal formControl and connect that ... but because the formControl is created a few levels higher up and also contains details like validators, I am a bit wary of the side effects of that.

PR #2227, opened 2024-02-09
refactor: remove school-block and always display icon if available

sleidig:
merge this before #2221 as a partial refactoring


tomwwinter (approved):
LGTM

Issue #2222, opened 2024-02-06
Multi-status filter dropdown should offer "select all" / "select none"


sleidig:
_As a user, I want to quickly change what gets filtered in a list in different ways: See almost all values and just exclude a few categories; Or see only one or two selected categories and hide all others._

We recently implemented filters to allow multiple values to be selected:
![image](https://github.com/Aam-Digital/ndb-core/assets/1682541/dcd9a3be-e1c3-4f73-94a3-2fea24d1985d)

If the user wants to switch between selecting many and few filter values, this can become cumbersome. Other systems offer some convenience functions to "select all" and "unselect" or "clear" selected values. e.g.: 
![image](https://github.com/Aam-Digital/ndb-core/assets/1682541/b1d4ee3e-3504-479b-806b-c851de404858)

Extend the `ListFilterComponent` to show two small buttoms above the list of filter options to "select all" and "clear".

PR #2245, opened 2024-02-19
Added clear and select-all button

sadaf895:
closes: #2222


sleidig (changes requested):
Works well and has nice, simple code :+1: 

The UI looks a bit bulky still ... 
I would try using a ["flat" button](https://material.angular.io/components/button/overview) (wihtout the borders the interface would probably be calmer) and letting them fill the whole width. Maybe also experiment with making the font-size a little smaller and/or adding `color="accent"` and check if that would look good (to make a bigger visual difference of the buttons with the actual options below)

sadaf895:
I have given some style to the buttons...Is it looking good or any specific modification required or you want to give any suggestion?

sleidig (approved):

PR #2248, opened 2024-02-20
fix: correct fallback value for loadFile on errors

sleidig:


src/app/features/file/couchdb-file.service.spec.ts
```diff
@@ -66,6 +68,12 @@ describe("CouchdbFileService", () => {
           provide: SyncStateSubject,
           useValue: of(SyncState.COMPLETED),
         },
+        {
+          provide: DomSanitizer,
+          useValue: {
+            bypassSecurityTrustUrl: (val: string) => val,
```
tomwwinter, line 74: that sounds not good 🙈

tomwwinter (approved):

Issue #1942, opened 2023-07-24
Import Module: infer column mapping from column header names


sleidig:
The new ImportColumnMappingComponent does not yet automatically map columns to entity properties. We used to infer mappings from the column headers of the imported data. When the column header matches the property label or id, that should be pre-selected as a mapping.

Suggested user story of functionality / requirements:
1. upload a csv file where (some) column headers in the first row exactly match the field name of our Child entity type
2. in the "column mapping" step of the Import Module, the system automatically pre-selects the field for those columns where the id or label of the entity type match
3. If more than one column matches the same field, only the first one is mapped

-----
### Implementation steps:

- [x] Create a simple service that can handle this logic and can be called from ImportMappingComponent
- [x] Write a set of unit tests for this scenario (imported data + entityType --> expected column mapping) for the service
- [x] Implement the automatic column mapping logic
- [x] test in app as part of the overall Import Module flow

sample import file: 
[sample-import.csv](https://github.com/user-attachments/files/16172395/sample-import.csv)

PR #2257, opened 2024-02-23
Import Column Mapping

sadaf895:
closes: #1942


src/app/core/import/import-column-mapping/import-column-mapping.service.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+    service.automaticallySelectMappings(testMapping, entitySchema);
+
+    // check
+    expect(testMapping).toEqual([
+      { column: "name", propertyName: "name" },
+      { column: "phone", propertyName: "phone" },
+      { column: "address", propertyName: "address" },
+      { column: "some other remarks" },
+    ]);
+  });
```
sleidig, line 46: each `it(...)` block is one test case, so it's enough if you duplicate this, rather than having a second `fdescribe(...)` block.

You can see that there are already two test cases within the fdescribe: `it("should be created" ...)` and `it("should map columns to their respective property names automatically" ...)`. Add your new test as a third it() block and clean up this file so it only has a single fdescribe().

src/app/core/import/import-column-mapping/import-column-mapping.service.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+
+  // TODO: test case for different capitalization, etc.
```
sleidig, line 48: one further test case to add here (and then implement the logic for it):

```
  it("should map columns also if they match the field label (instead of field id)", () => {
    // setup
    const testMapping: ColumnMapping[] = [
      { column: "phone number" },
      { column: "name" },
      { column: "address" },
      { column: "some other remarks" },
    ];
    // the schema in Aam Digital defines human-readable labels for each field in addition to the keys:
    const entitySchema: EntitySchema = new Map([
      ["name", { label: "Full Name" }],
      ["phone", { label: "Phone Number" }],
      ["address", { label: "Address" }]
    ]);

    // run
    service.automaticallySelectMappings(testMapping, entitySchema);

    // check
    expect(testMapping).toEqual([
      { column: "name", propertyName: "name" },
      { column: "phone", propertyName: "phone" },
      { column: "address", propertyName: "address" },
      { column: "some other remarks" },
    ]);
  });
```

src/app/core/import/import-column-mapping/import-column-mapping.service.ts
```diff
@@ -0,0 +1,35 @@
+import { Injectable } from "@angular/core";
+import { ColumnMapping } from "../column-mapping";
+import { EntitySchema } from "app/core/entity/schema/entity-schema";
+
+@Injectable({
+  providedIn: "root",
+})
+export class ImportColumnMappingService {
```
sleidig, line 8: I am adding a short comment (JsDoc), so that other developers quickly understand the purpose of this service. That comment is also displayed by IDEs as a tooltip, when written in the `/**` style.

```suggestion
/**
 * Infer most likely entity fields that the column from an imported file could be mapped to.
 * This is used by the Import Module to give users smart pre-selections in the user interface.
 */
@Injectable({
  providedIn: "root",
})
export class ImportColumnMappingService {
```

src/app/core/import/import-column-mapping/import-column-mapping.service.ts
```diff
@@ -0,0 +1,35 @@
...
+  constructor() {}
+
+  automaticallySelectMappings(
+    columnMapping: ColumnMapping[],
+    entitySchema: EntitySchema,
+  ) {
+    const allPropertyNames = Array.from(entitySchema.keys());
+    console.log("property names", allPropertyNames);
```
sleidig, line 16:
```suggestion
```

Let's remove this, leftover from debugging.

sleidig (approved):
Great work :100:

PR #2263, opened 2024-02-27
fix: display form correctly when creating new with permission conditions

### LINKED ISSUE #2262: Permission with condition hides all fields from form
**Describe the bug**
If a user has permissions for an entity that involve a condition, the details form is not showing any form fields at all.

(suspected regression from field-level permissions?)

**To Reproduce**
1. Set a user role permission: e.g. 
```
{
  "subject": "School",
  "action": "manage",
  "conditions": {
    "language": {
      "$in": [
        "Bengali"
      ]
    }
  }
}
```
3. Create a new entity of that type ("Add New" button)
4. The form is completely blank

for a sample, check dev.aam-digital.net with user "demo"

sleidig:
closes #2262


tomwwinter (approved):
lgtm

Issue #2213, opened 2024-02-02
Creating new records from a entity select field


sleidig:
*As a user I want to create a new _school_ directly from the "entity select" field in my _activity_ details view. Currently I can only select existing schools to link them here. To create a new record, I have to navigate to the school list, create it there and then come back to link it here again.*

**Objective**
(Sometimes) users should have an easier way to create new records while editing a related entity, as in the example described here. This should maybe be a flag to enable/disable for the entity-select component through config, however.

The functionality should be supported by entity and entity-array components in a similar way.

**Possible Approach / Steps**
- [x] check implementations of `EditSingleEntity` and `EditEntityArray` and consider aligning there implementations better, to avoid having to implement this feature separately for each component (see #2175)?
- [x] implement a similar approach to the EnumDropdownComponent / EditConfigurableEnumComponent to allow creation of a new record?
- [ ] consider aligning our different approaches to show popup and fullscreen forms, to allow use of the EntityDetailsComponent in a popup here (see #2069)?

---
this is the **second** case in the visualization (first case, see #2201):
![image](https://github.com/Aam-Digital/ndb-core/assets/1682541/bcaa486e-6547-42ef-b794-2e74e7c1d179)

PR #2293, opened 2024-03-14
Create new entity from entity-select fields in modal form

sleidig:
closes: #2213

Allow users to create new entity records from the entity-select field. At the moment only basic popup form supported, in the future this should be extended to the full details view in a modal.

This PR includes a small cleanup that should be merge separately in https://github.com/Aam-Digital/ndb-core/pull/2294

### DIFF
--- a/src/app/core/common-components/entity-select/entity-select.component.ts
+++ b/src/app/core/common-components/entity-select/entity-select.component.ts
@@ -235,7 +235,7 @@ export class EntitySelectComponent<
     if (this._entityType?.length < 1) {
       return;
     }
-    if (this._entityType.length > 0) {
+    if (this._entityType.length > 1) {
       this.logger.warn(
         "EntitySelect with multiple types is always creating a new entity of the first listed type only.",
       );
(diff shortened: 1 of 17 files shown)


src/app/core/common-components/entity-select/entity-select.component.ts
```diff
@@ -226,10 +230,53 @@ export class EntitySelectComponent<
...
+  createNewEntity = async (input: string): Promise<E> => {
+    if (this._entityType?.length < 1) {
+      return;
+    }
+    if (this._entityType.length > 0) {
+      this.logger.warn(
+        "EntitySelect with multiple types is always creating a new entity of the first listed type only.",
+      );
+      // TODO: maybe display an additional popup asking the user to select which type should be created?
+    }
```
tomwwinter, line 243: This always logs an warning, is this intended? Should be ```(this._entityType.length > 1)``` or am i wrong?

tomwwinter (changes requested):

src/app/core/common-components/entity-select/entity-select.component.ts
sleidig: oh, yes!

```suggestion
    if (this._entityType.length > 1) {
      this.logger.warn(
        "EntitySelect with multiple types is always creating a new entity of the first listed type only.",
      );
      // TODO: maybe display an additional popup asking the user to select which type should be created?
    }
```

tomwwinter (approved):
LGTM

PR #2294, opened 2024-03-14
refactor: remove deprecated User fields and functionalities

sleidig:
remove code from User entity class that is not used anymore since we have separated user account logic from the business logic entities.


tomwwinter (approved):
LGTM!

Issue #2151, opened 2024-01-03
Export entity references both as id and as human-readable toString


sleidig:
*When downloading an export to work with data in a spreadsheet, I want to understand entity links easily, reading the names rather than cryptic ids (although those may be needed as well.*

**Proposed Solutions & Alternatives**
for fields with dataType = "entity", export two columns:
- the id (if the data needs to be reliably linked)
- the toString representation of the referenced entity (so that users can intuitively understand the exported data)

--- 
### Implementation steps:
- [ ] write a unit test for DownloadService for the above requirement
- for implementation, check the DownloadService and the DataTransformationService it uses to create the exportable data ...

PR #2295, opened 2024-03-14
Export entity references both as id and as human-readable toString

christophscheuing:
closes: #2151


src/app/core/export/download-service/download.service.ts
```diff
@@ -149,4 +156,31 @@ export class DownloadService {
...
+  private async mapEntity(item: Entity, columnLabels): Promise<Object> {
+    const newItem = {};
+    for (const key in item) {
+      if (columnLabels.has(key)) {
+        newItem[key] = item[key];
+      }
+      if (columnLabels.has(key + "_readable")) {
+        const relatedEntitiesIds: string[] = Array.isArray(item[key])
+          ? item[key]
+          : item[key].split();
```
sleidig, line 169: Why split here?

sleidig (approved):
Good job!
Works well and looks clean.

src/app/core/export/download-service/download.service.ts
christophscheuing: The `item[key].split()` was meant to build an array containing the string as single value. But `[item[key]]` is simpler and shorter.

Issue #2269, opened 2024-03-01
Improve Form Builder UX (Entity Details drag&drop)


sleidig:
The basic drag & drop editor to build entity details view (https://demo.aam-digital.com/admin/entity/Child/details) can be improved further:

- [ ] Sidebar with available fields should not scroll with main content (otherwise user has to drag the "new field" block long distance and scroll while dragging in long forms)
- [ ] form field groups (the boxes containing some fields) should allow to be drag&drop reordered
- [x] allow re-ordering of tabs

![image](https://github.com/Aam-Digital/ndb-core/assets/1682541/04b47bf5-a222-47d8-a364-1f3c370930a0)

### Implementation steps
- adapt `AdminEntityFormComponent` (for field groups and side bar)
- adapt `AdminTabsComponent` (for tabs)

PR #2305, opened 2024-03-19
Added drag functionality in admin tabs

Abhinegi2:
closes: #2269 

### Visible/Frontend Changes
- [x] Added drag icon on the tabs
- [x] We can now drag tabs horizontally
- [x] Created a sort function to drag tabs
- [x] overall self testing

### Description of changes

1. Import Drag and Drop Modules: Added imports for drag and drop functionality from @angular/cdk/drag-drop. This allows you to use Angular's built-in drag-and-drop features for reordering tabs.
2. Added cdkDropList and cdkDrag directives to enable drag-and-drop functionality.
3. HTML file changes - 
- Wrapped the tab content inside a div with cdkDropList directive to make it a drop target for dragging tabs.
- Added drag handle icon for each tab
- Added cddDropListConnectedTo directive to specify which drop lists are considered connected. This allows tabs to be dragged between different instances of the component if needed.
- Utilized cdkDragLockAxis to lock the dragging movement along the x-axis.

4. Added Drag and Drop Logic to the Component.ts file
5. Created a method getAllTabs to generate an array of IDs for all tabs.
6. Implemented a drop method to handle the drop event and update the order of tabs.
7. Ensured that the selected tab is still active after dropping the active tab.
8. Used JSON.stringify and JSON.parse for Angular Ivy render bugs that are occurring when updating the tabs array.


src/app/core/admin/building-blocks/admin-tabs/admin-tabs.component.ts
```diff
@@ -76,4 +78,22 @@ export class AdminTabsComponent<
...
+        allTabs.push("tabs-" + i);
+      }
+    }
+
+    return allTabs;
+  }
+
+  drop(event: CdkDragDrop<string[]>) {
+    const previousIndex = parseInt(event.previousContainer.id.replace("tabs-", ""), 10);
+    const currentIndex = parseInt(event.container.id.replace("tabs-", ""), 10);
```
sleidig, line 94: I think you do not actually need this workaround using ids. The drop event also gives you the previous and current index directly, for example see https://github.com/Aam-Digital/ndb-core/blob/9303825461f4b4bf350e05ea4327a2142dc854c6/src/app/core/admin/admin-entity-details/admin-entity-form/admin-entity-form.component.ts#L142-L151

sleidig (changes requested):
Works fairly well :-)

- [ ] The handle is not shown properly for the active tab yet: Screenshot from 2024-03-20 09-12-27

- [ ] Also, the drop preview (how the element is displayed during dragging, while not lifting the mouse yet) is showing the moved element on top of the next one rather than to the side, but I am not sure how difficult this is to fix as all of that is in the mat-tab layouts.
- [ ] when I drag the currently active tab, after dropping it suddenly another (the new first tab) is selected. The selected tab should ideally remain active

A thought about usability overall, now that I see it in action: What do you think about only making the current active tab draggable (with the handle next to the title form field)? Then there would be a bit less complexity in the UI overall, with less drag handles

src/app/core/admin/building-blocks/admin-tabs/admin-tabs.component.ts
Abhinegi2: Hi @sleidig when I tried this `event.previousContainer.data` and event.container.data` I am not getting the previous and current container data so that's why I am using the `ids` approach for mat tabs.

sleidig: A pity this needs such a workaround but I couldn't find any simpler way either. Good solution then!

Issue #2292, opened 2024-03-14
Filters on "date" fields do not work


sleidig:
**To Reproduce**
1. Go to list view: https://demo.aam-digital.com/child
2. Edit Data Structure and add a filter for the "admission" field to the list (i.e. some field of date datatype); save Data Structure
3. Try to select a value to filter for -> no options shown

**Expected behavior**
similar to the Notes List, a date range filter should be offered

PR #2307, opened 2024-03-19
fix(filter): date-range filter available for any field

sleidig:
and refactoring of filter-component

closes: #2292


tomwwinter (approved):
LGTM

PR #2312, opened 2024-03-20
fix(sync): manually manage continuous sync

### LINKED ISSUE #935: Sync does not work after user is inactive for some time
**Describe the bug**
The remote authentication is not refreshed automatically so if a user is not making changes for 10 minutes, the remote session expires. Any local changes after that are not synced (Network tab of developer console will show a 401 response).

**To Reproduce**
1. Log in
2. Edit an entity --> changes are synced as expected
3. Wait 10 minutes
4. Edit another entity --> changes are not synced to server, request receives 401

**Expected behavior**
As long as the user is logged in the frontend, the remote auth should be automatically kept alive so that syncing always works.

### LINKED ISSUE #2100: Periodical instead of live sync
We currently use the PouchDB/CouchDB live sync functionality. This is very easy to setup and offers very quick entity updates.

The downside of this approach is that we have very little information about which data is already synchronized and in case of an error, how the app should handle it. At the moment we just restart the live sync and hope nothing has been lost on the way.

An alternative approach would be to periodically (e.g. every 15s) trigger a sync ourselves. This would allow us to check which documents havent been synced yet. This also makes error handling much easier because we can then visualize to the user what data hasn't been synced and what might have been the issue.

The issue in our backend where it [keeps using more RAM](https://github.com/Aam-Digital/replication-backend/issues/116) might also be linked to the live sync and could be resolved with this change.

## Approach

1. As a first prototype the live sync should be replaced with a simple periodical sync leaving everything else as it is. This can be used to check whether it resolves the mentioned backend issue.
2. If that resolved the problem the next step would be to add visualizers about unsynced changes and improve the error handling. (#412)

sleidig:
for better error handling
also prevents aborted sync after expired access token

closes #2100, closes #935


tomwwinter (approved):
LGTM

PR #2315, opened 2024-03-22
Added Field Groups drop functionality

Abhinegi2:
closes: #2269 

### Visible/Frontend Changes
- [x] Added drag icon in front of Groups Title
- [x] We can now drag groups.
- [x] Created a function to drag form groups
- [x] overall self-testing

### Description of changes
1.  HTML file changes - 
- Wrapped the group's content inside a div with cdkDropList directive to make it a drop target for dragging groups.
- Added drag handle icon for each group title.
- Added cddDropListConnectedTo directive to specify which drop lists are considered connected. This allows fields to be dragged between different group components if needed.

2. Implement ngAfterViewInit method responsible for initializing and configuring DropLists in the component, including logging their IDs, managing their order, and setting up interactions between them.


sleidig (changes requested):
I have minimized the required code changes, by using/moving cdkDropListGroup instead of calculating connectedTo arrays.

The preview has not been ideal yet (showing on top of each other in your implementation - and now it is even worse ...). 
@Abhinegi2 Could you research how cdk drop previews can be done in grid layouts (if at all)? Changing `admin-grid-layout` class to `flew-row` gives us a nice preview - but breaks the layout otherwise ...

If there is no feasible way for previews in grid layout, we can revert back to a helper container, similar to what you had. And make that a flex-row, so that the preview is display side-by-side instead of on top.

Abhinegi2:
> I have minimized the required code changes, by using/moving cdkDropListGroup instead of calculating connectedTo arrays.
> 
> The preview has not been ideal yet (showing on top of each other in your implementation - and now it is even worse ...). @Abhinegi2 Could you research how cdk drop previews can be done in grid layouts (if at all)? Changing `admin-grid-layout` class to `flew-row` gives us a nice preview - but breaks the layout otherwise ...
> 
> If there is no feasible way for previews in grid layout, we can revert back to a helper container, similar to what you had. And make that a flex-row, so that the preview is display side-by-side instead of on top.

Sure @sleidig  let me do some research on how we can give nice preview

Abhinegi2:
> I have minimized the required code changes, by using/moving cdkDropListGroup instead of calculating connectedTo arrays.
> 
> The preview has not been ideal yet (showing on top of each other in your implementation - and now it is even worse ...). @Abhinegi2 Could you research how cdk drop previews can be done in grid layouts (if at all)? Changing `admin-grid-layout` class to `flew-row` gives us a nice preview - but breaks the layout otherwise ...
> 
> If there is no feasible way for previews in grid layout, we can revert back to a helper container, similar to what you had. And make that a flex-row, so that the preview is display side-by-side instead of on top.

@sleidig can you please check the drop preview now and let me know its good now? - https://pr-2315.aam-digital.net/admin/entity/Child

sleidig:
The grid layout drag&drop is not supported natively, as discussed here: https://github.com/angular/components/issues/13372

There is a working sample for a custom implementation mentioned there that we may implement for ourselves here as well.

sleidig (commented):
Functionality (drag & drop) for filter reordering works very well now. Let's undo the changes to admin-entity-form and finalize + merge that part only. Then open a second PR for the other.

Can we extract the code of the onDropListEnter/Dropped (and maybe even programmatically creating the placeholder element) into an angular directive?

Abhinegi2:
> Functionality (drag & drop) for filter reordering works very well now. Let's undo the changes to admin-entity-form and finalize + merge that part only. Then open a second PR for the other.
> 
> Can we extract the code of the onDropListEnter/Dropped (and maybe even programmatically creating the placeholder element) into an angular directive?

I undo the changes from admin-entity-form and also opened another PR for filter reordering.

sleidig (approved):
:+1:

sleidig (changes requested):
sorry ... wrong PR approved :see_no_evil:

PR #2320, opened 2024-03-25
fix(export): more resilient export queries for uncleaned data

sleidig:
issue exporting ".schoolId:toEntities(School).name" in one client's setup


sleidig (commented):
successfully tested

PR #2323, opened 2024-03-26
fix(users): load config before initializing user entity

### LINKED ISSUE #2306: Configured entity types not available yet when setting user entity
**Describe the bug**
If a user account is linked to an entity of an entity type that is only defined in config, this leads to errors and breaks the app because that type may not be available yet during user initialization:

> Requested item "Client" is not registered. See dynamic-registry.ts for more details.
>     at o.get (6013.30a67cfff49c469c.js:2:298353)
>     at y.resolveConstructor (6013.30a67cfff49c469c.js:10:5478)
>     at y.receiveUpdates (6013.30a67cfff49c469c.js:10:4667)
>     at x.initUserEntity (6013.30a67cfff49c469c.js:13:15262)
>     at 6013.30a67cfff49c469c.js:13:15062

sleidig:
to ensure the required entity type is available

fixes #2306


sleidig (commented):
successfully tested in PR deployment

Issue #2324, opened 2024-03-27
Admin UI should accept queryParam to route to details or list editor view


sleidig:
![image](https://github.com/Aam-Digital/ndb-core/assets/1682541/1cc62ea8-9bb7-4252-abda-1b74a2a5c1ca)
pre-select the "right" section depending on whether the Admin UI is opened from a list or details view by the user (using a queryParam in URL route?)

PR #2327, opened 2024-03-29
 #2324/ Admin UI should accept queryParam to route to details or list editor view

Abhinegi2:
closes:  #2324

### Description of changes

-  Added query parameters into the routes to entity details and list screen.
-  Create a function in the AdminEntityComponent to retrieve details from the query parameters.
-  Utilize the mode property to set the modes.
-  Empty the params when users coming from specific screen


sleidig (changes requested):
works well, just needs fixed test dependencies

Abhinegi2:
> works well, just needs fixed test dependencies

@sleidig  fixed the test dependencies.

sleidig (approved):

PR #2331, opened 2024-04-02
fix(forms): reset anonymized fields immediately

### LINKED ISSUE #2285: anonymized fields that get deleted are not updated until reload
**Describe the bug**
Using "anonymize" from the menu of Entity Details clears most fields. The banner marking the record as anonymized appears immediately after the action but those fields that are cleared still hold their full values until the record is opened again. This is confusing to users. The form with all fields should immediately update.

sleidig:
fixes #2285

### Visible/Frontend Changes
- [x] when clicking "anonymize" for an entity, the fields that are removed are immediately shown as empty in the form

- [x] store "anonymized" removed properties as `null` to mark them explicitly as reset
- [x] retain `null` values during entity-schema transformations, to allow such explicitly reset values
- [x] prevent normal forms to save all empty fields as 'null'


tomwwinter (approved):
LGTM

PR #2334, opened 2024-04-08
added multi select functionality

Abhinegi2:
closes: #XX

### Visible/Frontend Changes
- [x] 
- [ ] 

- [x] 
- [ ]

### DIFF
--- a/src/app/core/admin/admin-entity/admin-entity-general-settings/admin-entity-general-settings.component.ts
+++ b/src/app/core/admin/admin-entity/admin-entity-general-settings/admin-entity-general-settings.component.ts
@@ -26,6 +26,7 @@ import { FontAwesomeModule } from "@fortawesome/angular-fontawesome";
 import { MatTooltipModule } from "@angular/material/tooltip";
 import { BasicAutocompleteComponent } from "../../../common-components/basic-autocomplete/basic-autocomplete.component";
 import { EntityConfig } from "../../../entity/entity-config";
+import { EntitySchemaField } from "app/core/entity/schema/entity-schema-field";
 
 @Component({
   selector: "app-admin-entity-general-settings",
@@ -47,25 +48,27 @@ import { EntityConfig } from "../../../entity/entity-config";
     BasicAutocompleteComponent,
   ],
 })
-export class AdminEntityGeneralSettingsComponent implements OnChanges, OnInit {
+export class AdminEntityGeneralSettingsComponent implements OnInit {
   @Input() entityConstructor: EntityConstructor;
   @Output() generalSettingsChange: EventEmitter<EntityConfig> =
     new EventEmitter<EntityConfig>();
   @Input() config: EntityConfig;
   form: FormGroup;
   basicSettingsForm: FormGroup;
-
+  toStringAttributess: SimpleDropdownValue[] = [];
+  private originalEntitySchemaFields: [string, EntitySchemaField][];
   constructor(private fb: FormBuilder) {}
 
   ngOnInit(): void {
     this.init();
   }
 
-  ngOnChanges(changes: SimpleChanges): void {
-    if (changes.config) {
-      this.init();
-    }
-  }
+  // ngOnChanges(changes: SimpleChanges): void {
+  //   if (changes.config) {
+  //     console.log(changes.config)
+  //     // this.init();
+  //   }
+  // }
 
   private init() {
     this.basicSettingsForm = this.fb.group({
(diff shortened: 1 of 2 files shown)


src/app/core/admin/admin-entity/admin-entity-general-settings/admin-entity-general-settings.component.ts
```diff
@@ -77,29 +73,40 @@ export class AdminEntityGeneralSettingsComponent implements OnChanges, OnInit {
...
-    toStringAttributesControl.setValue(toStringAttributesValue, {
-      emitEvent: false,
+    const basicInformationPanel = array.panels.find(
+      (panel) => panel.title === "Basic Information",
+    );
+    basicInformationPanel.components.forEach((component) => {
+      component.config.fieldGroups.forEach((fieldGroup) => {
+        allUsedFields.push(...fieldGroup.fields);
+      });
     });
```
sleidig, line 94: This won't work, because panel titles can be renamed to anything. Also, while it may be useful to only offer used fields, I would rather display all fields of this entity type here.
Use entityConstructor.schema to get the fields here (and hide those without label, because these are internal, technical fields): `allFields = [...this.entityConstructor.schema.values()].filter((field) => field.label)`

src/app/core/common-components/basic-autocomplete/basic-autocomplete.component.html
```diff
@@ -33,25 +33,60 @@
   autoActiveFirstOption
   [hideSingleSelectionIndicator]="multi"
 >
-  <mat-option
-    *ngFor="let item of autocompleteSuggestedOptions | async"
-    [value]="item"
+  <div *ngIf="!reorder">
```
sleidig, line 36: I think we could avoid copying the template block if we use `cdkDropListDisabled` (https://stackoverflow.com/a/54180096/1473411). Did you try that yet?

src/app/core/common-components/basic-autocomplete/basic-autocomplete.component.ts
```diff
@@ -101,6 +107,8 @@ export class BasicAutocompleteComponent<O, V = O>
    * Whether the user should be able to select multiple values.
    */
   @Input() multi?: boolean;
+  @Input() reorder?: boolean;
+  autocompleteDraggableOptions: SelectableOption<O, V>[] = [];
```
sleidig, line 111: can we not use the same autocompleteSuggestedOptions array also if drag&drop is availalbe?

sleidig (changes requested):
Looks quite good already :slightly_smiling_face: 

From functional / UI testing:
- [x] dropdown items checkbox and text should be vertically aligned: 
image
- [x] should only allow string or number fields currently. Other types (schemaField.dataType) will not be displayed properly if used as "toStringAttributes

Abhinegi2:
> Looks quite good already 🙂
> 
> From functional / UI testing:
> 
> * [ ]  dropdown items checkbox and text should be vertically aligned:

Hi @sleidig I have fixed this on my local is this looks fine or do we need some more modifications here?
image

src/app/core/common-components/basic-autocomplete/basic-autocomplete.component.ts
Abhinegi2: The drop method expects an array as input to perform operations such as moving items within the array. However, autocompleteSuggestedOptions is an Observable that emits values asynchronously, and it's not directly usable as an array in the drop method. That is why I am using different arrays to drag and drop.

Abhinegi2: The drop method expects an array as input to perform operations such as moving items within the array. However, autocompleteSuggestedOptions is an Observable that emits values asynchronously, and it's not directly usable as an array in the drop method. That is why I am using different arrays to drag and drop.

src/app/core/admin/admin-entity/admin-entity-general-settings/admin-entity-general-settings.component.ts
Abhinegi2: updated!

sleidig (changes requested):
UI looks good now. Did some minor code cleanup.

One issue I still discovered: After switching tab and coming back (i.e. re-initializing the options), the order in the form field is still correct but in the dropdown it is back to original order not matching the reordered sorting anymore.
I would suggest to put all selected items at the top of the dropdown, in the order they are selected. And then drop all the remaining fields in their default order.

Abhinegi2:
> UI looks good now. Did some minor code cleanup.
> 
> One issue I still discovered: After switching tab and coming back (i.e. re-initializing the options), the order in the form field is still correct but in the dropdown it is back to original order not matching the reordered sorting anymore. I would suggest to put all selected items at the top of the dropdown, in the order they are selected. And then drop all the remaining fields in their default order.

I have updated the functionality a little bit. Can you please check now that it's looking fine or do you have any suggestions?

sleidig (approved):

Issue #2120, opened 2023-12-04
Setup Wizard: Configure static entity details like label and icon


sleidig:
Make some general details about an entity type configurable.

- build simple config UI form to edit the "static" properties of an entity type (see entity.ts static properties: label + labelPlural, icon, toStringAttributes, hasPII). Implemented as a plain Angular form, similar to the code & UI of `AdminEntityFieldComponent`
- add this component into `AdminEntityComponent` as the case for mode = "general"
- extend `AdminEntityComponen.setViewConfig()` method to save the properties returned from the popup form to the config
  - (i.e. `newConfig.data[entityConfigKey].label = configResultFromNewComponent.label`, etc.)

---
### phase 2: better toString UX
- beyond the simple text field to configure "toStringAttributes", the user should be able to select from existing fields
  - --> use our BasicAutocompleteComponent
- allow re-ordering of selected fields
  - --> extend BasicAutocompleteComponent to (optionally, if flag is enabled) allow reordering of selected items

---
### phase 3: Configuring PII / Anyonymization
- add a tab to configure PII / anonymization:
  - checkbox "has personal information (PII)"
  - list of all fields of the entity type's schema with dropdown option for the anonymize setting (retain / partially anonymize / remove = default)
- tooltip on the anonymization type values to explain what "retain" or "retain anonymized" means
![image](https://github.com/Aam-Digital/ndb-core/assets/1682541/05155f3d-e702-4dbd-8eab-c99825090da5)

PR #2335, opened 2024-04-10
Feature/Configuring PII /Anyonymization

Abhinegi2:
closes:#2120

### Visible/Frontend Changes
- [x] Added a New tab to configure PII settings of Entity type.
- [x] Added checkbox to change the Anonymize settings of Entity.
- [x] Added Mat-Table to show the fields related to anonymized settings.
- [x] While switching between tabs PII changes will not lost until.


src/app/core/admin/admin-entity-details/admin-entity-field/admin-entity-field.component.html
```diff
@@ -145,7 +145,7 @@ <h2 mat-dialog-title i18n>Configure Field "{{ entitySchemaField.label }}"</h2>
       <mat-tab
         label="Advanced Options & Validation [COMING SOON]"
         i18n-label
-        [disabled]="true"
+        [disabled]="false"
```
sleidig, line 148:
```suggestion
```

sleidig:
- [ ] "has PII" checkbox needs to be saved and loaded from entity

src/app/core/admin/admin-entity/admin-entity-general-settings/admin-entity-general-settings.component.html
```diff
@@ -70,8 +74,53 @@ <h2 i18n>General Settings of "{{ entityConstructor.label }}" Records</h2>
...
       -->
-    <mat-tab label="Configure PII / Anonymization" i18n-label [disabled]="true">
-      <div class="grid-layout margin-top-regular"></div>
+    <mat-tab
+      label="Configure PII / Anonymization"
+      i18n-label
+      [disabled]="false"
+    >
+      <div class="grid-layout margin-top-regular">
+        <mat-checkbox [checked]="showTable" (change)="toggleTable($event)"
```
sleidig, line 83: add `i18n` here and for other parts of template where missing

src/app/core/admin/admin-entity/admin-entity.component.ts
```diff
@@ -86,6 +85,10 @@ export class AdminEntityComponent implements OnInit {
     });
   }
 
+  updateCheckboxState(checkboxStatus: boolean) {
+    this.checkboxState = checkboxStatus;
```
sleidig, line 89: why do you need to interact with checkbox PII status in the parent component? This should be completely encapsulated within AdminGeneralSettingsComponent only and saved via the configEntitySettings together with other config.

--> see/map to EntityConfig.hasPII

sleidig (changes requested):

sleidig:
@Abhinegi2 , experimenting with the UI I have decided on some adjustment and still see some polishing to do:
- [x] removed pagination, to give user a full, seamless overview of all fields
- [ ] limit the height, so that the scroll is within the "Configure PII" tab, not for the whole screen if many fields are listed
- [ ] fields need to display the field.label in the table, not the key/id

src/app/core/admin/admin-entity/admin-entity.component.ts
Abhinegi2: Previously  am using a custom checkbox to open the PII configuration, But now I know that we can use has PII from entity, So will change all occurrences.

Abhinegi2:
> @Abhinegi2 , experimenting with the UI I have decided on some adjustment and still see some polishing to do:
> 
> * [x]  removed pagination, to give user a full, seamless overview of all fields
> * [ ]  limit the height, so that the scroll is within the "Configure PII" tab, not for the whole screen if many fields are listed
> * [ ]  fields need to display the field.label in the table, not the key/id

sure, will do these adjustment

Abhinegi2:
> @Abhinegi2 , experimenting with the UI I have decided on some adjustment and still see some polishing to do:
> 
> * [x]  removed pagination, to give user a full, seamless overview of all fields
> * [x]  limit the height, so that the scroll is within the "Configure PII" tab, not for the whole screen if many fields are listed
> * [x]  fields need to display the field.label in the table, not the key/id

Hi @sleidig I have implemented the following feedback can you please confirm whether UI is looking good or not - 
1. Added height limit on PII fields.
2. Display the field name instead of the field key.

sleidig:
@Abhinegi2 , the scroll is still not working very well:
image


I would like to 
- always have only on scroll bar (only scroll within the tabs' contents, i.e. the inner scrollbar)
- use the full height of the (remaining) screen before having to scroll (i.e. avoid the white space at bottom currently)

src/app/core/admin/admin-entity/admin-entity-general-settings/admin-entity-general-settings.component.ts
```diff
@@ -38,19 +48,36 @@ import { StringDatatype } from "../../../basic-datatypes/string/string.datatype"
...
 export class AdminEntityGeneralSettingsComponent implements OnInit {
   @Input() entityConstructor: EntityConstructor;
   @Output() generalSettingsChange: EventEmitter<EntityConfig> =
     new EventEmitter<EntityConfig>();
   @Input() generalSettings: EntityConfig;
+  allPIIFields: any[];
+  @Input() showPIIDetails: boolean;
+  entitySchemaField: EntitySchemaField;
+  dataSource: MatTableDataSource<any>;
+  anonymizOptionList: string[] = ["Retain", "Partially Anonymize", "Remove"];
```
sleidig, line 71: Let's add the tooltips here: `anonymizOptionList: { label: string; tooltip: string }[] =`
(calling methods in the template is not ideal, as they are executed on every change detection cycle of angular. And it's easier to maintain the texts, if they are all in one place here).

sleidig: i.e. remove the getTooltip method for this then

sleidig: actually, we can also hard code the three options including tooltip directly in the template html. Anyway these are not dynamically generated currently.

src/app/core/admin/admin-entity/admin-entity-general-settings/admin-entity-general-settings.component.ts
```diff
@@ -38,19 +48,36 @@ import { StringDatatype } from "../../../basic-datatypes/string/string.datatype"
...
 })
 export class AdminEntityGeneralSettingsComponent implements OnInit {
   @Input() entityConstructor: EntityConstructor;
   @Output() generalSettingsChange: EventEmitter<EntityConfig> =
     new EventEmitter<EntityConfig>();
   @Input() generalSettings: EntityConfig;
+  allPIIFields: any[];
+  @Input() showPIIDetails: boolean;
+  entitySchemaField: EntitySchemaField;
+  dataSource: MatTableDataSource<any>;
```
sleidig, line 70: more specific naming, please: e.g. fieldAnonymizationDataSource

src/app/core/admin/admin-entity/admin-entity-general-settings/admin-entity-general-settings.component.ts
```diff
@@ -38,19 +48,36 @@ import { StringDatatype } from "../../../basic-datatypes/string/string.datatype"
...
   ],
 })
 export class AdminEntityGeneralSettingsComponent implements OnInit {
   @Input() entityConstructor: EntityConstructor;
   @Output() generalSettingsChange: EventEmitter<EntityConfig> =
     new EventEmitter<EntityConfig>();
   @Input() generalSettings: EntityConfig;
+  allPIIFields: any[];
+  @Input() showPIIDetails: boolean;
+  entitySchemaField: EntitySchemaField;
```
sleidig, line 69: this seems to never get assigned to anything?

src/app/core/admin/admin-entity/admin-entity-general-settings/admin-entity-general-settings.component.ts
```diff
@@ -65,18 +92,92 @@ export class AdminEntityGeneralSettingsComponent implements OnInit {
...
+      case "Partially Anonymize":
+        return "Partially Anonymize: Anonymize the value but retain some information.";
+      case "Remove":
+        return "Remove: Completely remove the value or anonymize it entirely.";
+      default:
+        return "";
+    }
+  }
+
+  selectionOnChange(value: any, selectFielddata: any) {
```
sleidig, line 154: clearer naming please, e.g. changeFieldAnonymization

also, can you figure out the actual types? `any` is something we try to avoid at all cost.

src/app/core/admin/admin-entity/admin-entity-general-settings/admin-entity-general-settings.component.ts
```diff
@@ -65,18 +92,92 @@ export class AdminEntityGeneralSettingsComponent implements OnInit {
...
+    } else if (selectFielddata.anonymize == "Retain") {
+      selectedFielddetails.anonymize = "retain";
+    } else if ((selectFielddata.anonymize = "Remove")) {
+      selectedFielddetails.anonymize = undefined;
+    }
+    const updatedEntitySchemaMerged = Object.assign(
+      { _isCustomizedField: true },
+      this.entitySchemaField,
+      selectedFielddetails,
+    );
```
sleidig, line 172: It should be enough to simply do `this.entityConstructor.schema.get(selectedFieldId).anonymize = newAnonymizeValue` 

also, you can use `<mat-option [value]="..."` to set the exact key (e.g. "retain" instead of "Retain") and avoid having to map it here in the method again.

sleidig (changes requested):

src/app/core/admin/admin-entity/admin-entity.component.scss
```diff
@@ -26,4 +26,9 @@
   ::ng-deep app-view-title {
     color: colors.$muted;
   }
+
+  &.general-mode {
+    overflow-y: hidden;
+  }
```
sleidig, line 32: This is not a feasible solution as it will make parts of the UI completely unreachable on smaller windows.

sleidig: I have removed this. If we cannot figure out the correct height for the mat-tab-group , it's better to scroll the full part rather than hide completely

Abhinegi2:
> @Abhinegi2 , the validator PR has been merged. Can you please replace the simple input field in the AdminEntityField with the new dropdown component here also

sure working on this part

sleidig (approved):

PR #2378, opened 2024-05-07
Refactor admin-entity-list component to enable drag and drop function…

Abhinegi2:
### Visible/Frontend Changes

- [x] Filters in the admin-entity-list form are now support reordering.

### DIFF
--- a/src/app/core/admin/admin-entity-list/admin-entity-list.component.html
+++ b/src/app/core/admin/admin-entity-list/admin-entity-list.component.html
@@ -4,15 +4,22 @@
 
 <div>
   <div
-    cdkDropList
-    cdkDropListOrientation="horizontal"
-    (cdkDropListDropped)="drop($event, config?.filters)"
+    cdkDropListGroup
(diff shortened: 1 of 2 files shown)


src/app/core/admin/admin-entity-list/admin-entity-list.component.html
```diff
@@ -4,15 +4,22 @@
 
 <div>
   <div
-    cdkDropList
-    cdkDropListOrientation="horizontal"
-    (cdkDropListDropped)="drop($event, config?.filters)"
```
sleidig, line 9: I think the `drop()` method is not used anymore then. If so, please remove it from the component.ts also

sleidig (approved):
:+1:

src/app/core/admin/admin-entity-list/admin-entity-list.component.html
sleidig: oh, right! Thanks

PR #2390, opened 2024-05-21
feat: support for user linked project permissions

tomwwinter:
related: https://github.com/Aam-Digital/internal/issues/21


src/app/core/permissions/ability/ability.service.ts
```diff
@@ -74,10 +75,20 @@ export class AbilityService extends LatestEntityLoader<Config<DatabaseRules>> {
...
-  private interpolateUser(
+  private async interpolateUser(
     rules: DatabaseRule[],
-    user: SessionInfo,
-  ): DatabaseRule[] {
+    sessionInfo: SessionInfo,
+  ): Promise<DatabaseRule[]> {
+    const user = await this.entityMapper
+      .load(User, sessionInfo.entityId)
+      .catch(() => {});
```
sleidig, line 84: Didn't we deprecate the `User` entity and instead are using `SessionInfo` and for business logic the `CurrentUserSubject` to allow any entity type as the "profile" of a user account?

sleidig (changes requested):

src/app/core/permissions/ability/ability.service.ts
tomwwinter: switched to CurrentUserSubject

src/app/core/entity/default-field-value/handle-default-values.usecase.ts
```diff
@@ -93,6 +93,10 @@ export class HandleDefaultValuesUseCase {
         return;
       }
 
+      if (change.length != null && change.length === 1) {
+        change = change[0];
+      }
+
```
sleidig, line 99: can you add a comment here, what the reason behind this is?
Seems counter-intuitive to me. Doesn't that potentially break some cases of "isArray" type fields?

sleidig (approved):
looks good :+1: 

Tested for regressions on dev server. Did not test new project rules (have been tested on client system, I think?)

src/app/core/entity/default-field-value/handle-default-values.usecase.ts
tomwwinter: Good finding, i added two more tests and a code comment for clarification.

sleidig (approved):
LGTM now :+1: 


Maybe this PR can also include the update to docs that still have to be moved from here: https://github.com/Aam-Digital/replication-backend/pull/147

doc/compodoc_sources/concepts/permissions.md
```diff
@@ -111,6 +111,70 @@ The `admin_app` role simpy allows user with this role to do everything, without
...
+    "role_1": [
+      { "subject":  "all", "action":  "manage"},
+      ...
+    ],
+    "role_2": [],
+    ...
+  }
+}
+```
+Important is the exact `_id` as this is how the backend can find this document and that the rules config has the correct structure.
```
sleidig, line 138: This is duplicating part of the above "Permission Structure" section?

doc/compodoc_sources/concepts/permissions.md
```diff
@@ -111,6 +111,70 @@ The `admin_app` role simpy allows user with this role to do everything, without
...
+  "conditions": {
+    "name": "${user.name}",
+    "projects": {
+      "$in": "${user.projects}"
+    }
+  }
+}
+```
+This allows users to update the `password` property of their *own* document in the `_users` database.
+Another available value is `${user.roles}` which is an array of rules which the user has.
```
sleidig, line 174:
```suggestion
Placeholders can currently access properties that the _replication-backend_ explicitly adds to the auth user object. 
Other available values are `${user.roles}` (array of roles of the user) and  `${user.projects}` (the "projects" attribute of the user's entity that is linked to the account through the "exact_username" in Keycloak).
```

sleidig (changes requested):

doc/compodoc_sources/concepts/permissions.md
tomwwinter: removed the duplicating part

sleidig (approved):

PR #2410, opened 2024-06-12
fix: ensure file upload works on newly created records

### LINKED ISSUE #2398: Error uploading attachment on newly created record (e.g. attachment to new note)
**Describe the bug**
When I select a file while creating a new entity (e.g. in the form for a new note), the upload fails (console shows a 404 error). When editing an existing entity and uploading a file, it works.

(check if this overlaps with https://github.com/Aam-Digital/ndb-core/issues/1985 or https://github.com/Aam-Digital/ndb-core/issues/1878)

sleidig:
fixes #2398

The replication-backend needs to load the entity (from the "app" db) for permission checks during the file upload (put on the "app-attachments" db). For newly created documents we need to ensure the sync (nowadays at scheduled intervals) is completed before the upload.


tomwwinter (approved):
LGTM

PR #2412, opened 2024-06-12
fix: use short label for filters of list

### LINKED ISSUE #2399: use labelShort for label of filter components

sleidig:
when available

fixes #2399


christophscheuing (approved):
Makes things shorter and LGTM.

PR #2413, opened 2024-06-12
fix: correctly map multi values during import

### LINKED ISSUE #2408: Import of multi-select enum values transforms values wrong (array within array)

sleidig:
fixes #2408


tomwwinter (approved):
LGTM

Issue #2217, opened 2024-02-05
Automatic status field changes on related entities upon changes of an entity (inferred fields)


sleidig:
(1) _Als Eventkoordinator plane ich Events innerhalb eines Projekts, wobei die Datenstrukturen hierarchisch sind - das Projekt mit seinem Sektor wird als "Parent" ausgewählt. Dabei will ich allgemeine Felder wie den "Sektor" vererben und nicht im neuen Formular selbst neu ausfüllen müssen._

(2) _Als Projektkoordinator möchte ich mich nicht um die Pflege des Status der Mentee-Datensätze kümmern müssen. Wenn ich ein Matching durchführe (eine Patenschaft erstelle), soll sich der Status der verknüpften Mentee und/oder Mentor automatisch aktualisieren und auf “in aktiver Patenschaft” wechseln. Momentan muss ich selbst daran denken und das manuell ändern._

**Requirements**
- [x] configure rules of inheritance of a field's value from a referenced entity
- [x] (1) Initial values on creation:
  - [x] on selecting a "parent" entity in a form, auto-fill inheriting fields with values from that parent
  - [x] When applying inherited fields to a form field that is dirty, ask for confirmation (overwrite or keep custom? → auto-mark as manually overwritten)
- [x] Show indicator next to inherited field to show whether this is “up-to-date”
- [x] Reset a field to again inherit the latest values from parent
- [ ] (2B) on opening “child”, check if parent’s inherited values have changed and ask user if changes should be re-applied / re-inherited
- [ ] Lock/Overwrite inherited field:
  - [x] Overwrite an inherited value (and prevent further updates)
  - [ ] Lock an inherited value (to prevent any future updates)
- [ ] (2A) on changing a “parent” entity, update child entities that inherit a changed field
  - [ ] ? Ask for confirmation whether to apply or not?
  - [ ] Manually re-apply a value from to all children (for debug/admin tasks), optionally also resetting overwritten values
  - [ ] (!) avoid conflicts - ensure only the first device applies those rules

Open Points:

- [ ] applies inherited values also when editing an existing entity, when the user changes the reference/parent

**Initial thoughts (outdated)**
- Automatically update a field in one entity, based upon the change of a field in a different entity that has some relationship to it.
  - this would also allow to inherit values from a hierarchy / referenced entity (see #2219 )
- top-down inheritance of a value
  - primarily initial default value when creating new sub-record
  - also works across multiple levels of hierarchy
[...]

PR #2437, opened 2024-06-26
feat: show ui for inherited default field values

tomwwinter:
see #2217 

### Remaining Tasks
- [x] UI only visible if the localAttribute (parent ref) field `isArray`
- [x] boolean inherited shows as "manually overwritten" if `false` is inherited from parent
- [x] clear values after removing entity-ref to parent? (works if parent is single value, but not if parent-ref isArray)

UX considerations
- [x] show inheritance icon more often?
  - [x] also on disabled form to indicate the link?
  - [x] also while "parent" is not selected, to indicate in edit mode that it can be inherited?
- [x] indicate when inheritance not possible (multiple parents selected)
- [x] more speaking icon?
- [x] label instead of id in hint


sleidig:
UX considerations:
- [ ] show inheritance icon more often?
  - [ ] also on disabled form to indicate the link?
  - [ ] also while "parent" is not selected, to indicate in edit mode that it can be inherited?
- [ ] indicate when inheritance not possible (multiple parents selected)
- [ ] more speaking icon?
- [ ] label instead of id in hint

src/app/core/common-components/entity-form/entity-form.service.ts
```diff
@@ -19,12 +19,27 @@ import { Subscription } from "rxjs";
...
-export type EntityForm<T extends Entity> = TypedForm<Partial<T>>;
+export type TypedFormGroup<T> = FormGroup<{
+  [K in keyof T]: ɵElement<T[K], null>;
+}>;
+
+export type PartialTypedFormGroup<T extends Entity> = TypedFormGroup<
+  Partial<T>
+>;
+
+export interface ExtendedEntityForm<T extends Entity> {
```
sleidig, line 35: suggested naming:
- EntityFormGroup (instead of PartialTypedFormGroup)
- EntityForm (instead of ExtendedEntityForm - it has extended details but we don't have a simple/unextended one to contrast and shorter is better :wink: )
- accordingly  createExtendedEntityForm() -->  createEntityForm

src/app/core/common-components/entity-form/entity-form.service.ts
```diff
@@ -19,12 +19,27 @@ import { Subscription } from "rxjs";
...
+  formGroup: PartialTypedFormGroup<T>;
+  entity: T;
+  defaultValueConfigs: Map<string, DefaultValueConfig>;
+  inheritedParentValues: Map<string, any>;
+  inheritedSyncStatus: Map<string, boolean>;
```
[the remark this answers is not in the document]
sleidig: and let's add JsDoc comments to each element. The names are not always completely intuitive

src/app/core/common-components/entity-form/entity-form.service.ts
```diff
@@ -147,7 +199,9 @@ export class EntityFormService {
       this.subscriptions.push(statusChangesSubscription);
     }
 
-    this.defaultValueService.handle(group, entity);
+    // if (handleDefaultValues) {
+    //   this.defaultValueService.handle(group, entity);
+    // }
```
sleidig, line 204: is this parameter obsolete or needs to be implemented still?

src/app/core/config/config-fix.ts
```diff
@@ -327,7 +325,8 @@ export const defaultJsonConfig = {
               "component": "Form",
               "config": {
                 "fieldGroups": [
-                  { "fields": ["name", "privateSchool"] },
+                  { "fields": ["name", "privateSchool", "parentSchool"] },
+                  { fields: [ "testSchools", "bool", "string", "enum", "refSingle", "refMulti" ]},
```
sleidig, line 329: remove before merge (+ entity config block)
```suggestion
```

src/app/core/entity/default-datatype/edit-component.ts
```diff
@@ -29,6 +30,16 @@ export abstract class EditComponent<T> implements OnInit, OnChanges {
...
 
+  /**
+   *
+   */
+  @Input() defaultValueConfig: DefaultValueConfig | undefined;
+
+  /**
+   *
+   */
+  @Input() defaultParentValue: any | undefined;
```
sleidig, line 41: JsDocs?

src/app/core/form-dialog/row-details/row-details.component.ts
```diff
@@ -69,26 +69,33 @@ export class RowDetailsComponent {
     @Inject(MAT_DIALOG_DATA) public data: DetailsComponentData,
     private formService: EntityFormService,
   ) {
-    this.init(data);
+    this.init(data)
+      .then()
+      .catch((reason) => console.log(reason)); // todo this should not called in constructor?
```
sleidig, line 74: yes, let's move to ngOnInit?

what are possible causes for this to throw?

src/app/core/entity/default-field-value/handle-default-values.usecase.ts
```diff
@@ -20,41 +22,103 @@ export class HandleDefaultValuesUseCase {
...
+      inheritedConfigs,
+      form,
+    );
+
+    await this.loadLinkedEntitiesIntoForm(
+      Array.from(linkedEntityRefs.values()).flat(),
+      form,
+    );
+
+    this.enableChangeListener(form);
```
sleidig, line 43: do we need to enable change listeners for cases other than "inherited"?

src/app/core/entity/default-value.service.ts
```diff
@@ -1,46 +1,42 @@
...
-import { FormGroup } from "@angular/forms";
 import { HandleDefaultValuesUseCase } from "./default-field-value/handle-default-values.usecase";
+import { EntitySchema } from "./schema/entity-schema";
+import { DefaultValueConfig } from "./schema/default-value-config";
+import { ExtendedEntityForm } from "../common-components/entity-form/entity-form.service";
 
 @Injectable({
   providedIn: "root",
 })
 export class DefaultValueService {
```
sleidig, line 11: Does this really add value or could we combine the use-case and this service into one for simplicity?

sleidig (commented):
Looks pretty good already! :smiley:

tsconfig.json
```diff
@@ -20,7 +20,7 @@
       "jasmine"
     ],
     "lib": [
-      "es2018",
+      "es2019",
```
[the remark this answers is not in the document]
tomwwinter: This was needed for the for (...of..) notation on Maps, when im not wrong.

src/app/core/common-components/entity-form/entity-form.service.ts
tomwwinter: Yes, this was just the naming for the refactoring.

src/app/core/common-components/entity-form/entity-form.service.ts
tomwwinter: This method will be replaced by createExtendedEntityForm after this refactoring, so this will be removed

tsconfig.json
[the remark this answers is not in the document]
sleidig: Do we need polyfills or other measures to ensure support in older browsers for this?

src/app/core/basic-datatypes/entity/edit-entity/edit-entity.component.html
```diff
@@ -7,3 +7,9 @@
   [form]="formControl"
 >
 </app-entity-select>
+
+<br />
+<div>
+  <code>{{ defaultValueConfig | json }}</code>
+  <code>{{ defaultParentValue }}</code>
+</div>
```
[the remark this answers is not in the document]
tomwwinter: Yes, already removed.

src/app/core/common-components/entity-form/entity-form.service.ts
tomwwinter: I renamed some of the suggestions already, feel free to rename the other ones

src/app/core/form-dialog/row-details/row-details.component.ts
tomwwinter: moved to ngOnInit, lets double check if everything works as expected... tests are green

src/app/core/entity/default-field-value/handle-default-values.usecase.ts
tomwwinter: For loading the linked parent entities it would be fine to just listen to all properties from `defaultValues.map(dv -> localField)`

For updating the sync state, we need to listen to all `defaultValues.keys()`

src/app/core/entity/default-value.service.ts
tomwwinter: Im fine with merging this together.

sleidig:
@tomwwinter I did some polishing for the UX and checked off the open points regarding that. Once we fix the last issue (with the reset of inherited fields when removing the isArray parent value), this should probably be ready.

tomwwinter:
> @tomwwinter I did some polishing for the UX and checked off the open points regarding that. Once we fix the last issue (with the reset of inherited fields when removing the isArray parent value), this should probably be ready.

Found and fixed this. Should be ready now.

sleidig (approved):

Issue #2290, opened 2024-03-14
bulk actions should allow to select range of rows by holding "shift" key


sleidig:
The "Bulk Actions" option on list views allows the user to select multiple rows to perfom an action on all of them at once. Currently each row has to be selected one by one.

- [x] allow the common approach to hold "shift" key to select a range of rows between start and end item
- [x] add buttons to "select all" and "clear" the full list (all pages but only the currently filtered rows)

PR #2451, opened 2024-07-08
Added shift and select all logic to Bulk Actions

Abhinegi2:
closes: #2290 

### Visible/Frontend Changes

- [x] Add checkbox buttons to "select all" and "clear" the full list (all pages but only the currently filtered rows)
- [x] Added the common approach to hold the "shift" key to select a range of rows between the start and end item.


tomwwinter:
I found an issue when selecting single elements:
 
https://github.com/Aam-Digital/ndb-core/assets/16425328/42fa0bd4-f6d2-439e-8971-5bd95b3f4477

Abhinegi2:
> I found an issue when selecting single elements:
> 
>  Screen.Recording.2024-07-11.at.14.51.58.mov

let me check and fix this, will update you once done

Abhinegi2:
@tomwwinter it's fixed can you please check now?

tomwwinter:
Looks better now. I just saw some CSS glitches, when the bulk overlay is visible:

image
image

Maybe we can add some white background here?

Abhinegi2:
> Looks better now. I just saw some CSS glitches, when the bulk overlay is visible:
> 
> image image
> Maybe we can add some white background here?

@tomwwinter just found that bulk action UI is not properly working for small screens(bulk actions are visible two times). I am also attaching an image for reference, we can create a new ticket/issue for the overall UI fixes -
image

tomwwinter (approved):
Looks good! I did some refactoring with @sleidig to solve some very tricky bugs with the templating logic.

sleidig (approved):

PR #2454, opened 2024-07-11
Fix/location custom address

sleidig:
closes: #1844 

- reworked EditLocation component, splitting it into more modular components and implementing as CustomFormControl
- changed datatype to hold manual address + lookup object
- moved address lookup into popup for more clearly separated UX
- some logic for smartly keeping manual address text and looked up address in sync

---
remaining TODOs:
- [x] fix tests
- [x] maybe: give hint to user after entering manual address, explaining that a map lookup has to be done manually?


src/app/features/location/location-input/location-input.component.ts
```diff
@@ -0,0 +1,183 @@
...
+      ngControl,
+      parentForm,
+      parentFormGroup,
+    );
+  }
+
+  updateLocationString(value: string) {
+    const manualAddress: string = value ?? "";
+    if (manualAddress === "" && this.value?.geoLookup) {
+      // TODO ask user if they want to remove the mapped location also?
```
tomwwinter, line 71: Should we do this right now a did you already created a ticket?

src/app/features/location/location-input/location-input.component.ts
```diff
@@ -0,0 +1,183 @@
...
+          results.some((r: GeoResult) =>
+            matchGeoResults(r, this.value.geoLookup),
+          )
+        ) {
+          // all good, most likely the current location is still correct
+          return;
+        }
+
+        // multiple options ... what do we do with that?
+        // TODO
```
tomwwinter, line 175: What about this open TODOs?

tomwwinter (commented):
I always get just "[selected on map: XY]" when i select something on the map. Would be nice, when i got the actually address?

tomwwinter:
When selecting an address over the search, the format of the address will be updated to something like: "Street #Nr, 12345 Berlin", but when i select an address over the map, by clicking somewhere, the original (very long) address is filled in.

tomwwinter:
The edit button of the Address, is a little out of focus in the top right corner, maybe we can limit the width of the input field to move it more center.

tomwwinter:
We should add some matomo tracking for the nominatim requests for better monitoring data.

src/app/features/location/location-input/location-input.component.ts
```diff
@@ -0,0 +1,99 @@
...
+      selectedLocation: this.value,
+      disabled: this._disabled,
+    };
+
+    const ref = this.dialog.open(MapPopupComponent, {
+      width: "90%",
+      autoFocus: ".address-search-input",
+      restoreFocus: false,
+      data: config,
+    });
```
tomwwinter, line 81: Maybe worth to check, if the dialog is actually unsubscribed, when the host component is destroyed?

See: https://stackoverflow.com/questions/58198544/angular-dialogref-unsubscribe-do-i-need-to-unsubscribe-from-afterclosed

src/environments/environment.ts
```diff
@@ -32,7 +32,7 @@ export const environment = {
   remoteLoggingDsn: undefined, // only set for production mode in environment.prod.ts
   /** The following settings can be overridden by the `config.json` if present, see {@link AppSettings} */
   demo_mode: true,
-  session_type: SessionType.mock,
+  session_type: SessionType.local,
```
tomwwinter, line 35:
```suggestion
  session_type: SessionType.mock,
```

tomwwinter (approved):

src/app/features/location/location-input/location-input.component.ts
sleidig: also browser navigate back triggers this finalize of the observable. Can't think of a reasonable case where we have to manually unsubscribe

Issue #2337, opened 2024-04-11
Do not clear autocomplete text after selecting item for multi-select


sleidig:
1. when selecting dropdown options in a multi-select field, if the user writes a word, options get filtered.
2. If the user picks one option, the text she/he had written disappears
3. so if the user wanted to select more than one option related to the word she/he had written, she/he has to write it again. and then select the next item.
4. It would be better if the word is kept and the person can click as many options as needed.

---
for test setup, change the `config-fix.ts` default config, making the School "language" field a multi-value enum:
```
"entity:School": {
    "attributes": {
      "language": {
        "dataType": "configurable-enum",
        "additional": "genders",
        "isArray": true,
        "label": $localize`:Label for the language of a school:Language`
      },
```

PR #2456, opened 2024-07-12
Do not clear autocomplete text after selecting item for multi-select 

Abhinegi2:
closes: #2337

### Visible/Frontend Changes

- [x] Retain the search field value typed by the user.
- [x] After selecting an option the user-typed word will be retained.


tomwwinter (approved):

src/app/core/config/config-fix.ts
```diff
@@ -818,7 +818,10 @@ export const defaultJsonConfig = {
         "label": $localize`:Label for if a school is a private school:Private School`
       },
       "language": {
-        "dataType": "string",
+        // ToDo - Remove multi-select option from this
+        "dataType": "configurable-enum",
+        "additional": "genders",
+        "isArray": true,
```
sleidig, line 824:
```suggestion
        "dataType": "string",
```

Issue #2419, opened 2024-06-17
Admin UI form field to set Default Value shows [Object object]


sleidig:
Since the default value config format was changed in https://github.com/Aam-Digital/ndb-core/pull/2362, the Admin UI field config popup form is broken, as it does not support the new, complex config format yet.

PR #2469, opened 2024-07-19
Fix/admin default value

sleidig:
closes: #2419


src/app/core/admin/admin-entity-details/admin-entity-field/default-value-options/default-value-options.component.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+    expect(component.valueChange.emit).not.toHaveBeenCalled();
+
+    component.form.setValue({
+      mode: "inherited",
+      value: null,
+      localAttribute: "foo",
+      field: null,
+    });
+    // TODO: why is this a valid form with the required "field" value missing?
+    expect(component.valueChange.emit).not.toHaveBeenCalled();
```
sleidig, line 109: for some reason the form (both in tests and in actual app) first is valid and in a second run invalid, so one change is emitted ...

tomwwinter (approved):

Issue #2461, opened 2024-07-15
Make "Create account & send invitation" button more visible in


sleidig:
When trying to invite a new user, creating a new account in `UserSecurityComponent`, users have sometimes overlooked the button to "Create account & send invitation". This maybe needs to visually stand out more prominently.

Idea: As soon as form is valid (email field is filled), change the button color to "accent" (blue) and do this with a small animation, to catch the eye of users.

PR #2489, opened 2024-07-30
Given blue accent color and slighlty animation effect to the button w…

sadaf895:
…henever the form is valid.

closes: #2461


sadaf895:
Ok I will do that.

src/app/core/user/user-security/user-security.component.scss
```diff
@@ -7,3 +7,22 @@
...
+    transform: scale(1);
+  }
+}
+
+.invite-button {
+  margin: 0 12px;
+}
+.invite-button-animate {
+  animation: buttonPulse 1s forwards;
+}
```
sleidig, line 28: actually the animation was still cut off, @sadaf895 . using margin rather than padding seems to help.

sleidig (approved):
Nice work! Thanks :-)

Issue #2229, opened 2024-02-12
apply some styling to table rows of inactive records


sleidig:
We allow users to include "inactive" (archived) records in list using a toggle below the table. In some views (where we use "ChildBlock", for example) the archived rows can be identified because the have a grey instead of black icon. In other views (e.g. the Schools List) it is impossible to identify which are inactive records.

--> Make the text color for the whole row of an inactive record always dark grey, to help users distinguish these.

This is probably to be set in the entities-table component. For implementation, you could use a conditional css class ("number one" here: https://stackoverflow.com/a/57196493/1473411), checking that rowEntity.isActive is false

![image](https://github.com/Aam-Digital/ndb-core/assets/1682541/cb855147-f2b8-4dd6-823e-e36370919836)

PR #2505, opened 2024-08-08
Applied grey color to the table rows of inactive records

sadaf895:
closes: #2229


src/app/core/basic-datatypes/entity/entity-block/entity-block.component.scss
```diff
@@ -11,3 +12,7 @@
 .block-height {
   line-height: sizes.$icon-block;
 }
+
+.inactive-row {
```
sleidig, line 16: this is not a row in a table anymore, better name it just "inactive"?

sleidig (approved):
works well, great job! :slightly_smiling_face:

Issue #2494, opened 2024-08-01
Import enum matching popup should be wider (and dropdown box use full width of the field)


sleidig:
![image](https://github.com/user-attachments/assets/de5cda91-8a30-4b29-bd4a-dfc10e0bb099)

PR #2508, opened 2024-08-13
Enum matching popup should be wider

Abhinegi2:
closes: #2494 

### Visible/Frontend Changes

- [x] Enum popup will now be using fill wdth and wider.


sleidig (changes requested):
We have a similar issue for entity-select dropdowns, e.g. see http://localhost:4200/attendance/recurring-activity -> details view -> assigned users

If there is no way to handle this in the underlying app-basic-autocomplete, we may need to implement something similar here as well ...

Abhinegi2:
> We have a similar issue for entity-select dropdowns, e.g. see http://localhost:4200/attendance/recurring-activity -> details view -> assigned users
> 
> If there is no way to handle this in the underlying app-basic-autocomplete, we may need to implement something similar here as well ...

updated entity select component by using same approach

src/app/core/common-components/entity-select/entity-select.component.ts
```diff
@@ -20,11 +20,13 @@ import { asArray } from "../../../utils/utils";
...
 import { FormDialogService } from "../../form-dialog/form-dialog.service";
 import { EntityRegistry } from "../../entity/database-entity.decorator";
-
 @Component({
   selector: "app-entity-select",
   templateUrl: "./entity-select.component.html",
-  styleUrls: ["./entity-select.component.scss"],
+  styleUrls: [
+    "./entity-select.component.scss",
+    "../../basic-datatypes/configurable-enum/enum-dropdown/enum-dropdown.component.scss",
```
sleidig, line 28: you can move this stylesheet to the folder of the underlying basic-autocomplete

sleidig (changes requested):
Your approach with positioning absolute has the disadvantage that the text (if long) overlaps with the icons, making both unreadable:
image

 Maybe we need to limit the width of the input then. Or work with content projection ...?
 Can you find out how `matIconSuffix` usually does that?

src/app/core/common-components/entity-select/entity-select.component.ts
Abhinegi2: it`s update please have a look now

sleidig (approved):
The icons are still overlapped in some edge cases but this seems good enough for now to close the issue :+1:

Issue #2388, opened 2024-05-17
table filter with many selected options leads to HTTP 431 error


tomwwinter:
**Describe the bug**
When selecting filters, the state of the filter will persist to the url. When filtering for many entities, all uuids are added to the url, wich leads to a very large url. This will lead to an 431 error later or sooner.

**To Reproduce**

1. Go to instance with many entities in database
2. Click on 'School Results'
3. Open the 'Student' filter and click on 'select all'
4. reload page
5. See error

**Expected behavior**
The app should not go to error state, when filter entities. Filter options may need a limit for selecting options.

**Screenshots**
![image](https://github.com/Aam-Digital/ndb-core/assets/16425328/42a97b6e-5d51-41e5-9a6d-bf015e55ecb6)

**Desktop (please complete the following information):**

- OS: all
- Browser: all
- Version: all

sleidig:
Approach:
- [ ] check reasonable max length for query params (research)
- [ ] write unit test that fails if too long, if this can be reproduced in tests, or at least a test that expects that the params are not longer than a hard-coded length that was researched
- [ ] skip adding the longest filter option to the query params URL in case the max length is overshot otherwise

see `FilterComponent` for the code that applies filter params to URL

Abhinegi2:
**Maximum URL Length:** The maximum length for URLs varies across browsers and server configurations. The general maximum limit is around 2,000 characters for [reference](https://www.baeldung.com/cs/max-url-length).

PR #2513, opened 2024-08-15
Fix the HTTP 431 error

Abhinegi2:
closes: #2388

### Visible/Frontend Changes

- [x] Limit max query params as 2000.
- [x] Added Test cases.

### DIFF
--- a/src/app/core/basic-datatypes/entity/entity-block/entity-block.component.ts
+++ b/src/app/core/basic-datatypes/entity/entity-block/entity-block.component.ts
@@ -11,10 +11,12 @@ import { FaIconComponent } from "@fortawesome/angular-fontawesome";
 import { DisplayImgComponent } from "../../../../features/file/display-img/display-img.component";
 import { EntityBlockConfig } from "./entity-block-config";
 import { EntityFieldViewComponent } from "../../../common-components/entity-field-view/entity-field-view.component";
+import { DynamicComponent } from "../../../config/dynamic-components/dynamic-component.decorator";
 
 /**
  * Display an inline block representing an entity.
  */
+@DynamicComponent("EntityBlock")
 @Component({
   selector: "app-entity-block",
   templateUrl: "./entity-block.component.html",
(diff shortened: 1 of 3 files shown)


src/app/core/filter/filter/filter.component.spec.ts
```diff
@@ -98,6 +98,23 @@ describe("FilterComponent", () => {
     expect(component.filterSelections[0].selectedOptionValues[1]).toBe("bar");
   });
 
+  it("should fail if the URL length exceeds the safe limit", async () => {
```
sleidig, line 101: this test is somewhat pointless, because it didn't fail before your code changes either.

Can you work on a better unit test? I would make a call to `filterOptionSelected` in the test action, simulating a loooot of generated selected options (to ensure it'll be >2000 characters). Then spy on the router.navigate call and expect (a) that the url is shorter than 2000 and (b) that the longest param was removed but others stay in the url

sleidig (changes requested):

src/app/core/filter/filter/filter.component.spec.ts
Abhinegi2: updated the test cases can you please check now

src/app/core/filter/filter/filter.component.ts
```diff
@@ -113,11 +113,67 @@ export class FilterComponent<T extends Entity = Entity> implements OnChanges {
...
+    let potentialUrl = this.router
+      .createUrlTree([], {
+        relativeTo: this.route,
+        queryParams: { ...this.route.snapshot.queryParams, ...params },
+        queryParamsHandling: "merge",
+      })
+      .toString();
+
+    if (potentialUrl.length <= 2000) {
+      this.router.navigate([], {
```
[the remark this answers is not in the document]
Abhinegi2: It is also updated

sleidig (approved):
I have rewritten the test to more closely reflect what actually happens when a user selects filters and removed a lot of unused code from the implementation. Seems ready to merge now for me

Issue #1874, opened 2023-05-05
deleting user entity does not delete user account


sleidig:
Deleting a user entity in the Aam Digital UI does not delete the linked user account in Keycloak, this may lead to confusion or inadvertently keeping access for outdated users.

TheSlimvReal:
We could use a similar approach as we have for the photos. There we listen for deletions of entities and if one of them has attachments, we delete the attachments from the DB.

PR #2514, opened 2024-08-15
feat: delete keycloak user when deleting user entity

tomwwinter:
closes: https://github.com/Aam-Digital/ndb-core/issues/1874

### Visible/Frontend Changes

- [ ] show warning popup, when keycloak user could not removed after user deletion 

- [ ] see https://github.com/Aam-Digital/account-backend/pull/38

### DIFF
--- a/src/app/core/entity-list/duplicate-records/duplicate-records.service.spec.ts
+++ b/src/app/core/entity-list/duplicate-records/duplicate-records.service.spec.ts
@@ -15,10 +15,13 @@ import { UpdateMetadata } from "../../entity/model/update-metadata";
 import { MatDialog } from "@angular/material/dialog";
 import { MatSnackBar } from "@angular/material/snack-bar";
 import { FileService } from "../../../features/file/file.service";
+import { KeycloakAuthService } from "../../session/auth/keycloak/keycloak-auth.service";
 
 describe("DuplicateRecordsService", () => {
   let service: DuplicateRecordService;
   let entityMapperService: EntityMapperService;
+
+  let mockAuthService: jasmine.SpyObj<KeycloakAuthService>;
 
   @DatabaseEntity("DuplicateTestEntity")
   class DuplicateTestEntity extends Entity {
@@ -41,6 +44,7 @@ describe("DuplicateRecordsService", () => {
         { provide: MatDialog, useValue: {} },
         { provide: MatSnackBar, useValue: {} },
         { provide: FileService, useValue: {} },
+        { provide: KeycloakAuthService, useValue: mockAuthService },
         ComponentRegistry,
       ],
     });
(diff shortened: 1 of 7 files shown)


src/app/core/entity-list/duplicate-records/duplicate-records.service.spec.ts
```diff
@@ -41,6 +44,7 @@ describe("DuplicateRecordsService", () => {
         { provide: MatDialog, useValue: {} },
         { provide: MatSnackBar, useValue: {} },
         { provide: FileService, useValue: {} },
+        { provide: KeycloakAuthService, useValue: mockAuthService },
```
sleidig, line 47: does this really need the service?

src/app/core/entity/entity-actions/entity-actions.service.ts
```diff
@@ -102,9 +104,22 @@ export class EntityActionsService {
...
+          next: () => {},
+          error: async () =>
+            await this.confirmationDialog.getConfirmation(
+              $localize`:delete account in keycloak related error title:Keycloak User could not be deleted`,
+              $localize`:delete account in keycloak related error dialog:User Account could not be deleted in Keycloak. Please delete user manually in Keycloak.`,
+              OkButton,
+            ),
+        });
+      }
       result.mergeResults(await this.entityDelete.deleteEntity(entity));
```
sleidig, line 120: wouldn't this code logically rather belong in the EntityDeleteService, one layer down?

src/app/core/entity/entity-actions/entity-actions.service.ts
```diff
@@ -102,9 +104,22 @@ export class EntityActionsService {
...
     let result = new CascadingActionResult();
+
     for (let entity of entities) {
+      if ("User" === entity.getType()) {
+        this.keycloakAuthService.deleteUser(entity.getId()).subscribe({
+          next: () => {},
+          error: async () =>
+            await this.confirmationDialog.getConfirmation(
+              $localize`:delete account in keycloak related error title:Keycloak User could not be deleted`,
+              $localize`:delete account in keycloak related error dialog:User Account could not be deleted in Keycloak. Please delete user manually in Keycloak.`,
```
sleidig, line 115: What happens if there actually is no user to be deleted? Do we get feedback from the API?

tomwwinter: Nothing, it's save to call this endpoint multiple times. If something there, we delete it, if not, nothing to delete - also fine.

src/app/core/entity-list/duplicate-records/duplicate-records.service.spec.ts
tomwwinter: Yes, some other dependency needs Keycloak.

sleidig (approved):

PR #2534, opened 2024-08-26
fix: liveSync just starts one sync job at a time

tomwwinter:
See #https://github.com/Aam-Digital/internal/issues/58


sleidig (approved):

sleidig (changes requested):
@tomwwinter , it seems like on the dev.aam-digital.net deployed version of this there always is a re-sync when I reload/re-login? Is this cause by our changes here?

sleidig (approved):
works well. the re-sync issue seems to be unrelated of this and only occurs with the "demo" user on dev, not with "demo-admin" ...

Issue #2266, opened 2024-02-28
Feature: PDF generation with entity placeholders


sleidig:
_As a user I want to download a pdf of a certificate with the name and address from the entity's fields, so that I can quickly, directly from the system create files for certain templates._

From various clients and use cases we have observed the need to create PDFs or print certain documents (e.g. a workshop certificate, school mark sheet, invoice, etc.). At the moment users create these by hand using Word or possibly export data from Aam Digital as a .csv table and then use a Mail Merge functionality to generate individual documents from a Word template. This is a lot of extra effort and requires advanced skills with Word.

#### Managing templates
1. [x] User can create a new template
    1. [x] define the valid entity type(s) for the template
    2. [x] provide a name for the template
    3. [x] upload template (Word) file
2. [x] User can view list of all templates in an admin interface
    1. [x] filter templates by the entity type (record type, e.g. "Child", "School", "Note", ...) they are available for.
3. [x] User can edit an existing template
    1. [x] edit the name and entity type(s) for which the template can be used
    2. [x] edit the template design (i.e. re-upload a new template file?)
4. [x] User can delete an existing template

#### Generating PDFs from templates
1. [x] User can create a PDF from a template for a specific entity record
    1. [x] use the action "Print Documents" from the entity menu on the detail view of a record (three dots menu)
    3. [x] select a template in the dialog from all the available templates for this entity type
    4. [x] (generate PDF filled with current data via API)
    5. [x] download the generate PDF document, which is generated from the template by filling in details (e.g. name, address, etc.) of the current record for the placeholders of the template.
2. [ ] Generated PDF is automatically added as file attachment to the entity to be stored.
4. [x] User can add a new template from the UI while trying to generate a PDF
    1. [x] The UI for selecting an existing template guides the user to the option of creating a new template (_see "Managing templates"_), if the required template doesn't exist yet.

Further extensions of the functionality:
- User can create PDFs from a template for a batch of selected entity records (bulk action)
    1. The same flow to generate PDFs is available as a bulk action from list views for a selection of multiple records at once.
    3. The UX works the same as for an individual PDF from a details view

-----
## Implementation Tasks
https://docs.google.com/spreadsheets/d/1PecYL2UgN71P8sh8wXnKALKJCBYDngWV_0L8WEX5MY4/edit?gid=0#gid=0

[...]

PR #2554, opened 2024-08-30
Feature: PDF Generator

sleidig:
see #2266

### TODO:
- Bugs & remaining improvements:
  - [x] Applicable Entity Types change causes errors
  - [x] tooltip for entity-block to display description of file-template
  - [x] filter file-types allowed to select: docx, doc, odt, ods, xlsx, pptx, ppt, odp
  - [x] unit tests ...
  - [x] integrate with backend
  - [x] tooltip in selection?
  - ~delete template doc in carbone API (via backend tracking / cron?) when ... FileTemplate entity is deleted ... new template doc is uploaded, replacing an existing file~
  - ~undo entity create if template upload fails?~
  - [x] show more explanation on new Template form about the file to use
  - [x] get filename from API
  - [x] store template file name for better UX
  - [x] translations
  - [x] download template to review as admin `GET template/:id`

Open tasks / questions:
- [ ] should we use ExportTemplate entity ID instead of the templateId for showing the template file (`GET template/:id`)?


sleidig:
@tomwwinter what kind of files are supported as templates? .doc, docx, .odt?

src/app/features/pdf-generator/entity-2-file/file-template-selection-dialog/file-template-selection-dialog.component.ts
```diff
@@ -0,0 +1,79 @@
...
+      .generatePdfFromTemplate(templateId, this.entity)
+      .pipe(finalize(() => (this.loadingRequestedFile = false)))
+      .subscribe((file) => {
+        this.downloadService.triggerDownload(
+          file,
+          "pdf",
+          this.entity.toString(),
+        );
+        this.dialogRef.close(true);
+      });
```
tomwwinter, line 77: Can we use a pipe switchMap() here, instead of subscribing?

src/app/features/pdf-generator/pdf-generator-api/pdf-generator-api.service.ts
```diff
@@ -0,0 +1,115 @@
...
+    if (!this.navigator.onLine) {
+      return throwError(() => new NotAvailableOfflineError("File Attachments"));
+    }
+
+    const formData = new FormData();
+    formData.append("template", file, file.name);
+
+    return this.httpClient.post(this.BACKEND_URL + "template", formData).pipe(
+      switchMap(async (res: TemplateUploadResponseDto) => {
+        // TODO: maybe store template filename + templateId (API) in entity, to show the user better meta-information?
```
tomwwinter, line 65: The filename of the rendered file can also be defined via variables (reportName). There is a "Content-Disposition" Header from carbone.io, wich we can forward to the frontend.

tomwwinter (commented):
Looks good so far, i added some notes.

proxy.conf.json
```diff
@@ -9,7 +9,7 @@
     }
   },
   "/query": {
-    "target": "http://localhost:3000",
+    "target": "http://localhost:9000",
```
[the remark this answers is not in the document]
sleidig: I think this was originally at port 3000 when the reporting backend was still a NestJS project. Now 9000 to point to local aam-services-backend makes more sense and I would commit this.

But thanks for double checking!

src/app/features/pdf-generator/entity-2-file/file-template-selection-dialog/file-template-selection-dialog.component.ts
sleidig: this is called onClick, so we do need to subscribe somewhere at the end of the chain, don't we?

tomwwinter (approved):

Issue #2291, opened 2024-03-14
bulk action: edit fields


sleidig:
We support various bulk actions from list views, like deleting or archiving all selected records. Users also sometimes need to update one or more fields across all selected entities (e.g. changing the "status" of multiple records without having to edit each one by opening it individually).

The bulk action button "edit" could open a modal form that allows to set one (or more) fields to the same new value for all selected records.

as an example: Hubspot's UI for a similar functionality looks like this:
![image](https://github.com/Aam-Digital/ndb-core/assets/1682541/934bf841-3969-4d39-b77f-e3e5b93dd670)

### Possible Implementation Steps:
- Add button "Edit Field" to `#bulkActions` in EntityListComponent
- Implement component to allow selecting one field (dropdown of all attributes of the entity type's schema) + form field to set a new value as a user
- implement a service "EntityBulkEditService" (take `EntityDeleteService` as a template to build a similar structure), which sets a given field to a given value for all selected records in bulk
- add that new service to `EntityActionsService` and add integrate it to support the new bulk action via click of the button

PR #2565, opened 2024-09-10
Feature/ Bulk edit

Abhinegi2:
closes: #2291

### Visible/Frontend Changes

- [x] Bulk Edit button was added.
- [x] Warning message about the bulk edit.
- [x] New dialog box added from where users can update multiple entities


src/app/core/entity/entity-actions/entity-bulk-edit/entity-bulk-edit.component.html
```diff
@@ -0,0 +1,33 @@
+<h2 mat-dialog-title i18n>Bulk edit {{ selectedRows }} tasks</h2>
```
sleidig, line 1: try using entityConstructor.labelPlural here instead of calling it "tasks" generically?

src/app/core/entity/entity-actions/entity-actions.service.ts
```diff
@@ -54,6 +61,14 @@ export class EntityActionsService {
...
         tooltip: $localize`:entity context menu tooltip:Remove the record completely from the database.`,
       },
+      {
+        action: "edit",
+        execute: (e, nav) => this.edit(e, nav),
+        permission: "edit",
+        icon: "edit",
+        label: $localize`:entity context menu:Edit`,
+        tooltip: $localize`:entity context menu tooltip:Update the record.`,
+      },
```
sleidig, line 71: This adds it to the context menu of a single entity details view. I think we don't need to add "bulk edit" feature here at all. (not very clearly documented code however ...)

sleidig (changes requested):
Nice work!

From UX testing:
- I like the "warning" / explanation text. But I think we could skip one step for users by displaying that above the bulk edit form in that "second" dialog directly.
- "Property to update" is wider than the field to select the new value. That could also take full width (you'll probably need something like `:host ::ng-deep mat-form-field` to get to the css of the child component)
-

src/app/core/entity/entity-actions/entity-actions.service.ts
```diff
@@ -168,6 +175,58 @@ export class EntityActionsService {
...
+        entities[0].getConstructor().labelPlural;
+    } else {
+      textForEditEntity =
+        $localize`:Definite article singular:the` +
+        " " +
+        entities[0].getConstructor().label +
+        ' "' +
+        entities[0].toString() +
+        '"';
+    }
```
sleidig, line 209: is this still used? I think can be removed?

src/app/core/entity/entity-actions/entity-actions.service.ts
```diff
@@ -168,6 +175,58 @@ export class EntityActionsService {
...
+
+    const dialogRef = this.matDialog.open(EntityBulkEditComponent, {
+      width: "30%",
```
sleidig, line 212: unusable on small screens with this. Why not leave it without explicit width?

src/app/core/permissions/permission-types.ts
```diff
@@ -9,6 +9,7 @@ const actions = [
   "create",
   "update",
   "delete",
+  "edit",
```
sleidig, line 12:
```suggestion
```

use "update" for this

src/app/core/entity/entity-actions/entity-actions.service.ts
Abhinegi2: i added width earlier because the field and dynamic field looked so lengthy, now let me remove the explicit width

src/app/core/entity/entity-actions/entity-bulk-edit/entity-bulk-edit.component.ts
```diff
@@ -0,0 +1,148 @@
...
+  styleUrl: "./entity-bulk-edit.component.scss",
+})
+export class EntityBulkEditComponent<E extends Entity> implements OnInit {
+  formField: FormFieldConfig;
+  schemaFieldsForm: FormGroup;
+  entityConstructor: EntityConstructor;
+  selectedRows: any;
+  entityData: any;
+  showDynamicFields: boolean = false;
+  _field: FormFieldConfig;
```
sleidig, line 61: why is there a `formField` and `_field`?

src/app/core/entity/entity-actions/entity-edit.service.ts
```diff
@@ -0,0 +1,47 @@
...
+  constructor(
+    protected override entityMapper: EntityMapperService,
+    protected override schemaService: EntitySchemaService,
+    private unsavedChanges: UnsavedChangesService,
+  ) {
+    super(entityMapper, schemaService);
+  }
+
+  async editEntity<E extends Entity>(
+    result: any,
```
sleidig, line 25: what is "result" supposed to contain? as it is an input parameter, this name doesn't seem very well fitting?

Can you define the type of this also?

src/app/core/entity/entity-actions/entity-edit.service.ts
```diff
@@ -0,0 +1,47 @@
...
+    entityParam: E | E[],
```
sleidig, line 26: entities or entitiesToEdit is maybe more inuitive as a name here?

src/app/core/entity/entity-actions/entity-actions.service.ts
```diff
@@ -168,6 +175,58 @@ export class EntityActionsService {
     return true;
   }
 
+  /**
```
sleidig, line 178: let's move this whole block to the EntityEditService. You can inject EntityActionsService there to call the showSnackbarConfirmationWithUndo() if needed.

src/app/core/entity/entity-actions/entity-bulk-edit/entity-bulk-edit.component.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  private init() {
+    this.initForm();
+    this.fetchEntityFieldsData();
+  }
+
+  private initForm() {
+    this.schemaFieldsForm = this.fb.group({
+      selectedField: ["", Validators.required],
+    });
```
sleidig, line 93: if this only has a single field, it's probably simpler to create a FormControl without the FormGroup. I don't think you are strictly required to have a <form> in the template also and can just use `<mat-select [formControl]="selectedSchemaFieldFormControl"` directly.

sleidig (changes requested):
A few remarks for code cleanups remaining. Otherwise works well

src/app/core/entity/entity-actions/entity-edit.service.ts
Abhinegi2: @sleidig I have renamed "result" to "updatedEntity" to reflect its purpose better. Additionally, since the label can hold various types of values (like a string or an array), I updated the type definition to:`updatedEntity: {selectedField: string, label: E| E[]},`
Please let me know if this approach makes sense or if you have further suggestions.

Abhinegi2: did some changes - 
1.  rename the label.
2. changed the type to "any"

src/app/core/entity/entity-actions/entity-edit.service.ts
```diff
@@ -0,0 +1,84 @@
...
+import { CascadingEntityAction } from "./cascading-entity-action";
+import { UnsavedChangesService } from "app/core/entity-details/form/unsaved-changes.service";
+import { lastValueFrom } from "rxjs";
+import { EntityBulkEditComponent } from "./entity-bulk-edit/entity-bulk-edit.component";
+import { MatDialog } from "@angular/material/dialog";
+import { EntityActionsService } from "./entity-actions.service";
+
+/**
+ * Safely edit an entity including handling references with related entities.
+ * This service is usually used in combination with the `EntityActionsService`, which provides user confirmation processes around this.
```
sleidig, line 14:
```suggestion
 * Bulk edit fields of multiple entities at once.
```

sleidig (approved):
Made some final cleanup of typings and names myself. Good to go now.

Good work, @Abhinegi2 :pray:

PR #2574, opened 2024-09-12
ci: update download source for chrome images

tomwwinter:


sleidig (approved):
I am still not really convinced of pinning (+ having to maintain) a specific chrome version here, though.

Issue #2249, opened 2024-02-20
nested / hierarchical menu items


sleidig:
When there are a lot of different menu items on the left (many different entities), the menu becomes hard to navigate. Nested items, expanding upon hovering on an entry, will be useful.

![image](https://github.com/Aam-Digital/ndb-core/assets/1682541/c54d78ac-5f4b-4ad5-bd5f-5e0c19d44926)

- work in `NavigationComponent` 
- maybe create a storybook file for this component for easier testing and design
- check Angular Material [menu](https://material.angular.io/components/menu/examples) and [nav-list](https://material.angular.io/components/list/overview#navigation-lists) components and experiment with them to figure out the best way to handle this
- adapt `MenuItem` config structure to allow nested items (in a backwards-compatible way; see comments below in discussion)
- implement code for nested items

PR #2609, opened 2024-10-09
nested navigation menu items

sadaf895:
#2249


src/app/core/config/dynamic-routing/route-permissions.service.ts
```diff
@@ -21,7 +21,7 @@ export class RoutePermissionsService {
   async filterPermittedRoutes(items: MenuItem[]): Promise<MenuItem[]> {
     const accessibleRoutes: MenuItem[] = [];
     for (const item of items) {
-      if (await this.isAccessibleRouteForUser(item.link)) {
+      if (await this.isAccessibleRouteForUser(item.link as string)) {
```
sleidig, line 24: This is slightly tricky: We check and remove menu items here if the current user doesn't have permissions to access the link.

For submenu sections we'll have to check the subMenu items instead of the item itself (which doesn't have a link, right?) and
- [ ] add the parent menu item only if at least one of the submenu items is accessible
- [ ] check each submenu item and only show those that are accessible to the current user

sleidig: maybe instead of doing this here, we can move it into the individual `MenuItemComponent`.

sleidig (commented):
@sadaf895 , let us look at the permission aspects together in a joint session

src/app/core/config/dynamic-routing/route-permissions.service.ts
```diff
@@ -20,15 +20,27 @@ export class RoutePermissionsService {
...
-      if (await this.isAccessibleRouteForUser(item.link)) {
+      if (item.link && (await this.isAccessibleRouteForUser(item.link))) {
         accessibleRoutes.push(item);
+      } else if (item.subMenu) {
+        const accessibleSubItems: MenuItem[] = await this.filterPermittedRoutes(
+          item.subMenu,
+        );
+
+        if (accessibleSubItems.length > 0) {
+          const filteredParentItem: MenuItem = Object.assign({}, item);
```
sleidig, line 33:
```suggestion
          // only adding the item if there is at least one accessible subMenu item
          const filteredParentItem: MenuItem = Object.assign({}, item);
```

src/app/core/ui/navigation/menu-item.component.ts
```diff
@@ -0,0 +1,41 @@
...
+    RouterLink,
+    Angulartics2Module,
+    NgForOf,
+    MatMenuModule,
+  ],
+  standalone: true,
+})
+export class MenuItemComponent {
+  @Input() item: MenuItem; // Receives each individual menu item as input
+  activeLink: string;
```
sleidig, line 30: what is "activeLink" used for?

src/app/core/ui/navigation/navigation/navigation.component.ts
```diff
@@ -26,9 +26,11 @@ import { NgForOf } from "@angular/common";
 import { Angulartics2Module } from "angulartics2";
 import { FaDynamicIconComponent } from "../../../common-components/fa-dynamic-icon/fa-dynamic-icon.component";
 import { RoutePermissionsService } from "../../../config/dynamic-routing/route-permissions.service";
+import { MatMenuModule } from "@angular/material/menu";
+import { CommonModule } from "@angular/common";
+import { MenuItemComponent } from "../menu-item.component";
 
-/**
- * Main app menu listing.
+/* Main app menu listing.
```
sleidig, line 33:
```suggestion
/**
 * Main app menu listing.
```

src/app/core/ui/navigation/navigation/navigation.component.ts
```diff
@@ -82,16 +88,16 @@ export class NavigationComponent {
...
-      newUrl.startsWith(item.link),
+    const items: MenuItem[] = this.menuItems.filter(
+      (item) => typeof item.link === "string" && newUrl.startsWith(item.link),
     );
     switch (items.length) {
       case 0:
         return "";
       case 1:
-        const link = items[0].link;
+        const target = items[0].link as string;
```
sleidig, line 98: @sadaf895 , I think we don't need to rename this anymore. Do you want to undo these changes and explicit types ("as string") to keep the changes to a minimum?

src/app/core/ui/navigation/menu-item.component.ts
sadaf895: To keep track of the currently active route or link within the menu.

sadaf895: Yes, right  I should pass the activeLink value as an @Input to MenuItemComponent.

sadaf895:
It is working for me now, @sleidig Let me know that whether it is working well for you also.

src/app/core/config/dynamic-routing/route-permissions.service.spec.ts
```diff
@@ -20,4 +38,86 @@ describe("RoutePermissionsService", () => {
...
+
+Nested:
+  item 1
+    1.1
+    1.2   x
+
+  item 2  x
+    2.1   x
+
+ */
```
sleidig, line 123:
```suggestion
 */

```

src/app/core/ui/navigation/menu-item.component.ts
sleidig: > It is working for me now, Let me know that whether it is working well for you also.

Nope, not working for me @sadaf895 . Did you run the app and see the menu item highlighted correctly? We need to get back to a design where the currently visited menu is white background (see screenshots)

In the code I don't see you passing in any value to the `activeMenuItem` input also.

Issue #2401, opened 2024-06-06
Public Forms should support dynamic prefilled / default values


sleidig:
The "default" prefilled values of a field in a Public Form (`PublicFormConfig.prefilled`) currently only allows fixed values. This should support the same dynamic placeholders that `EntitySchemaField.defaultValue` allows to be configured.

PR #2635, opened 2024-10-24
Feature/public config should suport dynamic values

Abhinegi2:
closes: #2401 

### Visible/Frontend Changes

- [x] Default value will be visible in public forms


Abhinegi2:
let me resolve the conflicts.

src/app/features/public-form/public-form.component.spec.ts
```diff
@@ -73,17 +73,6 @@ describe("PublicFormComponent", () => {
...
-    testFormConfig.prefilled = { name: "new", category: "M" };
-    initComponent();
-    tick();
-
-    expect(component.entity.name).toBe("new");
-    expect(component.entity.category).toBe(
-      genders.find(({ id }) => id === "M"),
-    );
-  }));
-
```
tomwwinter, line 86: Can you please write a new test for the changed input parameters?

src/app/features/public-form/public-form.component.ts
```diff
@@ -101,9 +96,13 @@ export class PublicFormComponent<E extends Entity> implements OnInit {
     Object.entries(this.prefilled).forEach(([prop, value]) => {
       this.entity[prop] = value;
     });
```
tomwwinter, line 98: I think, this can be removed.

src/app/features/public-form/demo-public-form-generator.service.ts
```diff
@@ -18,8 +18,10 @@ export class DemoPublicFormGeneratorService extends DemoDataGenerator<PublicForm
     form.title = $localize`Example form`;
     form.description = $localize`This is a form that can be shared as a link or embedded in a website. It can be filled by users without having an account. For example you can let participants self-register their details and just review the records within Aam Digital.`;
     form.entity = "Child";
-    form.prefilled = { status: "new" };
-    form.columns = [["name", "phone", "gender", "dateOfBirth", "center"]];
+    form.columns = {
+      fields: ["name", "phone", "gender", "dateOfBirth", "center"],
```
tomwwinter, line 22: You can add a defaultValue config here for better testing: 

```js

fields: [ 
{
  id: "name",
  defaultValue: {
    mode: "static",
    value: "static name example",
  },
},
"phone", 
"gender",
"dateOfBirth",
"center"
]

```

tomwwinter (commented):
Looks good so far. However, somehow the default values are not applied yet. 

Next step would be to take a deeper look at EntityFormService.createEntityForm() and analyse, why the default value is not applied correctly.

src/app/core/config/config.service.ts
```diff
@@ -68,6 +68,7 @@ export class ConfigService extends LatestEntityLoader<Config> {
       migrateEntityBlock,
       migrateGroupByConfig,
       addDefaultNoteDetailsConfig,
+      migratePublicFormConfig,
```
sleidig, line 71: This will probably not work yet because these migrations are only applied to the "big" Config:CONFIG_ENTITY doc and not to other Config entities ...

maybe we should move this migration code to the PublicForm component and execute it there?

Abhinegi2: As we removed the `PublicFormConfig.prefilled` from the PublicConfig and its usages from component file, so still we need to use this migration in PublicComponent?

Abhinegi2:
> Looks good so far. However, somehow the default values are not applied yet.
> 
> Next step would be to take a deeper look at EntityFormService.createEntityForm() and analyse, why the default value is not applied correctly.

@tomwwinter I have updated the EntityFormService now the default value is working fine. You can test this further.

src/app/core/config/config.service.ts
sleidig: There migration will be needed for backward compatibility. Some clients system config will still have this old format. But let's finish the basic implementation and then think how the migration would look

tomwwinter: This migrations are just handling the Config:CONFIG_ENTITY document. For this kind of migration we need here, is currently no approach implemented.

What we should do, is to do this migration directly in the public-form.component. To be able to test it, you can use this real config from another instance:

```json
{
  "_id": "PublicFormConfig:1",
  "_rev": "2-d0073d2edeb0f1bf419c299603263dba",
  "title": "Mentee Test Formular",
  "description": "Dies ist nur ein Beispiel Formular für Mentees. Interessierte können dieses Formular ausfüllen und die Daten werden danach durch die Mitarbeiter:innen eingesehen.",
  "entity": "Child",
  "columns": [
    [
      "vorname"
    ],
    [
      "name"
    ],
    [
      "gender"
    ],
    [
      "dateOfBirth"
    ],
    [
      "email"
    ],
    [
      "phone"
    ]
  ],
  "prefilled": {
    "projectStatus": "AUSWAHL"
  }
}
```

tomwwinter (changes requested):

tomwwinter (approved):
LGTM :)

src/app/features/public-form/public-form-config.ts
```diff
@@ -1,12 +1,13 @@
...
 import { DatabaseField } from "../../core/entity/database-field.decorator";
+import { FieldGroup } from "app/core/entity-details/form/field-group";
 
 @DatabaseEntity("PublicFormConfig")
 export class PublicFormConfig extends Entity {
   @DatabaseField() title: string;
   @DatabaseField() description: string;
   @DatabaseField() entity: string;
-  @DatabaseField() columns: string[][];
+  @DatabaseField() columns: FieldGroup;
```
sleidig, line 11: I actually had thought to get `columns: FieldGroup[]` here (aligning exactly with `FormConfig` interface for the details views). Do you want to still quickly adapt this, @Abhinegi2 ?

src/app/features/public-form/public-form-config.ts
```diff
@@ -1,12 +1,13 @@
...
   @DatabaseField() prefilled: { [key in string]: any };
```
sleidig, line 12:
```suggestion
```

This is not supported anymore, is it?

Abhinegi2: Still we are using this on the migration which we have added for backward compatibility in publicformcomponent, by checking if this is present remove this from config.

src/app/features/public-form/public-form-config.ts
Abhinegi2: Sure Sebastian will work on this.

src/app/core/config/config.service.ts
sleidig: @Abhinegi2 , I have added a unit test for migration because we also need support to migrate from the old "prefilled" config field (this seems to not work yet - neither in the app when I test on an old public form config stored in my db, nor in the new unit test).

Have you implemented migration for the default values that are defined as "prefilled" yet? Can you please look into this?

src/app/features/public-form/public-form-config.ts
sleidig: ah, I understand. But if the current/new config format doesn't allow this anymore without migration, let's remove it from this class. (for the migration method, you can still access the old field, e.g. using `formConfig["prefilled"]`)

sleidig:
- [ ] we probably need a feature to hide a field from the forms to complement this (because until now, we used "prefilled" to set some default value on an invisible field/property in the background)

Abhinegi2:
> * [ ]  we probably need a feature to hide a field from the forms to complement this (because until now, we used "prefilled" to set some default value on an invisible field/property in the background)

Thank you for the feedback!  I assume we might need to introduce a "Hide" property for fields as part of the overall system configuration, similar to how we manage properties like "Is required" and "default value." 
Additionally, I wanted to confirm if we should implement this feature in the current PR or if we can create a separate issue to handle these changes.

sleidig:
> > * [ ]   we probably need a feature to hide a field from the forms to complement this (because until now, we used "prefilled" to set some default value on an invisible field/property in the background)
> 
> Thank you for the feedback! I assume we might need to introduce a "Hide" property for fields as part of the overall system configuration, similar to how we manage properties like "Is required" and "default value." Additionally, I wanted to confirm if we should implement this feature in the current PR or if we can create a separate issue to handle these changes.

Yes, exactly. Let's add a property `hidden` to the SchemaFieldConfig interface. You can use this for migrations here if a field appeared only in "prefilled" but not in "columns".

However, let us implement the "hidden field" logic in a separate branch and PR.

src/app/features/public-form/public-form-config.ts
Abhinegi2: made the requested changes please have a look now-

- [ ] update the type columns: FieldGroup[]
- [ ] Added migration for the default values that are defined as "prefilled"

src/app/core/config/config.service.ts
sleidig: When testing I still notice some problems even though the unit test passed (not sure what I missed in the test case). I have adjusted the logic myself a bit and got it working now.

src/app/core/common-components/entity-form/entity-form.service.ts
```diff
@@ -87,6 +87,10 @@ export class EntityFormService {
     forTable = false,
   ): FormFieldConfig {
     const fullField = toFormFieldConfig(formField);
+
+    if (fullField.defaultValue) {
+      entityType.schema.get(fullField.id).defaultValue = fullField.defaultValue;
+    }
```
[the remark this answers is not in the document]
Abhinegi2: While adding a default value for a field, I noticed it doesn't work when added only to the field group array. It works correctly only when included in the schema. 
Therefore, I added it here to ensure the default value is passed, even if it is within a field group, and applied correctly in the form.

sleidig:
This is ready to merge as soon as #2668 is available

src/app/core/common-components/entity-form/entity-form.service.ts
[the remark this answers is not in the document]
sleidig: Then this is actually something we need to change and fix here.

Currently, this PR leads to the following behavior:
1. if I set up "default name" in the public form for some outside users (or more realistically, a "needs confirmation" status)
2. any authorized user using the normal entity details view (Children List --> Add New) will get those same default values

This is not feasible for the kind of flexible public forms we require. The defaultValue config must be locally overwritable for a single view only, the same way this is already possible for the label for example.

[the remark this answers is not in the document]
Abhinegi2: Thank you for the feedback
I will review how the label configuration is currently implemented to allow local overrides and then work on applying a similar approach for default values.

src/app/features/public-form/public-form-config.ts
sleidig: Sorry, turns out we need this afterall - otherwise the EntityMapperService just skips this property when loading old configs from the DB. I added it again as "deprecated"

src/app/features/public-form/public-form.component.ts
```diff
@@ -147,6 +148,25 @@ export class PublicFormComponent<E extends Entity> implements OnInit {
...
+  
+            let defaultValue = null;
+              defaultValue = value;
+  
+            if (defaultValue !== null) {
+              const targetFormControl = this.form.formGroup.get(fieldName);
+              targetFormControl.setValue(defaultValue);
+            }
+          }
+        });
```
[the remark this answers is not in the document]
Abhinegi2: I have investigated this and found that `await this.defaultValueService.handleEntityForm(entityForm, entity) `is responsible for setting the default values. 
However, the current implementation of **defaultValueConfigs** only checks for default values if they are defined directly in the entity schema, and does not work/is resposible inside fieldGroups.

PR #2640, opened 2024-10-26
fix: prevent duplicate keys in filter options

tomwwinter:
prevent angular warnings for duplicate keys in options drop down.

### DIFF
--- a/src/app/core/filter/filters/filters.ts
+++ b/src/app/core/filter/filters/filters.ts
@@ -79,15 +79,17 @@ export class SelectableFilter<T extends Entity> extends Filter<T> {
     attributeName: string,
   ): FilterSelectionOption<T>[] {
     let keys = new Set();
-    return valuesToMatchAsOptions
-      .filter((k) => !!k)
-      .map((k) => ({
-        key: k.toString().toLowerCase(),
-        label: k.toString(),
-        filter: { [attributeName]: k } as DataFilter<T>,
-      }))
-      // remove duplicates:
-      .filter((value) => !keys.has(value.key) && keys.add(value.key));
+    return (
+      valuesToMatchAsOptions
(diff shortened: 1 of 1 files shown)


src/app/core/filter/filters/filters.ts
```diff
@@ -78,13 +78,15 @@ export class SelectableFilter<T extends Entity> extends Filter<T> {
...
+    let keys = new Set();
     return valuesToMatchAsOptions
       .filter((k) => !!k)
       .map((k) => ({
         key: k.toString().toLowerCase(),
         label: k.toString(),
         filter: { [attributeName]: k } as DataFilter<T>,
-      }));
+      }))
+      .filter((value) => !keys.has(value.key) && keys.add(value.key));
```
sleidig, line 89: wouldn't it be slightly more readable to just convert the result to a Set to get unique values rather than this filter function?
`return [...new Set(resultWithDuplicates)];`
https://dev.to/clairecodes/how-to-create-an-array-of-unique-values-in-javascript-using-sets-5dg6

sleidig (approved):

src/app/core/filter/filters/filters.ts
sleidig: oh, you are probably right! :blush: 
```suggestion
      // remove duplicates:
      .filter((value) => !keys.has(value.key) && keys.add(value.key));
```

(this is probably also possible without the Set as helper with somethign like `filter((value, index, array) => array.find(e => e.key) === index);`  but either one works)

sleidig:
@tomwwinter , we have some lint errors here now

sleidig:
@sadaf895 , would you be able to fix the `ng lint` errors for us in this branch? :blush:

Issue #2615, opened 2024-10-14
Attendance status sometimes not saved


sleidig:
**To Reproduce**

1. Create a new note and select category "school class" (or another type allowing attendance participants)
2. Add new children and select an attendance for each in the drop-down 
3. Save and reopen that note 
4. Often/sometimes the drop-down for attendance status of the children is empty again

tomwwinter:
When opening Add Note Dialog, this error is logged to console: 

![Image](https://github.com/user-attachments/assets/a37218cc-428c-47e5-b9a4-b55a3d339301)

PR #2641, opened 2024-10-26
fix: attendance status sometimes not saved

tomwwinter:
closes: #2615


src/app/core/common-components/entity-select/entity-select.component.ts
```diff
@@ -69,7 +69,7 @@ export class EntitySelectComponent<
     }
 
     this._entityType = Array.isArray(type) ? type : [type];
-    this.loadAvailableEntities();
+    this.loadAvailableEntities().then((_) => {});
```
sleidig, line 72: Is this required? (and what effect does it have? Promises should always execute, not only upon subscription, I think?)

sleidig (approved):
Seems to work now :+1:

PR #2644, opened 2024-10-30
feat: add multi-query support

tomwwinter:
image


src/app/features/reporting/report-config.ts
```diff
@@ -24,15 +24,46 @@ class ReportConfig extends Entity {
...
+   *  (sql v2 only) ReportDefinitionItem, ether ReportQuery or ReportGroup
+   *
+   *  Can be ReportQuery:
+   *  {query: "SELECT * FROM foo"}
+   *
+   *  Can be ReportGroup:
+   *  {groupTitle: "This is a group", items: [...]}
+   *
+   */
+  @DatabaseField() reportDefinition: ReportDefinitionDto[];
```
sleidig, line 60: maybe better "reportDefinitions" as this is an array?

src/app/features/reporting/report-config.ts
```diff
@@ -24,15 +24,46 @@ class ReportConfig extends Entity {
...
+  /** (reporting/exporting only, in browser reports) the definitions to calculate the report's aggregations */
   @DatabaseField() aggregationDefinitions: any[] = [];
 
-  /** (sql only) the definition to calculate the report */
+  /** (sql v1 only) the definition to calculate the report */
   @DatabaseField() aggregationDefinition: string | undefined = undefined;
+
+  /**
+   *  (sql v2 only) transformations that are applied to input variables (e.g. startDate, endDate)
+   *  example: {startDate: ["SQL_FROM_DATE"], endDate: ["SQL_TO_DATE"]}
```
sleidig, line 44:
```suggestion
   *  (sql v2 only) transformations that are applied to input variables (e.g. startDate, endDate)
   * the keys can be used as placeholders (`$key`) in SQL queries
   *  example: {startDate: ["SQL_FROM_DATE"], endDate: ["SQL_TO_DATE"]}
```

Is this the correct format or ${key}?

src/app/features/reporting/report-config.ts
```diff
@@ -24,15 +24,46 @@ class ReportConfig extends Entity {
...
+   *  {groupTitle: "This is a group", items: [...]}
+   *
+   */
+  @DatabaseField() reportDefinition: ReportDefinitionDto[];
+}
+
+export interface ReportDefinitionDto {
+  query?: string;
+  groupTitle?: String;
+  items: ReportDefinitionDto[];
```
sleidig, line 66: shouldn't "items" also be optional? (not required when just having query)

sleidig, line 66:
```suggestion
  /** an SQL query */
  query?: string;
  
  /** title (human-readable) for a set of hierarchically grouped sub-items */
  groupTitle?: String;
  
  /** hierarchical child items, building a recursive set of report groups display in an indented way */
  items: ReportDefinitionDto[];
```

src/app/features/reporting/report-config.ts
```diff
@@ -69,13 +100,39 @@ export interface ExportingReport extends ReportConfig {
...
 
   /**
-   * a list of arguments, passed into the sql statement
+   * a list of arguments, passed into the sql statement (only v1)
    */
   neededArgs: string[];
+
+  /**
+   *  (sql v2 only) transformations that are applied to input variables (e.g. startDate, endDate)
+   *  example: {startDate: ["SQL_FROM_DATE"], endDate: ["SQL_TO_DATE"]}
```
sleidig, line 121: maybe we remove the detailed code comment in the class (or here in the interface; which one is imported in other services?) and replace it with a "see SqlReport interface", so that we do not have to keep the comments updated in two places?

src/app/features/reporting/reporting/select-report/select-report.component.ts
```diff
@@ -87,10 +87,16 @@ export class SelectReportComponent implements OnChanges {
   private checkDateRangeReport(): void {
     if (this.selectedReport.mode !== "sql") {
       this.isDateRangeReport = true;
-    } else {
+    } else if (this.selectedReport.version == 1) {
```
sleidig, line 90: or undefined?

src/app/features/reporting/reporting/sql-v2-table/sql-v2-table.component.html
```diff
@@ -0,0 +1,33 @@
...
+      mat-cell
+      *matCellDef="let row"
+      [style.padding-left.px]="16 + row.level * 20"
+    >
+      {{ row.key }}
+    </td>
+  </ng-container>
+
+  <ng-container matColumnDef="Anzahl">
+    <th mat-header-cell *matHeaderCellDef>Anzahl</th>
```
sleidig, line 20:
```suggestion
  <ng-container matColumnDef="Count">
    <th mat-header-cell *matHeaderCellDef i18n>Count</th>
```

i18n maybe needed in some of the other elements also

src/app/features/reporting/reporting/sql-v2-table/sql-v2-table.component.ts
```diff
@@ -0,0 +1,72 @@
...
+            value: value
+              .flatMap((value) => {
+                return Object.values(value).filter(
+                  (value) => typeof value === "number",
+                );
+              })
+              .reduce(
+                (previousValue: number, currentValue: number) =>
+                  previousValue + currentValue,
+              ),
```
sleidig, line 60: can we put this in a named helper function for readability? I guess this sums up all sub-item values?

sleidig (commented):
Code looks very good :+1: 

Is this deployed for testing already?

src/app/features/reporting/report-config.ts
tomwwinter: It's just $key. e.g. 
```sql
SELECT * FROM WHERE c.date BETWEEN $startDate AND $endDate
```

src/app/features/reporting/sql-report/sql-report.service.ts
```diff
@@ -76,7 +85,144 @@ export class SqlReportService {
...
+   *       "status": "PATENSCHAFT",
+   *       "foerderungDurch": null
+   *     }
+   *   ]
+   * ]
+   *
+   * @param data raw ReportCalculationData or sub-array
+   * @param level each level will add some left-padding to visualize hierarchy
+   */
+  flattenData(data: any[], level = 0): SqlReportRow[] {
```
sleidig, line 158: strictly speaking we should add some unit tests for such methods ...

src/app/features/reporting/sql-report/sql-report.service.ts
```diff
@@ -185,4 +331,29 @@ export class SqlReportService {
...
+  getCsv(sqlData: SqlReportRow[]): any {
+    let deepestLevel: number = sqlData.reduce(
+      (previousValue, currentValue) =>
+        Math.max(previousValue, currentValue.level),
+      0,
+    );
+
+    let csv = "";
+
+    csv += "Name" + ",".repeat(deepestLevel + 1) + "Value\n";
```
sleidig, line 344: any particular reason why we don't use PapaParse library for this?

Are we sure values are always escaped properly to not break CSV formating / delimiters?

tomwwinter: Was not that easy to handle the nesting properly with Papa. Key and Value are properly escaped.

Issue #2460, opened 2024-07-15
New Datatype: url


sleidig:
In addition to a normal text datatype it would be useful to be able to display a field as a clickable link (URL).

- Implement a new "Datatype" by copying the `LongTextDatatype` and `EditLongTextComponent` and `DisplayTextComponent`
- adjust the template to display an clickable link instead of simple text

PR #2646, opened 2024-10-30
Url datatype

sadaf895:
closes: #2460


src/app/core/basic-datatypes/string/display-url/display-url.component.ts
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * This component displays a URL attribute as a clickable link.
+ */
+@DynamicComponent("DisplayUrl")
+@Component({
+  selector: "app-display-url",
+  template: `
+    <a *ngIf="value" [href]="value" target="_blank">{{ value }}</a>
+    <span *ngIf="!value">No URL available</span>
```
sleidig, line 14:
```suggestion
    <span *ngIf="!value">-</span>
```

src/app/core/basic-datatypes/string/edit-url/edit-url.component.html
```diff
@@ -0,0 +1,17 @@
...
+    <a [href]="formControl.value" target="_blank" matInput>
+      {{ formControl.value || "No URL available" }}
+    </a>
+  </ng-container>
+  <ng-template #inputField>
+    <input [formControl]="formControl" matInput [title]="label" type="url" />
+  </ng-template>
+
+  <mat-error *ngIf="formControl.invalid && formControl.touched">
+    <span *ngIf="formControl.errors?.['pattern']">Invalid URL format</span>
```
sleidig, line 14:
```suggestion
    <span *ngIf="formControl.errors?.['pattern']" i18n>Invalid URL format</span>
```
marking this to be translatable

sleidig (changes requested):
It seems like you didn't test this in the app yet :wink:   It didn't work for me because we forgot to register the components (see my commit now).

The component still has some issues. Test it while keeping the Developer Console open and see various errors. I think it is not "allowed" to mark a `<a` with matInput, this only works for `<input` as far as I know.

sleidig:
`EditFileComponent` might have some logic applied that allows users to click even on the disabled field. Maybe you, @Abhinegi2 and @sadaf895, can look at this together to finish this PR

Abhinegi2:
@sadaf895 I made some modifications to allow user clicks even when the form is disabled. I also added test cases and updated the cursor styling. Please review these changes and let me know if any functionality is missing.  

To enable user clicks while the form is disabled, I followed this approach(which we already following in `EditFileComponent ` ):  
[(https://stackoverflow.com/questions/3100319/event-on-a-disabled-input/32925830#32925830)](https://stackoverflow.com/questions/3100319/event-on-a-disabled-input/32925830#32925830)

src/app/core/basic-datatypes/string/edit-url/edit-url.component.html
```diff
@@ -0,0 +1,22 @@
...
+  <mat-label>{{ label }}</mat-label>
+
+  <!-- Single input field for both editable and disabled states -->
+  <input
+    [formControl]="formControl"
+    matInput
+    [title]="label"
+    type="url"
+    [readonly]="formControl.disabled"
+    placeholder="Enter URL"
```
sleidig, line 14:
```suggestion
    placeholder="Enter URL"
    i18n-placeholder
```

src/app/core/basic-datatypes/string/edit-url/edit-url.component.html
sleidig: this i18 tag is still missing.

Also, is "pattern" used anywhere to check validity?

sleidig (changes requested):
Code and functionality looks pretty good already :slightly_smiling_face: 

From functional testing a few suggestions to polish this further. Can you take these points up again, @sadaf895 ?

- [x] in the list view, the link is not working and instead the details view of record is loaded. If clicking exactly on that link, I would like to instead open the external url in a new tab (@Abhinegi2 has implemented something similar only last week)
- [x] The form field's link text should look a bit more like a clickable link. I suggest we give this an underline styling or maybe underline on hover (i.e. maybe giving this our global "clickable" css class)
- [x] if the value is not a valid url (e.g. just some random text), this anyway opens a new tab to our app. In such a case the link should not be clickable (and not styled like a link
- [ ] if I put a URL without leading http or https it is not opened as an external link. Probably we should prefix that automatically if it is required to work as links. Maybe we can use a similar design to the screenshot below, where "https://" is always automatically prefixed and cannot be changed?

image

If unsure about design, do ask for feedback on our Slack

src/app/core/basic-datatypes/string/edit-url/edit-url.component.ts
```diff
@@ -0,0 +1,52 @@
...
+  ],
+  standalone: true,
+})
+export class EditUrlComponent extends EditComponent<string> {
+  
+  /**
+   * Checks if the current formControl value is a valid URL.
+   * Returns true only if it is a properly formatted http/https URL.
+   */
+  isValidUrl(value: string | null): boolean {
```
sleidig, line 31: unit test cases to implement for this functionality:
- `it("should be invalid for value that is not a complete URL"`:  input value: `"example"`, expected result: `false`
- `it("should be valid for empty value"`:  input value: `""` or input value: `null` (test both), expected result: `true`
- `it("should be valid for url-like value"`:  input value: `"https://example.com"`, expected result: `true`

sadaf895:
@george-neha Can you please check the functionality as a user's perspective and share your feedback on it.

sadaf895:
> As a user, I will do either of the two things:
> 
> 1. type out the website link (more probable),
> 2. copy and paste the link the URL.
> 
> I feel adding a prefix to the website field with https:// and letting the user type in the website name would be more user-friendly. (see image below) image
> 
> In scenarios where a user will copy-paste the URL. Would it be possible to truncate the additional https://? What do you think @sadaf895

@george-neha Can you now check the functionality? Any improvement or modification is needed?
@sleidig @Abhinegi2 Can you both also please check the updated functionality or any feedback you want to give regarding this?

Abhinegi2:
@sadaf895 While testing, I found a UI glitch. Here’s how to reproduce it:

- Enter or paste any website URL in the field (e.g., https://www.talisman.org/tao/).
- Try deleting the URL one character at a time. The `https://` prefix doesn’t get removed and keeps reappearing in a loop. 

However, if you select(complete url) and delete the entire url at once, it works correctly.

Abhinegi2:
@george-neha could you test it again, and let me know if you have any suggestions

sleidig:
This is not properly usable yet, unfortunately ...
Adding and deleting parts of the prefix creates a number of weird results and issues.

I think we should maybe spell out sensible behavior and test cases for these edge cases of editing:
- trying to delete part of the prefix as a user
- supporting something other than https (just http, for example)


-----
Clicking links seems to work very well now, however :+1: 

Only the cursor here should be a pointer to indicate clickability:
image

sadaf895:
Test Cases for Editing a Website URL:

[ ] Entering a valid URL
- Input: http://example.com
- Expected Output: URL is saved and displayed as a clickable link.

[ ] Editing an existing URL
- Given: http://old-site.com
- Action: Change to http://new-site.com and save.
- Expected Output: New URL is displayed and clickable.

[ ] Clearing an existing URL
- Given: http://example.com
- Action: Remove the URL and save.
-Expected Output: The field is empty after saving.

sadaf895:
list of issues need to be fixed
- [ ] for any changes made it need to click the save button twice
- [ ]  if I delete any character/letter of "https://," then the system is auto-creating the precursor of http:// resulting in a website link http://https:/snuniv.in
- [ ] invalid URL ka functionality is now working only after i click on save and click on edit again.

sleidig:
After discussing the problems giving users confusing results when changing the prefix, we decided the following should be the behaviour:

- _typing / pasting something: if without "http://" or "https://" prefix -> immediately add "https://" automatically_
  - if the first entered letter is "x" (for example) -> show "https://x"
  - if the first entered letter is "h" -> show "https://" (not "https://h")
  - if the first letter is a whitespace " " -> trim that and show "https://"
  - if the value (pasted) is a url starting with "http://" -> do not add a prefix and keep the value as it is
  - if the value (pasted) is a url starting with "https://" -> do not add a prefix and keep the value as it is
- _changing an existing value: do not "accept" changes that "break" the valid prefix_
  - changing any part after the "://" -> just do the change without changes (trim whitespaces at the end)
  - deleting any letter of the "https://" (e.g. new value "htps://google.com") -> revert back to the valid "https://" (e.g. "https://google.com")
  - changing from "https://" to "http://" (e.g. new value "http://google.com") -> accept new value (e.g. "http://google.com")
  - adding/changing any letter to the prefix: e.g. "httpx://google.com" -> "https://google.com"
  - "httpS://" -> "https://" (but not important because URLs are case insensitive)

sleidig:
@sadaf895 , can you still check about the cursor pointer style (see comment above)? In EditFile field the cursor does change, for example.

Otherwise this is good to go, I think :-)

sleidig (approved):
Seems to work well now and looks like clean code :100: 

Nice work, @sadaf895 . This certainly was more complex than expected and took a long time. But the new dataType will be a really good feature for users, giving much better UX there.
Thanks for your support here, @george-neha , @Abhinegi2

PR #2648, opened 2024-11-04
fix: flatten result data before display it

tomwwinter:


sleidig (approved):
Not sure what this fixes and how to test it - but the code change looks simple enough :wink: 

I assume you check this with the different possible report results

Issue #2612, opened 2024-10-11
e2e Tests for all common app scenarios


sleidig:
We currently do not have e2e tests (i.e. "end-to-end" testing of whole scenarios how the user interacts with the app). This means to check if a change breaks existing functionally in other places, we currently click through the app ourselves which takes time and is unreliable. Instead, we should establish automated tests for this again which will also run in the CI of every PR.

- [x] choose a e2e test framework
  - Playwright seems to gain traction, is supported by angular [[>]](https://angular.dev/tools/cli/end-to-end#more-information-on-end-to-end-testing-tools) and has a VS Code plugin to record tests easily [[>]](https://playwright.dev/docs/codegen) 
  - Cypres might be an alternative
- [x] set up the testing framework in our repository
  - i.e. install npm packages, etc.
- [x] record some basic scenarios as samples
- [x] discuss with core team and create a list of ~all~ the first / most relevant scenarios
- [x] implement e2e tests for the first scenario

After the foundational work is good and merged:
- [ ] #2309
- [ ] #2889
- maybe set up Cucumber / Gherkin for non-technical test definitions [[see guide](https://www.browserstack.com/guide/playwright-cucumber), but also [these benefits and limitations](https://blog.cronn.de/en/testing/2023/09/12/cucumber-playwrigth-1-en.html)]
- ... then continue to gradually add e2e tests for _all_ relevant scenarios: [GoogleSheet](https://docs.google.com/spreadsheets/d/16woa2AYUjhkwD9D4hKWjzhaP0wI0PaR0WL8sudzSw74/edit?usp=sharing)

george-neha:
@sleidig Please review the test cases: https://docs.google.com/spreadsheets/d/16woa2AYUjhkwD9D4hKWjzhaP0wI0PaR0WL8sudzSw74/edit?usp=sharing.

sleidig:
Nice work, @george-neha :+1: 
Let's focus on implementing the "Attendance" scenario as a first automated e2e test and then take the learnings from that back to the other scenarios.

This looks pretty comprehensive. I guess we might at some point discover an obscure step/feature that is not covered in the scenario yet but it'll be easy to add then and this seems to cover the key functionality.

PR #2650, opened 2024-11-06
test(e2e) : for all common scenarios

sadaf895:
closes: #2612

----
- [x] can we init the app (open it and generate demo data) only once and then run all the tests there (instead of generating data every time)?
- [x] Breaking down the large test into smaller tests
- [x] Locators that does not depend on the exact text
- [x] adapt expect to not rely on the exact date in demo data (Mocking date)


tomwwinter:
@sadaf895 Hey,  i just pushed some updates. I restructured some files and removed cypress. Can you please take a look into the failing tests? It's not necessary to start the app anymore, just run `npm run e2e`

e2e/tests/dashboard-tests.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    // Check for the "Tasks due" widget
+    const tasksDueElement = page.locator("text=Tasks due");
+
+    await expect(tasksDueElement).toBeVisible();
+
+    // Verify that the task names and due dates are displayed
+    const taskNames = await page
+      .locator("app-widget-content")
+      .filter({ hasText: "get signed agreement Nov 19," })
```
sleidig, line 50: Does this test pass, even now (when 19th Nov is a long time in the past)?

sadaf895: I have restructured it.

e2e/tests/dashboard-tests.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+    const tasksDueElement = page.getByText("Tasks due");
+    await expect(tasksDueElement).toBeVisible();
+
+    // Verify at least one task is listed
+    const taskElements = page.locator("app-widget-content");
+    const taskCount = await taskElements.count();
+    expect(taskCount).toBeGreaterThan(0);
+  });
+
+  test("Navigate to Attendance page and verify sections", async ({ page }) => {
```
sleidig, line 60: The file here is called "Dashboard Tests". This case should probably be moved to the attendance tests file, if it is not already tested there somewhere in different tests?

e2e/tests/attendance-tests.spec.ts
```diff
@@ -0,0 +1,335 @@
...
+        }
+      }
+    }
+    const __DateNowOffset = ${fakeNow} - Date.now(); // Offset for Date.now()
+    const __DateNow = Date.now;
+    Date.now = () => __DateNow() + __DateNowOffset; // Override Date.now()
+  }`);
+}
+
+test.describe("Dashboard Page Tests", () => {
```
sleidig, line 23:
```suggestion
test.describe("Attendance Module", () => {
```

sleidig (changes requested):
The structure and overall test files look quite good now, @sadaf895  :slightly_smiling_face: 

The following are still consistently failing on my device when running `npm run e2e` currently. Are they working for you? Did my cleanups break anything? But not all were passing for me even before I made changes.

image

Abhinegi2:
@sleidig, I discussed the issue with the team regarding the Playwright integration failing in CI. After debugging, we found that in our `StartApp` function (inside `core-e2e-utils`), `backgroundTasksDetailsElement` is returning `null`, causing a timeout issue. Additionally, the following lines are not working as expected and also result in a timeout:  

```
await page.locator(`[aria-label="${startDateLabel}"]`).click();
await page.locator(`[aria-label="${endDateLabel}"]`).click();
```  
It seems like the elements are not being properly located or rendered in the test environment.

Abhinegi2:
@sleidig, I discussed the issue with the team regarding the Playwright integration failing in CI. 
After debugging, we found that the Angular app is taking longer to start in the CI environment, which is causing unexpected failures. To test this, we temporarily increased the timeout to **5 minutes**, and it is now working. We are currently analyzing the **idle start time**; it's around 2 minutes. 

1. Additionally, we found that certain test actions are **taking too long** and exceeding the **12,000ms timeout**. Specifically:  
- In our `StartApp` function (inside `core-e2e-utils`), `backgroundTasksDetailsElement` is returning `null`, leading to a timeout.  
- The following actions are not working as expected and also result in a timeout:  
  ```
  await page.locator(`[aria-label="${startDateLabel}"]`).click();
  await page.locator(`[aria-label="${endDateLabel}"]`).click();
  ```  
  Those elements may not be properly located or rendered in the test environment.  

2. Another issue is that **Playwright CI checks are currently taking nearly 30 minutes**, which is not ideal.

sleidig (approved):

Issue #1579, opened 2022-12-08
use GPS sensor data to set a location field


sleidig:
*As a social field worker I want to record the location of a participant's home with the GPS of my device to quickly document it even if no clear address exists.*

**Objective**
The "location" field datatype currently allows the user to either type a written address and search its geolocation or open an (OpenStreetmap) visual map and manually place a pin on that map. If no accurate streetnames exist or users are mostly doing field visits, this is not possible or convenient. Instead, there should be the option to use GPS location also.

With this new feature, the location field in the app will show an additional location-crosshairs icon next to the address input/search box. When the user clicks on that icon, the device's GPS sensor is accessed (if available) and the current location is loaded and set for the location field. The user sees a preview of the location in the map visualization and confirms this.

The behaviour and result of using the GPS sensor should be the same as clicking on the map to manually place a pin/location. (e.g. reverse-lookup of address should be supported to suggest a written address for display or otherwise a label is displayed that explains the location was fixed through GPS; see #1563 / #139).

![image](https://github.com/user-attachments/assets/83fae62f-609e-48c0-a96e-61f3ab002de8)

**Proposed Solutions & Alternatives**
- use PWA (progressive web app) GPS sensor APIs to access the current device location
- allow user to set these coordinates similar to how they can set coordinates by clicking on the map popup
- encapsulate the GPS access logic into a new Angular Service for easier testing and reuse

PR #2651, opened 2024-11-06
[Feat]: Use GPS sensor data to set a location field 

Ayush8923:
closes: https://github.com/Aam-Digital/ndb-core/issues/1579

### Visible/Frontend Changes
 - [x] Implemented a new location-crosshairs icon button next to the address input/search box, enabling users to capture their current GPS location and automatically populate the selected location field with the retrieved coordinates.


src/app/features/location/address-edit/address-edit.component.html
```diff
@@ -65,7 +65,21 @@
...
+      ></app-address-search>
+      <button
+        mat-icon-button
+        (click)="updateLocationFromGps()"
+        matTooltip="To accurately log your current address, please allow GPS access."
+        i18n-matTooltip
+      >
+        <fa-icon *ngIf="!gpsLoading" icon="location-crosshairs"></fa-icon>
+        <mat-spinner *ngIf="gpsLoading" diameter="24"></mat-spinner>
+      </button>
```
pankaj-ag, line 82: Should we create a component for this? Let's check what patterns we are following

Ayush8923, line 82: I think that if we can use the GPS button in multiple places, we could create a component for it. This way, we could use that component directly to get the user's location coordinates via GPS. Please let me know if creating a new component for this is necessary.

src/app/features/location/gps.service.spec.ts
```diff
@@ -0,0 +1,88 @@
...
+      (success) => {
+        success(mockGeolocationPosition() as GeolocationPosition);
+      },
+    );
+
+    try {
+      await service.getGpsLocationCoordinates();
+    } catch (error) {
+      expect(error).toBe("Geolocation is not supported by this browser.");
+    }
```
sleidig, line 86: I think this test also passes if no error is thrown at all (then that expectation is skipped). Is that what you intended?

I also don't see how you mock a failure or rejection in the test setup here. Can you explain what you have in mind here?

src/app/features/location/gps.service.ts
```diff
@@ -0,0 +1,72 @@
...
+    longitude: number;
+    accuracy: number;
+  }> {
+    if (!("geolocation" in navigator)) {
+      return;
+    }
+
+    const permissionStatus = await navigator.permissions.query({
+      // eslint-disable-next-line prettier/prettier
+      name: "geolocation", //NOSONAR geolocation is necessary
```
sleidig, line 22: why is disabling prettier necessary here?

src/app/features/location/gps.service.ts
```diff
@@ -0,0 +1,72 @@
+import { Injectable } from "@angular/core";
+import { GeoService } from "./geo.service";
+
+@Injectable({
+  providedIn: "root",
+})
+export class GpsService {
+  constructor(private readonly geoService: GeoService) {}
+  location: { lat: number; lon: number } | null = null;
```
sleidig, line 9: What is the purpose of storing the location in the service here? Please add a JsDoc comment with a short explanation.

src/app/features/location/address-edit/address-edit.component.ts
```diff
@@ -74,6 +83,27 @@ export class AddressEditComponent {
     });
   }
 
+  async updateLocationFromGps() {
+    this.gpsLoading = true;
+    try {
+      const location = await this.gpsService.getGpsLocationCoordinates();
+      const address = await this.gpsService.getGpsLocationAddress();
```
sleidig, line 90: To ensure the exact same behavior, let's somehow reuse the logic for lookup between MapPopupComponent `mapClicked()` handler and this case here where the raw coordinates are coming from GPS (rather than the mouse click as in mapClicked).

I am not completely sure what's the best way to do this. Maybe moving the lookupCoordinates method from MapPopupComponent into the service?

We could also keep GpsService independent of any address lookup, just focusing on the coordinates

sleidig (changes requested):
Really clean, well-structured code. Thanks!

I made just a few remarks to align it even better with our code base.

Functional testing:
- seems to work well, overall!
- [x] for explicit user feedback we could show a Toast message (using AlertService) after successfull address lookup saying something like "Selected address based on GPS coordinate lookup as "X Street, London""
- [ ] If the user denies GPS access (or another error occurs) we should display feedback to the user. Maybe also using AlertService 
  - [ ] is it possible to understand why errors occur (on first rejecting permission, I see that in the console. But after having blocked GPS and retrying, I then see a more generic error. Can we tell the user even in that case, that it failed because of missing permissions?)

Ayush8923:
Thanks, @sleidig for sharing your suggestions and feedback. I will take a look at this and make the changes accordingly ASAP.

src/app/features/location/address-edit/address-edit.component.html
sleidig: yes, I think that's a good suggestion. Let's turn this into a component with a simple Output property

src/app/features/location/gps.service.ts
Ayush8923: I had to disable Prettier for this specific line because it was causing issues with the SonarCloud CI checks. The problem was that Prettier's automatic indentation adjustments were moving the `// NOSONAR geolocation is necessary` comment to the next line, which led to the SonarCloud analysis failing. By disabling Prettier on this line, I was able to ensure that both Prettier and SonarCloud CI checks passed successfully without any conflicts.

sleidig: ah, I see. I think we, as admins, can review and confirm SonarCloud's security hotspot warnings on their platform also. Can you remove these comments once and let me check that?

src/app/features/location/gps.service.ts
Ayush8923: The purpose of storing the location in the location property is to cache the coordinates retrieved from the `getGpsLocationCoordinates()` method and then directly use the `location` property in the `getGpsLocationAddress` method. This helps to reduce repeated geolocation API calls. I will add a JsDoc comment in the code.

src/app/features/location/address-edit/address-edit.component.ts
```diff
@@ -74,6 +83,27 @@ export class AddressEditComponent {
...
+      this.updateLocation({
+        locationString: address,
+        geoLookup: {
+          lat: location.latitude,
+          lon: location.longitude,
+          display_name: address,
+        },
+      });
+    } catch (error) {
+      console.error("Error updating location:", error);
```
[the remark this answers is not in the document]
Ayush8923: Thanks for this feedback. I have updated the code accordingly.

src/app/features/location/gps.service.ts
sleidig: I have marked this as reviewed and safe now :+1:

src/app/features/location/address-edit/address-edit.component.ts
Ayush8923: I've updated the code to reuse the mapClicked() function for retrieving the address based on coordinates and to update the location accordingly. Could you please review the changes and let me know if any improvements are needed?

Ayush8923:
@sleidig I've updated the code as per your review suggestions and have also responded to your questions. However, I'm currently facing an issue with the `gps.service.spec.ts` test cases. The error message I'm getting `Uncaught ReferenceError: Cannot access 'MapPopupComponent' before initialization`. I'm actively working on resolving this.

src/app/features/location/gps.service.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+  let fixture: ComponentFixture<MapPopupComponent>;
+  let location = { lat: 51.5074, lon: -0.1278 };
+
+  beforeEach(async () => {
+    const geoServiceSpy = jasmine.createSpyObj("GeoService", ["lookup"]);
+
+    TestBed.configureTestingModule({
+      declarations: [MapPopupComponent],
+      providers: [GpsService, { provide: GeoService, useValue: geoServiceSpy }],
+    })
```
sleidig, line 35: Any specific reason why you want to use the component in this test setup, @Ayush8923 ?

For a classic unit test, I would only create an instance of the service itself and test the service in isolation without any components.

Ayush8923:
@sleidig I have updated the test cases for the `gps.service` file, and all the test cases are working fine. However, there is an issue in the `address-edit.component.spec.ts` file, where it's showing the error Cannot access `AddressEditComponent` before initialization. I'm not sure why this issue is occurring, as I haven't made any changes to this file. I read a few articles, and it seems like it might be related to circular dependencies. I am currently investigating the issue and trying to fix it.

image

src/app/features/location/address-edit/address-edit.component.html
```diff
@@ -65,7 +65,13 @@
 </div>
 
 @if (!disabled) {
-  <app-address-search
-    (locationSelected)="updateFromAddressSearch($event)"
-  ></app-address-search>
+  <div class="margin-bottom-regular container">
+    <div class="flex-row gap-regular align-start full-width">
```
sleidig, line 69: is there a benefit from having two divs rather than one here?

src/app/features/location/address-edit/address-edit.component.ts
```diff
@@ -16,6 +18,7 @@ import { FaIconComponent } from "@fortawesome/angular-fontawesome";
   selector: "app-address-edit",
   standalone: true,
   imports: [
+    NgIf,
```
sleidig, line 21: I think ngIf is not used anymore now, as the gps component was extracted? (the new `@if` syntax doesn't require an import)

src/app/features/location/address-gps-location/address-gps-location.component.ts
```diff
@@ -0,0 +1,57 @@
...
+          lat: location.latitude,
+          lon: location.longitude,
+        };
+        await this.mapPopupComponent.mapClicked(this.location);
+        this.alertService.addInfo("Selected address based on GPS.");
+      }
+    } catch (error) {
+      Logging.error(error);
+      this.alertService.addInfo(
+        "Location permission denied. Please enable it in your device settings.",
```
sleidig, line 51: Are you sure this always fails due to permissions? Or can there be other causes and we are possibly sending the user in a wrong direction here?

src/app/features/location/address-gps-location/address-gps-location.component.ts
```diff
@@ -0,0 +1,57 @@
...
+  async updateLocationFromGps() {
+    this.gpsLoading = true;
+    try {
+      const location = await this.gpsService.getGpsLocationCoordinates();
+      if (location) {
+        this.location = {
+          lat: location.latitude,
+          lon: location.longitude,
+        };
+        await this.mapPopupComponent.mapClicked(this.location);
```
sleidig, line 45: In order to reuse this in a clean way, we shouldn't inject the component but instead move the method into a service so that it is properly encapsulated (Single Responsibility Principle --> a component shouldn't provide general methods in addition to defining UI).

I suggest you use GeoService.reverseLookup here and change that reverseLookup method, so that it includes the fallback logic that currently is done in MapPopupComponent.lookupCoordinates (https://github.com/Aam-Digital/ndb-core/blob/85e527475199e26d0b17c868d0acf1acec1e26b9/src/app/features/location/map-popup/map-popup.component.ts#L111-L118).
Ideally mapClicked there can then directly use geoService.reverseLookup. And updateLocationFromGps can also use that same method without extra transformations.

src/app/features/location/address-edit/address-edit.component.ts
Ayush8923: Yes, you are right. Now, we don't need to import it. I will remove it from this component. Thanks!

src/app/features/location/address-gps-location/address-gps-location.component.ts
Ayush8923: I have updated the `reverseLookup` method as per your suggestion and added fallback logic, which was previously happening in `MapPopupComponent.lookupCoordinates`. I have one question here, do I also need to add the `updateLocation` logic in this `address.gps.location` component to help update the selected location field with the address based on coordinates?

sleidig: changes look good, thanks!

sleidig: As the GPS button is now a component, it should have an `@Output` property to emit whenever a new location was selected through that. Then you'll need to use that output in adress-edit.component to set the value. Maybe you can use the updateFromAddressSearch method. Please test that and see if it would need any adjustments

Ayush8923: Thanks, @sleidig for the suggestion. I will make the changes accordingly.

src/app/features/location/address-edit/address-edit.component.ts
```diff
@@ -114,8 +116,12 @@ export class AddressEditComponent {
       geoLookup: value?.geoLookup,
     });
   }
-}
 
-function matchGeoResults(a: GeoResult, b: GeoResult) {
-  return a.lat === b.lat && a.lon === b.lon;
```
Ayush8923, line 120: @sleidig I removed this unused function as part of a cleanup. I double-checked, and it doesn't appear to be used anywhere in the code.

src/app/features/location/address-gps-location/address-gps-location.component.ts
Ayush8923: @sleidig I reviewed this scenario and noticed that when the user denies the permission, an alert message is shown. The same alert message also appears if any code within the try block fails.

src/app/features/location/address-edit/address-edit.component.html
Ayush8923: I refactored the code by eliminating an extra div and combining its class with another div.

Ayush8923:
@sleidig Did you get a chance to check this PR? If the `MapPopupComponent.spec.ts` file needs to be updated with additional test cases, I can update the test cases accordingly.

src/app/features/location/address-gps-location/address-gps-location.component.ts
sleidig: So that means the most likely explanation is the permissions but it could also be any other kind of error. Then we should change that message to users a bit

```suggestion
        $localize`Failed to access device location. Please check if location permission is enabled in your device settings`,
```

src/app/features/location/geo.service.ts
```diff
@@ -87,9 +87,19 @@ export class GeoService {
    * @param coordinates of a place (`lat` and `lon`)
    */
   reverseLookup(coordinates: Coordinates): Observable<GeoResult> {
+    if (!coordinates) {
+      return undefined;
+    }
+
+    const fallback: GeoResult = {
+      display_name: $localize`[selected on map: ${coordinates.lat} - ${coordinates.lon}]`,
```
sleidig, line 95:
```suggestion
      display_name: $localize`[selected coordinates: ${coordinates.lat} - ${coordinates.lon}]`,
```

generalizing this a bit to not confuse people who used GPS location

src/app/features/location/gps.service.ts
sleidig: @Ayush8923 , I think in the current state of the implementation this is not used at all anymore?

src/app/features/location/address-gps-location/address-gps-location.component.ts
sleidig: nice, looks good!

src/app/features/location/gps.service.ts
Ayush8923: Yes, we are not using it anymore.

sleidig (approved):
Did some final polishing and testing. Good to go now. Thanks, @Ayush8923 !

Do have a look at the last small commits I made, to understand why your test was failing

Issue #2653, opened 2024-11-11
Date field can't be deleted / saved empty


sleidig:
**Describe the bug**
Date fields cannot be reset (after saving with an empty field, the previous date just reappears in that field again).

**To Reproduce**

1. Go to 'Notes'
2. Click on one note to open the details form
3. Select the date value and delete it; the form field appear empty now
4. Save the form
5. --> The date reappears and cannot be reset (deleting a dropdown value in another field works as expected)

PR #2664, opened 2024-11-15
fix: store null values from form fields

tomwwinter:
closes: #2653

@sleidig Is there a reason, why null values are not supposed to be stored in here?


sleidig (approved):
We didn't want to clutter the docs in the DB with a ton of properties holding "null" values. I adjusted your fix a bit to avoid that while retaining your solution :blush:

Issue #2542, opened 2024-08-27
Some permission rules can cause re-sync after every reload of the app


sleidig:
**Describe the bug**
On dev system the following rule causes the sync to re-start completely every time the app is reloaded or the user logs in anew.

**To Reproduce**
The following permission rule seems to cause the issue:
```
{
  "subject": "Child",
  "action": "manage",
  "conditions": {
    "assignedSchoolProject": {
      "$in": "${user.projects}"
    }
  }
}
```

--> demo.aam-digital.net with user `demo` (pass)

-----
affects: SSF!, GE?

PR #2665, opened 2024-11-15
fix: prevent re-sync after reloading the page if user related permission is set

tomwwinter:
closes: #2542


src/app/core/permissions/ability/ability.service.ts
```diff
@@ -86,33 +86,38 @@ export class AbilityService extends LatestEntityLoader<Config<DatabaseRules>> {
   private interpolateUser(
     rules: DatabaseRule[],
     sessionInfo: SessionInfo,
-  ): DatabaseRule[] {
-    const user = this.currentUser.value;
+  ): Promise<DatabaseRule[]> {
+    return firstValueFrom(
```
sleidig, line 90: let's use async/await to make this complex method slightly less nested?

sleidig (changes requested):
This works to resolve the resync bug (although I don't completely understand yet what exactly happened causing it ...)

However, users without a profile entity linked (which we do officially support) do not get any access now, even with a role that provides "manage all" without project-based rules.

We should have unit tests for this central functionality of project-based permissions. Not sure how to write those ... if in doubt, let's do this together in a pair programming session.

sleidig (approved):

Issue #2655, opened 2024-11-12
Edit location dialog save button not reachable on mobile


sleidig:
![Screenshot_20241112-083320.png](https://github.com/user-attachments/assets/0472840a-bf7e-4348-8880-ffd79b7eeb46)

If opening the app in browser on mobile (without installing PWA app, which would hide the address bar and give more space), the save button may not be visible for the location edit dialog (address/location form field type) and the user cannot confirm the selection)

PR #2666, opened 2024-11-15
fix: the bug edit location dialog save button not reachable on mobile

sadaf895:
closes: #2655

### DIFF
--- a/src/app/features/location/location-input/location-input.component.ts
+++ b/src/app/features/location/location-input/location-input.component.ts
@@ -75,7 +75,7 @@ export class LocationInputComponent extends CustomFormControlDirective<GeoLocati
 
     const ref = this.dialog.open(MapPopupComponent, {
       width: "90%",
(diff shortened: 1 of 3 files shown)


sleidig (changes requested):
The save button is always accessible now :+1: 

However, the map part is so big that even on large screens we have scroll bars now. This should be limited so that it only fills the dialog window but doesn't need scrolling (at least for large windows)

image

src/app/features/location/location-input/location-input.component.ts
```diff
@@ -75,6 +75,7 @@ export class LocationInputComponent extends CustomFormControlDirective<GeoLocati
 
     const ref = this.dialog.open(MapPopupComponent, {
       width: "90%",
+      height: "100vh",
```
sleidig, line 78:
```suggestion
      height: "95vh",
```
let's give this a small margin at least, for better design

PR #2668, opened 2024-11-20
feat(core): allow hiding of fields through config

sleidig:
while still including the angular form field for default values etc.

see #2635
required there to support hidden default value fields


tomwwinter (approved):
LGTM

Issue #2642, opened 2024-10-29
Manage translations outside of repository using a specialized software


sleidig:
Currently, UI translation keys are extracted using an npm script, manually translated and then commited in our git repository. This requires the translator to have git access and be comfortable with a large, technical xml (.xliff) file. Instead, we should use some cloud-based system to manage translations to outsource these tasks and decouple them from software development.

https://poeditor.com/ is free for open source projects and could be a good solution

EDIT: our project has been approved for the free FOSS licence
**_https://poeditor.com/projects/view?id=740876_**

-----
An ideal process would be:

Automatic updates of tokens for translation:
- whenever there is an update to "master" branch, the translation platform automatically receives updated translation tokens
- (status quo: we need to manually run `npm run extract-i18n` and then commit the updated files to git

Automatic updates of translations for app users:
- ~whenever someone adds/updates a translation for a token in some language on the translation platform, the translations for deployed systems automatically update for users~
  - ~this would make corrections to translations faster and easier~
  - ~how can different app versions be handled with this? (some systems might run an older version, where certain translation tokens are different - so the system would need to have version-related translations?)~
  - ~if translations are loaded by devices directly, we need to ensure they are cached by service worker so that offline-functionality is fully supported~
  - ~this approach will require code changes to dynamically load translation files instead of delivering them as static angular assets (which is the current approach)~
- **OR**: we can easily open a PR in GitHub to commit new translations that were done on the translation platform
  - this would allow us to keep the existing architecture where translation files are part of the code base and packaged with the app directly

--> read about approaches recommended by POEditor; they support CI integration with GitHub etc.

sleidig:
For now, it seems best to _not_ change the architecture how translation files are delivered to users (i.e. stick with the translations being part of our code repository; option 2 above) and get the integration with the translation platform set up. We can always come back to the question of delivering translation files to users dynamically (independent of a new app release) at a later time.

tomwwinter:
> I’ve documented the current behaviour and the proposed solution in this [Document](https://docs.google.com/document/d/1Th2UYISml-Q8flXO4yrwX49kgkN8LvkdH0mMWm-N66E/edit?tab=t.0) for better clarity and understanding.

Can you please mask all api_tokens in this document?

Ayush8923:
[...]

PR #2672, opened 2024-11-22
[Feat]: Create the Github Action for Sync Translations with POEditor

Ayush8923:
closes: https://github.com/Aam-Digital/ndb-core/issues/2642

### Description of changes
- This PR introduce a GitHub Actions workflow that integrates with POEditor to automate the synchronization of translations. The workflow performs the following actions:
1. Extraction of Translation Keys
2. Upload Translations to POEditor
3. Export Translations from POEditor
4. Commit and Push Changes
5. Automated Pull Request Creation


Ayush8923:
This is the Automated **[PR](https://github.com/Aam-Digital/ndb-core/pull/2680)** generated by the `sync-translations` workflow.

.github/workflows/sync-translations-poeditor.yml
```diff
@@ -0,0 +1,185 @@
+name: Sync Translations with POEditor
+
+on:
+  push:
+    branches:
+      - feat/translation-poeditor-integration
```
Ayush8923, line 6: We need to update the branch name before merging this PR to the master.

.github/workflows/sync-translations-poeditor.yml
```diff
@@ -0,0 +1,185 @@
...
+    paths:
+      - "src/assets/locale/**"
```
Ayush8923, line 8: This workflow will run only if changes are made to files within this specific directory.

.github/workflows/sync-translations-poeditor.yml
```diff
@@ -0,0 +1,185 @@
...
+              -F "language=$lang" \
+              -F "updating=terms_translations" \
+              -F "overwrite=1" \
+              -F "sync_terms=1" \
+              -F "file=@${files[$lang]}")
+            status=$(echo $response | jq -r '.response.status')
+            if [ "$status" != "success" ]; then
+              echo "Error uploading file for $lang: $response"
+              exit 1
+            fi
```
Ayush8923, line 77: This is the POEditor upload API, where need to send the language `xlf` file. For reference, here is the documentation of the [POEditor Upload API](https://poeditor.com/docs/api#projects_upload).

.github/workflows/sync-translations-poeditor.yml
```diff
@@ -0,0 +1,185 @@
...
+            echo "Waiting for 30 second to respect rate limits..."
+            sleep 30
```
Ayush8923, line 79: Added the sleep interval, ensuring compliance with the rate limit of one request every 20 seconds.

.github/workflows/sync-translations-poeditor.yml
```diff
@@ -0,0 +1,185 @@
...
+                echo "Downloading translations for $lang from $url"
+                if [ "$lang" == "en" ]; then
+                  curl -s -o src/assets/locale/messages.xlf $url
+                else
+                  curl -s -o src/assets/locale/messages.$lang.xlf $url
+                fi
+              else
+                echo "No URL found for downloading translations for $lang"
+                exit 1
+              fi
```
Ayush8923, line 105: This is the POEditor export API, where we can export the updated language files. For reference, here is the documentation of the [POEditor Export API](https://poeditor.com/docs/api#projects_export).

Ayush8923:
This is the Automated **[PR](https://github.com/Aam-Digital/ndb-core/pull/2680)** generated by the `sync-translations` workflow.

Ayush8923:
@sleidig As we discussed over the call, it is currently not creating automated PRs. Last time, it created this [Automated PR](https://github.com/Aam-Digital/ndb-core/pull/2680). But now, it is creating a branch and pushing it to the origin, but it is not creating a Pull Request. I need to check with @tomwwinter once to see if there have been any changes to the workflow permissions.

.github/workflows/sync-translations-poeditor.yml
sleidig: This should be automated via CI as well:

when I merge a new change on master ...
... `npm run extract-i18n` is automatically executed
... and the resulting latest .xlf files are then sent to POEditor (as is happening here already, I guess)

Ayush8923: @sleidig If we want `npm run extract-i18n` to execute automatically, we need to set up a separate GitHub Action workflow. This workflow will run whenever there is an update to the master branch. It will:

1. Automatically execute the `npm run extract-i18n` command to update the language files.
2. Commit and push the updated language files to the master branch.
3. Trigger the `Sync Translations with POEditor` workflow.

Is my understanding correct based on your suggestions?

sleidig: Yes, potentially it could run like this. However, this would mean that after almost every commit to master, there will be another (automatic) commit to update the results from "npm run exxtract-i18n" (this is what I pointed out last week during our call). So that would be far from ideal.

Maybe instead the CI could work like this?
- run "npm run extract-i18n within CI without committing it
- check the result of this and trigger the push action to POEditor (if there were changes)
- open a PR with the new xlf files (rather than immediately committing them to master)
- update that branch/PR whenever POEditor users update a translation?

Ayush8923: Got it. So maybe insteaf of creating the new flow, we can update this workflow. And need to made these changes which I mentioned below.

Ayush8923: @sleidig Understood. Instead of creating a new flow, I think we can update the existing workflow. Here are the small changes I propose:

1. Remove the condition that restricts the workflow to run only when specific files change.
2. Add a step after the `npm run extract-i18n` command to check if there are any updates on the branch; the workflow should only run if there are updates, otherwise, it should be skipped.
3. Continue sending the updated language files to POEditor and follow the remaining steps as they are already implemented in this workflow.

.github/workflows/sync-translations-poeditor.yml
```diff
@@ -0,0 +1,200 @@
...
+
+      - name: Check if any files changed
+        run: |
+          if git diff --quiet; then
+            echo "No changes detected."
+            echo "changed=false" >> $GITHUB_ENV
+          else
+            echo "Changes detected."
+            echo "changed=true" >> $GITHUB_ENV
+          fi
```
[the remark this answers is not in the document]
Ayush8923: @tomwwinter Yes, this can be managed at the GitHub Action level, and I have implemented it that way before. However, following Sebastian's suggestion, we can handle this through CI. For reference, please check this [comments](https://github.com/Aam-Digital/ndb-core/pull/2672#discussion_r1856692659). Please let me know, if I need to changes this.

.github/workflows/sync-translations-poeditor.yml
```diff
@@ -0,0 +1,200 @@
...
+
+      - name: Configure Git
+        run: |
+          git config user.name ${{ secrets.GIT_BOT_USERNAME }}
+          git config user.email ${{ secrets.GIT_BOT_EMAIL }}
+          git remote set-url origin https://${{ secrets.GIT_BOT_ACCESS_TOKEN }}@github.com/Aam-Digital/ndb-core.git
+
+      - name: Create or update branch for translations
+        run: |
+          branch_name="chore/update-poeditor-translations"
```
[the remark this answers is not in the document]
Ayush8923: Yes, this action will run after every commit on the master branch. I believe we don’t need to create another PR if the translation automated PR has already been created. Instead, we can update the language files in the same PR. What do you suggest in this?

.github/workflows/sync-translations-poeditor.yml
```diff
@@ -0,0 +1,200 @@
...
+            fi
+            echo "Waiting for 30 seconds to respect rate limits..."
+            sleep 30
+          done
+
+      - name: Commit updated translations
+        if: env.changed == 'true'
+        run: |
+          git add src/assets/locale/*.xlf
+          git commit -m "Update translations from POEditor"
```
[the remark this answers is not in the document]
Ayush8923: I have updated the changes according to the suggestions.

.github/workflows/sync-translations-poeditor.yml
```diff
@@ -0,0 +1,200 @@
...
+              body: [
+                'This PR updates translations from POEditor.',
+                'Generated by GitHub Actions.'
+              ].join('\n'),
+            });
+            await github.rest.issues.addLabels({
+              owner,
+              repo,
+              issue_number: result.data.number,
+              labels: ['feature', 'automated pr'],
```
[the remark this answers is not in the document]
Ayush8923: I have removed the `feature` label.

.github/workflows/sync-translations-poeditor.yml
```diff
@@ -0,0 +1,200 @@
...
+          git push origin HEAD
+
+      - name: Create Pull Request
+        if: env.changed == 'true'
+        uses: actions/github-script@v6
+        with:
+          script: |
+            const { repo, owner } = context.repo;
+            const headBranch = 'chore/update-poeditor-translations';
+            const baseBranch = 'master';
```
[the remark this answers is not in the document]
Ayush8923: I have moved this in the env variable and then used the env variable key here.

.github/workflows/sync-translations-poeditor.yml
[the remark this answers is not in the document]
tomwwinter: Thanks, missed that conversation.

sleidig (approved):
I did quite a lot of experimentation and tuning, observing what effects things have a POEditor's interface. This seems to be the most standard approach now and works okay, although if we change translated texts in the code, POEditor recognizes them only as new and will discard previous translations.

Issue #2271, opened 2024-03-01
Setup Wizard: Edit public forms


sleidig:
Public forms (e.g. https://demo.aam-digital.com/public-form/test) need a drag&drop admin interface similar to the way users can already edit the entity details view (with the `AdminEntityDetailsComponent`).

From a user's perspective, the following requirements / user stories are involved:
As an admin user, I can ...
- [x] list all available "Public Forms" in an admin view
- [x] create/read/update/delete PublicFormConfigs
- [x] use an intuitive, visual UI to preview & edit the config for a selected Public Form
  - [x] editing fields using the same drag&drop editor available for designing an Entity Details View (in AdminEntityDetailsComponent)
  - [x] editing other details like title, url, description and entity type in additional config form fields (similar to the AdminEntityGeneralSettingsComponent)
- [ ] #2708

-----
From a technical perspective, the new Admin UI feature needs to manage valid [`PublicFormConfig`](https://github.com/Aam-Digital/ndb-core/blob/ac595571d9454edc72d23ae22db8769b3a727e69/src/app/features/public-form/public-form-config.ts#L6) entities. The `PublicFormComponent` displays these and should ideally not require any changes for this Admin UI implementation. However, if you identify ways to make the config format more simple or suitable, adaptions can be discussed.

Suggested implementation steps (to be adjusted & extended):
- [x] improve `PublicFormConfig` format to use FieldGroup rather than string[]: #2401 
- [x] ~Create a new `AdminPublicFormComponent` component that displays a form to edit a `PublicFormConfig` object~ (may not be needed if we can rely on the EntityDetails system in a smart way)
- [x] make the new admin component available from the UI as another section in AdminEntity, which can list all PublicForms for that entity type, maybe similar to AdminEntityTypesComponent

sleidig:
This has some overlap with
- https://github.com/Aam-Digital/ndb-core/issues/2401

Suggested there to adjust the config format for PublicFormConfig to be more similar and consistent with Entity Details. Maybe we should first implement those changes, so that the Admin UI can be designed for the future-proof config format then, @Abhinegi2

PR #2682, opened 2024-11-25
Feature/setup wizard public forms

Abhinegi2:
closes: #2271

### Visible/Frontend Changes

- [x] Added All public forms listed in Admin view.
- [x] Crud operation can be performed for public form.


src/app/features/public-form/public-form-config.ts
```diff
@@ -1,15 +1,33 @@
...
 
 @DatabaseEntity("PublicFormConfig")
 export class PublicFormConfig extends Entity {
-  @DatabaseField() title: string;
-  @DatabaseField() description: string;
-  @DatabaseField() entity: string;
-  @DatabaseField() columns: FieldGroup[];
+  static override label = $localize`:PublicFormConfig:Public Form`;
+  static override labelPlural = $localize`:PublicFormConfig:Public Forms`;
+  static override route = "admin/public-forms";
```
sleidig, line 11:
```suggestion
  static override route = "admin/public-form";
```

our standard is to use singular in routes

sleidig (commented):
base functionality and code looks good so far.

just some further improvements and then the integration of the drag&drop "columns" editing :smiley:

src/app/features/public-form/public-form-config.ts
```diff
@@ -1,15 +1,33 @@
...
+  title: string;
+  @DatabaseField({
+    label: $localize`:PublicFormConfig:Description`,
+    dataType: LongTextDatatype.dataType,
+  })
+  description: string;
+  @DatabaseField({
+    label: $localize`:PublicFormConfig:Entity`,
+  })
+  entity: string;
```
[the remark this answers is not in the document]
Abhinegi2: Sure, @sleidig ! I'll create a separate branch for this and implemet it on [AdminEntityFieldComponent](https://github.com/Aam-Digital/ndb-core/blob/8c9c39ac746f152962a850d0b045440a455072da/src/app/core/admin/admin-entity-details/admin-entity-field/admin-entity-field.component.html#L123-L141) and [ImportEntityTypeComponent](https://github.com/Aam-Digital/ndb-core/blob/8c9c39ac746f152962a850d0b045440a455072da/src/app/core/import/import-entity-type/import-entity-type.component.html#L4-L18)

src/app/features/public-form/public-form.module.ts
```diff
@@ -0,0 +1,68 @@
...
+      entityType: PublicFormConfig.ENTITY_TYPE,
+      panels: [
+        {
+          components: [
+            {
+              component: "Form",
+              config: {
+                fieldGroups: [
+                  {
+                    fields: ["title", "description", "entity", "columns"],
```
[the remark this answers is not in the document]
Abhinegi2: added the routeing logic! please have a look now.

src/app/features/public-form/public-form-config.ts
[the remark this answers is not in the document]
sleidig: Nice! Thanks for finding that component - as you see, I had already forgotten about that :see_no_evil:

src/app/features/public-form/public-form-config.ts
```diff
@@ -1,15 +1,41 @@
...
+  @DatabaseField({
+    label: $localize`:PublicFormConfig:Entity`,
+    editComponent: "EditEntityTypeDropdown",
+  })
+  entity: string;
+  @DatabaseField({
+    label: $localize`:PublicFormConfig:Columns`,
+    isArray: true,
+  })
+  columns: FieldGroup[];
```
sleidig, line 38:
```suggestion
  @DatabaseField({
    label: $localize`:PublicFormConfig:Title`,
  })
  title: string;

  @DatabaseField({
    label: $localize`:PublicFormConfig:Form Link ID`,
    description: $localize`:PublicFormConfig:The identifier that is part of the link (URL) through which users can access this form (e.g. demo.aam-digital.com/public-form/MY_FORM_LINK_ID)`,
  })
  route: string;

  @DatabaseField({
    label: $localize`:PublicFormConfig:Description`,
    dataType: LongTextDatatype.dataType,
  })
  description: string;

  @DatabaseField({
    label: $localize`:PublicFormConfig:Entity`,
    description: $localize`:PublicFormConfig:The type of record that is created when a someone submits the form (e.g. if you select "Note" here, the form will create new entries in your "Notes List" and you can select only fields of your "Note" data structure for this form)`,
    editComponent: "EditEntityTypeDropdown",
  })
  entity: string;

  @DatabaseField({
    label: $localize`:PublicFormConfig:Columns`,
    isArray: true,
  })
  columns: FieldGroup[];

```

src/app/features/public-form/public-form.component.ts
```diff
@@ -83,7 +83,19 @@ export class PublicFormComponent<E extends Entity> implements OnInit {
...
+    const publicForm = publicForms.find(
+      (form: PublicFormConfig) => form.route === id,
+    );
+    const publicFormId = publicForm ? publicForm["_id"] : id;
+
+    this.formConfig = await this.entityMapper.load(
+      PublicFormConfig,
+      publicFormId,
+    );
+
```
sleidig, line 98: if you already found the form matching the route, you don't have to load it yet again from the DB. Just use the `publicForm` entity directly?

And event to match with the "_id", you could use an in-memory `publicForms.find(x => x.getId(true) === id)`, so we save one db request

src/app/features/public-form/public-form.module.ts
```diff
@@ -0,0 +1,84 @@
...
+        {
+          title: "General Setting",
+          components: [
+            {
+              component: "Form",
+              config: {
+                fieldGroups: [
+                  {
+                    fields: ["route", "title", "description", "entity"],
+                  },
```
sleidig, line 60:
```suggestion
                  {
                    fields: ["route", "title"],
                  },
                  {
                    fields: ["description", "entity"],
                  },
```

two columns (if width allows) may be a bit nicer to look at

sleidig (commented):
Looks and works well :smiley: 

Only the drag&drop integration remaining, I think.

Abhinegi2:
@sleidig I did some cleanup and added the test cases please have a look now.

src/app/features/public-form/public-form.module.ts
```diff
@@ -0,0 +1,108 @@
...
+  }
+}
+
+const dynamicComponents: [string, AsyncComponent][] = [
+  [
+    "EditPublicFormField",
+    () =>
+      import(
+        "app/features/public-form/edit-public-form-field/edit-public-form-field.component"
+      ).then((c) => c.EditPublicFormFieldComponent),
```
sleidig, line 41: I would rename this to EditPublicFormFields or maybe EditPublicFormColumns. We use the term "form field" for each individual form control already (also within the entity details etc.)

src/app/features/public-form/public-form.module.ts
```diff
@@ -0,0 +1,108 @@
...
+            {
+              component: "Form",
+              config: {
+                fieldGroups: [
+                  {
+                    fields: [
+                      {
+                        id: "columns",
+                        editComponent: "EditPublicFormField",
+                        dynamicComponentInput: {
```
sleidig, line 94: We use `additional` as a standard within the FormFieldConfig / EntitySchemaField type for any additional config that is specific for some component rather than relevant for every implementation.

Let's rename `dynamicComponentInput` to `additional` in that regard

src/app/features/public-form/public-form-config.ts
```diff
@@ -1,15 +1,46 @@
...
-  @DatabaseField() columns: FieldGroup[];
+  static override label = $localize`:PublicFormConfig:Public Form`;
+  static override labelPlural = $localize`:PublicFormConfig:Public Forms`;
+  static override route = "admin/public-form";
 
+  @DatabaseField({
+    label: $localize`:PublicFormConfig:Title`,
+  })
+  title: string;
+  @DatabaseField({
```
sleidig, line 17: can you keep an empty line between these, please? I find it much easier to read which annotation refers to which line with some more spacing
```suggestion
  title: string;

  @DatabaseField({
```

src/app/features/public-form/public-form.module.ts
```diff
@@ -0,0 +1,108 @@
...
+          components: [
+            {
+              component: "Form",
+              config: {
+                fieldGroups: [
+                  {
+                    fields: ["route", "title"],
+                  },
+                  {
+                    fields: ["description", "entity"],
```
sleidig, line 75: The issue during testing made me realize that admins may face problems creating new forms still because the `Config:Permissions` need to allow public role to create records of that entity ...

Ideally we would ask the user upon saving the PublicFormConfig if the Permissions shall be updated to include that rule and then edit the Config:Permissions doc automatically.
However, we don't really have a good entrypoint for this functionality when we only use the standard EntityDetails and fields :thinking: 

For v1 let's at least add a explainer text. Something like this:
```
{ id: "permissions_remark", editComponent: "EditDescriptionOnly", label: $localize:`:PublicFormConfig admin form:If you want external people filling this form without logging in, the Permission System also has to allow "public" users to create new records of this type. If you are seeing problems submitting the form, please contact your technical support team.` }
```

src/app/features/public-form/public-form.module.ts
sleidig: also, we can use these details (editComponent, additional) to the schema annotation in the entity class directly.

src/app/features/public-form/public-form-config.ts
```diff
@@ -1,15 +1,46 @@
 import { Entity } from "../../core/entity/model/entity";
 import { DatabaseEntity } from "../../core/entity/database-entity.decorator";
 import { DatabaseField } from "../../core/entity/database-field.decorator";
+import { LongTextDatatype } from "app/core/basic-datatypes/string/long-text.datatype";
 import { FieldGroup } from "app/core/entity-details/form/field-group";
 
 @DatabaseEntity("PublicFormConfig")
```
sleidig, line 7:
```suggestion
/**
 * Each entity of this type defines a new publicly accessible form
 * that can be reached through the given route even by users without being logged in.
 */
@DatabaseEntity("PublicFormConfig")
```

src/app/features/public-form/edit-public-form-field/edit-public-form-field.component.ts
```diff
@@ -0,0 +1,43 @@
...
+export class EditPublicFormFieldComponent<T extends Entity = Entity>
+  extends EditComponent<T>
+  implements OnInit
+{
+  entityConstructor: EntityConstructor;
+  publicConfigDetailsView: any;
+
+  constructor(private entities: EntityRegistry) {
+    super();
+  }
```
sleidig, line 24: The new angular "inject()" function helps to do DI without contructor. That way we can skip that annoying call to super() constructor.

```suggestion
  private entities = inject(EntityRegistry);
```
(will need an additional import, check Quick Fix suggestions)

sleidig (changes requested):
Code looks fairly good already :+1:   Some small remarks for further cleanup.

I still see some important UX issues for the editing of fields:
1. the Edit Button is super confusing if I can still drag & drop things outside edit mode
2. navigating away without saving column changes (both with and without edit mode) doesn't give any warning and discards changes
3. Editing a field's details (like the label) changes that for the whole entity, also reflecting in the Entity Details view. It can be helpful to overwrite these things only in the individual public form (and then save the whole FormFieldConfig in the PublicFormConfig instead of just the entity field id).

(3.) is maybe an advanced feature for v2 improvements

(1.) + (2.) should be solvable by making the component "more properly" an EditComponent, i.e. only editing the value of the formControl and _not_ affecting the entity object directly as a side effect.
We'll have to change the implementation of AdminEntityForm to work with an Output. This may be complex for you, let's do a pair programming session for this.

src/app/features/public-form/public-form.module.ts
sleidig: can you also rename the file and folder names accordingly?

src/app/features/public-form/public-form-config.ts
```diff
@@ -1,14 +1,58 @@
...
 import { FieldGroup } from "app/core/entity-details/form/field-group";
 
+/**
+ * Each entity of this type defines a new publicly accessible form
+ * that can be reached through the given route even by users without being logged in.
+ */
+/**
+ * Each entity of this type defines a new publicly accessible form
+ * that can be reached through the given route even by users without being logged in.
+ */
```
sleidig, line 14:
```suggestion
```

src/app/features/public-form/public-form-config.ts
```diff
@@ -1,14 +1,54 @@
...
+    editComponent: "EditEntityTypeDropdown",
+    validators: {
+      required: true,
+    },
+  })
+  entity: string;
+
+  @DatabaseField({
+    label: $localize`:PublicFormConfig:Columns`,
+    isArray: true,
```
sleidig, line 49:
```suggestion
    isArray: true,
    editComponent: "EditPublicFormColumns",
    additional: {
      entity: PublicFormConfig.ENTITY_TYPE,
    },
```

adding it here, you can remove it from the module and simplify the config there

src/app/features/public-form/public-form.module.ts
```diff
@@ -0,0 +1,116 @@
...
+              config: {
+                fieldGroups: [
+                  {
+                    fields: [
+                      {
+                        id: "columns",
+                        editComponent: "EditPublicFormColumns",
+                        additional: {
+                          entity: PublicFormConfig.ENTITY_TYPE,
+                        },
```
sleidig, line 104: do you even us this "additional" anymore in the component?

Abhinegi2: We are extending EditComponent now, so we no longer need this.

Abhinegi2:
@sleidig, To disable the Drag and Drop UI in Publicforms, I found two methods:

1. Using the cdkDropListDisabled Method: It can be applied to each section where cdkDropList is used. It disables drag-and-drop functionality based on the boolean. However, it doesn't disable other interactive elements like buttons (e.g., "Edit Fields"), so additional steps would be needed to handle those cases.
2. Using Angular Class Binding: This approach involves applying a class binding to disable the entire form. It seems more suitable for our use case as it disables all actions, not just drag-and-drop functionality.

So,  currently, I am proceeding with the second approach. Let me know if you have any questions or suggestions.

src/app/features/public-form/public-form-config.ts
Abhinegi2: I tried the suggested approach, but it didn’t work as expected. After making these changes in Public Forms, the edit view displays a regular text field for [config.columns] instead of the drag-and-drop UI. I also added the `EditPublicFormColumns` import in `coreComponents` like other dynamic components, but it’s still not working.

george-neha:
Some comments:
1. The edited label name (while configuring the fields) is not displayed in the public form.
image
image
2. Currently, there is no warning message when navigating from the "Configure fields" tab to the "General Settings" tab. Displaying a warning message would be helpful for the user.

Abhinegi2:
@sleidig updated disable logic, now grabbing the edit field will not work when it is in read-only mode.
please have a look now and then deploy these changes in the Dev site for testing.

src/app/features/public-form/edit-public-form-columns/edit-public-form-columns.component.html
```diff
@@ -0,0 +1,18 @@
+<div class="hint-banner" i18n>
+  You can edit how users will see the details of this public form. To edit
+  fields, click on the Edit button, make your changes, and then save. These
+  changes will reflect on the public form.
+  <br />
+  Drag and drop fields and sections in this preview of a profile view. The
+  editor below closely resembles how the form will look for users later.
```
sleidig, line 7:
```suggestion
  editor below closely resembles how the form will look for users later.
  Forms show all fields below each other not in multiple columns, however.
```

src/app/features/public-form/edit-public-form-columns/edit-public-form-columns.component.html
```diff
@@ -0,0 +1,18 @@
...
+  <br />
+</div>
+<div [class.disabled]="formControl.disabled">
```
sleidig, line 10: can we move this class.disabled into the AdminEntityForm component and make it part of the logic for the isDisabled input? Then developers only need to toggle this with one single variable

src/app/features/public-form/edit-public-form-columns/edit-public-form-columns.component.html
```diff
@@ -0,0 +1,18 @@
+<div class="hint-banner" i18n>
```
sleidig, line 1: This still needs a margin at the bottom:
image

src/app/features/public-form/public-form-config.ts
```diff
@@ -1,14 +1,54 @@
...
+ */
 @DatabaseEntity("PublicFormConfig")
 export class PublicFormConfig extends Entity {
-  @DatabaseField() title: string;
-  @DatabaseField() description: string;
-  @DatabaseField() entity: string;
-  @DatabaseField() columns: FieldGroup[];
+  static override label = $localize`:PublicFormConfig:Public Form`;
+  static override labelPlural = $localize`:PublicFormConfig:Public Forms`;
+  static override route = "admin/public-form";
```
sleidig, line 15:
```suggestion
  static override route = "admin/public-form";
  static override toStringAttributes = ["title"];
```

adding human-readable "name" to entity

sleidig (changes requested):
Seems to work very well now. Great job, @Abhinegi2 

Just some very last, final polishing needed.

sleidig (approved):

Issue #2623, opened 2024-10-21
SkillLab API integration


sleidig:
### Scenarios for matching:
1. A record for a person already exists in Aam Digital and needs to be “enriched” with data from skillmap.
2. No record for the person exists in Aam Digital, a completely new record is created using the data from skillmap.
3. (?) Person exists (is created) in Aam Digital - and is created in SkillMap?

We assume that, in most cases, scenario 1 will apply.

### Workflows for importing data:
two main functionalities / processes to pull the data:
- 1.1 Import data for individual records: For every record created, there could be a button that called “import profile from skillmap”
- 1.2 Import data in bulk: On the display of the people’s list, there is a button “import profiles from skillmap”. Aam then parses the data set from skillLab, finds matches, and completes missing data.
- 1.3 Import data for selected items: the user either selects groups of items or filters for a group of items and then clicks on the button “import profiels from skillLab”.

--> We will first implement (1.1) and then extend that core method to the other bulk flows

### Default Use Case (for user perspective):
1. User opens Individual's profile in Aam
2. User clicks "Load Skills" button
3. Aam connects to SkillLab:
   1. Aam Server connects & auth
   2. Aam Server requests data
   3. Aam Server finds match (or asks user?)
   4. (?) Aam Server create / update entity in Aam DB; or server sends data to user who then saves the entity
5. User sees skill records (or some message if nothing is found)

-----
## Implementation Steps:

**Critical paths:**
- matching of SkillMap record to Aam record
  - --> experiment manually with API
  - if many unmatchable records we may need some UI to manually match/search?
- UI / data structure to present skills in Aam
- Data structure to store skill-individual-relation (embedded in Individual? as separate relationship entities?)

**Questions**
- Do GIZ team have access to SkillMap?
- How are users using skills in Aam?
- How/when/whether to update skills of a participant?

-----
see 
- [GoogleDrive](https://drive.google.com/drive/u/0/folders/1Aaq1m4XJv0X09bpqNPBPH-OnAAtRK3MU)
- [Implementation Plan (GoogleSheet)](https://docs.google.com/spreadsheets/d/1qPOp0HEQz9P2vnFfAEv52H2Lz9UWIIB1q8DPR_Hi6_s/edit?usp=sharing)
- [ESCO Skill docs](https://esco.ec.europa.eu/de/classification/skill_main)

-----
## Remaining Sub-Tasks:

- Backend:
   - [x] SkillLab connection, data import, endpoints 
[...]

PR #2687, opened 2024-11-26
SkillLab integration

sleidig:
see #2623 

To test the prototype / mock, visit the Details View of a Child record use the "Link external profile" button in the form.

## TODOs
- [x] load ESCO skill details (name, description) via API
- [x] loading indicator for ESCO Skill requests
- [x] Skills field doesn't update immediately yet after clicking "Update data"
- [x] "Unlink" is not saved (resets to previous external ID after reload)
- [x] linking an user-profile is possible in "non-edit" mode (?)
- [x] search parameters (e.g. name) are only used if the entity was saved (should take the current form value immediately)
- [x] make details for matching configurable (Entity Type, fields for search (fullName, email, ...), etc.)
- [x] implement a bulk import/link workflow
- [x] import skill values in bulk action also
- [x] tests 
   - [x] escoApi
   - [x] skillApi
   - [x] components: if api throws error
   - [x] bulkLink component
- [x] ui for 403 response if role is missing
- [x] pagination UI
- [x] correctly pre-select during bulk link when exactly one match + UI to distinguish existing previous match vs. newly pre-selected match
- [x] make skill target field configurable (instead of hard-coding to import into `.skills`)
- [x] create a dataType to wrap the edit component?
- [ ] wrap into mat-form-field for more consistent UI within form views?
- [x] clear in-code //TODO

### Refactorings:
- [ ] implement a re-usable `ErrorDetailsComponent` (for LinkExternalProfileDialogComponent as well as ReportingComponent and other places where we need to both display a simple message and give access to technical error details for experts)
- [ ] make the Bulk Actions configurable through a service instead of hard-coding actions into EntityListComponent (--> similar or combined (?) system to the `EntityActionsMenuService`)
- [ ] better way to close/return when using the generic `app-view-actions` (without using mat-dialog-close explicitly)


src/app/core/common-components/entity-select/entity-select.component.ts
```diff
@@ -135,6 +135,14 @@ export class EntitySelectComponent<
...
   @Input() additionalFilter: (e: E) => boolean = (_) => true;
 
+  ngOnChanges(changes: SimpleChanges): void {
+    if (changes["form"]) {
+      this.form.valueChanges.subscribe((value) => {
+        this.updateAvailableOptions().then((_) => {});
+      });
+    }
+  }
+
```
sleidig, line 145: @tomwwinter @Abhinegi2  is this a change one of you did somewhere for a bugfix? I can't remember why this should be here anymore :see_no_evil:

Abhinegi2: As far as I remember 😅, I did not make these changes.

src/app/core/config/config-fix.ts
```diff
@@ -586,6 +587,14 @@ export const defaultJsonConfig = {
...
                     header: $localize`:Header for form section:Personal Information`
                   },
+                  {
+                    fields: [
+                      "externalProfileMockResults",
+                      "externalProfile",
+                      "skills"
+                    ],
+                    header: "Skill API Integration (under development)"
+                  },
```
sleidig, line 597: remove from demo config before merging into master

tomwwinter (approved):

PR #2700, opened 2024-12-03
fix(public forms): correct permission checks to allow unauthenticated…

sleidig:
… users access to forms again


tomwwinter (approved):
LGTM

Issue #2661, opened 2024-11-14
View list of notifications in the app


sleidig:
_As a user I want to see details of all my notifications in the user interface of the app in order to not rely on the system push messages that can be discarded quickly._

Within the app itself users should be able to view a list of notifications with some details about each notification (e.g. title, description, at a later stage also an action button).

- should be accessible from the #2660 
- This task includes:
  - [x] To show the list of notifications, we can show the same notification bell icon, when the user click on that icon the list of the notification list will be shown. For reference, Added the screenshot of the notification list (this list shown after clicking the notification bell icons).
![image](https://github.com/user-attachments/assets/cc7ef569-0391-488f-9171-de78b7dc590e)
- [x] Whenever the user clicks on the bell icon then the list of all the notifications will be displayed corresponding to the user, so user can read the notification from the list.

- Next tasks building on this:
  - #2690

Ayush8923:
@sleidig So far, I have implemented the notification UI by referring to the screenshot mentioned in the issue. For reference, I have added the screenshot below. I still need to do code cleanups, which I am gradually doing during development.

<img width="808" alt="image" src="https://github.com/user-attachments/assets/2f3ba15d-eaf2-419d-a440-2557eaef46d6">

1. When the user clicks the settings button, a toggle option will appear, allowing them to turn their notifications ON or OFF. For reference, I have added the screenshot below.

<img width="372" alt="image" src="https://github.com/user-attachments/assets/38ded848-7aa3-4dc5-a663-48cd89b44296">

If the notification list is very long, how many notifications should we display initially before allowing the user to scroll through the rest? Please feel free share your thoughts/suggestions on this.

Ayush8923:
@sleidig I am conducting research to improve the notification list UI. We can add the two tabs inside the notification list: 
1. All: The **All** tab displays all notifications.
2. Unread: The **Unread** tab displays only unread notifications.

Additionally, when the user clicks on the three dots (menu icon), multiple options should appear, such as:
- Notification Settings
- Enable Notifications

From this menu, users should also be able to navigate directly to the **Notification Settings** page inside the profile. This enhancement aims to improve user experience and provide quick access to relevant notification management options. Please let me know your suggestions and thoughts on this.

<img width="353" alt="image" src="https://github.com/user-attachments/assets/2b538cea-2487-4e03-ac5c-5bc6e08f711e">

sleidig:
Thanks. This looks like good additions. 

[...]

Issue #2690, opened 2024-11-28
Store and access notifications & mark them "read"


sleidig:
For the "notification centre" UI, it is necessary to store notification events in the same way as the user-facing functionalities:
- #2660
- #2661 

DISCUSS:
- Do we store events in CouchDB as "normal" entities, reusing our existing architecture? Or do we store these completely different?
- Is an event record specific for a single user or are these "global" events, that are viewed by all users who have activated that type of notification?

### We need to implement the below-mentioned functionality in the notification list.
- [x] Display all notifications in the `All` tab.
- [x] Show only unread notifications in the `Unread` tab.
- [x] When a user clicks on the three dots and selects "Mark as Read," the notification's `readStatus` should be updated to true for that user.
- [x] When a user selects `Delete Notification`, the notification should be removed and no longer displayed in the notification list.
- [x] In the three-dot menu, selecting `Enable notifications` should activate the notification for the user.
- [x] In the three-dot menu, selecting `Notification Settings` should redirect the user to the Notifications tab within the Profile section. This is handled in this [PR](https://github.com/Aam-Digital/ndb-core/pull/2712).
- [x] When the user clicks on the notification then the user will be redirect to the corresponding notification action URL.
- [x] When a user marks a notification as read or deletes it, the unread notification count should be updated accordingly.

Ayush8923:
> Is an event record specific for a single user or are these "global" events, that are viewed by all users who have activated that type of notification?

As far as I understand, event records are user-specific. Each user will have the ability to select the types of notifications they wish to receive. A new sidebar tab will be introduced, allowing users to define their preferences for different event types. From this tab, users can opt-in to notifications for specific event types, ensuring they are only notified about the events they have selected.

In the CouchDB user document, we can to add the following fields to manage this customization:
1. **notificationsEnabled:** A boolean field indicating whether the user has enabled notifications.
2. **enabledNotificationTypes:** An array field listing the event types for which the user has chosen to receive notifications.

@sleidig @tomwwinter Please let me know what you think/suggest?

Ayush8923:
> > Is an event record specific for a single user or are these "global" events, that are viewed by all users who have activated that type of notification?
> 
> As far as I understand, event records are user-specific. Each user will have the ability to select the types of notifications they wish to receive. A new sidebar tab will be introduced, allowing users to define their preferences for different event types. From this tab, users can opt-in to notifications for specific event types, ensuring they are only notified about the events they have selected.
> 
[...]

PR #2711, opened 2024-12-11
[Feat]: Notification List UI Implementation

Ayush8923:
closes: https://github.com/Aam-Digital/ndb-core/issues/2661

### Visible/Frontend Changes
-  Implement the Notification List UI as per this [comment](https://github.com/Aam-Digital/ndb-core/issues/2661#issuecomment-2531246221).
- Create the different components for the `Notification-items`.
- Create Two tabs for all notifications and unread notifications.
- Update the notification settings icon to display three dots instead of the settings icon, and show the settings menu items upon clicking the dots.


Ayush8923:
For reference, here is the notification list UI.

1. When the All notification tab is active, then we can show all the notifications.
image

2. If there is no notification for the user.
image

3. When the user clicks on the three dots, the menu will offer options to mark the notification as read or delete it.
image

4. When the user clicks on the header settings menu, they can mark the notification as active and also be redirected to the notification settings page.
image

Ayush8923:
In this PR, the notification list is dynamic. I have removed the hardcoded notification list.

src/app/features/notifications/notifications.component.ts
```diff
@@ -94,18 +85,51 @@ export class NotificationsComponent implements OnInit {
...
+  private countUnreadNotifications(notifications: NotificationActivity[]) {
+    return notifications.filter(
+      (notification) => notification.readStatus === false,
+    ).length;
+  }
+
+  markAllRead($event: Event) {
+    $event.stopPropagation();
+    Logging.log("All notifications marked as read");
+  }
```
Ayush8923, line 111: Need to update the logic so that when the user click on the mark all as read button then all the notification mark as read.

src/app/features/notifications/notifications.component.ts
```diff
@@ -94,18 +85,51 @@ export class NotificationsComponent implements OnInit {
...
+
+  markAsRead(notification: NotificationActivity): void {
+    // Need to add/update this logic to mark as read the notification from the CouchDB
+    notification.readStatus = true;
+    this.hasUnreadNotificationCount = this.countUnreadNotifications(
+      this.allNotifications,
+    );
+    Logging.log("Notification marked as read");
+  }
```
Ayush8923, line 120: This function is for when the user mark as read in single notification.

src/app/features/notifications/notifications.component.ts
```diff
@@ -94,18 +85,51 @@ export class NotificationsComponent implements OnInit {
...
+    this.allNotifications = this.allNotifications.filter(
+      (n) => n !== notification,
+    );
+    this.unreadNotifications = this.unreadNotifications.filter(
+      (n) => n !== notification,
+    );
+    this.hasUnreadNotificationCount = this.countUnreadNotifications(
+      this.allNotifications,
+    );
+    Logging.log("Notification deleted");
```
Ayush8923, line 133: This function is used for if user need to delete any notification.

Ayush8923:
@sleidig This PR is ready for review. Can you please review this PR and let me know if there any changes needed.

src/app/features/notifications/notifications.component.ts
```diff
@@ -74,16 +66,15 @@ export class NotificationsComponent implements OnInit {
...
         NotificationActivity,
       );
     // The user is hardcoded for testing purposes, need to remove this.
-    this.notifications = this.filterUserNotifications(
-      allNotifications,
-      "User:demo",
+    this.filterUserNotifications(notifications, "User:demo");
+    this.hasUnreadNotificationCount = this.countUnreadNotifications(
+      this.allNotifications,
     );
```
Ayush8923, line 77: This is to get the dynamic notification list corresponding to the user. But the `this.currentUser.value?.getId()` getting undefined that's why I have added the hardcoded value just for testing purposes. I will remove this when I implement the notification list logic.

src/app/features/notifications/model/notifications-activity.ts
```diff
@@ -0,0 +1,18 @@
+import { Entity } from "../../../core/entity/model/entity";
+import { DatabaseField } from "../../../core/entity/database-field.decorator";
+import { DatabaseEntity } from "../../../core/entity/database-entity.decorator";
+
+@DatabaseEntity("Notification")
+export class NotificationActivity extends Entity {
```
sleidig, line 6:
```suggestion
/**
 * This represents one specific notification event for one specific user,
 * displayed in the UI through the notification indicator in the toolbar.
 */
@DatabaseEntity("NotificationEvent")
export class NotificationEvent extends Entity {
```

Let's discuss the most clear naming. I am thinking:
- NotificationEvent = single, specific notification entry
- NotificationRule (or maybe NotificationSubscription or ....) = one configuration of a certain notification type for which the user wants to see events
- ... maybe there are other entities or interfaces also in the overall picture but I can't think of any right now.

What do you think? Please suggest alternative namings also.

src/app/features/notifications/model/notifications-activity.ts
```diff
@@ -0,0 +1,18 @@
...
+  @DatabaseField({ dataType: "string" }) title: string;
+  @DatabaseField({ dataType: "string" }) body: string;
+  @DatabaseField({ dataType: "string" }) actionURL: string;
+
+  @DatabaseField({ dataType: "string" }) sentBy: string;
+  @DatabaseField({ dataType: "string" }) fcmToken: string;
+  @DatabaseField({ dataType: "boolean" }) readStatus: boolean;
```
sleidig, line 13: You can skip the dataType (just put `@DatabaseField()`) unless it is something special. The annotation infers it automatically from the Typing of the property

src/app/features/notifications/notifications.component.ts
sleidig: I think you can simple replace the Logging with a call to EntityMapperService saving the updated notification entity?
`await this.entityMapper.save(notification);`

src/app/features/notifications/notifications.component.ts
sleidig: Can you implement that now already to finalize this component? (I guess updating all unread notifications and then saving those using the entityMapper?)

sleidig (commented):
Overall approach looks good :+1:

src/app/features/notifications/notifications.component.ts
Ayush8923: For these types of functionalities, I have created this [issue](https://github.com/Aam-Digital/ndb-core/issues/2690). Can we implement this functionality under that issue?

src/app/app.component.ts
```diff
@@ -25,11 +25,7 @@ import { filter } from "rxjs/operators";
  */
 @Component({
   selector: "app-root",
-  template: `@if (configFullscreen) {
-      <router-outlet></router-outlet>
-    } @else {
-      <app-ui></app-ui>
-    }`,
```
Ayush8923, line 32: For this, I have created the `app.component.html` file and used it directly here.

Ayush8923:
@sleidig The PR deployment is failing, possibly due to insufficient Docker space. Could you please take a look into this? Once resolved, Neha will test the UI and share any feedback/suggestions.

Ayush8923:
@george-neha I have added the mocked data for UI testing purposes, please test the notification list UI and share your feedback/suggestions.

Ayush8923:
> Deployed to https://pr-2711.aam-digital.net/

@george-neha I have added the mocked data for UI testing purposes, please test the notification list UI in the above-mentioned URL and share your feedback/suggestions.

src/app/app.component.ts
sleidig: Is this directly related to your PR?

If not, let's not do it as part of this branch :blush:
Separate template file may be preferable and this is a simple change but having unrelated code has sometimes bloated PRs or even led to side-effects previously, so let's be strict with "one task, one PR"

src/app/features/notifications/notifications.component.ts
sleidig: I think in our case we should work with `SessionInfo` and its `id` property to identify and filter for users. Every authenticated user will have this (it's the keycloak user account ID).

While the demo should give you a defined user entity with `CurrentUserSubject` (not sure why you are getting undefined ...), this is not available for every user in every system, depending on the custom setup.

george-neha:
The UI for this notification feature is working as per functionality. However, the timestamp for each notification is a little confusing. 
image
**Suggestion:** The notifications can be segregated into "Now/New" and "Earlier." 
image

The timestamp can also be displayed in this manner: 
image

sleidig:
> The UI for this notification feature is working as per functionality. However, the timestamp for each notification is a little confusing.

Agreed :+1:   Let's display the time since event as suggested + a tooltip on that to see the actual date and time (in human-readable, locale formatting)

src/app/features/notifications/notifications.component.html
```diff
@@ -0,0 +1,91 @@
+<button
+  mat-icon-button
+  matTooltip="Notifications"
+  i18n-matTooltip="Notifications"
```
sleidig, line 4:
```suggestion
  i18n-matTooltip="notifications toolbar icon tooltip"
```

src/app/features/notifications/notifications.component.ts
```diff
@@ -0,0 +1,134 @@
...
+  ) {}
+
+  ngOnInit(): void {
+    this.loadAndProcessNotifications();
+  }
+
+  onNotificationBellClick() {
+    // this.firebaseNotificationService.sendNotification();
+    Logging.log("notificationBellClicked");
+  }
```
sleidig, line 60: does this still / in the future do anything? Or will the click simply open the menu (done directly in the template)?

src/app/features/notifications/notifications.component.html
```diff
@@ -0,0 +1,91 @@
...
+    [matBadge]="
+      hasUnreadNotificationCount > 0 ? hasUnreadNotificationCount : null
+    "
+    matBadgeColor="accent"
+  >
+    <fa-icon class="white" icon="bell"></fa-icon>
+  </span>
+</button>
+
+<mat-menu #data>
```
sleidig, line 18: maybe clearer?
```suggestion
<mat-menu #notificationList>
```

src/app/features/notifications/notifications.component.html
```diff
@@ -0,0 +1,91 @@
...
+  <div class="notification-panel">
+    <div
+      class="flex-row justify-space-between align-center notification-panel-header"
+    >
+      <h1 class="notification-title" i18n>Notifications</h1>
```
sleidig, line 23:
```suggestion
      <h2 class="notification-title" i18n>Notifications</h2>
```

src/app/features/notifications/notifications.component.html
```diff
@@ -0,0 +1,91 @@
...
+      (click)="$event.stopPropagation()"
+    >
+      <mat-tab label="All" i18n-label></mat-tab>
+      <mat-tab label="Unread" i18n-label></mat-tab>
+    </mat-tab-group>
+
+    <ng-container *ngIf="selectedTab === 0">
+      <!-- All Notifications -->
+      <app-notification-item
+        *ngFor="let notification of allNotifications"
```
sleidig, line 47: shall we switch to the new syntax of `@if` and `@for` here?

src/app/features/notifications/notifications.component.html
```diff
@@ -0,0 +1,91 @@
...
+      <div
+        class="flex-column no-notification-message"
+        *ngIf="allNotifications.length === 0"
+      >
+        <fa-icon class="no-notification-icon" icon="bell"></fa-icon>
+        <span class="no-notification-text" i18n>You have no notifications</span>
+      </div>
+    </ng-container>
+
+    <ng-container *ngIf="selectedTab === 1">
```
sleidig, line 61: rather than copying the (exactly same?) template, we can probably do something like `let ... of (selectedTab === 1 ? unreadNotifications : allNotifications)`?

src/app/features/notifications/notifications.component.html
```diff
@@ -0,0 +1,91 @@
...
+        [notification]="notification"
+        [onMarkAsRead]="markAsRead"
+        [onDeleteNotification]="deleteNotification.bind(this, notification)"
+      ></app-notification-item>
+      <div
+        class="flex-column no-notification-message"
+        *ngIf="allNotifications.length === 0"
+      >
+        <fa-icon class="no-notification-icon" icon="bell"></fa-icon>
+        <span class="no-notification-text" i18n>You have no notifications</span>
```
sleidig, line 57: The icon feels very massive. Lets make this a bit smaller (maybe half the size but with margin around) and remove the bold
image

src/app/features/notifications/notifications.component.html
```diff
@@ -0,0 +1,91 @@
...
+      </div>
+    </ng-container>
+  </div>
+</mat-menu>
+
+<!-- New mat-menu for ellipsis -->
+<mat-menu #settingsMenu>
+  <button mat-menu-item (click)="markAllRead($event)">
+    <fa-icon icon="bell"></fa-icon>
+    <span class="menu-option" i18n>Mark as notification active</span>
```
sleidig, line 85:
```suggestion
    <span class="menu-option" i18n>Mark all notifications as read</span>
```

src/app/features/notifications/notifications.component.ts
```diff
@@ -0,0 +1,134 @@
...
+    MatTabsModule,
+    NotificationItemComponent,
+  ],
+  templateUrl: "./notifications.component.html",
+  styleUrl: "./notifications.component.scss",
+})
+export class NotificationsComponent implements OnInit {
+  public allNotifications: NotificationEvent[] = [];
+  public unreadNotifications: NotificationEvent[] = [];
+  public hasUnreadNotificationCount = 0;
```
sleidig, line 43: simply use `unreadNotifications?.length` instead?

src/app/features/notifications/notifications.component.ts
```diff
@@ -0,0 +1,134 @@
...
+  public isEnableNotification = false;
```
sleidig, line 44: is this used?

src/app/features/notifications/notifications.component.ts
```diff
@@ -0,0 +1,134 @@
...
+  public selectedTab = 0;
+
+  constructor(
+    private entityMapper: EntityMapperService,
+    private mockNotificationsService: MockNotificationsService,
+    private currentUser: CurrentUserSubject,
```
sleidig, line 50: --> use SessionInfo as discussed?

sleidig (changes requested):
Regardless of my many remarks, the UI and code looks pretty good!

The UI updates (update indicator upon marking read, etc.) doesn't fully work yet. Do you still want to implement this here or only later?

src/app/features/notifications/notifications.component.ts
Ayush8923: Yes. Now, I have used the `SessionInfo`.

src/app/features/notifications/notifications.component.ts
Ayush8923: Yes, this used when the user enable the notification for their account. But for now, I have removed this variable. I will implement this later when I implement the logic for the notification list.

src/app/features/notifications/notifications.component.html
Ayush8923: The "Mark as notification active" option is used when the user needs to enable notifications for their account. However, to mark all notifications as read in one action, I have added a one more menu item labeled "Mark all as read." When the user clicks on this option, the readStatus of all notifications will be set to true.

src/app/features/notifications/notifications.component.html
Ayush8923: Thanks for this suggestion. I have update the code accordingly. Please take a look once.

src/app/features/notifications/notifications.component.ts
Ayush8923: Not needed. I have used this for testing purpose. But now I have removed this.

src/app/features/notifications/notification-item/notification-item.component.html
```diff
@@ -0,0 +1,36 @@
...
+
+  <!-- Blue indicator for unread notifications -->
+  <span
+    *ngIf="!notification.readStatus"
+    class="unread-indicator"
+    matTooltip="Unread notification"
+    i18n-matTooltip="Unread notification"
+    matBadgeColor="accent"
+    [matMenuTriggerFor]="notificationMenu"
+    (click)="$event.stopPropagation()"
```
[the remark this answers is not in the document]
Ayush8923: When I remove ` $event.stopPropagation()`, clicking on any menu item causes the menu to close automatically. For example, when I click the blue dot, the new menu item list opens, but the previous one closes. In some cases, however, I don't want the menu item to close automatically. That's why I added `$event.stopPropagation()` to prevent the menu item from closing on its own.

src/app/features/notifications/notification-item/notification-item.component.html
```diff
@@ -0,0 +1,36 @@
...
+>
+  <div class="notification-details">
+    <strong i18n class="notification-item">{{ notification.title }}</strong>
+    <p i18n class="notification-item">{{ notification.body }}</p>
+    <p i18n class="notification-time">{{ notification.created.at }}</p>
+  </div>
+
+  <!-- Blue indicator for unread notifications -->
+  <span
+    *ngIf="!notification.readStatus"
```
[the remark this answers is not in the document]
Ayush8923: I have added an empty circle for read notifications. When the user clicks on the icon, they will be able to delete the notification or mark it as unread.

src/app/features/notifications/notification-item/notification-item.component.html
```diff
@@ -0,0 +1,36 @@
...
+      <span class="menu-option" i18n>Mark as Read</span>
+    </button>
+    <button
+      mat-menu-item
+      (click)="deleteNotification(); $event.stopPropagation()"
+    >
+      <fa-icon icon="trash"></fa-icon>
+      <span class="menu-option" i18n>Delete Notification</span>
+    </button>
+  </mat-menu>
```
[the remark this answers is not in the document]
Ayush8923: I have added the condition-based `mark as read` and `mark as unread` options.

src/app/features/notifications/NotificationTimePipe.ts
```diff
@@ -0,0 +1,43 @@
+/**
```
Ayush8923, line 1: @sleidig Need your suggestion on this. I have created this file for the notification time. Is this the correct way to implement it, or are there any changes you would recommend?"

Ayush8923:
> > The UI for this notification feature is working as per functionality. However, the timestamp for each notification is a little confusing.
> 
> Agreed 👍 Let's display the time since event as suggested + a tooltip on that to see the actual date and time (in human-readable, locale formatting)

Thanks for this suggestion @george-neha @sleidig.
I have updated the UI accordingly, also I have added the tooltip so that when the user hovers over the date, a tooltip will appear displaying the date and time.

src/app/features/notification/NotificationTimePipe.ts
```diff
@@ -0,0 +1,43 @@
+/**
```
Ayush8923, line 1: @sleidig, Please share your suggestion on this, I have created this file for the notification time. Is this the correct way to implement it, or are there any changes you would recommend?

src/app/features/notification/notification.component.ts
```diff
@@ -0,0 +1,113 @@
...
+    // await this.entityMapper.loadType<NotificationEvent>(NotificationEvent);
+
+    // TODO: Need to remove this line once the notification list UI tests are completed.
+    const notifications =
+      this.mockNotificationsService.getNotifications() as unknown as NotificationEvent[];
+
+    // The user is hardcoded for testing purposes, need to remove this.
+    this.filterUserNotifications(
+      notifications,
+      this.sessionInfo.value?.entityId,
```
sleidig, line 66: needs to be id, not entityId (which again may not be defined for every user account)
```suggestion
    this.filterUserNotifications(
      notifications,
      this.sessionInfo.value?.id,
```

src/app/features/notifications/notification-item/notification-item.component.html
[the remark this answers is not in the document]
sleidig: Ah, I see ... Seems a bit tedious to put it on almost every event handler but if we need it, that's okay. (might be any way to stop this more centrally without all the repeated  stopPropagation calls?)

src/app/features/notification/notification.component.html
```diff
@@ -0,0 +1,89 @@
...
+        <div class="flex-column no-notification-message">
+          <fa-icon
+            class="no-notification-icon"
+            [icon]="['far', 'bell']"
+          ></fa-icon>
+          <span class="no-notification-text" i18n
+            >You have no notifications</span
+          >
+        </div>
+      }
```
sleidig, line 70: we can probably use `@empty` here to save some code: https://angular.dev/api/core/@for

src/app/features/notifications/notifications.component.html
sleidig: Thanks :-)

src/app/features/notifications/notifications.component.html
sleidig: Ah, I looked at the code to try and understand what you mean by "Mark as notification active". 
So this is something like "Enable notifications"?

sleidig (commented):
Looking pretty good already.

A few more optimizations for UI & UX:
- [x] menu icons should have some width to cleanly align: 
image
- [x] that menu (mark unread / delete upon blue dot) should auto-close after clicking one option, in my opinion
- [x] same for the top "..." menu. "Mark all read" should close the action menu (but keep the notifications list open).

src/app/features/notification/notification.component.html
```diff
@@ -0,0 +1,89 @@
...
+    <!-- Tabs -->
+    <mat-tab-group
+      [(selectedIndex)]="selectedTab"
+      (click)="$event.stopPropagation()"
+    >
+      <mat-tab label="All" i18n-label></mat-tab>
+      <mat-tab label="Unread" i18n-label></mat-tab>
+    </mat-tab-group>
+
+    @if (selectedTab === 0 || selectedTab === 1) {
```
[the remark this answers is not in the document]
Ayush8923: I checked the code again, and now we don't need this condition. The `selectedTab` will either be 0 or 1. I have updated it.

src/app/features/notification/NotificationTimePipe.ts
```diff
@@ -0,0 +1,43 @@
...
+    } else if (hours < 24) {
+      return `${hours}h`;
+    } else if (days === 1) {
+      return "Yesterday";
+    } else if (days < 7) {
+      return `${days}d`;
+    } else if (days >= 7) {
+      const dayOfWeek = notificationTime.toLocaleString("en-US", {
+        weekday: "short",
+      });
```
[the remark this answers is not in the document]
Ayush8923: I have modified the code to implement this:
1. For notifications 7 or more days ago, show the number of days (e.g. "10d", "15d").
2. For even longer timespans (30+ days), we can display the month and year (e.g. "Jan 2024").

Ayush8923:
> that menu (mark unread / delete upon blue dot) should auto-close after clicking one option, in my opinion

@sleidig I tried to implement the functionality in this way, but the `mat-menu` is not closing due to the use of `stopPropagation`. When I use `stopPropagation` in a `mat-menu` item, the `mat-menu` doesn't close. However, when I remove `stopPropagation`, both `mat-menu` close. So, I implemented it in such a way that whenever the user selects any `mat-menu` item, the `mat-menu` does not close. If the user wants to close it, they can do so. Could you please suggest a solution if you have one?

> same for the top "..." menu. "Mark all read" should close the action menu (but keep the notifications list open).

Same for this.

Ayush8923:
> Deployed to https://pr-2711.aam-digital.net/

@george-neha I have made some updates to the notification list UI. Please take a look and let us know if you have any suggestions. Also, I have modified the notification time and when the user hovers in the notification time then in the tooltip the actual date and time will be shown.
1. For notifications 7 or more days ago, show the number of days (e.g. "10d", "15d").
2. For even longer timespans (30+ days), we can display the month and year (e.g. "Jan 2024").

george-neha:
Working as per discussed functionality and UI. Can be moved to tech review.
@Ayush8923- Can this be aligned?
image

src/app/features/notification/notification.component.html
```diff
@@ -0,0 +1,97 @@
...
+      </button>
+    </div>
+
+    <!-- Tabs -->
+    <mat-tab-group [(selectedIndex)]="selectedTab">
+      <mat-tab label="All" i18n-label></mat-tab>
+      <mat-tab label="Unread" i18n-label></mat-tab>
+    </mat-tab-group>
+
+    <div class="flex-column gap-small">
```
sleidig, line 41: Can you design the divs in a way that we only have a scrollbar within the items and the title + tabs are always visible outside of the scroll box?

image

sleidig (approved):
Made some small improvements. Looks fine now :+1:

src/app/features/notification/notification.component.html
Ayush8923: Yes, we can do this. I have updated the code accordingly. but slight text is shown in the background at the top when we scroll the notification list.

Ayush8923: It's working properly now. When the user scrolls through the notification list, only the notification list will scroll, while the title and tabs remain fixed.

sleidig: somehow, if I scroll all the way to the bottom the top bar eventually scrolls out, although initially it remains fixed as intended, @Ayush8923

Ayush8923: @sleidig  I have fixed this. Now, when the user scrolls the notification list then the title and tabs remain fixed.

sleidig: your fixed caused a horizontal scrollbar for me and displayed the scroll across the whole menu rather than just the part that is actually scrollable. I made some further adjustments to fix this.

Issue #2696, opened 2024-12-02
Configure/Disable certain notifications for a user


sleidig:
To allow users to enable or disable various notification types in the UI:
- [x] Create the multi-select option on the user [profile page](https://demo.aam-digital.com/user-account), allowing users to select the notification types they want to receive.
- [x] Implement the Notification setting UI according to this suggested [UI](https://github.com/Aam-Digital/ndb-core/issues/2696#issuecomment-2531195353)

For details on the required database updates, here is the discussion:
https://github.com/Aam-Digital/ndb-core/issues/2690#issuecomment-2511143287

sleidig:
This looks great! I like the individual toggling for each type but having the "where you receive" as a central setting separately.

For the "What notifications" section, I could imagine a listing if different active and standard notification types and then a button to "add new notification rule" for configuring an additional custom trigger (e.g. a very special filter to get informed about a sub-type of certain entities). A bit like this (as a sample of the logic, not necessarily for the exact UI):
![image](https://github.com/user-attachments/assets/4e1d04c1-063c-4402-9f2f-e2b6a1593be3)

Such custom notification triggers (in contrast to default general notification types) should also offer to get deleted rather than just deactivated.

Abhinegi2:
@Ayush8923  
can we add some text like this -
<img width="822" alt="Image" src="https://github.com/user-attachments/assets/716694ea-4de8-47dc-87b3-2ede323dd838" />

Ayush8923:
# TODOs for Notification UI Updates  

## Descriptions  
- [ ] Update the description for the Notification Settings top title.  
- [ ] Add a description to the `What notifications you receive` title.  
- [ ] Add a description to the `Where You Receive Notifications` title.  

## Dynamic Saving  
- [ ] Remove the Save and Cancel buttons and implement dynamic saving functionality.  

## Information Icons  
Add an information icon to the following:  
- [ ] Select Entity Type  
- [ ] Notification Condition  
- [ ] Toggle Button  
- [ ] Push Notification inside the `Where you receive notifications` section

## Default Settings  
- [ ] If no notification settings are selected by the user, add a default entry for each field.  

## Test Trigger Button  
- [ ] Add a `Test Trigger` button to each notification rule so users can preview how the notification will appear.  

## Section Updates  
- [ ] Update the `Where You Receive Notifications` section:  
  - [ ] Add Email Notifications with a `Coming Soon` label and keep it disabled.  

---

[...]

PR #2712, opened 2024-12-12
[Feat]: Notification Setting Page

Ayush8923:
closes: https://github.com/Aam-Digital/ndb-core/issues/2696

### Visible/Frontend Changes

- [x] Create the notification setting page inside the profile page.
- [x] Create the notification setting page according to this [suggested UI](https://github.com/Aam-Digital/ndb-core/issues/2696#issuecomment-2531195353).


Ayush8923:
> Deployed to https://pr-2712.aam-digital.net/

@george-neha Please test the notification setting page UI inside the profile page in the above-mentioned URL and share your feedback/suggestions. For reference: please check this issue's [comments](https://github.com/Aam-Digital/ndb-core/issues/2696#issuecomment-2531195353).

src/app/core/entity/entity-type-select/entity-type-select.component.html
```diff
@@ -0,0 +1,12 @@
+<mat-form-field>
```
Ayush8923, line 1: As per the discussion on Slack, for now we can use `mat-form-field` inside the EntityTypeSelectComponent. As for the approach of splitting `mat-form-field` from the component would require us to implement extends `CustomFormControlDirective` on the `EntityTypeSelectComponent`. For this, we can create the issue for that and implement it later.

george-neha:
UI working as per functionality. 
**Suggestion:** The Save button should be enabled only if any changes are made to the notification settings. 
image

Ayush8923:
> UI working as per functionality. **Suggestion:** The Save button should be enabled only if any changes are made to the notification settings. image

Thanks for this suggestion @george-neha.
Yes, we will implement this logic in this [issue](https://github.com/Aam-Digital/ndb-core/issues/2723) and also, and I have also added this suggestion in this issue description.

Ayush8923:
> Deployed to https://pr-2712.aam-digital.net/

@george-neha As discussed regarding the [notification settings page UI](https://github.com/Aam-Digital/ndb-core/issues/2696#issuecomment-2553402972), I have updated the UI accordingly. You can test the changes using the link [provided above](https://github.com/Aam-Digital/ndb-core/pull/2712#issuecomment-2538211090). Please let me know if any further adjustments are needed or if you have any suggestions.

sleidig (commented):
UI looks good to me mostly (no code review yet, as we have to resolve merge conflicts since the base branch was merged and squashed recently).

- [ ] I think we should more prominently highlight if push notifications are disabled. Maybe show a banner in the "Where you receive notifications" or a warning icon next to the "Push Notification" entry if it is not enabled?
- [x] "Notification Method" dropdown should be enabled and allow multi-select of all option - even if that is only "Push Notification" currently and nothing else is supported. That will make the purpose clearer to the user. (and they can actually disable "push" and just see it in the "Notification Center", if they want)
- [x] The initial rule cannot be deleted. I would rather display an empty section with only the "add new" button in that case.

george-neha:
> * [ ]  I think we should more prominently highlight if push notifications are disabled. Maybe show a banner in the "Where you receive notifications" or a warning icon next to the "Push Notification" entry if it is not enabled?

@sleidig @Ayush8923 for this I had initially suggested an approach where the user will first notice "where you receive notifications" before they take action on individual notifications here under this [issue](https://github.com/Aam-Digital/ndb-core/issues/2696#issuecomment-2559685167)

> 1.The "where you receive notifications" section should be moved up. In situations where multiple notification conditions are added, selecting where the notifications should be received requires scrolling down to enable or disable them. 

I honestly feel the page can get a hierarchy refresh which would benefit the user from not missing crucial information. 

Secondly, for additional highlight we can use status badges using text "Enabled/ Disabled" next to the toggle button. 

Here is an example I saw from Material Design:

image

The banner may not scale in the future if we need to highlight about email notifications too. 

What do you think?

sleidig:
I somewhat disagree about changing the order, moving "where  you receive notifications" to the top: While it is important for users to understand if "push" is not enabled, I expect this setting to hardly ever be changed while users may regularly change the individual notification rules. Therefore, I believe that section is much more likely to be relevant and should stay on top.

Ayush8923:
> I think we should more prominently highlight if push notifications are disabled. Maybe show a banner in the "Where you receive notifications" or a warning icon next to the "Push Notification" entry if it is not enabled?

@sleidig @george-neha I have added a warning icon next to the "Where you receive notifications" section title if the user's notifications are not enabled. Can you please check it once and let me know if any changes or updates are needed in the Notification Settings page UI?
image

2.  As per the @george-neha suggestion, the UI is looking like this, please check this also and please share your suggestion/thoughts. 
image

sleidig:
> > I think we should more prominently highlight if push notifications are disabled. Maybe show a banner in the "Where you receive notifications" or a warning icon next to the "Push Notification" entry if it is not enabled?
> 
> @sleidig @george-neha I have added a warning icon next to the "Where you receive notifications" section title if the user's notifications are not enabled. Can you please check it once and let me know if any changes or updates are needed in the Notification Settings page UI?
> image
> 
> 2.  As per the @george-neha suggestion, the UI is looking like this, please check this also and please share your suggestion/thoughts. 
> image
> 
> 

I like it. If we have the enabled badge there, maybe turn the warning icon into a red "disabled " badge for consistency and better visibility?

Ayush8923:
> I like it. If we have the enabled badge there, maybe turn the warning icon into a red "disabled " badge for consistency and better visibility?

Thanks for this suggestion, I have updated the UI according. So if the push notification is enabled then the `Enabled` badge is showing next to the `Where you receive notifications` section and if the push notification is disabled then the `Disabled` badge is showing.

Ayush8923:
> Deployed to https://pr-2712.aam-digital.net/

@george-neha I'm Moving this to Functional Review. You can test the changes using the link [provided link above](https://github.com/Aam-Digital/ndb-core/pull/2712#issuecomment-2538211090). Please let me know if any further adjustments are needed or if you have any suggestions.

src/app/features/notification/notification-settings/notification-settings.component.scss
```diff
@@ -0,0 +1,133 @@
...
+  margin-top: 30px;
+}
+
+.browser-notifications-container {
+  margin-top: 32px;
+}
+
+.notification-setting-title {
+  margin-bottom: 10px;
+}
```
sleidig, line 22: Why is this sometimes 30px sometimes 32px? We should probably work with our standard sizes from `@use "variables/sizes";`

src/app/features/notification/notification-settings/notification-settings.component.scss
```diff
@@ -0,0 +1,133 @@
...
+
+.panel-wrapper {
+  box-shadow: 0px 0px 10px 0px rgba(0, 0, 0, 0.1);
```
sleidig, line 25: use mat-elevation for consistency?
`@use "@angular/material/core/style/elevation" as mat-elevation;`

sleidig (changes requested):
Some last glitches, otherwise both code and UI look good :+1:

src/app/core/entity/entity-type-select/entity-type-select.component.ts
```diff
@@ -14,22 +28,51 @@ import { MatFormField, MatLabel } from "@angular/material/form-field";
...
+    {
+      provide: NG_VALUE_ACCESSOR,
+      useExisting: forwardRef(() => EntityTypeSelectComponent),
+      multi: true,
+    },
+  ],
   standalone: true,
 })
+
+// TODO: Need to use the CustomFormControlDirective abstract class instead of ControlValueAccessor.
```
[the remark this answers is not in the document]
Ayush8923: @sleidig The reason for making changes in this component was that earlier we were splitting the `mat-form-field`. We have an existing component for the entity field but we couldn't import the `EntitySelectComponent`, which is a dynamic component, into an HTML file. As a result, we were encountering an error like `No provider for NgControl found in NodeInjector.` Based on previous discussions, we realized that extending the `CustomFormControlDirective` is necessary, as this is the long-term solution to implement special components as form controls. However, for now, the approach might be too complex and time-consuming. So that's why we did not extend the `CustomFormControlDirective`. Here is the discussion [link](https://aam-digital.slack.com/archives/D080VLA5WUB/p1734366972998179?thread_ts=1734366484.204119&cid=D080VLA5WUB).

Additionally, in this component, the test cases were failing, so to resolve the issues with the test cases, I added the required functions.

src/app/features/notification/notification-settings/notification-settings.component.scss
```diff
@@ -0,0 +1,133 @@
...
+.notifications-options-container {
+  @extend .panel-wrapper;
+  background-color: colors.$background;
+  margin-top: 20px;
+  gap: 40px;
+
+  @media screen and (min-width: breakpoints.$xl) {
+    flex-direction: row;
+  }
+}
```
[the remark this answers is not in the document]
Ayush8923: Yes, I added the breakpoint because, in the `notification-panel-wrapper`, we have two sections: one for the notification rule input field and another for the notification rule action items. On large screens (1400px, I updated the breakpoint to fix the UI breaking issue you mentioned), we need all the action items and input fields in a single row. For screens smaller than 1400px, we can display the notification rules in a column layout. However, if I remove the breakpoints and use `flex-wrap`, the entire UI breaks because each input field has its own defined width.

src/app/features/notification/notification-settings/notification-settings.component.scss
Ayush8923: Thanks, I update this code according to your suggestion.

sleidig (approved):
Looks good :slightly_smiling_face:

PR #2714, opened 2024-12-12
fix: add type check before date parsing

tomwwinter:


sleidig (approved):
LGTM :+1: 

maybe a small unit test?

PR #2718, opened 2024-12-13
[Feat]: Entity Type Selector Component

Ayush8923:
closes: #XX

### Visible/Frontend Changes

- [x]
- [ ]

### DIFF
--- a/src/app/core/entity/entity-type-selector/entity-type-selector.component.html
+++ b/src/app/core/entity/entity-type-selector/entity-type-selector.component.html
@@ -1,0 +1,8 @@
+<mat-form-field>
+    <app-basic-autocomplete
+        [options]="entityTypes"
+        [optionToString]="optionToLabel"
+        [valueMapper]="optionToId"
+        [multi]="isMulti"
+    ></app-basic-autocomplete>
+</mat-form-field>
--- a/src/app/core/entity/entity-type-selector/entity-type-selector.component.spec.ts
+++ b/src/app/core/entity/entity-type-selector/entity-type-selector.component.spec.ts
@@ -1,0 +1,23 @@
+import { ComponentFixture, TestBed } from '@angular/core/testing';
+
+import { EntityTypeSelectorComponent } from './entity-type-selector.component';
+
+describe('EntityTypeSelectorComponent', () => {
+  let component: EntityTypeSelectorComponent;
+  let fixture: ComponentFixture<EntityTypeSelectorComponent>;
+
+  beforeEach(async () => {
+    await TestBed.configureTestingModule({
+      imports: [EntityTypeSelectorComponent]
+    })
+    .compileComponents();
+
+    fixture = TestBed.createComponent(EntityTypeSelectorComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+});
--- a/src/app/core/entity/entity-type-selector/entity-type-selector.component.ts
+++ b/src/app/core/entity/entity-type-selector/entity-type-selector.component.ts
@@ -1,0 +1,42 @@
+import { Component, OnInit, Input } from '@angular/core';
+import { DynamicComponent } from "../../config/dynamic-components/dynamic-component.decorator";
+import { BasicAutocompleteComponent } from "../../common-components/basic-autocomplete/basic-autocomplete.component";
+import { FaIconComponent } from "@fortawesome/angular-fontawesome";
+import { MatFormField, MatLabel } from "@angular/material/form-field";
+import { MatTooltip } from "@angular/material/tooltip";
+import { NgIf } from "@angular/common";
+import { ReactiveFormsModule } from "@angular/forms";
+import { MatInput } from "@angular/material/input";
+import { EntityConstructor } from "../model/entity";
+import { EntityRegistry } from "../database-entity.decorator";
+
+/**
+ * Edit component for selecting an entity type from a dropdown.
+ */
+@DynamicComponent("EditEntityTypeDropdown")
+@Component({
+  selector: 'app-entity-type-selector',
+  templateUrl: './entity-type-selector.component.html',
+  imports: [
+    BasicAutocompleteComponent,
+    MatFormField,
+    ReactiveFormsModule,
+  ],
+  standalone: true,
+})
+export class EntityTypeSelectorComponent implements OnInit
+{
+  @Input() isMulti = false;
+
+  entityTypes: EntityConstructor[];
+  optionToLabel = (option: EntityConstructor) => option.label;
+  optionToId = (option: EntityConstructor) => option.ENTITY_TYPE;
+
+  constructor(private entityRegistry: EntityRegistry) {}
+
+  ngOnInit() {
+    this.entityTypes = this.entityRegistry
+      .getEntityTypes(true)
+      .map(({ value }) => value);
+  }
+}


src/app/core/entity/entity-type-selector/entity-type-selector.component.ts
```diff
@@ -0,0 +1,32 @@
...
+  imports: [BasicAutocompleteComponent, MatFormField, ReactiveFormsModule],
+  standalone: true,
+})
+export class EntityTypeSelectorComponent implements OnInit {
+  @Input() isMulti = false;
+
+  entityTypes: EntityConstructor[];
+  optionToLabel = (option: EntityConstructor) => option.label;
+  optionToId = (option: EntityConstructor) => option.ENTITY_TYPE;
+  placeholder = "Select an entity type";
```
sleidig, line 23: Let's  move this into the html template (and add i18n-placeholder there to the html element, so that it is translatable)

src/app/core/entity/entity-type-selector/entity-type-selector.component.html
```diff
@@ -0,0 +1,9 @@
+<mat-form-field>
```
sleidig, line 1: Do we need this mat-form-field to wrap around it here?

src/app/core/entity/entity-type-selector/entity-type-selector.component.spec.ts
```diff
@@ -0,0 +1,22 @@
+import { ComponentFixture, TestBed } from "@angular/core/testing";
+import { EntityTypeSelectorComponent } from "./entity-type-selector.component";
+import { MockedTestingModule } from "../../../utils/mocked-testing.module";
+
+describe("EditConfigurableEnumComponent", () => {
```
sleidig, line 5:
```suggestion
describe("EntityTypeSelectorComponent", () => {
```

src/app/core/entity/entity-type-selector/entity-type-selector.component.ts
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Component for selecting an entity type from a dropdown.
+ */
+@Component({
+  selector: "app-entity-type-selector",
+  templateUrl: "./entity-type-selector.component.html",
+  imports: [BasicAutocompleteComponent, MatFormField, ReactiveFormsModule],
+  standalone: true,
+})
+export class EntityTypeSelectorComponent implements OnInit {
```
sleidig, line 17: I think we are now duplicating the same code that is also in EditEntityTypeDropdownComponent?

Can you change the EditEntityType... component to use this new component now, instead of keeping duplicate code?

sleidig (changes requested):

src/app/core/entity/entity-type-selector/entity-type-selector.component.html
Ayush8923: I think yes, we need the`mat-form-field` to apply both styling and behaviour to the form controls.

src/app/core/entity/entity-type-selector/entity-type-selector.component.ts
Ayush8923: Yes, I have made a few changes to the `EditEntityTypeDropdownComponent`. Please take a look and let me know if there are any other changes needed.

Ayush8923:
@sleidig I made the changes according to your suggestions. Please take a look and let me know if there are any changes needed. Also, I take the latest pull from the master branch then the deployment and CI checks fail.

src/app/core/entity/edit-entity-type-dropdown/edit-entity-type-dropdown.component.ts
```diff
@@ -36,20 +19,12 @@ export class EditEntityTypeDropdownComponent
...
   multi = false;
 
-  entityTypes: EntityConstructor[];
-  optionToLabel = (option: EntityConstructor) => option.label;
-  optionToId = (option: EntityConstructor) => option.ENTITY_TYPE;
-
-  constructor(private entityRegistry: EntityRegistry) {
+  constructor() {
     super();
   }
```
sleidig, line 24: can completely skip constructor here now

src/app/core/entity/edit-entity-type-dropdown/edit-entity-type-dropdown.component.html
```diff
@@ -1,11 +1,4 @@
...
-  <app-basic-autocomplete
-    [formControl]="formControl"
-    [options]="entityTypes"
-    [optionToString]="optionToLabel"
-    [valueMapper]="optionToId"
-    [multi]="multi"
-  ></app-basic-autocomplete>
-</mat-form-field>
+<app-entity-type-selector
+  [isMulti]="multi"
```
sleidig, line 2: let's maybe call this "allowMultiSelect"

src/app/core/entity/entity-type-selector/entity-type-selector.component.html
sleidig: I understand. But I think it may be better here to put the mat-form-field wherever we use app-entity-type-select, instead of inside the component itself. That's the model we chose for app-basic-autocomplete also (and I hope we had some reasons for that :sweat_smile: ). This would also ensure the EditEntityTypeDropdown will define its own mat-form-field including label, which you have currently lost now

src/app/core/entity/entity-type-selector/entity-type-selector.component.ts
```diff
@@ -0,0 +1,32 @@
...
```
sleidig, line 17: I would like to name this "EntityTypeSelect" not ...selector because we also have an EntitySelectComponent. So we don't introduce different terms for similar functions

sleidig (changes requested):

src/app/core/entity/edit-entity-type-dropdown/edit-entity-type-dropdown.component.ts
```diff
@@ -18,16 +11,7 @@ import { EntityRegistry } from "../database-entity.decorator";
...
-    MatLabel,
-    MatTooltip,
-    NgIf,
-    ReactiveFormsModule,
-    MatInput,
-  ],
+  imports: [EntityTypeSelectComponent, MatFormField, MatLabel],
   standalone: true,
 })
 export class EditEntityTypeDropdownComponent
```
sleidig, line 17: Sorry, one more rename request, to also exactly align this with the other components' naming: 
EditEntityTypeDropdown --> EditEntityType (including test and folder, as usual)

sleidig (commented):
Something seems to have broken the PublicFormConfig form now, where we also use the edit component: https://pr-2718.aam-digital.net/admin/public-form/test
Can you check that?

Ayush8923:
> Something seems to have broken the PublicFormConfig form now, where we also use the edit component: https://pr-2718.aam-digital.net/admin/public-form/test Can you check that?

Sure, let me check this and will add the necessary corrections.

src/app/core/entity/entity-type-select/entity-type-select.component.html
```diff
@@ -1,11 +1,12 @@
 <mat-form-field>
   <mat-label>{{ label }}</mat-label>
```
Ayush8923, line 2: As per the discussion on Slack, for now we can use `mat-form-field` inside the EntityTypeSelectComponent. As for the approach of splitting `mat-form-field` from the component would require us to implement extends `CustomFormControlDirective` on the `EntityTypeSelectComponent`. For this, we can create the issue for that and implement it later.

Ayush8923:
> Something seems to have broken the PublicFormConfig form now, where we also use the edit component: https://pr-2718.aam-digital.net/admin/public-form/test Can you check that?

@sleidig I've made some changes and resolved the issue. Please review this PR and let me know if any further adjustments are required.

sleidig (approved):

Issue #2657, opened 2024-11-12
public-form view throws exception if the form-config does not exist locally


tomwwinter:
**Describe the bug**
When a user open a public form with an id that does not exist in the local database, an unhandled exception is thrown.

**To Reproduce**
Start application locally

1. Go to some non existing public-form: http://localhost:4200/public-form/123f
2. See error

**Expected behavior**
We should handle all errors and show a 404 error page.

**Screenshots**
![image](https://github.com/user-attachments/assets/b2d4b5cf-fc5c-4343-88ab-4069b364110e)

**Desktop (please complete the following information):**

- OS: all
- Browser: all 
- Version: 3.42.2

**Additional context**

sleidig:
Is this still happening after the latest changes, @Abhinegi2 ?

Abhinegi2:
Yes, @sleidig It is still happening let me look into this issue

Abhinegi2:
@sleidig
While reviewing the current approach for handling public forms, I found two possible ways to address errors:

1. **Redirect to a 404 Page**
Use a try-catch block to detect errors and redirect to the existing 404 page. However, if we choose this option, we may need to modify the HTML of the 404 page, as it currently includes a "Go to Dashboard" button, which might not be relevant in this context.
2. **Handle Errors Within the Component**
 If the public form configuration does not match any valid entry, we can set an error state within the component itself and display an error message (e.g., "Form not found") directly in the template, instead of redirecting to a 404 page. 

Let me know your thoughts on these approaches

sleidig:
Thanks for working this out!

In this specific context (the component is used by external visitors also, in contrast to all other parts of the app) I think it'll be more clear and flexible to go with (2), i.e. adding error handling to the component itself

george-neha:
Changes are working according to the functionality. Can be pushed for review.

Abhinegi2:
@george-neha, could you please test the following cases on the [dev site](https://dev.aam-digital.net/)?

1. Non-existent Public Form
- Try to access a public form that does not exist.
- Verify that the message displayed is:
"The form you are looking for is either unavailable or doesn't exist. Please check the link and try again."
Access without Required Permissions

2. Access a public form that exists but requires permissions you do not have( Currently, only child-related public forms are accessible.)
- Try to access a public form with an entity type other than "Child.
- Verify that the message displayed is:
"You do not have the required permissions to submit this form. Please log in or contact your system administrator."

george-neha:
[...]

PR #2726, opened 2024-12-16
added error handling template for Public forms

Abhinegi2:
closes: [#2657](https://github.com/Aam-Digital/ndb-core/issues/2657)

### Visible/Frontend Changes

- [x] Showcase Public Form Not Found whenever any error occurs in the loading form from config.


src/app/features/public-form/public-form.component.ts
```diff
@@ -86,20 +86,25 @@ export class PublicFormComponent<E extends Entity> implements OnInit {
   }
 
   private async loadFormConfig() {
-    const id = this.route.snapshot.paramMap.get("id");
+    try {
```
sleidig, line 89: While a general try-catch around everything works, of course, this is not very fine-grained error handling :wink:  

 Can you refine this to only "try" the part that can actually throw errors and also distinguish between "not found" (--> just display feedback to user) and any truly unexpected error (--> use Logging to track this in our monitoring)?

Abhinegi2: i have updated the error handling to use Logging to track the unexpected error and for expected (Public form config not found ) displaying the feedback to user

sleidig: The way this is implemented now, I do not see any "expected" way why this should throw an error. Even without permissions, the loadAll would give an empty array, I think. So we can skip the try block completely - or am I missing something?

Generally, try-blocks should only wrap 1-2 lines of code, in my opinion. It should be very clear what kind of error can happen from exactly which line and then catch that specifically only.

Abhinegi2: Previously, on my local setup(without docker usage), when the database was not connected, only the title and description were loading, and the fields were not visible. That’s why I thought adding a try-catch block to handle such cases.

However, I agree that we can skip this since no errors have occurred after establishing the database connection(using aam-services).

Also thanks for sharing the insights on try-catch usage, I'll keep this in mind going forward.

sleidig: those errors (preventing the fields to load) were caused by server not being reachable? Or what were the specific errors?
There could be some cases that we do want to catch. If so, let's add a try-catch but trying to be rather specific what we think could fail.

sleidig: Oh, right! That's a good point that we could maybe check and add an explicit error message for. I bet that'll save us a lot of time troubleshooting, the next time we run into this :laughing: 

How about injecting `EntityAbility` and using `this.ability.cannot("create", entityType)` to explicitly check the required permissions in our init method here, possibly setting an error like "You do not have the required permissions to submit this form. Please log in or contact your system administrator."?

Abhinegi2: I initially injected `EntityAbility and AlertService` to check permissions and set errors within the init method. However, this approach always displayed an error, regardless of whether the permissions were set correctly or not. To resolve this, I moved the logic `this.ability.cannot("create", entityType)` into the loadFromConfig method, where it now works as expected.

src/app/features/public-form/public-form.component.ts
```diff
@@ -93,10 +98,20 @@ export class PublicFormComponent<E extends Entity> implements OnInit {
...
+      return;
+    }
 
     this.entityType = this.entities.get(
       this.formConfig.entity,
     ) as EntityConstructor<E>;
+    if (this.ability.cannot("create", this.entityType)) {
+      this.alertService.addDanger(
+        "You do not have the required permissions to submit this form. Please log in or contact your system administrator",
+      );
```
sleidig, line 112: let's make this a text like the "not found" in the template also. Alerts tend to be easily overlooked by users.

(maybe turn the "publicFormNotFound" into a general "error" property that can hold different values?)

sleidig (commented):
Code looks good! Just one small UI change suggestion.

Do you know how to write unit tests for angular? Would be great to write two small tests for the two error conditions (not found + no permissions).

Let's then set up dev server with scenarios for both errors and let Neha test one last time

src/app/features/public-form/public-form.component.ts
Abhinegi2: sure, let me make these changes

Abhinegi2:
> Code looks good! Just one small UI change suggestion.
> 
> Do you know how to write unit tests for angular? Would be great to write two small tests for the two error conditions (not found + no permissions).
> 
> Let's then set up dev server with scenarios for both errors and let Neha test one last time

Sure Sebastian, Let me try to add test cases for both scenarios, will let you know if I face any issues.

Abhinegi2:
> I've added the test case for 'no permissions' and am now working on the 'not found' scenario. I'm facing some issues, but I'll try to resolve them. If I get stuck for too long, I may need some assistance.

Implemented all the requested changes, Please have a look now.

src/app/features/public-form/public-form.component.ts
sleidig: Heading is wrong in this case (claiming "form not found"). If the user has permission to the form config but not to create a new record, we can display the actual form title and description, I think, and only below that display the permission error. Then the user has more context?

Test Page: https://dev.aam-digital.net/public-form/schoolform

What do you think from a user's perspective, @george-neha ?

george-neha: As a user, the message should be empathetic and should also indicate the next/other steps the user could take. 
The wording for the message could be :

**Access Denied to Public Form-Title**

Unfortunately, access has been denied due to insufficient permissions to access this Public Form.
Please contact your team for further assistance.

Abhinegi2: @george-neha it's looking better now?
image

src/app/features/public-form/public-form.component.ts
```diff
@@ -93,10 +98,24 @@ export class PublicFormComponent<E extends Entity> implements OnInit {
...
+        error: $localize`The form you are looking for is either unavailable or doesn't exist. Please check the link and try again.`,
+      };
+      return;
+    }
 
     this.entityType = this.entities.get(
       this.formConfig.entity,
     ) as EntityConstructor<E>;
+    if (this.ability.cannot("create", this.entityType)) {
+      this.publicFormNotFound = {
```
sleidig, line 113: rename to "error"?

sleidig: and to simplify the component code, maybe it's easier to just add a `error: "not_found" | "no_permissions"` and move the texts to the html template file

src/app/features/public-form/public-form.component.ts
```diff
@@ -93,10 +98,24 @@ export class PublicFormComponent<E extends Entity> implements OnInit {
...
+        title: $localize`Access Denied to Public Form - ${this.formConfig.title}`,
+        error: $localize`Unfortunately, access has been denied due to insufficient permissions to access this Public Form. Please contact your system administrator.`,
```
sleidig, line 115:
```suggestion
        title: $localize`No access to public form "${this.formConfig.title}"`,
        error: $localize`Unfortunately, you do not have the required permissions to submit data through this form. Please contact the person who shared this link with you. In case you have a user account, please make sure you are logged in.`,
```

sleidig (commented):
Final tuning. I think this will be the last iteration.

PR #2738, opened 2024-12-19
[Feat]: Notification list logic updates

Ayush8923:
closes: #2690

### Visible/Frontend Changes

- Display all notifications in the `All` tab.
- Display unread notification in the `Unread` tab.
- When a user clicks on the `Enable` & `Disable` notifications then the corresponding user enabled and disabled the notification accordingly.
- When a user clicks on the `Mark all as read` then all notifications `readStatus` `true` and all the notifications will be readed.
- When the user clicks on the `Mark as read` and `Mark as unread` options than the `readStatus` boolean value is updated accordingly.
- When the user clicks on the `Delete Notification` option then the chooses notification will be deleted.
- When the user clicks on any notification user will be redirected to the corresponding notification action URL.
- When the user clicks on the `Notification Setting` option, they are redirected to the notification setting page (when the notification setting page is merged). This is also handled in this [PR](https://github.com/Aam-Digital/ndb-core/pull/2712).
- When the user marks the notification read & unread and deletes any notification then the notification count will be updated accordingly.


Ayush8923:
@tomwwinter Currently, the notification list database schema is structured as follows, and the notification list logic has been implemented based on this schema:
```json
{
  "_id": "NotificationEvent:7440a5e539b7a97f459e820467041486",
  "_rev": "25-faeee9c667c7cc6dbc34e6605881c48e",
  "title": "Dummy Notification Title",
  "body": "This is a dummy notification body.",
  "actionURL": "/ayush",
  "notificationFor": "a146a21f-93ee-438e-ba3d-f2bc469568a6",
  "fcmToken": "f6dmHHeKUM1Q_DfSCLmtl9:APA91bEgFmu2yUcMTzOr8CusfjWHTUPPvsVivuOX5pGAInhmPsLrEdAWsuPFBHjwNaLnrQWAjBXohvn1Mkb_fYHwIpIi3aBiOy9HudBWzytXDwwlbLwd66o",
  "readStatus": false,
  "created": {
    "at": "2024-12-03T19:10:07.647Z"
  },
  "updated": {
    "at": "2024-12-26T07:02:26.832Z"
  }
}
```

Could you please confirm if the `notificationEvent` will be saved in the database in this manner after the notification is triggered? Additionally, let me know if there are any updates or changes to the `NotificationEvent` database schema so I can adjust the notification list logic accordingly.

src/app/features/notification/model/notification-config.ts
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Represents a specific notification type configuration.
+ */
+export class NotificationType {
+  @DatabaseField() notificationType: string;
+  @DatabaseField() enabled: boolean;
+  @DatabaseField() channels: { [key: string]: boolean };
+  @DatabaseField() entityType: string;
+  @DatabaseField() conditions: any;
+}
```
sleidig, line 28:
```suggestion
export class NotificationConfig extends Entity {
  /**
   * The default mode(s) through which all notifications are sent to the user.
   *
   * Can be overwritten for each notificationType to disable a channel for certain notifications.
   * If the channel is not activated globally here, the individual override has no effect, however.
   */
  @DatabaseField() channels: { [key: string]: boolean };

  /**
   * Specific rules to trigger notifications.
   */
  @DatabaseField() notificationTypes: NotificationType[];

  constructor(id = null) {
    super(id);
  }
}

/**
 * Represents a specific notification type configuration.
 */
export class NotificationType {
  /** The general type of notification (e.g. changes to entities, etc.) */
  @DatabaseField() notificationType: string;
  
  /** whether this notification is enabled or currently "paused" */
  @DatabaseField() enabled: boolean;
  
  /**
   * override for the global notification channel(s).
   * e.g. define here if this specific notification rule should not show as email/push notification
   */
  @DatabaseField() channels: { [key: string]: boolean };
  
  /** (for "entity_change" notifications only): type of entities that can trigger notification */
  @DatabaseField() entityType: string;
  
  /** (for "entity_change" notifications only): conditions which changes cause notifications */
  @DatabaseField() conditions: any;
}
```

src/app/features/notification/model/notification-config.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Represents a specific notification type configuration.
+ */
+export class NotificationType {
+  @DatabaseField() notificationType: string;
+  @DatabaseField() enabled: boolean;
+  @DatabaseField() channels: { [key: string]: boolean };
+  @DatabaseField() entityType: string;
+  @DatabaseField() conditions: any;
```
sleidig, line 27:
```suggestion
  @DatabaseField() conditions: DataFilter;
```

(see type already defined in the code base and used for list filters)

src/app/features/notification/model/notification-config.ts
```diff
@@ -0,0 +1,28 @@
+import { Entity } from "../../../core/entity/model/entity";
+import { DatabaseField } from "../../../core/entity/database-field.decorator";
+import { DatabaseEntity } from "../../../core/entity/database-entity.decorator";
+
+/**
+ * This represents one specific notification config for one specific user,
+ */
+@DatabaseEntity("NotificationConfig")
+export class NotificationConfig extends Entity {
+  @DatabaseField() channels: { [key: string]: boolean };
```
sleidig, line 10:
```suggestion
  @DatabaseField() channels: { [key: string]: boolean };
```

Can we introduce a `type NotificationChannel = "push"; ` instead of allowing any string for the keys? (later to add other options beyond push, was implemented.

src/app/features/notification/notification-item/notification-item.component.ts
```diff
@@ -44,5 +46,7 @@ export class NotificationItemComponent {
     this.deleteClick.emit();
   }
 
-  protected readonly closeOnlySubmenu = closeOnlySubmenu;
+  onNotificationClick(notification: any) {
```
sleidig, line 49: no `: any` typing, please

src/app/features/notification/notification.component.ts
```diff
@@ -35,80 +37,157 @@ import { closeOnlySubmenu } from "./close-only-submenu";
...
+    this.router.navigate([notification.actionURL]);
   }
 
-  protected readonly closeOnlySubmenu = closeOnlySubmenu;
+  /**
+   * Redirect the user to Notification setting page
+   */
+  onRedirectToNotificationsSetting() {
+    this.router.navigate(["/user-account"], { queryParams: { tabIndex: 1 } });
+  }
```
sleidig, line 211: I think you could directly use `[routerLink]="[...]"` directive in the html so that we don't need an extra method in the .ts here

sleidig: (same for notificationListener above?)

src/app/features/notification/notification.component.ts
```diff
@@ -35,80 +37,157 @@ import { closeOnlySubmenu } from "./close-only-submenu";
...
   }
 
+  /**
+   * Updates the read status of a single notification.
+   */
   async updateReadStatus(notification: NotificationEvent, newStatus: boolean) {
     notification.readStatus = newStatus;
     await this.entityMapper.save(notification);
     this.filterUserNotifications(this.allNotifications);
   }
```
sleidig: instead of another method, we can just call `updateReadStatusForNotifications([notification])` (wrapped in array instead of passing single item as parameter) and avoid the code duplication.

src/app/features/notification/notification.component.ts
```diff
@@ -35,80 +37,157 @@ import { closeOnlySubmenu } from "./close-only-submenu";
...
+    } else {
+      const newNotificationConfig = new NotificationConfig(this.userId);
+      newNotificationConfig.channels = notificationChannels;
+      await this.entityMapper.save(newNotificationConfig);
+    }
+
+    this.alertService.addInfo(
+      `Notification ${this.hasNotificationEnabled ? "enabled" : "disabled"}.`,
+    );
+  }
```
sleidig, line 164: Let's move this to the NotificationSettings component only and remove the menu entry for it also. Code (and maybe even UI) will be much easier if this is only in the full settings page and not also in the notification panel directly.

src/app/features/notification/notification.component.ts
```diff
@@ -35,80 +37,157 @@ import { closeOnlySubmenu } from "./close-only-submenu";
...
   public allNotifications: NotificationEvent[] = [];
   public unreadNotifications: NotificationEvent[] = [];
+  private notificationsSubject = new Subject<NotificationEvent[]>();
   public selectedTab = 0;
   public hasNotificationEnabled = false;
+  protected readonly closeOnlySubmenu = closeOnlySubmenu;
 
   constructor(
     private entityMapper: EntityMapperService,
-    private mockNotificationsService: MockNotificationsService,
```
sleidig: can completely delete that file now, right?

sleidig (changes requested):
Looks good and seems to work pretty well. Just some code cleanup requests.

When clicking on an event, that should also mark it as read, in my opinion.

Items that have no action still appear clickable but nothing happens for the user. In that case, we could just mark the item as read upon click.

src/app/features/notification/notification.component.ts
```diff
@@ -0,0 +1,230 @@
...
+    private entityMapper: EntityMapperService,
+    private sessionInfo: SessionSubject,
+    private alertService: AlertService,
+    private router: Router,
+  ) {}
+
+  ngOnInit(): void {
+    this.notificationsSubject.subscribe((notifications) => {
+      this.filterUserNotifications(notifications);
+    });
```
sleidig, line 58: why did you decide to work with a subject rather than a simple array?

src/app/features/notification/notification.component.ts
Ayush8923: I have removed this code from this component and I will implement this code in the notification setting UI PR.

src/app/features/notification/notification.component.ts
Ayush8923: As per my understanding, I decided to use a `Subject` instead of a simple array because
1. The `Subject` enables pushing new notifications into the stream and ensures that all subscribers are updated instantly without additional logic to manually refresh the array.
2. Using a `Subject` helps centralize the flow of notifications. This ensures consistent behavior and minimizes the chances of missing updates, as all consumers rely on the same reactive data source.
3. `Subject` decouples the notification generation and consumption logic. For example, the component can subscribe to notifications and handle them as needed without directly modifying the underlying notification array.

sleidig (approved):

Issue #2743, opened 2024-12-20
Ensure New Fields from Public Forms are also saved in the Related Entity Schema


Abhinegi2:
- [x] Create a reusable service to handle this synchronization logic, avoiding duplication in both the Public Forms Admin and Entity Details Admin sections.
- [x] Implement logic to apply changes from the Entity schema to the configuration document whenever PublicFormConfig is saved.
- [x] Ensure that canceling changes in Public Form columns actively reset the Entity schema to its previous state, similar to the behavior in the AdminEntity component.

Abhinegi2:
@george-neha Could you please test the following scenarios on the [dev site](https://dev.aam-digital.net/)?

- [ ] Adding a new field to the Public Form:
 Verify that adding a new field in the Public Form automatically reflects in the corresponding entity. For example, if you add a field related to the "Child" entity in the Public Form, ensure the field appears in the "Child" entity's details view.

- [ ] Save and Cancel actions:
Test the Save and Cancel actions across the system, including entities like "Child" and "School" and dialog components (e.g., when editing existing field configurations). Confirm that these actions are functioning as expected after the recent changes.

george-neha:
@Abhinegi2 I tested the functionality of the Child entity, and it is functioning as required. However, I was unable to test the school/notes entity due to a lack of permissions.

Abhinegi2:
@george-neha I have updated the permissions. Please test them and let me know.

PR #2745, opened 2024-12-23
Ensure New Fields from Public Forms are also saved in the Related Entity Schema 

Abhinegi2:
closes: #2743

### Visible/Frontend Changes

- [x]
- [ ]

- [x]
- [ ]


src/app/core/admin/admin-entity.service.spec.ts
```diff
@@ -1,12 +1,21 @@
...
 import { AdminEntityService } from "./admin-entity.service";
+import { EntityMapperService } from "../entity/entity-mapper/entity-mapper.service";
+import { mockEntityMapper } from "../entity/entity-mapper/mock-entity-mapper-service";
 
 describe("AdminEntityService", () => {
   let service: AdminEntityService;
 
   beforeEach(() => {
-    TestBed.configureTestingModule({});
+    TestBed.configureTestingModule({
```
sleidig, line 11: can you move the test for saving the schema and view config from admin-entity.component.spec to here?

src/app/core/entity-details/form/form.component.ts
```diff
@@ -41,6 +41,8 @@ export class FormComponent<E extends Entity> implements FormConfig, OnInit {
 
   form: EntityForm<E> | undefined;
 
+  // @Output() onFormSaved = new EventEmitter<void>();
+
```
sleidig, line 45:
```suggestion
```

sleidig (changes requested):
When I change a field config (in the popup) or the field positions (drag&drop) in the Public Form Admin and then press "cancel", the changes are not saved (disappear after reload) but the UI still shows the changes after cancelling. The UI should reload or reset in this case, to immediately show users that changes have been discarded.

Otherwise seems to work well.

Issue #2723, opened 2024-12-16
Enhance Notification Settings Logic for User Entity Updates in CouchDB


Ayush8923:
**Objective**
This issue involves refining the notification settings logic and ensuring seamless synchronization with CouchDB. The following tasks are outlined:

**Proposed Solutions & Alternatives**
- [x] **Update Notification Logic:**
    - Implement logic to allow users to select any entity and define the corresponding notification conditions. These conditions should be saved and updated in the `User` entity in CouchDB.

---

- [x] **Database Schema Example**
Below is an example of the expected schema for a `NotificationConfig` entity after notification settings are applied:

```json
{
    "_id": "NotificationConfig:<user-uuid>",
    "_rev": "2-f9e14ee88f8107bd3a7bd5ec82f40fa4",
    "created": {
        "at": "2024-01-08T17:45:16.529Z",
        "by": "User:demo"
    },
    "updated": {
        "at": "2024-01-08T17:45:16.529Z",
        "by": "User:demo"
    },
    "notificationTypes": [
        {
            "entityType": "User",
            "conditions": {
                "field": "name",
                "operator": "eq",
                "value": "Anilaabh"
            },
            "isNotificationEnabled": true
        },
        {
            "entityType": "Task",
            "conditions": {
                "field": "name",
                "operator": "eq",
                "value": "Anilaabh"
            },
            "isNotificationEnabled": true
        }
    ]
}
```

---

- [x] **Push Notification Toggle Logic:**
    - Ensure that when the user enables or disables push notifications from the "Where you receive notifications" section in the notification settings page, the notificationsEnabled field in CouchDB's `User` entity is updated accordingly.

- [x] **Add/Delete Notification Rules:**
    - Implement logic to handle the addition and deletion of notification rules. When rules are added or removed, ensure the data is consistently updated in the `User` entity in CouchDB.
- [x] The Save button should be enabled only if any changes are made to the notification settings.

- [x] **Retrieve and Update UI:**
    - Fetch all notification rules associated with the user and update the UI dynamically to reflect these rules.
    
- [ ] Update the Notification rule condition UI(maybe need to add the operator field) and based on that user can set the condition accordingly. https://github.com/Aam-Digital/ndb-core/issues/2775

sleidig:
Just one remark here, @Ayush8923 : The configuration should be stored as its own Entity (maybe "NotificationConfig:<user_uuid>" or whatever name you suggest as more fitting).
[...]

PR #2752, opened 2024-12-26
[Feat]: Notification setting logic updates

Ayush8923:
closes: #2723

### Visible/Frontend Changes
- [x] Implement the logic to save the notification rules in CouchDB
- [x]  Implement the logic that when the user enables or disables push notifications from the `Where you receive notifications` section in the notification settings page, the value is updated accordingly in the CouchDB.
- [x] Retrieve the existing notification rules from the CouchDB and then show all the notification rules.


src/app/features/notification/notification-settings/notification-settings.component.ts
```diff
@@ -45,79 +48,271 @@ import { ConfirmationDialogService } from "app/core/common-components/confirmati
...
+  async appendNewNotificationRule() {
+    const newRule = this.initializeNotificationRuleFormGroup();
+    this.notificationRules.push(newRule);
+  }
+
+  async onEnableNotification(event: MatSlideToggleChange) {
+    // TODO: If the user to not allow the permission then don't need to update the value.
+    const NotificationToken = this.getNotificationToken();
+    const notificationConfig = await this.loadNotificationConfig();
+    this.isPushNotificationEnabled = event.checked;
```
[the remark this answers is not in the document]
Ayush8923: Yes, I will create a different PR for this. Once the Firebase Integration PR will merged and this notification setting logic PR and Firebase integration PR are in sync then I will implement this logic in the different PR. In this, we just need to call the notification service function and need to do a few small changes in this file.

src/app/features/notification/notification-settings/notification-settings.component.ts
```diff
@@ -45,79 +48,271 @@ import { ConfirmationDialogService } from "app/core/common-components/confirmati
   templateUrl: "./notification-settings.component.html",
   styleUrl: "./notification-settings.component.scss",
 })
-export class NotificationSettingsComponent {
-  hasPushNotificationEnabled: boolean = false;
+export class NotificationSettingsComponent implements OnInit {
```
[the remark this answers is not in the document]
Ayush8923: I have updated the code and refactored the notification setting logic component. Please take a look once and let me know if there are any changes and updates in logic.

sleidig (commented):
I did a fairly major further refactoring to simplify the layers of formControls and methods. Please have a look at the latest code and check if you find any errors or have further suggestions, @Ayush8923

Ayush8923:
Thank you, @sleidig, for refactoring the formControl. It looks much better now. However, there was a minor issue where, if the user had already chosen `push` as the notification method, it wasn't being auto-selected. To fix this, I made some final changes and removed the hardcoded values from the notification method. This way, in the future, if we need to add an `email` option to the notification methods, we can simply add a new key to the notification method without requiring any code-level changes.

I have checked the functionality, and it's working well. Could you please review the final changes once? If any further modifications are needed, please let me know. Otherwise, the PR is ready to merge.

src/app/features/notification/notification-rule/notification-rule.component.ts
```diff
@@ -89,9 +96,13 @@ export class NotificationRuleComponent implements OnChanges {
...
     [key: string]: boolean;
   }): string[] {
-    return Object.entries(channels ?? [])
+    const parsedChannels = Object.entries(channels ?? [])
       .filter(([key, value]) => value === true)
       .map(([key, value]) => key);
+
+    return parsedChannels
+      .map((index) => this.notificationMethods[index]?.key)
+      .filter((key) => key !== undefined);
```
sleidig, line 105: which case makes this mapping necessary? Don't we already assign the notificationMethods' key as the value in each mat-option?

src/app/features/notification/notification-rule/notification-rule.component.html
```diff
@@ -0,0 +1,96 @@
...
+          <mat-select
+            formControlName="channels"
+            multiple
+            placeholder="Choose notification channels"
+            i18n-placeholder="Notification Method Select placeholder"
+          >
+            @for (notificationMethod of notificationMethods; track $index) {
+              <mat-option
+                [value]="notificationMethod.key"
+                i18n="notification method option"
```
sleidig, line 53:
```suggestion
```

sleidig (commented):
Thanks for testing and improving! Unfortunately I had to remove the notifcation-method-select component to make all the properties work with formControls easily but taking the options out into a variable is a good approach

src/app/features/notification/notification-rule/notification-rule.component.ts
Ayush8923: Yes, we can assign the notificationMethods' key as the value in each mat-option. I have updated the code.

sleidig: but why do will still need that line `parsedChannels.map((index) => this.notificationMethods[index]?.key);` at all? Isn't that the same as the creating of parsedChannels above?

In any way, we should actually have some simple unit tests for this. Could you write one to check that a component.value is correctly parsed into the formControls after calling component.ngOnChanges({ value: true })  and another one that checks that the valueChange is emited (and has the correct parsed format) after one of the `form`'s formControls is being used with setValue()?

Ayush8923: The line `parsedChannels.map((index) => this.notificationMethods[index]?.key)` is needed because `parsedChannels` contains the indices where the values are true, but in `notificationMethods`, the key are in strings like as: `push`. So, first, in `parsedChannels`, we have the indices where the values are true, and then using this line, I extracted the `keys` from `notificationMethods` corresponding to the true indices and displayed them in the field.

Yes, sure. I will update the test cases for this.

sleidig: > The line `parsedChannels.map((index) => this.notificationMethods[index]?.key)` is needed because `parsedChannels` contains the indices where the values are true, but in `notificationMethods`, the key are in strings like as: `push`. So, first, in `parsedChannels`, we have the indices where the values are true, and then using this line, I extracted the `keys` from `notificationMethods` corresponding to the true indices and displayed them in the field.

no, this is not how the `MatSelect` component works and outputs values.
You can put a breakpoint or console.log into our updateValue() method to see how the form's value actually looks (it's an array of the keys of the active mat-options, i.e. `["push"]` or `[]` for us)

src/app/features/notification/notification-rule/notification-rule.component.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  it("should create", () => {
+    expect(component).toBeTruthy();
+  });
+
+  it("should parse component.value into formControls on ngOnChanges", () => {
+    component.value = mockValue;
+    component.ngOnChanges({ value: { currentValue: mockValue } } as any);
+
+    expect(component.value).toEqual(mockValue);
```
sleidig, line 48:
```suggestion
    expect(component.form.getRawValue()).toEqual({
      entityType: "entityType1",
      enabled: true,
      channels: ["push"], // expect channels value to be parsed into an array
      conditions: "some condition",
      notificationType: "entity_change",
    });
```

src/app/features/notification/notification-rule/notification-rule.component.ts
Ayush8923: I rechecked the notification method `mat-select`. The issue in this line `for (let option in options ?? [])` Now we’re using `of` instead of `in`. Now, it's working fine.
I have update the code. Please take a look once.

Issue #2756, opened 2024-12-27
Field group titles are not visible or updating correctly in the drag-and-drop UI.


Abhinegi2:
### Description:
Currently, editing, updating, or adding a field group title does not work as expected.

### Steps to Reproduce:
Navigate to the Details View of any entity.
Attempt to update or add a field group title.
Save the changes.
### Current Behavior:
After saving, the updated or newly added field group title is not visible in the UI.
### Expected Behavior:
Field group titles should be correctly updated and visible in the UI after making and saving changes.

PR #2765, opened 2025-01-03
field-group titles fixes

Abhinegi2:
closes: [#2756](https://github.com/Aam-Digital/ndb-core/issues/2756)


sleidig (approved):
Thanks for troubleshooting this. I changed the fix to properly trigger a change output instead of relying on editing the config object in place via reference. (We want to avoid any side effects and the component should not affect objects except through an Output)

PR #2767, opened 2025-01-03
[Feat]: Notification Rule Condition Logic

Ayush8923:
closes: #2775

### Visible/Frontend Changes

- Create the UI for the notification condition.
- Create the own component that handles each notification condition.
- Implement the logic to update the notification condition in CouchDB.
- If the user does not choose any entity in the notification rule then the user is not able to add the notification condition.
- The entityType field is disabled if the user adds the notification condition corresponding to the entity.


Ayush8923:
@sleidig I am currently saving the notification conditions in the CouchDB in the following format. Could you please share your suggestions on whether this format is appropriate or if any changes are required?
```
"conditions": [
  {
    "entityTypeField": "subject",
    "operator": "$eq",
    "condition": "English"
  },
  {
    "entityTypeField": "text",
    "operator": "",
    "condition": ""
  }
]
```

sleidig:
> @sleidig I am currently saving the notification conditions in the CouchDB in the following format. Could you please share your suggestions on whether this format is appropriate or if any changes are required?
> 
> ```
> "conditions": [
>   {
>     "entityTypeField": "subject",
>     "operator": "$eq",
>     "condition": "English"
>   },
>   {
>     "entityTypeField": "text",
>     "operator": "",
>     "condition": ""
>   }
> ]
> ```

As discussed previously and noted [here](https://github.com/Aam-Digital/ndb-core/issues/2723#issuecomment-2559409075) (although a bit hidden in a comment of the code sample ...), the conditions should be of type `DataFilter` (i.e. following https://www.mongodb.com/docs/manual/reference/operator/query/)

Let's have a look at this together before further work, however. If we are able to use our app filter components, these may already provide that format.

Ayush8923:
> As discussed previously and noted [here](https://github.com/Aam-Digital/ndb-core/issues/2723#issuecomment-2559409075) (although a bit hidden in a comment of the code sample ...), the conditions should be of type `DataFilter` (i.e. following https://www.mongodb.com/docs/manual/reference/operator/query/)
> 
> Let's have a look at this together before further work, however. If we are able to use our app filter components, these may already provide that format.

Yes, currently, the condition filter we are using in the notification rule is built using the reference from MongoDB query operator. If we can use the existing filter component, it would be great. In that case, we wouldn't need to create an additional condition operator.

Ayush8923:
> Deployed to https://pr-2767.aam-digital.net/

Can you please test the notification condition UI in the above-mentioned link? Please share your suggestions/thoughts on this if you have any.

george-neha:
I am currently able to add a notification rule without being prompted to specify any criteria. The notification criteria section is collapsed by default, which makes it unclear that I need to add additional criteria. Suggestion: either expand this section or prompt the user to add more criteria.

image

src/app/features/notification/model/notification-config.ts
```diff
@@ -61,7 +60,15 @@ export class NotificationRule {
...
+  @DatabaseField() conditions: NotificationCondition[];
+}
+
+export class NotificationCondition {
+  @DatabaseField() entityTypeField: string;
+
+  @DatabaseField() operator: string;
+
+  @DatabaseField() condition: string;
 }
```
sleidig, line 72: I would indeed like the "conditions" to be of type `DataFilter`, for example

```json
{
  "user": { "$eq": "my-user-id" },
  "status": { "$in": ["NEW", "TO_BE_APPROVED" ] }
}
```

This makes it identical/compatible with our other filters. Although you will need to parse it differently to display one condition component per object property then.

Ayush8923:
> @Ayush8923 I am currently able to add a notification rule without being prompted to specify any criteria. The notification criteria section is collapsed by default, which makes it unclear if I need to add additional criteria. Suggestion: either expand this section or prompt the user to add more criteria.
---

Thanks for this suggestion. I have updated this and it's deployed on the same link. Now, when the user clicks on the entity field then the corresponding accordion will expand.

Ayush8923:
@sleidig I am implementing the JSON editor component in this separate PR #2785 so that we can review and merge this notification condition, and then test the entire notification settings feature in the dev environment.

src/app/features/notification/model/notification-config.ts
Ayush8923: I will update the notification condition logic accordingly.

george-neha:
@Ayush8923 @sleidig As a User, I feel the dropdown value for "notification condition" is very confusing and forces the user in a maze to select different conditionalities, such as greater than/less than/,and/or,etc. for each sub-entity.  Instead of selecting conditionalities, multiple actions are listed that can trigger notifications. (see image below)
image
 What do you think?

src/app/features/notification/notification-rule/notification-rule.component.ts
```diff
@@ -65,24 +84,57 @@ export class NotificationRuleComponent implements OnChanges {
...
     this.form.valueChanges.subscribe((value) => this.updateValue(value));
   }
 
+  /**
+   * Disable the entityType field if there are notification conditions.
+   */
+  private updateEntityTypeControlState() {
+    this.form.get("conditions").valueChanges.subscribe(() => {
+      const conditionsLength = (this.form.get("conditions") as FormArray)
+        .length;
```
tomwwinter, line 121: This will crash, if parsing is not possible. Please catch this.

Ayush8923: Thanks for this suggestion. I have updated the code accordingly. Please check and let me know if there are any suggested changed needed.

src/app/features/notification/notification-rule/notification-rule.component.ts
```diff
@@ -124,4 +176,89 @@ export class NotificationRuleComponent implements OnChanges {
...
+   */
+  private parseConditionsObjectToArray(
+    conditions: DataFilter<any> | undefined,
+  ): NotificationRuleCondition[] {
+    if (!conditions) {
+      return [];
+    }
+
+    return Object.entries(conditions).map(([entityField, condition]) => {
+      const operator = Object.keys(condition)[0];
```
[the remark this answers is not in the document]
Ayush8923: I haven't encountered any case so far where the `condition` array would be empty, and if `conditions` is empty, it won't run here.

tomwwinter (approved):

src/app/features/notification/notification-rule/notification-rule.component.html
```diff
@@ -88,6 +83,59 @@
...
+  >
+    <div class="flex-row align-center notification-condition-section">
+      <strong i18n>Define Notification Criteria</strong>
+      <app-help-button
+        text="You can choose a notification based on the selected entity. For example, if you select 'task', you can create a notification for when a task is assigned to a user."
+        i18n-text
+      ></app-help-button>
+    </div>
+    @for (
+      notificationCondition of form.get("conditions")["controls"];
```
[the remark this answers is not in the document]
Ayush8923: Thanks for this suggestion. I have updated the code accordingly and handled this type of error. Can you please check and let me know if there are any suggested changes needed.

PR #2740, opened 2024-12-19
initial setup of public form listing

Abhinegi2:
closes #2708

PR #2771, opened 2025-01-03
refactor(core): add option to entities-table to open full details vie…

sleidig:
Add option in core components to open an entity list item in a full Details View popup as an alternative to the simplified RowDetails form popup or navigation to the details route.

required for #2740


tomwwinter (approved):
LGTM

PR #2773, opened 2025-01-06
[Feat]: Entity Field Select Component

Ayush8923:
### Visible/Frontend Changes

- [x] Entity field select component
- [x] Fetch the field corresponding to the entity


sleidig (commented):
- [ ] Let's try the same approach as the newly refactored EntityTypeSelect, as that can be used in slightly more flexible ways: https://github.com/Aam-Digital/ndb-core/pull/2772/files#diff-adf899ddfdc3f5e3362af2578421c77b244d05532b6cd86cd42f328561cc4f79
- [ ] Can you remove the other implementation of this logic and use the new component there instead (I think that was in ImportColumnMappingComponent)?

sleidig:
Apparently the issue in import column mapping component is unrelated to these changes and already a bug in master. Will be handled separately: https://github.com/Aam-Digital/ndb-core/issues/2782

sleidig (approved):

Issue #2766, opened 2025-01-03
administer "pre-filled" values for Public Forms


Abhinegi2:
... how to easily configure a "prefilled" status etc. for a form, to filter the results in the entity lists? (that's not obvious how to configure currently)

- [x] add new property "(something like prefilled)" to PublicFormConfig to store those values separate from displayed columns
- [x] alllow to add any field (not only the ones added in the "columns")
- [x] PublicFormComponent needs to add those separately stored fields as "hideFromForm" fields to the public form
- [x] what happens if the field is both in columns and in the "prefilled"?
   - --> field should be visible and show the default value
   - :question: maybe such a field has to be readonly (because the admin wants it to always have that defined default value)

Abhinegi2:
@sleidig I am thinking of the following approach -
1. Added a new select dropdown in the Public form General settings tab
- New database field added
-  will create a component to showcase all fields related to that entity.
2. Will the Selected field be saved in the new database field?
- do we need to store the field key or whole formfieldconfig?
3. In the Public form component add a check if the field is both in columns and "prefilled" Then the field should be visible and show the default value

Abhinegi2:
Hi @george-neha,

Can you please test the following functionality on the dev site: https://dev.aam-digital.net/?

1. **New Tab:** Configure Default/Pre-filled Values

We have added a new tab called Configure Default/Pre-filled Values in the Public Form. This tab is used to set hidden fields and their default values for the Public Form.

**Test Scenarios:**

- Default Value Behavior

- [ ] If a selected field is present in both the columns and prefilled configuration, the field should be visible in the Public Form and display the default value.
- [ ] If a selected field is only in the prefilled configuration, the field should not be visible in the Public Form.

- Form Submission

- [ ] After submitting the form (e.g., for a Child Public Form), a new entry should be created in the system.
- [ ] The new entry should include all the submitted data, including the prefilled values that were not visible in the Public Form.

**I have added tooltips and additional information to assist users in understanding this feature. However, if you have suggestions for better wording, phrasing, or syntax for the tooltips or instructions, please feel free to share your ideas.**

PR #2780, opened 2025-01-08
administer "pre-filled" values for Public Forms

Abhinegi2:
closes: [#2766](https://github.com/Aam-Digital/ndb-core/issues/2766)

### Visible/Frontend Changes

- [x]
- [ ]


Abhinegi2:
currently not using entity-field-select for the field dropdown as it wasn't implemented earlier. I will now integrate this new component into the dropdown.

src/app/features/public-form/public-form.module.ts
```diff
@@ -109,6 +116,26 @@ const viewConfigs: ViewConfig[] = [
...
+            {
+              component: "Form",
+              config: {
+                fieldGroups: [
+                  {
+                    fields: [
+                      {
+                        id: "restrictedPrefilled",
+                        editComponent: "EditPrefilledValuesComponent",
+                      },
```
sleidig, line 131: can we move this into the entity annotation, as for the other fields? :blush:

src/app/features/public-form/edit-prefilled-values/edit-prefilled-values.component.scss
```diff
@@ -0,0 +1,49 @@
...
+  padding: 22px;
+  background-color: aliceblue;
+}
+
+.component-info {
+  padding: 10px;
+}
+
+.new-defaultvalue-rule {
+  margin-left: 5px;
```
sleidig, line 48: let's use sizes.$small and sizes.$regular etc.?

src/app/features/public-form/public-form-config.ts
```diff
@@ -59,4 +59,10 @@ export class PublicFormConfig extends Entity {
 
   /** @deprecated use ColumnConfig directly in the columns array instead */
   @DatabaseField() prefilled: { [key in string]: any };
+
+  @DatabaseField({
+    label: $localize`:PublicFormConfig:Restricted Prefilled`,
+    isArray: true,
+  })
+  restrictedPrefilled: FieldGroup[];
```
sleidig, line 67: in what way are these "restricted"?

src/app/features/public-form/edit-prefilled-values/edit-prefilled-values.component.html
```diff
@@ -0,0 +1,60 @@
...
+  not impact other forms or configurations.
+</p>
+<div class="defaultvalue-container" [class.disabled]="formControl.disabled">
+  <form [formGroup]="prefilledValueSettings">
+    <div formArrayName="prefilledvalue">
+      <div
+        *ngFor="let fieldGroup of prefilledValues.controls; let i = index"
+        [formGroupName]="i"
+      >
+        <div class="flex-row justify-space-between align-center container">
```
sleidig, line 17: The fields have line-breaks very early, even on a fairly wide screen. I think the form field could reduce its width a bit more before switching the layout. (Maybe there is an unnecessarily large min-width on the entity-field-select?)
image

src/app/features/public-form/edit-prefilled-values/edit-prefilled-values.component.html
```diff
@@ -0,0 +1,60 @@
...
+          </div>
+          <app-default-value-options
+            [value]="fieldGroup.get('defaultValue').value"
+            (valueChange)="fieldGroup.get('defaultValue').setValue($event)"
+            [entityType]="entityConstructor"
+          ></app-default-value-options>
+
+          <button
+            mat-icon-button
+            (click)="removeRestrictedPrefilled(i)"
```
sleidig, line 39: delete doesn't work

src/app/features/public-form/edit-prefilled-values/edit-prefilled-values.component.ts
```diff
@@ -0,0 +1,183 @@
...
+})
+export class EditPrefilledValuesComponent
+  extends EditComponent<FieldGroup[]>
+  implements OnInit
+{
+  entityConstructor: EntityConstructor;
+  formConfig: FormConfig;
+  defaultValue: DefaultValueConfig;
+  prefilledValueSettings = this.fb.group({
+    prefilledvalue: this.fb.array([]),
```
sleidig, line 51:
```suggestion
    prefilledValue: this.fb.array([]),
```

src/app/features/public-form/edit-prefilled-values/edit-prefilled-values.component.ts
```diff
@@ -0,0 +1,183 @@
...
+  });
+
+  private entities = inject(EntityRegistry);
+
+  constructor(private fb: FormBuilder) {
```
sleidig, line 56: can probably do `private fb = inject(FormBuilder)` to avoid having to call super constructor?

src/app/features/public-form/edit-prefilled-values/edit-prefilled-values.component.ts
```diff
@@ -0,0 +1,183 @@
...
+    super();
+  }
+
+  override ngOnInit(): void {
+    if (!this.entity) return;
+
+    this.entityConstructor = this.entities.get(this.entity["entity"]);
+    this.formConfig = { fieldGroups: this.formControl.getRawValue() };
```
sleidig, line 64: Do we need to work with FieldGroup[] here (and in the PublicFormConfig)? Isn't this more directly just an array of FormFieldConfig s?

sleidig (changes requested):

src/app/features/public-form/public-form-config.ts
Abhinegi2: my initial thought was that users should not be able to see this field if it's not available in columnConfig, so I considered the name "restrictedPrefilled." 
However, after thinking, "conditionalPrefilled" might be a better fit. This name indicates that the field is prefilled conditionally, based on specific configurations. 
whats your thought on this?

sleidig: maybe just "prefilledFields" in a very general way, then?

src/app/features/public-form/edit-prefilled-values/edit-prefilled-values.component.ts
```diff
@@ -0,0 +1,126 @@
...
+  implements OnInit
+{
+  entityConstructor: EntityConstructor;
+  formConfig: FormConfig;
+  FormFieldConfig: FormFieldConfig;
+  defaultValue: DefaultValueConfig;
+
+  private entities = inject(EntityRegistry);
+
+  private fb = inject(FormBuilder);
```
sleidig, line 53: Currently, the tab/form can be saved even when some of the fields within this component show as invalid (e.g. required field empty). This should integrate with the EditComponent's formControl and set it invalid when necessary

src/app/features/public-form/public-form.module.ts
```diff
@@ -112,6 +119,21 @@ const viewConfigs: ViewConfig[] = [
             },
           ],
         },
+        {
+          title: $localize`:PublicFormConfig admin form panel:Configure Default/Pre-filled Fields`,
```
sleidig, line 123:
```suggestion
          title: $localize`:PublicFormConfig admin form panel:Configure Pre-filled Values`,
```

src/app/features/public-form/edit-prefilled-values/edit-prefilled-values.component.ts
```diff
@@ -0,0 +1,126 @@
...
+  removePrefilledFields(index: number): void {
+    if (index < 0 || index >= this.prefilledValues.length) {
+      return;
+    }
+
+    this.prefilledValues.removeAt(index);
+    this.formControl.markAsDirty();
+  }
+
+  private updateFieldGroups(value): void {
```
sleidig, line 113: can you add typing for the value parameter?

I am still not fully clear on your structure here. Isn't the single property "prefilledValue" somewhat irrelevant and we could just directly put objects with id+defaultValue+hideFromForm within the formGroup and array?

src/app/features/public-form/public-form.component.ts
```diff
@@ -110,9 +110,35 @@ export class PublicFormComponent<E extends Entity> implements OnInit {
...
+          hideFromForm: item.hideFromForm ?? true,
+        };
+        const fieldGroup = this.fieldGroups.find((group) =>
+          group.fields.some((field) => field === item.id),
+        );
+        if (fieldGroup) {
+          fieldGroup.fields.push(newField);
+        } else {
+          this.fieldGroups.push({ fields: [newField] });
+        }
```
sleidig, line 137: I don't think it makes a difference where we add the field.

Currently, even when finding the field is already there, you still add it a second time, just to the same group, don't you? This seems to work, although I think overwriting only the defaultValue for an existing field would be cleaner.

Please also add two unit tests:
- a) prefilled has a field that is not part of visible fields yet (should get added hidden)
- b) prefilled has a field that is also in visible fields (should take on the new defaultValue)

sleidig: I have adjusted the tests a bit for clarity (the new PublicFormConfig was not being used anyway, for example  ...)

sleidig (approved):

PR #2781, opened 2025-01-08
Feat: Conditional Filter Component

Ayush8923:
### Visible/Frontend Changes

- [x] Create the reusable conditional filter component to apply to specific conditions wherever needed.


Ayush8923:
@sleidig I have created a separate component for the conditional filter, which allows the user to choose conditional operators, and this component can be reused. I have used this component on the notification settings page. For reference: you can check this component in the [notification rule section](https://pr-2767.aam-digital.net/), after expanding the rule, when you choose to add a new condition, and then check the notification condition filter. Can you please review this and share your suggestions/thoughts on this conditional filter.

sleidig (commented):
This looks great, nice work @Ayush8923 .

As we are currently needing more and more special select components, I have been thinking if there is any easier approach. Would something like this also work for your context:

instead of own component:
```<app-conditional-filter />```

maybe we can use an improved version of basic-autocomplete:
```html
<app-basic-autocomplete
  [options]="[{ id: '$eq', toString: () => 'Equal To' }, ...]"
/>
```

Do you think this would work for your context?

sleidig:
> Yes @sleidig, this will work correctly in my case. We just need to move the `conditionMappings` variable into the component where we use the `auto-basic-autocomplete` with conditional operators.

Great. I am removing the custom component here then and you will specify the array of options in your settings component in the other branch :+1:

sleidig (approved):

PR #2785, opened 2025-01-10
[Feat]: Json Editor Component

Ayush8923:
closes: #XX

### Visible/Frontend Changes

- [x] Init the Json Editor component

### DIFF
--- a/src/app/features/notification/notification-rule/notification-rule.component.ts
+++ b/src/app/features/notification/notification-rule/notification-rule.component.ts
@@ -32,6 +32,7 @@ import { CdkAccordionItem, CdkAccordionModule } from "@angular/cdk/accordion";
 import { NotificationConditionComponent } from "./notification-condition/notification-condition.component";
 import { MatDialog } from "@angular/material/dialog";
 import { YesNoButtons } from "app/core/common-components/confirmation-dialog/confirmation-dialog/confirmation-dialog.component";
+import { JsonEditorComponent } from "../json-editor/json-editor.component";
 
 /**
  * Configure a single notification rule.
@@ -69,7 +70,7 @@ export class NotificationRuleComponent implements OnChanges {
   @Output() notificationConditionValueChange =
     new EventEmitter<NotificationRule>();
 
-  notificationConditions: string[] = []
+  notificationConditions: string[] = [];
 
   form: FormGroup;
   readonly dialog = inject(MatDialog);
@@ -161,6 +162,33 @@ export class NotificationRuleComponent implements OnChanges {
   removeCondition(conditionIndex: number) {
     (this.form.get("conditions") as FormArray).removeAt(conditionIndex);
     this.removeNotificationCondition.emit();
+  }
+
+  openConditionsInJsonEditorPopup() {
+    // TODO
+    // open MatDialog, passing in the current conditions property
+    // dialog displays a new component with json editor
+    // on dialogclose, update the conditions property with the dialogs return value
+    const buttons = YesNoButtons;
+
+    const dialogRef = this.dialog.open(JsonEditorComponent, {
+      data: {
+        title: "Notification Condition",
+        conditions: {
+          operator: "$gt",
+          condition: "",
+          entityTypeField: "date",
+        },
+        buttons: buttons,
+        closeButton: true,
+      },
+    });
+
+    dialogRef.afterClosed().subscribe((result) => {
+      if (result) {
+        console.log({ result });
+      }
+    });
   }
 
   handleToggleAccordion(notificationRuleItem: CdkAccordionItem) {
--- a/src/app/features/notification/json-editor/json-editor.component.html
+++ b/src/app/features/notification/json-editor/json-editor.component.html
@@ -1,0 +1,14 @@
+<mat-dialog-content style="width: 90vh">
+  @if (data.closeButton) {
+    <app-dialog-close mat-dialog-close></app-dialog-close>
+  }
+  <mat-form-field style="width: 80vh; margin-top: 30px">
+    <mat-label>Json Condition</mat-label>
+    <textarea matInput [value]="jsonData" rows="5"></textarea>
+  </mat-form-field>
+</mat-dialog-content>
+
+<mat-dialog-actions>
+  <button mat-stroked-button mat-dialog-close i18n>Cancel</button>
+  <button mat-raised-button mat-dialog-close i18n>Save</button>
+</mat-dialog-actions>
--- a/src/app/features/notification/json-editor/json-editor.component.spec.ts
+++ b/src/app/features/notification/json-editor/json-editor.component.spec.ts
@@ -1,0 +1,22 @@
+import { ComponentFixture, TestBed } from "@angular/core/testing";
+
+import { JsonEditorComponent } from "./json-editor.component";
+
+describe("JsonEditorComponent", () => {
+  let component: JsonEditorComponent;
+  let fixture: ComponentFixture<JsonEditorComponent>;
+
+  beforeEach(async () => {
+    await TestBed.configureTestingModule({
+      imports: [JsonEditorComponent],
+    }).compileComponents();
+
+    fixture = TestBed.createComponent(JsonEditorComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it("should create", () => {
+    expect(component).toBeTruthy();
+  });
+});
--- a/src/app/features/notification/json-editor/json-editor.component.ts
+++ b/src/app/features/notification/json-editor/json-editor.component.ts
@@ -1,0 +1,60 @@
+import { Component, EventEmitter, Inject, Input, Output } from "@angular/core";
+import {
+  MatDialogRef,
+  MAT_DIALOG_DATA,
+  MatDialogModule,
+} from "@angular/material/dialog";
+import { CommonModule } from "@angular/common";
+import { FormsModule } from "@angular/forms";
+import { DialogCloseComponent } from "app/core/common-components/dialog-close/dialog-close.component";
+import { MatButtonModule } from "@angular/material/button";
+import { MatFormField, MatLabel } from "@angular/material/form-field";
+import { MatInput } from "@angular/material/input";
+
+@Component({
+  selector: "app-json-editor",
+  standalone: true,
+  imports: [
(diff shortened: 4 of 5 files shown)


src/app/core/common-components/json-editor/json-editor.component.ts
```diff
@@ -0,0 +1,79 @@
...
+   */
+  ngAfterViewInit(): void {
+    this.jsonEditor = createJSONEditor({
+      target: this.json.nativeElement,
+      props: {
+        content: { json: this.value || {} },
+        onChange: (updatedContent: any, { contentErrors }: any) => {
+          if (contentErrors?.length === 0) {
+            this.value = updatedContent.json;
+          }
```
tomwwinter, line 53: What happen when contentErrors > 0? Should we display a notice here to the user?

tomwwinter (approved):

src/app/core/common-components/json-editor/json-editor.component.ts
Ayush8923: If `contentErrors > 0,` then something is wrong in JSON. I have made a few changes in the code to handle this error. Can you please take a look and let me know if any changes are needed?

Ayush8923:
@sleidig Can you please review this PR and let me know if there are any changes needed.

src/app/core/common-components/json-editor/json-editor.component.ts
```diff
@@ -0,0 +1,71 @@
...
+  imports: [CommonModule],
+  templateUrl: "./json-editor.component.html",
+})
+export class JsonEditorComponent {
+  /**
+   * JSON value to be edited by the user.
+   */
+  @Input() value: object = {};
+
+  @Input() height = "65vh";
```
sleidig, line 32: can you design this component so that we can set any height through css on the `<app-json-editor style="height: 65vh"` and the json editor will fill whatever height it receives? That will make use much easier than having a custom input for the formatting here.

src/app/features/notification/notification-settings/notification-settings.component.ts
```diff
@@ -116,7 +116,9 @@ export class NotificationSettingsComponent implements OnInit {
   }
 
   private async saveNotificationConfig(config: NotificationConfig) {
-    await this.entityMapper.save(config);
+    await this.entityMapper.save(config).catch((err) => {
+      Logging.error(err.message);
+    });
     this.alertService.addInfo($localize`Notification settings saved.`);
```
sleidig, line 122: No!
This tells the user "settings saved" even if an error occured while saving, doesn't it?

sleidig (commented):
Seems to work fairly well, except when invalid json is entered (I added my thoughts to the comment thread for this). Can we open the "text" instead of the "tree" view by default?

I think ideally json-editor.component would extend `CustomFormControlDirective`, so that we can use this anywhere within a mat-form-field.
Not sure if implementing that will turn out too challenging for you @Ayush8923  but maybe you can give it a try.

src/app/core/common-components/json-editor/json-editor.component.ts
Ayush8923: Thanks for this suggestion, it makes sense. I have updated it based on your suggestions. Now, the save button can only be pressed if the JSON is valid. Otherwise, the save button will remain disabled if the JSON is invalid.

**For reference:** I made some changes to the JSON to make it invalid, and the JSON editor component displayed an error indicating that the JSON is invalid and disabled the save button accordingly.

image

src/app/core/common-components/json-editor/json-editor.component.ts
Ayush8923: Yes, I have updated the styling of the JSON editor component based on your suggestion. Now, by adding the height in the style from the parent component, we can manage the height of the JSON editor component directly from the parent. Please review the updated code and let me know if any changes are needed.

Ayush8923:
> Seems to work fairly well, except when invalid json is entered (I added my thoughts to the comment thread for this). Can we open the "text" instead of the "tree" view by default?
> 
> I think ideally json-editor.component would extend `CustomFormControlDirective`, so that we can use this anywhere within a mat-form-field. Not sure if implementing that will turn out too challenging for you @Ayush8923 but maybe you can give it a try.

Yes, by default, we can open the `text` tab instead of the `tree` tab. I have also updated this in the code. Please take a look and let me know if any changes are needed.

Sure, I am trying to extend the `CustomFormControlDirective` component into this JSON editor component.

src/app/features/notification/notification-settings/notification-settings.component.ts
Ayush8923: I have updated the code where the `Notification settings saved` alert was showing even if the data wasn't saved or if there was an error during the saving process. Now, if any error occurs while saving the data, it will display the error in the log.

Ayush8923:
@sleidig I have extended the `CustomFormControlDirective` component into this JSON editor component. And also, I have made the changes according to your suggestions. Please take a look once and let me know if there are any other changes required. 

Additionally, I was trying to resolve the document conflict issue in this PR. The data is being updated correctly in the CouchDB document, and no conflicts are being shown in CouchDB. However, sometimes when saving data to CouchDB, a document conflict error is displayed.

src/app/core/common-components/json-editor/json-editor.component.ts
sleidig: :+1: 

You can remove the alert. The red error bar in the editor is very visible and should be sufficient.

src/app/core/common-components/json-editor/json-editor.component.ts
sleidig: Great :slightly_smiling_face: 

The width of the popup jumps from narrow to wide and back (when errors are displayed it becomes wide). Can you fix an good explicit width for the MatDialog (maybe some "80vw"? if it works well on both desktop and mobile)

src/app/features/notification/notification-rule/notification-condition/notification-condition-editor/notification-condition-editor.component.html
```diff
@@ -0,0 +1,26 @@
...
+      </button>
+      <button
+        [disabled]="!isValidJson"
+        mat-raised-button
+        (click)="onJsonValueSave()"
+        i18n
+      >
+        Save
+      </button>
+    </div>
```
sleidig, line 24: we generally have "Save" first, "Cancel" second in order

src/app/features/notification/notification-rule/notification-condition/notification-condition-editor/notification-condition-editor.component.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+      MatDialogRef<NotificationConditionEditorComponent>
+    >;
+    fixture.detectChanges();
+  });
+
+  it("should create", () => {
+    expect(component).toBeTruthy();
+  });
+
+  it("should close dialog when json is changed", () => {
```
sleidig, line 39: really? This doesn't seem to happen (I have to click save, not only change the json - which is good this way, I think). Or do I misunderstand this test title?

src/app/features/notification/notification-rule/notification-condition/notification-condition-editor/notification-condition-editor.component.ts
```diff
@@ -0,0 +1,65 @@
...
+    JsonEditorComponent,
+    MatDialogModule,
+    DialogCloseComponent,
+    MatButtonModule,
+  ],
+  schemas: [CUSTOM_ELEMENTS_SCHEMA],
+  templateUrl: "./notification-condition-editor.component.html",
+  styleUrl: "./notification-condition-editor.component.scss",
+})
+export class NotificationConditionEditorComponent {
```
sleidig, line 24: Let's rename "JsonEditorDialogComponent" and move this to core. This can be reused for any kind of json editing

sleidig: Can you extract this into its separate branch directly based into "master", also :blush: ?

Ayush8923: Sure, I have updated the component name according to your suggestion. 
Also, I have created the new [PR](https://github.com/Aam-Digital/ndb-core/pull/2798) for this which is based on `master`.

src/app/features/notification/notification-rule/notification-condition/notification-condition-editor/notification-condition-editor.component.html
Ayush8923: I have updated it and made the changes accordingly.

src/app/core/common-components/json-editor/json-editor.component.ts
Ayush8923: I have fixed this, and added the width for the `MatDialog` which works well on both desktop and mobile. Can you please check and let me know if any further changes are needed?

sleidig: works well now, thanks!

src/app/features/notification/notification-rule/notification-rule.component.ts
```diff
@@ -190,17 +191,57 @@ export class NotificationRuleComponent implements OnChanges {
...
+    const conditionsFormArray = this.form.get("conditions") as FormArray;
+    conditionsFormArray.clear();
+    parsedObject.forEach((condition) => {
+      const conditionGroup = new FormGroup({
+        entityTypeField: new FormControl(condition.entityTypeField),
+        operator: new FormControl(condition.operator),
+        condition: new FormControl(condition.condition),
+      });
+      conditionsFormArray.push(conditionGroup);
+    });
```
sleidig, line 238: shouldn't this be more or less the same as the initForm() method that we call onChanges?

sleidig (approved):
JSON Editor seems to work well now :+1:

src/app/features/notification/notification-rule/notification-rule.component.ts
Ayush8923: I have updated the code and reused the init form control. Please take a look on the updated changes and let me know if there are any further changes needed.

sleidig: Thanks!

Issue #2754, opened 2024-12-27
click on Notification entry navigates to the relevant view (e.g. details of the changed entity)


sleidig:
The basic functionality is already implemented in PR #2738.
To make this less error-prone, we should maybe dynamically generate the actionUrl in the NotificationEvent class based on the notificationType , e.g. for entity_change the link can be generated by looking up the current route for that entityType instead of hard-coding the link into each NotificationEvent document (any maybe leading into nothing if the admin changes the route for that entity type).

PR #2786, opened 2025-01-14
[Feat]: Notification entry navigates to the relevant view 

Ayush8923:
closes: #2754

### Visible/Frontend Changes

- [x] Implement the logic to after clicking on the notification user will be redirected to the relevant view.

### DIFF
--- a/src/app/features/notification/model/notification-event.ts
+++ b/src/app/features/notification/model/notification-event.ts
@@ -9,12 +9,43 @@ import { NotificationType } from "./notification-config";
  */
 @DatabaseEntity("NotificationEvent")
 export class NotificationEvent extends Entity {
+  /*
+   * The title of the notification.
+   */
   @DatabaseField() title: string;
+
+  /*
+   * The body of the notification.
+   */
   @DatabaseField() body: string;
+
+  /*
+   * The URL to redirect the user to when the notification is clicked.
+   */
   @DatabaseField() actionURL: string;
+
+  /*
+   * The user ID for whom the notification is intended
+   */
   @DatabaseField() notificationFor: string;
+
+  /*
+   * The notification token to be used for the notification.
+   */
   @DatabaseField() notificationToken: string;
+
+  /*
+   * The type of notification.
+   */
   @DatabaseField() notificationType: NotificationType;
+
+  /*
+   * The entity type of the notification.
+   */
   @DatabaseField() entityType: string;
+
+  /*
+   * The status of the notification.
+   */
   @DatabaseField() readStatus: boolean;
 }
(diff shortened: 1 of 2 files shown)


Ayush8923:
@tomwwinter I have implemented logic for notification entries to navigate to the relevant view. Currently, we save the endpoint in the `actionUrl` for each notification event. To make this dynamic, we need to save `notificationType` and `entityType` in the notification event. This way, when a user clicks on any notification, the system can determine the corresponding action URL based on the entityType and redirect the user to the appropriate page. Can you please take a look at this logic and share your suggestions/inputs on this.

src/app/features/notification/model/notification-event.ts
```diff
@@ -1,17 +1,51 @@
...
+
+  /*
+   * The user ID for whom the notification is intended
+   */
   @DatabaseField() notificationFor: string;
+
+  /*
+   * The notification token to be used for the notification.
+   */
   @DatabaseField() notificationToken: string;
```
tomwwinter, line 35: What is this used for? What kind of token is this?

tomwwinter (approved):

Ayush8923:
@sleidig Currently, I have implemented the logic for a dynamic URL in this way (you can check it in the file changes of this PR). It is working correctly in most cases, but for some entities, it is not generating the action URL because we are not getting the `entitytype` inside the config. Please review the logic and share your suggestions so that I can update the code accordingly.

src/app/features/notification/notification.component.ts
```diff
@@ -135,15 +135,39 @@ export class NotificationComponent implements OnInit {
...
+    );
+
+    switch (notification.notificationType) {
+      case "entity_change":
+        if (entityRoute) {
+          actionURL = `/${entityRoute.path}`;
+        }
+        break;
+      default:
+        break;
```
sleidig, line 155: default: return notification.actionUrl?

src/app/features/notification/notification.component.ts
```diff
@@ -135,15 +135,39 @@ export class NotificationComponent implements OnInit {
...
 
+  private generateNotificationActionURL(
+    notification: NotificationEvent,
+  ): string {
+    const routes = this.router.config;
+    let actionURL = "";
+
+    const entityRoute = routes.find(
+      (route) => route.data?.config?.entityType === notification.entityType,
+    );
```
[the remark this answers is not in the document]
sleidig: I think we do not have the entity ID of the specific document that triggered this event stored in the NotificationEvent yet, however? (e.g. the _id of the Note entity that was created and for which the user is now notified) 
Maybe we need to add this as an additional property (not sure what to name that in a generic way ...), @tomwwinter ?

Ayush8923:
Here is the new draft [PR](https://github.com/Aam-Digital/ndb-core/pull/2796) that updates the logic notification event to redirect to the relevant view.

PR #2790, opened 2025-01-17
easier way to copy the link to the form (add some button?)

Abhinegi2:
closes: #2774 
### Visible/Frontend Changes

- [x]
- [ ]

### DIFF
--- a/src/app/features/public-form/edit-publicform-route/edit-publicform-route.component.html
+++ b/src/app/features/public-form/edit-publicform-route/edit-publicform-route.component.html
@@ -1,1 +1,7 @@
-<p>edit-publicform-route works!</p>
+<form [formGroup]="form" [class.disabled]="formControl.disabled">
+  <mat-form-field appearance="outline">
+    <mat-label>Form Link ID</mat-label>
+    <span class="prefix-text" matTextPrefix>{{ getPrefixValue() }}&nbsp;</span>
+    <input matInput formControlName="route" placeholder="Enter Form Link ID" />
+  </mat-form-field>
+</form>
--- a/src/app/features/public-form/edit-publicform-route/edit-publicform-route.component.html
+++ b/src/app/features/public-form/edit-publicform-route/edit-publicform-route.component.html
@@ -1,44 +1,42 @@
-<form [formGroup]="form">
-  <div class="flex-row">
-    <mat-form-field
-      appearance="outline"
-      [class.disabled]="formControl.disabled"
-      floatLabel="always"
+<div class="flex-row">
+  <mat-form-field
+    appearance="outline"
+    [class.disabled]="formControl.disabled"
+    floatLabel="always"
+  >
+    <mat-label>
+      <span i18n>Form Link ID </span>
+      <fa-icon
+        icon="question-circle"
+        matTooltip="The identifier that is part of the link (URL) through which users can access this form (e.g. demo.aam-digital.com/public-form/MY_FORM_LINK_ID)."
+        i18n-matTooltip
+      ></fa-icon>
+    </mat-label>
+    <span
+      i18n
+      matTextPrefix
+      [ngClass]="{ 'prefix-text': !formControl.disabled }"
     >
-      <mat-label>
-        <span i18n>Form Link ID </span>
-        <fa-icon
-          icon="question-circle"
-          matTooltip="The identifier that is part of the link (URL) through which users can access this form (e.g. demo.aam-digital.com/public-form/MY_FORM_LINK_ID)."
-          i18n-matTooltip
-        ></fa-icon>
-      </mat-label>
-      <span
-        i18n
-        matTextPrefix
-        [ngClass]="{ 'prefix-text': !formControl.disabled }"
-      >
-        {{ getPrefixValue() }}&nbsp;
-      </span>
-      <input
-        matInput
-        formControlName="route"
-        placeholder="Enter Form Link ID"
-      />
-      @if (formControl.invalid && formControl.touched) {
-        <mat-error>
-          <app-error-hint [form]="formControl"></app-error-hint>
-        </mat-error>
-      }
-    </mat-form-field>
-    <button
-      mat-icon-button
-      matTextSuffix
-      matTooltip="Copy the link to share or paste it somewhere else"
-      i18n-matTooltip
-      (click)="copyToClipboard()"
-    >
-      <fa-icon [icon]="['far', 'copy']" size="2x"></fa-icon>
-    </button>
-  </div>
-</form>
+      {{ getPrefixValue() }}&nbsp;
+    </span>
+    <input
+      matInput
+      [formControl]="formControl"
+      placeholder="Enter Form Link ID"
+    />
+    @if (formControl.invalid && formControl.touched) {
+      <mat-error>
+        <app-error-hint [form]="formControl"></app-error-hint>
+      </mat-error>
+    }
+  </mat-form-field>
+  <button
+    mat-icon-button
+    matTextSuffix
+    matTooltip="Copy the link to share or paste it somewhere else"
+    i18n-matTooltip
+    (click)="copyToClipboard()"
+  >
+    <fa-icon [icon]="['far', 'copy']" size="2x"></fa-icon>
+  </button>
+</div>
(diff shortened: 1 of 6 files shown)


Abhinegi2:
> Deployed to https://pr-2790.aam-digital.net/

@george-neha  Could you test the current functionality for copying the public form link? 
Alternatively(another approach), I can add a "Copy" button between the question icon and the input field for easier access?

george-neha:
@Abhinegi2  A Few Suggestions: 
 1. Can the font color for the highlighted part(below image) be grey, like the heading for the rest of the screen?  It would give the impression to the user that it's non-editable. 
image
2. The screen is very right heavy due to the text. Can the "Description" field be moved below the "Title" field?

Abhinegi2:
@george-neha I have implemented both suggested changes have a look now.

src/app/features/public-form/public-form.module.ts
```diff
@@ -76,7 +83,14 @@ const viewConfigs: ViewConfig[] = [
               config: {
                 fieldGroups: [
                   {
-                    fields: ["route", "title"],
+                    fields: [
+                      {
+                        id: "route",
+                        editComponent: "EditPublicformRouteComponent",
```
sleidig, line 89: does this also work if you add the editComponent to the PublicFormConfig entity class instead? Then everything is in one place

src/app/features/public-form/edit-publicform-route/edit-publicform-route.component.html
```diff
@@ -0,0 +1,32 @@
+<form [formGroup]="form">
+  <div class="flex-row">
+    <mat-form-field
+      appearance="outline"
+      [class.disabled]="formControl.disabled"
+      floatLabel="always"
+    >
+      <mat-label i18n>Form Link ID</mat-label>
```
sleidig, line 8: one thought about UX complexity:

We could show the tooltip explanation which is currently a big "?" button on the right as a small tooltip with the label here. Like in this example:
image

Then the component would have only one big icon on the right and is therefore maybe less overwhelming?

What do you think, @george-neha , @Abhinegi2 ?

sleidig (commented):
Love how this looks and feels now :heart_eyes:

src/app/features/public-form/public-form.module.ts
Abhinegi2: I already tried but this is not working when adding it to entity class

src/app/features/public-form/edit-publicform-route/edit-publicform-route.component.html
Abhinegi2: I think moving the tooltip explanation makes the interface cleaner.

src/app/features/public-form/public-form.module.ts
sleidig: It seems to work for me, if I move it to PublicFormConfig.

Could it be that your local database in the Config:... somehow contains a custom "entity:PublicFormConfig" definition and therefore the code change does not take effect?

src/app/features/public-form/public-form.module.ts
```diff
@@ -47,6 +47,13 @@ const dynamicComponents: [string, AsyncComponent][] = [
         "app/features/public-form/edit-prefilled-values/edit-prefilled-values.component"
       ).then((c) => c.EditPrefilledValuesComponent),
   ],
+  [
+    "EditPublicformRouteComponent",
```
sleidig, line 51: We have registered these without the "Component" suffix --> "EditPublicFormRoute"

I would also tend to capitalize it "PublicForm" (although I think I understand your idea to group the naming)

src/app/features/public-form/edit-publicform-route/edit-publicform-route.component.ts
```diff
@@ -0,0 +1,74 @@
...
+  ],
+  templateUrl: "./edit-publicform-route.component.html",
+  styleUrls: ["./edit-publicform-route.component.scss"],
+})
+export class EditPublicformRouteComponent
+  extends EditComponent<string>
+  implements OnInit
+{
+  form: FormGroup;
+  constructor(
```
sleidig, line 40:
```suggestion

  constructor(
```

src/app/features/public-form/edit-publicform-route/edit-publicform-route.component.ts
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  override ngOnInit(): void {
+    const publicFormConfig: PublicFormConfig = {
+      route: this.formControl.getRawValue(),
+    } as Partial<PublicFormConfig> as PublicFormConfig;
+
+    this.form = this.fb.group({
+      route: [publicFormConfig.route, Validators.required],
+    });
```
sleidig, line 54: Do we need this separate form to wrap around the single form-control? Could we work with `this.formControl` provided trhough the EditComponent base class directly? (e.g. `this.formControl.addValidators`, this.formControl.valueChanges, etc.)

src/app/features/public-form/public-form.module.ts
Abhinegi2: thanks, Sebastian, I checked my local database and found Config:CONFIG_ENTITY still has old definitions related to the route field so I updated that, and it's working for me too.

src/app/features/public-form/edit-publicform-route/edit-publicform-route.component.ts
Abhinegi2: updated, now using EditComponent formcontrol to set validators and the route value, and seems working as expected

src/app/features/public-form/edit-publicform-route/edit-publicform-route.component.ts
```diff
@@ -0,0 +1,59 @@
...
+import { CommonModule } from "@angular/common";
+import { MatFormFieldModule } from "@angular/material/form-field";
+import { MatInputModule } from "@angular/material/input";
+import { FontAwesomeModule } from "@fortawesome/angular-fontawesome";
+import { MatIconButton } from "@angular/material/button";
+import { MatTooltipModule } from "@angular/material/tooltip";
+import { AlertService } from "app/core/alerts/alert.service";
+import { ErrorHintComponent } from "app/core/common-components/error-hint/error-hint.component";
+
+@Component({
```
sleidig, line 14:
```suggestion
/**
 * Special Form Field to edit an ID and copy the full public-form URL generated based on this.
 */
@Component({
```

sleidig (changes requested):
Code looks good now :+1: 

Just one UI glitch I stumbled upon:
This line break (and not using the full with of field) is not ideal
Screenshot From 2025-01-22 15-34-04

src/app/features/public-form/edit-publicform-route/edit-publicform-route.component.scss
```diff
@@ -1,7 +1,15 @@
...
+}
 .disabled {
   opacity: 0.5;
   pointer-events: none;
 }
 .prefix-text {
+  display: flex;
+  width: max-content;
+}
+.prefix-text-disabled {
```
sleidig, line 13:
```suggestion
}

.disabled {
  opacity: 0.5;
  pointer-events: none;
}

.prefix-text {
  display: flex;
  width: max-content;
}

.prefix-text-disabled {
```

src/app/features/public-form/edit-publicform-route/edit-publicform-route.component.html
```diff
@@ -15,7 +15,8 @@
     <span
       i18n
       matTextPrefix
-      [ngClass]="{ 'prefix-text': !formControl.disabled }"
+      class="prefix-text"
+      [ngClass]="{ 'prefix-text-disabled': !formControl.disabled }"
```
sleidig, line 19: Now it's too wide :see_no_evil: 
image

sleidig (changes requested):

src/app/features/public-form/edit-publicform-route/edit-publicform-route.component.html
Abhinegi2: fixed this please have a look now

sleidig: this can still be wider than its parent:

image


Please test it thoroughly with a continuous change of screen widths

PR #2796, opened 2025-01-21
Update the Notification event to relevent view

Ayush8923:
closes: https://github.com/Aam-Digital/ndb-core/issues/2754

### Visible/Frontend Changes

- [x] Update the notification redirection logic.


sleidig (approved):
@Ayush8923 , @tomwwinter  I have generalized the NotificationEvent interface to give us more flexibility, identifying the exact entity that triggered an event (if applicable)

Issue #2721, opened 2024-12-13
Public Form should show a "successfully submitted" page


sleidig:
After submitting a Public Form, the user can see a small toast message at the bottom of the screen but also immediately gets a new, empty form again. The message is easy to overlook and people may get confused if the form was really saved. It will be better to display a full screen "Your data was successfully submitted" page (with a link to "submit another form" that re-opens the empty form view)

Abhinegi2:
@george-neha  and @sleidig After successfully submitting a Public Form, the following page will appear. Please take a look and share any suggestions and alternative approaches if you have- 

<img width="1512" alt="image" src="https://github.com/user-attachments/assets/369bcceb-7871-444f-9245-ba8971691d75" />

Abhinegi2:
@george-neha you can test the above-mentioned functionality here -https://pr-2797.aam-digital.net/
Also let me know if you have any siuggestions

PR #2797, opened 2025-01-21
added successfully submitted component

Abhinegi2:
closes: [#2721](https://github.com/Aam-Digital/ndb-core/issues/2721)


src/app/core/config/dynamic-routing/submission-success/submission-success.component.html
```diff
@@ -0,0 +1,16 @@
...
+    [icon]="['fas', 'circle-check']"
+    class="margin-bottom-regular success-icon"
+    size="6x"
+  ></fa-icon>
+  <h1 i18n>Your data was successfully submitted!</h1>
+  <p i18n>Thank you for your submission.</p>
+  <button
+    mat-stroked-button
+    class="margin-top-regular"
+    (click)="submitAnotherForm()"
```
sleidig, line 12:
```suggestion
    (click)="submitAnotherForm()"
    i18n
```

src/app/core/config/dynamic-routing/submission-success/submission-success.component.html
```diff
@@ -0,0 +1,16 @@
+<div class="success-container">
```
sleidig, line 1: Let's move this component to `features/public-form`. I think this is specific to that module only.

src/app/app.routing.ts
```diff
@@ -51,6 +52,7 @@ export const allRoutes: Routes = [
         (c) => c.PublicFormComponent,
       ),
   },
+  { path: "submission-success", component: SubmissionSuccessComponent },
```
sleidig, line 55: Could you extract the newly added and the route above (public-form/:id) into a separate Module routes structure, similar to how we do it for AdminModule routes below?

That would nicely encapsulate all related routes into its module.

sleidig (commented):
Looks and works well for me :+1:

src/app/features/public-form/public-form.component.spec.ts
```diff
@@ -83,19 +83,20 @@ describe("PublicFormComponent", () => {
...
+    const navigateSpy = spyOn(TestBed.inject(Router), "navigate");
     saveSpy.and.resolveTo();
     component.form.formGroup.get("name").setValue("some name");
 
     component.submit();
 
     expect(saveSpy).toHaveBeenCalledWith(component.form, component.entity);
     tick();
-    expect(openSnackbarSpy).toHaveBeenCalled();
+    expect(navigateSpy).toHaveBeenCalledWith(["/submission-success"]);
```
sleidig, line 99:
```suggestion
    expect(navigateSpy).toHaveBeenCalledWith(["/public-form/submission-success"]);
```
?

src/app/features/public-form/public-form-routing.ts
```diff
@@ -0,0 +1,11 @@
+import { Routes } from "@angular/router";
+import { SubmissionSuccessComponent } from "./submission-success/submission-success.component";
+import { PublicFormComponent } from "./public-form.component";
+
+export const publicFormRoutes: Routes = [
+  {
+    path: "public-form/:id",
```
sleidig, line 7:
```suggestion
    path: "form/:id", // added as subroute resulting in /public-form/form/:id
```

I think route changed because child-routes already have a prefix of public-form/ from the app.routing? But somehow both /public-form and public-form/public-form works ... Any idea why, @Abhinegi2 ?

Abhinegi2: I checked and found that both /public-form and /public-form/public-form are working 🙈. 
In the PublicFormComponent, we are searching for parameters(id), and the id we get in both cases is the same (e.g., "publicformid"). 
This happens because the /public-form route is already defined in-app.routing, so the page exists regardless of the redundant prefix. That could be why both routes work.

sleidig: now works as expected, thanks!

Don't actually quite understand why not before, but whatever :sweat_smile:

sleidig (approved):

PR #2798, opened 2025-01-22
[Feat]: JSON Editor Component with Dialog Integration

Ayush8923:
### Visible/Frontend Changes

- [x] Crete the JSon editor component 
- [x] Create the Mat Dialog with JSON Editor


src/app/core/common-components/json-editor/json-editor.component.ts
```diff
@@ -0,0 +1,111 @@
...
+  standalone: true,
+  imports: [CommonModule],
+  templateUrl: "./json-editor.component.html",
+  styleUrl: "./json-editor.component.scss",
+  providers: [
+    { provide: MatFormFieldControl, useExisting: JsonEditorComponent },
+  ],
+})
+export class JsonEditorComponent extends CustomFormControlDirective<object> {
+  @Output() isValidChange = new EventEmitter<boolean>();
```
sleidig, line 32: Ideally we would mark the FormControl as "invalid", rather than introducing our own isValidChange output ... But I am not sure myself how to implement that in a nice way within this JsonEditorComponent.

If you, @Ayush8923 , maybe with any of your seniors, have an idea for it. Let's try that. Otherwise, we'll stick with the current version, which works alright.

Ayush8923: @sleidig I have updated the code to handle everything using the formControl. Additionally, if the JSON is invalid, the corresponding `formControl` will be marked as invalid, and the `Save` button will be disabled. Please review the changes and let me know if any further adjustments are required.

You can test the functionality in this PR deployment link: https://pr-2785.aam-digital.net/

Also, if in this Notification JSON editor, [PR](https://github.com/Aam-Digital/ndb-core/pull/2785) changes look good then can we merge these changes on the final notification branch because I made some cleanups in this PR?

src/app/core/json-editor-dialog/json-editor-dialog.component.spec.ts
```diff
@@ -31,45 +31,17 @@ describe("JsonEditorDialogComponent", () => {
...
-    component.jsonData = jsonData;
+  it("should save the updated JSON value when the form is valid", () => {
+    component.jsonDataControl.setValue({ key: "updated value" });
     component.onJsonValueSave();
-    expect(dialogRef.close).toHaveBeenCalledWith(jsonData);
+    expect(dialogRef.close).toHaveBeenCalledWith({ key: "updated value" });
   });
 
-  it("should close dialog with null when onJsonValueCancel is called", () => {
+  it("should close dialog with null", () => {
```
sleidig, line 44:
```suggestion
  it("should close dialog with null when user cancels", () => {
```

src/app/core/common-components/json-editor/json-editor.component.ts
```diff
@@ -29,12 +32,11 @@ import { MatFormFieldControl } from "@angular/material/form-field";
   ],
 })
 export class JsonEditorComponent extends CustomFormControlDirective<object> {
-  @Output() isValidChange = new EventEmitter<boolean>();
-
   @ViewChild("json", { static: true }) json!: ElementRef<HTMLDivElement>;
 
+  @Input() formControl!: FormControl;
```
sleidig, line 37: The `CustomFormControlDirective` already has some logic to integrate with a formControl being passed. I think this should be available through `this.ngControl.control`.

To not interfere with other logic that set errors, I would also recommend that we try to integrate the json-valid check with CustomFormControlDirective's flow, see
 https://github.com/Aam-Digital/ndb-core/blob/0a23155a456ad87f128f7d2924c5432e41a661aa/src/app/core/common-components/basic-autocomplete/custom-form-control.directive.ts#L166

i.e. could you try in the constructor here to create a custom errorStateMatcher that makes an "or" combination of the this.errorStateMatcher.isErrorState() and your json-invalid check?

src/app/core/json-editor-dialog/json-editor-dialog.component.ts
```diff
@@ -0,0 +1,65 @@
...
+    JsonEditorComponent,
+    MatDialogModule,
+    DialogCloseComponent,
+    MatButtonModule,
+  ],
+  schemas: [CUSTOM_ELEMENTS_SCHEMA],
+  templateUrl: "./json-editor-dialog.component.html",
+  styleUrl: "./json-editor-dialog.component.scss",
+})
+export class JsonEditorDialogComponent {
```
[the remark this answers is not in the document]
Ayush8923: I have moved both components as per the suggested directory location.

src/app/core/common-components/json-editor/json-editor.component.ts
sleidig: Please check the new implementation and try to break it, @Ayush8923  :sweat_smile:     I think this could work

src/app/core/common-components/json-editor/json-editor.component.ts
sleidig: Okay, I have investigated the cause a bit further still:

Previously, you had defined `@Input() formControl` in JsonEditorComponent. This made it just a simple Input, unrelated to the ReactiveForms module of Angular and the special CustomFormControl's injected NgControl .

Now, we have removed the Input . `<app-json-editor [formControl]="jsonDataControl"` would now actually complain that there is no "formControl" input. But as we import the ReactiveFormsModule here (in the JsonEditorDialog), then this "formControl" directive becomes available from the ReactiveFormsModule and links into angular's system ... which in turn makes it available as NgControl through the dependency injection

sleidig (approved):
Works well now :smiley:

PR #2800, opened 2025-01-22
Fix: Update `isPushNotificationEnabled` Value Based on User Notification Preferences

Ayush8923:
### Visible/Frontend Changes
- [x] Update the `isPushNotificationEnabled` value if the user enables/diable the notification.


Ayush8923:
The changes were lost when we merged the notification rule PR with the final notification branch.

sleidig:
Covered now also as part of PR #2803, I think

PR #2803, opened 2025-01-23
[Fix]: Notifications: Internal testing updates

Ayush8923:
### Visible/Frontend Changes

- [x] Adjust the alignment of the test notification icon.
- [x] Modify the alignment of the enable/disable toggle button.
- [x] Update the alignment for the push notification method.
- [x] Update the alert message is shown when the user has not enabled notification permissions or has completely denied them.
- [x]  Implement a function to check whether the user has granted notification permissions to the browser, and based on that, enable/disable the push notification toggle.
- [x] Sort the notification list with the latest notifications at the top, followed by the older ones.
- [ ] disable the push toggle if there was an error registering
- [x] disable / communicate test notification if push is disabled


Ayush8923:
@george-neha As per our discussion yesterday, I have updated the UI accordingly. Can you please test this and please let me know if there are any further changes needed and if I missed something.

george-neha:
1. Can the tooltip here be displayed above the text?
image
2. The alert message is incomplete; rephrase it to " Aam Digital wants to send you notifications. 
image
3. Instead of an icon for "Test Notification," can we make it a button "Test/Sample Notification" 
image

4. Instead of " Could not register device in aam-digital backend. Enable the notification permission to receive push notifications.", rephrase the message to "Please enable notification permissions to receive important updates."

Ayush8923:
Thanks for these suggestions @george-neha. I made some updates according to your suggestion.

1.  I have updated the tooltip position, Now when the user hovers over the icon then the tooltips are shown above the text.

2. The behaviour of the permission popup is influenced by the domain prefix (e.g., dev or prod), so the initial text updates automatically based on the deployment environment. Since this is the default browser permission popup, I don't think we can change this but let me check this once.
Additionally, if the user opens the same link in another browser where notification permission hasn't been granted, they'll need to re-enable the push notification option.

3. For me, I think adding an icon here would be more appropriate and will keep the UI less cluttered. Otherwise, the UI will get overloaded with buttons like `add new condition`, JSON editor, test notification, etc. What do you think, @sleidig on this?

4. I have updated the test according to your suggestion. Please check.

sleidig:
For the error message, just to double check: Is this only shown when browser permissions are missing or also if a different error occured (the first sentence of the original message, "Could not register device in aam-digital backend.", seems to suggest a general problem).
We have to distinguish between the two cases and let the user know if the problem is with the server or other tech or if indeed they can fix the problem by granting permissions.

sleidig:
For the "Test Notification": I think the action is not common and users will not expect or intuitively understand the icon only. So in this case I would also prefer to give a text button "Test Notification".

Is the test notification in any way related to one specific rule? I think this is more a general demo of Push notification, right? In that case the test button should actually be in the bottom section, where the user can enable push notifications in general, I think.

Ayush8923:
> For the error message, just to double check: Is this only shown when browser permissions are missing or also if a different error occured (the first sentence of the original message, "Could not register device in aam-digital backend.", seems to suggest a general problem). We have to distinguish between the two cases and let the user know if the problem is with the server or other tech or if indeed they can fix the problem by granting permissions.

I double-checked, and the error message only appears when the user is prompted for notification permission and denies it. In such cases, the notification token doesn't get registered on the backend, resulting in a 400 error code from the backend and then this triggers the `Please enable the notification permission` alert to be displayed.

sleidig:
> I double-checked, and the error message only appears when the user is prompted for notification permission and denies it. In such cases, the notification token doesn't get registered on the backend, resulting in a 400 error code from the backend and then this triggers the `Please enable the notification permission` alert to be displayed.

Sorry @Ayush8923 , but what you are saying does not seem to be correct:
Please test in this PR branch https://pr-2803.aam-digital.net/ to enable push, allow in the browser and observe that (probably because the backend is not enabled here for the PR deployment) another error occurs.

Ayush8923:
> Sorry @Ayush8923 , but what you are saying does not seem to be correct: Please test in this PR branch https://pr-2803.aam-digital.net/ to enable push, allow in the browser and observe that (probably because the backend is not enabled here for the PR deployment) another error occurs.

Thanks, @sleidig for the help. I'm able to catch the error. I made a few changes to show this relevant alert message to the user. Let me check these changes on the PR deployment link.

george-neha:
Suggestions for the text below 
1. **Notification Settings :** "  Notifications alert you to important system events, such as new registrations via public forms or tasks assigned to you. They appear in the toolbar and can be sent via email or push notifications" 
2. **Notification Center (in app)**- "Click the bell icon in the top toolbar to access notifications anytime. The notification center shows all messages, including read ones, and cannot be disabled."
What do you think @Ayush8923 @sleidig ?

Ayush8923:
@george-neha I have updated the content in the notification settings as you suggested, please take a look once.

sleidig (approved):

PR #2808, opened 2025-01-23
fix(Admin UI): keep the focus when changing field group titles

### LINKED ISSUE #2805: Admin FieldGroup titles loses focus after typing one character only
**Describe the bug**
When editing FieldGroup titles, the field loses focus after typing only one character. After that I have to click into the title field again to continue typing.

**To Reproduce**

1. Go to Admin Entity Details (e.g. through "Children List" -> "Edit data structures" -> "Details View")
2. Click on a title for a group of fields and start typing (see screenshot)

Image

sleidig:
fixes #2805


Abhinegi2 (approved):
Fixes are working as expected in both the EntityDetails view and the PublicForm's 'Configure Fields' section.

PR #2818, opened 2025-01-28
Fix circular dependencies

sleidig:
run `npm run check-circular-deps`
and go through each of the detected issues:

1. Look at the imports of the files to understand what is imported where ...
2. if there is some export (a class, interface, variable, etc.) that can be moved to a new file, fix the circular dependency by creating a new file, moving it there and importing that new file in the affected files (so that they only depend on the third file but not on each other)
3. if it is not clear how the imports can be separated and fixed, write down here what you found
  - the files affected
  - what they import from each other
  - briefly describe why it is difficult to separate it 
4. make a git commit (for 2. above) or a github comment (for 3. above) for the circular dependency issue you worked on and then move on to the next line of the "npm run check-circular-deps"


sleidig:
@sadaf895 , some more background on circular dependencies (do some further research on your own also):
- https://dev.to/tahsinature/tired-of-circular-dependency-in-typescriptnodejs-4i0a
- https://medium.com/visual-development/how-to-fix-nasty-circular-dependency-issues-once-and-for-all-in-javascript-typescript-a04c987cf0de
- https://www.reddit.com/r/typescript/comments/ra4wah/psa_dont_make_the_same_mistake_i_did_avoid/?rdt=38099
- and maybe ask GitHub co-pilot chat for an explanation, including your follow up questions :blush:

src/app/core/basic-datatypes/configurable-enum/configurable-enum.interface.ts
```diff
@@ -1,47 +1,10 @@
...
-   */
-  isInvalidOption?: boolean;
-
-  /**
-   * optional styling class that should be applied when displaying this value
-   */
-  style?: string;
-}
-
 export const EMPTY: ConfigurableEnumValue = {
```
sleidig, line 7: I think now we have 2x EMPTY (here and in configurable-enum-types.ts)?

Looking at this, I am wondering if we should just make a single file with all those things instead of three separate files ... I really like that previously we tried to separate the special ordering code but it seems to be very complicated to split it without having circular deps.
What do you think, @sadaf895 ?

src/app/features/location/GeoLocation.ts
sleidig, line 1: our naming convention for filenames is to use "kebab-case", i.e. geo-location.ts rather than GeoLocation.ts

src/app/core/basic-datatypes/configurable-enum/configurable-enum.interface.ts
sadaf895: Ok I need to analyze and think about it.

src/app/core/filter/filters/dateFilter.ts
```diff
@@ -1,38 +1,30 @@
...
-import { DateRangeFilterComponent } from "../../basic-datatypes/date/date-range-filter/date-range-filter.component";
+import { Entity } from "../../entity/model/entity";
+import { DataFilter } from "./filters";
 
 /**
  * Represents a filter for date values.
- * The filter can either be one of the predefined options or two manually entered dates.
  */
-export class DateFilter<T extends Entity> extends Filter<T> {
-  override component = DateRangeFilterComponent;
```
sleidig, line 15: this broke the date filter functionality. I have reverted some of the changes to fix it

src/app/core/filter/filters/dateFilter.ts
```diff
@@ -1,15 +1,14 @@
...
-import { DataFilter, Filter } from "./filters";
+import { calculateDateRange } from "app/core/basic-datatypes/date/date-range-filter/date-range-filter-panel/date-range-utils";
 import { isValidDate } from "../../../utils/utils";
+import { Entity } from "../../entity/model/entity";
+import { DataFilter, Filter } from "./filters";
+import { DateRangeFilterConfigOption } from "../../entity-list/EntityListConfig";
 import { DateRangeFilterComponent } from "../../basic-datatypes/date/date-range-filter/date-range-filter.component";
 
 /**
  * Represents a filter for date values.
```
sleidig, line 11:
```suggestion
 * Represents a filter for date values.
 * The filter can either be one of the predefined options or two manually entered dates.
```

sleidig (commented):
Thanks for the extensive work, @sadaf895 .

I found some issues and fixed these but it looks good to me know

Issue #2427, opened 2024-06-20
Import: columns with numbers are not converted to string for importMapping lookups


sleidig:
During imports a value is imported as `number` if it is all digits in the csv file. However, in our entities the related field may be a string. If such a field is used to map to an entity reference, the lookup in the EntityDatatype does not work anymore, as string and number equivalents of a value are not equal.

sleidig:
Hi @kumarannathan , thanks for reaching out and finding us! It's great to see your initiative to help our system - and thereby all our NGOs and social workers using Aam Digital :smiley: 

Have you worked with Angular and TypeScript before and are familiar with unit tests written in jasmine? (sorry, many details!)

We'll be happy to guide you on the specifics of the Aam Digital code base and this concrete issue/task then.

sleidig:
Test File for importing "Recurring Activity" records and linking "assigned users" during the import with phone number:

- one case where that number is clearly a string (with a "+" prefix); this already works now
- one case where the number is a "number" only - this fails to look up the correct user entity during import

for testing, make sure you have two user entities with those numbers in the csv file

[import-numbers.csv](https://github.com/user-attachments/files/18077963/import-numbers.csv)

PR #2823, opened 2025-01-29
Entity Mapping Update

sadaf895:
closes: #2427


src/app/core/basic-datatypes/entity/entity.datatype.spec.ts
```diff
@@ -23,50 +23,76 @@ import { EntitySchemaField } from "../../entity/schema/entity-schema-field";
...
+    });
+
+    it("should handle numeric-string mismatches correctly", async () => {
+      const c3 = new TestEntity();
+      c3.other = "456"; // Ensure it's a string
+      entityMapper = mockEntityMapper([c3]);
+      dataType = new EntityDatatype(entityMapper, null as any);
+      
+      await expectAsync(dataType.importMapFunction("456", schema, "other"))
+        .toBeResolvedTo(c3.getId());
```
sleidig, line 67: I think the following test case should cover the requirements of this issue:

```suggestion
      // "simple" case: imported value is string already
      await expectAsync(dataType.importMapFunction("456", schema, "other"))
        .toBeResolvedTo(c3.getId());
      // "advanced" case: imported value is number but should match also
      await expectAsync(dataType.importMapFunction(456, schema, "other"))
        .toBeResolvedTo(c3.getId());
```

src/app/core/basic-datatypes/entity/entity.datatype.spec.ts
```diff
@@ -23,50 +23,76 @@ import { EntitySchemaField } from "../../entity/schema/entity-schema-field";
...
+  testDatatype(new EntityDatatype(null as any, null as any), "1", "1", "User");
 
-  it("should map to the referenced entity", async () => {
-    const c1 = TestEntity.create("first");
-    const c2 = new TestEntity();
-    c2.other = "123";
-    const entityMapper = mockEntityMapper([c1, c2]);
-    const dataType = new EntityDatatype(entityMapper, null);
-    const schema = TestEntity.schema.get("ref");
+  describe("importMapFunction", () => {
```
sleidig, line 28: these second level "describe(" blocks make it a bit hard to follow what actually changed in this ... For easier review, can you please undo these. And also undo the changes in entity.datatype.ts in the `anonymize()` method, as these are not related, I think?

Sorry for the inconvenience :blush:   but to keep our speed of code improvements and reviews high, this is an important aspect.

src/app/core/basic-datatypes/entity/entity.datatype.spec.ts
```diff
@@ -23,27 +23,66 @@ import { EntitySchemaField } from "../../entity/schema/entity-schema-field";
...
     ).toBeResolvedTo(c2.getId());
+  });
+
+  it("should return undefined when no matching entity is found", async () => {
     await expectAsync(
       dataType.importMapFunction("345", schema, "other"),
     ).toBeResolvedTo(undefined);
   });
 
+  it("should handle numeric-string mismatches correctly", async () => {
```
sleidig, line 62: what is the "numeric-string mismatch" here in the test code? I don't see any mismatch - it looks like the same logic as `it("should map to the referenced entity by other field` above?

If so, let's simplify test code and remove one of the (almost) identical test cases. I personally like these last two ("should handle numeric-string mismatches correctly" and "should handle numeric value correctly") because they have exactly the same structure and the only difference in the test code is that aspect that we really want to test, i.e. importing a string or number value.

So let's remove some of the other tests if they don't check for fundamentally different things

src/app/core/basic-datatypes/entity/entity.datatype.ts
```diff
@@ -46,19 +47,59 @@ export class EntityDatatype extends StringDatatype {
...
   }
 
-  override importMapFunction(
+  /**
+   * Maps a value from an import to an actual entity in the database by comparing the value with the given field of entities.
+   * Handles type conversion between numbers and strings to improve matching.
+   *
+   * @param val The value from an import that should be mapped to an entity reference.
+   * @param schemaField The config defining details of the field that will hold the entity reference after mapping.
+   * @param additional The field of the referenced entity that should be compared with the val.
```
sleidig, line 56:
```suggestion
   * @param additional The field of the referenced entity that should be compared with the val. (e.g. if we run importMapFunction for a field that is an entity-reference to a "School" entity, this could be "name" if the "School" entity has a "name" property and the import should use that name to match the correct school)
```

src/app/core/basic-datatypes/entity/entity.datatype.ts
```diff
@@ -46,19 +47,59 @@ export class EntityDatatype extends StringDatatype {
...
+    }
+
+    const normalizedVal = this.normalizeValue(val);
+
+    try {
+      const entities = await this.entityMapper.loadType(schemaField.additional);
+      const matchedEntity = entities.find((entity) => {
+        const entityFieldValue = this.normalizeValue(entity[additional]);
+        return entityFieldValue === normalizedVal;
+      });
```
sleidig, line 75:
```suggestion
    try {
      const entities = await this.entityMapper.loadType(schemaField.additional);
      const matchedEntity = entities.find((entity) => this.normalizeValue(entity[additional]) === this.normalizeValue(val));
```

this is a little shorter and easier to read, maybe.

(probably this change to be done in the IDE to fix lint / code formatting)

src/app/core/basic-datatypes/entity/entity.datatype.ts
```diff
@@ -46,19 +47,59 @@ export class EntityDatatype extends StringDatatype {
...
   }
 
+  /**
+   * Normalizes a value for comparison, converting it to a standardized string format.
+   * Ensures both numbers and strings are treated consistently.
+   *
+   * @param val The value to normalize.
+   * @returns The normalized value as a string.
+   */
+  private normalizeValue(val: any): string {
```
sleidig, line 91:
```suggestion
  function normalizeValue(val: any): string {
```

This doesn't need anything from the class, so we can make it a simple function and above, when using it, write `normalizeValue()` instead of `this.normalizeValue()` (which is even shorter).

sleidig (commented):
works well. Suggesting just some tiny code simplifications.

sleidig (approved):
Thanks, @sadaf895 ! And thanks, @kumarannathan ! :pray:

Issue #2806, opened 2025-01-23
Bug/Dropdown's selected position is shifted right and upwards on the page


Abhinegi2:
**Describe the bug**
Dropdown with more than 4 options exhibits unexpected behavior when switching tabs:
- Open a dropdown with more than available options.
- Scroll through the options.
- Switch to another browser tab.
- Return to the original tab.
- Observe: The dropdown's selected position is shifted right and upwards on the page.

**To Reproduce**

1. Navigate to: .Admin->Admin-overview->Export Templates->add new template
2. Click on the "Applicable Entity Types" dropdown.
3. Scroll down within the dropdown options.
4. Switch to a different browser tab.
5. Return to the original tab.

**Expected Behavior**
The dropdown's position should remain consistent after switching tabs.

**Screenshots**

<img width="1506" alt="Image" src="https://github.com/user-attachments/assets/07f09cce-5e5b-4e58-a089-574817c220d9" />

Abhinegi2:
@george-neha could you please test the issue described above? The dropdown's position should now remain consistent after switching tabs. This fix has been deployed to [PR-2836](https://pr-2836.aam-digital.net/).

PR #2836, opened 2025-02-05
Fix/dropdown position after tab switches

Abhinegi2:
closes: #2806

### Visible/Frontend Changes

- [x] The dropdown's positionnow remains consistent after switching tabs.


sleidig (approved):
Seems to help and looks simple enough. Thanks for investigating this! :pray:

Issue #2273, opened 2024-03-01
defaultValue for configurable-enum dropdowns


sleidig:
In the Entity Schema we can set a `defaultValue` for an attribute/field. However, this currently does not work for configurable-enum dataType, because values for that are transformed in contrast to simple string dataType.

For public forms we do support this:
https://github.com/Aam-Digital/ndb-core/blob/e8899916101ab955b1eeac8c04e3dbd61700e9ce/src/app/features/public-form/public-form.component.ts#L76-L79

Probably this approach can be used to improve the implementation of the default value assignment:
https://github.com/Aam-Digital/ndb-core/blob/e8899916101ab955b1eeac8c04e3dbd61700e9ce/src/app/core/common-components/entity-form/entity-form.service.ts#L197

sleidig:
@Abhinegi2 , you shouldn't have to implement a new method in EntitySchemaService but rather the `valueToEntityForm()` method should do that already if called with the relevant parameters: https://github.com/Aam-Digital/ndb-core/blob/5b9074415e995041206733777300b6bad815f52b/src/app/core/entity/schema/entity-schema.service.ts#L217

PR #2841, opened 2025-02-06
defaultValue for configurable-enum dropdowns

Abhinegi2:
closes: #2273

### DIFF
--- a/src/app/core/config/config-fix.ts
+++ b/src/app/core/config/config-fix.ts
@@ -1002,11 +1002,7 @@ export const defaultJsonConfig = {
         dataType: "configurable-enum",
         label: $localize`:Label for the gender of a child:Gender`,
         additional: "genders",
-        anonymize: "retain",
-        defaultValue: {
-          mode: "static",
-          value: "M"
-        },
+        anonymize: "retain"
       },
       admissionDate: {
         dataType: "date-only",
--- a/src/app/core/config/config-fix.ts
+++ b/src/app/core/config/config-fix.ts
@@ -1002,7 +1002,11 @@ export const defaultJsonConfig = {
         dataType: "configurable-enum",
         label: $localize`:Label for the gender of a child:Gender`,
         additional: "genders",
-        anonymize: "retain"
+        anonymize: "retain",
+        defaultValue: {
+          mode: "static",
+          value: "M"
+        },
       },
       admissionDate: {
         dataType: "date-only",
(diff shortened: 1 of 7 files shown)


src/app/core/default-values/default-value.service.ts
```diff
@@ -109,6 +113,28 @@ export class DefaultValueService {
...
+      const enumValues = this.enumService.getEnumValues(fieldConfig.additional);
+      const matchedEnum = enumValues.find(
+        (enumValue) => enumValue.id === staticDefaultValue,
+      );
+
+      transformedDefaultValue = matchedEnum
+        ? this.entitySchemaService.valueToEntityFormat(
+            staticDefaultValue,
+            fieldConfig,
+          )
```
sleidig, line 129: have you checked what `this.entitySchemaService.valueToEntityFormat(staticDefaultValue, fieldConfig)` returns in the cases that you handle specially here (i.e. when no matchedEnum exists and you run valueToEntityFormat; and if fieldConfig.isArray is set)? My expectation would be that `entitySchemaService.valueToEntityFormat()` already returns sensible results for these automatically and the extra logic here is not needed.

I also think we can run this for any fieldConfig.dataType - there should not be issues for other defaultValues (and this may even prepare for other special defaults)

sleidig (commented):
Works well and looks fairly good :+1: 

I wonder if we can't simplify this further, however (see comment)

src/app/core/default-values/default-value.service.ts
Abhinegi2: 1. I added the matchEnum check because, if matchEnum does not exist, the selected field value appears as [invalid option], which I felt was not a sensible result. That’s why I included this check. - image
2. Regarding fieldConfig.dataType, I agree with your point and will remove that.

src/app/core/config/config-fix.ts
```diff
@@ -1002,7 +1002,11 @@ export const defaultJsonConfig = {
         dataType: "configurable-enum",
         label: $localize`:Label for the gender of a child:Gender`,
         additional: "genders",
-        anonymize: "retain"
+        anonymize: "retain",
+        defaultValue: {
+          mode: "static",
+          value: "M"
+        },
```
sleidig, line 1009:
```suggestion
        anonymize: "retain"
```

revert before merge

sleidig (approved):
I think you had not tested this for the case of a field with isArray = true. I have fixed the functionality for this and adjusted the test of public form config to focus on the actually relevant aspect.

Looks good to me now

Issue #2703, opened 2024-12-03
Field " Applicable Entity type" should be marked as mandatory field in Add new export template screen


george-neha:
Field " Applicable Entity type" should be marked as mandatory field in Add new export template screen. 
Currently if this field is blank, the file attachment does not get updated.

Steps: 
1. Click on a child/beneficiary
2. Click on the three dot menu and select "Generate File"
![Image](https://github.com/user-attachments/assets/d554457e-72b3-4cb1-abc2-ec935f1e5858)
3. Click on "configure available templates"
4. In the "export template" window, click on "Add new"
5. The field  " Applicable Entity type" should be marked as mandatory.

Abhinegi2:
@george-neha could you test the above-mentioned issue here  https://pr-2845.aam-digital.net/

george-neha:
@Abhinegi2 Can you show a similar message (see the screenshot below) for mandatory fields? "This field is required"

<img width="960" alt="Image" src="https://github.com/user-attachments/assets/3ab5c805-01ca-463d-8bbf-d1b15e4f08d4" />

Abhinegi2:
@george-neha updated! can you check now?

PR #2845, opened 2025-02-10
fix: Add required validation for applicableForEntityTypes field

Abhinegi2:
closes: #2703 

### Visible/Frontend Changes

- [x] Applicable Entity type" marked as a mandatory field in the Add new export template screen


sleidig (approved):
Thank you for also fixing the missing error hint, @Abhinegi2 and @george-neha :slightly_smiling_face:

Issue #2782, opened 2025-01-09
Import UI auto-mapped columns cannot be cleared/reset


sleidig:
**Describe the bug**
The column mapping step of Import Module has issues when the user wants to change pre-selected mappings.

**To Reproduce**

1. Go to 'Import' (e.g. from list view, top right context menu -> Import from file)
2. Upload a file where a column has the exact matching column title as the field id or label of a Entity Field
3. The matching column is automatically pre-selected in the Column Mapping step in the UI
4. Try to clear or change that field:
   5. Changes are not taken
   6. The "configure value mapping" button (e.g. for fields of type configurable-enum) does not disappear after clearing/changing the mapping

PR #2846, opened 2025-02-11
Import UI auto-mapped columns cannot be cleared/reset

Abhinegi2:
closes: #2782

### Visible/Frontend Changes

- [x] Now the child component focuses on rendering a single column mapping.


src/app/core/import/edit-import-column-mapping/edit-import-column-mapping.component.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+import { EditImportColumnMappingComponent } from "./edit-import-column-mapping.component";
+import { MockedTestingModule } from "app/utils/mocked-testing.module";
+import { MatDialog } from "@angular/material/dialog";
+import { of } from "rxjs";
+import { ColumnMapping } from "../column-mapping";
+import { TestEntity } from "../../../utils/test-utils/TestEntity";
+import { componentRegistry, ComponentRegistry } from "app/dynamic-components";
+import { DiscreteImportConfigComponent } from "app/core/basic-datatypes/discrete/discrete-import-config/discrete-import-config.component";
+
+fdescribe("EditImportColumnMappingComponent", () => {
```
sleidig, line 11:
```suggestion
describe("EditImportColumnMappingComponent", () => {
```

src/app/core/import/edit-import-column-mapping/edit-import-column-mapping.component.ts
```diff
@@ -0,0 +1,94 @@
...
+  @Output() valueChange = new EventEmitter<ColumnMapping[]>();
+
+  currentlyMappedDatatype: DefaultDatatype;
+
+  /** warning label badges for a mapped column that requires user configuration for the "additional" details */
+  mappingAdditionalWarning: string;
+
+  hideOption = (option: FormFieldConfig) => this.usedColNames.has(option.id);
+
+  async openMappingComponent(col: ColumnMapping) {
```
sleidig, line 51: I think this will always open the dialog for `this.col`? Then we could skip this method parameter for simplicity.

src/app/core/import/edit-import-column-mapping/edit-import-column-mapping.component.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    fixture = TestBed.createComponent(EditImportColumnMappingComponent);
+    component = fixture.componentInstance;
+    component.col = columnMapping;
+    component.rawData = rawData;
+    fixture.detectChanges();
+
+    spyOn(component.valueChange, "emit");
+  });
+
```
sleidig, line 45: Let's also add one basic test for selecting a mapping.
(feel free to "simulate" the user action by directly calling this.updateMapping() instead of dealing with test harnesses for the html components, though)

```
it("should emit changes after selected entity-field is changed", async () => {
```

src/app/core/import/edit-import-column-mapping/edit-import-column-mapping.component.ts
```diff
@@ -0,0 +1,94 @@
...
+    MatButtonModule,
+    MatBadgeModule,
+  ],
+})
+export class EditImportColumnMappingComponent {
+  private dialog = inject(MatDialog);
+  private componentRegistry = inject(ComponentRegistry);
+  private schemaService = inject(EntitySchemaService);
+
+  @Input() col: ColumnMapping;
```
sleidig, line 36: Let's call this "value" or "columnMapping". "col" is not intuitive for a developer looking afresh at this code

src/app/core/import/edit-import-column-mapping/edit-import-column-mapping.component.ts
```diff
@@ -0,0 +1,94 @@
...
+import { EntitySchemaService } from "../../entity/schema/entity-schema.service";
+import { HelpButtonComponent } from "app/core/common-components/help-button/help-button.component";
+import { MatInputModule } from "@angular/material/input";
+import { EntityFieldSelectComponent } from "app/core/entity/entity-field-select/entity-field-select.component";
+import { FormsModule } from "@angular/forms";
+import { MatButtonModule } from "@angular/material/button";
+import { MatBadgeModule } from "@angular/material/badge";
+import { MappingDialogData } from "../import-column-mapping/import-column-mapping.component";
+
+@Component({
```
sleidig, line 17:
```suggestion
/**
 * Component to edit a single imported column's mapping to an entity field
 * (including special transformations, if applicable).
 */
@Component({
```

src/app/core/import/edit-import-column-mapping/edit-import-column-mapping.component.ts
```diff
@@ -0,0 +1,94 @@
...
+  private componentRegistry = inject(ComponentRegistry);
+  private schemaService = inject(EntitySchemaService);
+
+  @Input() col: ColumnMapping;
+  @Input() entityCtor: EntityConstructor;
+  @Input() usedColNames: Set<string>;
+  @Input() rawData: any[];
+  @Input() columnMapping: ColumnMapping[] = [];
+
+  @Output() valueChange = new EventEmitter<ColumnMapping[]>();
```
sleidig, line 42: Let the Output also only relate to that mapping that this component is responsible for, i.e. 

```suggestion
  @Output() valueChange = new EventEmitter<ColumnMapping>();
```

and then `this.valueChange.emit(this.col)`, I think.

sleidig: As far as I can see, we can then also completely remove the `@Input() columnMapping: ColumnMapping[]` for the component here.

src/app/core/import/edit-import-column-mapping/edit-import-column-mapping.component.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  currentlyMappedDatatype: DefaultDatatype;
+
+  /** warning label badges for a mapped column that requires user configuration for the "additional" details */
+  mappingAdditionalWarning: string;
+
+  hideOption = (option: FormFieldConfig) => this.usedColNames.has(option.id);
+
+  async openMappingComponent(col: ColumnMapping) {
+    this.updateDatatypeAndWarning();
```
sleidig, line 52: I don't think this can change here to something that wasn't already done when calling it in `updateMapping()`? Do we need to call it again here?

src/app/core/import/edit-import-column-mapping/edit-import-column-mapping.component.ts
```diff
@@ -0,0 +1,94 @@
...
+          entityType: this.entityCtor,
+        },
+        width: "80vw",
+        disableClose: true,
+      })
+      .afterClosed()
+      .subscribe(() => this.updateMapping(true));
+  }
+
+  updateMapping(settingAdditional = false) {
```
sleidig, line 74: might be a bit less confusing if we actually set the value to `this.col` here in this function (instead of doing that via two-way binding and assuming it's already done by hidden magic :wink: )

src/app/core/import/import-column-mapping/import-column-mapping.component.html
```diff
@@ -14,30 +14,12 @@
...
-        [matBadgeHidden]="!mappingAdditionalWarning?.[col.column]"
-        matBadgeColor="warn"
-      >
-        Configure value mapping
-      </button>
-    }
+    <app-edit-import-column-mapping
+      [col]="col"
+      [entityCtor]="entityCtor"
+      [usedColNames]="getUsedColNames(col)"
```
sleidig, line 20: let's store this in a variable instead. Calling functions in templates is not ideal for performance because they get called every change detection cycle

src/app/core/import/import-column-mapping/import-column-mapping.component.ts
```diff
@@ -69,17 +68,9 @@ export class ImportColumnMappingComponent implements OnChanges {
   /** properties that need further adjustments through a component */
   dataTypeMap: { [name: string]: DefaultDatatype };
```
sleidig, line 69: do we still need that dataTypeMap somewhere? Or can we delete this completely?

sleidig (changes requested):
A few small code cleanup suggestions still :slightly_smiling_face:

sleidig:
@Abhinegi2 I still found one functional bug:
1. download a csv from Children list as sample to re-import
2. Import that file as Children
3. In "Map Columns" step, switch the "gender" column directly to map to field "Center" (instead of emptying the field)
4. open the "Configure value mapping" dialog
5. --> it still shows the mapping details related to gender rather then "center"

src/app/core/import/edit-import-column-mapping/edit-import-column-mapping.component.ts
```diff
@@ -0,0 +1,97 @@
...
+})
+export class EditImportColumnMappingComponent {
+  private dialog = inject(MatDialog);
+  private componentRegistry = inject(ComponentRegistry);
+  private schemaService = inject(EntitySchemaService);
+
+  @Input() columnMapping: ColumnMapping;
+  @Input() entityCtor: EntityConstructor;
+  @Input() usedColNames: Set<string>;
+  @Input() rawData: any[];
```
sleidig, line 43:
```suggestion

  /**
   * the actually imported data
   * (to let this component configure special transformations, e.g. to map values to dropdown categories)
   */
  @Input() rawData: any[];
```

src/app/core/import/edit-import-column-mapping/edit-import-column-mapping.component.ts
```diff
@@ -0,0 +1,97 @@
...
+  ],
+})
+export class EditImportColumnMappingComponent {
+  private dialog = inject(MatDialog);
+  private componentRegistry = inject(ComponentRegistry);
+  private schemaService = inject(EntitySchemaService);
+
+  @Input() columnMapping: ColumnMapping;
+  @Input() entityCtor: EntityConstructor;
+  @Input() usedColNames: Set<string>;
```
sleidig, line 42:
```suggestion

  /**
   * Entity fields that are already mapped and should not be offered to the user for selecting here.
   */
  @Input() usedColumnNames: Set<string>;
```

sleidig:
Unfortunately there still seem to be issues in the preview stage, @Abhinegi2 : Mappings are not done on the newly selected details (column title here should be "center" and values should be valid):
image

src/app/core/import/import-column-mapping/import-column-mapping.component.spec.ts
```diff
@@ -20,50 +15,25 @@ describe("ImportColumnMappingComponent", () => {
...
-    await component.openMappingComponent(columnMapping);
+    spyOn(component.columnMappingChange, "emit");
+
+    component.updateColumnMapping(originalColumnMapping, newColumnMapping);
 
-    expect(component.columnMappingChange.emit).toHaveBeenCalled();
+    expect(originalColumnMapping.propertyName).toBe("Test2");
+    expect(component.columnMappingChange.emit).toHaveBeenCalledWith([
+      ...component.columnMapping,
+    ]);
```
sleidig: turns out this is actually confirming that emit is called with an empty array ...
I fixed the expectation to be more useful.

sleidig (approved):
Works well now and is clean code. Thanks, @Abhinegi2 :+1:

Issue #2840, opened 2025-02-05
Multiple dropdown values are selected, but not being displayed.


george-neha:
**Describe the bug**
In the Notes/task menu on selecting multiple dropdown values from the dropdown list, all the selected names are not getting displayed. Only a few selected values are being displayed.

**To Reproduce**

1. Go to " Notes" menu.
2. Go to "Children" and select multiple children's name from the dropdown list.
3. Only a few of the selected names are displayed in the field.

Same issue is seen even in "Groups", only few of the selected names are being displayed.

**Expected behavior**
All the selected children's name should be listed in the field.

**Screenshots**

<img width="960" alt="Image" src="https://github.com/user-attachments/assets/6726dd0a-aab4-4e9e-823d-71cc18587f37" />

<img width="960" alt="Image" src="https://github.com/user-attachments/assets/a8634bfb-d0a8-4a63-86dd-ecabc8a02819" />

Abhinegi2:
I found two issues with the ChipList:

1. Clicking on an option only works when clicking directly on the checkbox; clicking elsewhere on the chip doesn't work.
2. After selecting an option, the dropdown UI has a glitch where its position shifts to the right
(..attached the screen recording for reference)

https://github.com/user-attachments/assets/1eed8841-008e-4d38-b077-44c3a3285c2e

Abhinegi2:
@george-neha could you test this fix on the following site: https://pr-2853.aam-digital.net/ ?

PR #2853, opened 2025-02-13
Multiple dropdown values are selected, but not being displayed.

Abhinegi2:
closes: #2840


sleidig (approved):
Hmm, this component is really becoming a intimidating mix of complex logic and workarounds :sweat_smile:  

But the fix works well and these changes are simple enough, thanks @Abhinegi2  and @george-neha !

Issue #2783, opened 2025-01-09
Download of reports as .csv format broken


sleidig:
**Describe the bug**
When downloading a report result as .csv the output format is forcing all data into a single column, putting addtional headers "Name,Value" on top of everything. The download for the old client-side reports work correctly but the server-side calculated reports (ReportConfig.mode = "sql") have this issue

**Sample csv output (broken):**

> Name,Value
"eventStartDate: 2024-12-01, eventEndDate: 2024-12-03, eventStatus: Implemented, eventType: Traini..."

### To reproduce:
1. Visit the "Reports" section (in a system where server-side reports are enabled): https://dev.aam-digital.net/report
2. Execute a server-side (`mode: "sql"`) report (e.g. in dev system select "Beneficiary (Individuals) [SQL]" report)
3. Click "Download" and try to open the file in a spreadsheet application (does not show columns as expected)

-----
## Suggested Implementation Approach:

1. Write unit tests for `SqlReportService.getCsv()` to have an easy way for testing without having to run the whole backend report server
2. change the return type of `SqlReportService.getCsv()` to something more specific than `any`
3. Make corrections to the getCsv method
4. Check about the formats for the two version of sql reports that we have (v2 has some special structure for nesting - get support from @tomwwinter who originally implemented this, if needed)

PR #2854, opened 2025-02-14
Download of reports as .csv format broken

Abhinegi2:
closes: #2783

### Visible/Frontend Changes

- [x] download report as CSV are working now


Abhinegi2:
@sleidig I have fixed this and tested the fix with v1 and v2 SQL reports. The downloaded CSV format is now working as expected.
You can test the functionality now, and once confirmed, I will revert the mocked_report file.

sleidig (commented):
The functionality with mocks here works well :+1: 

Can you please revert to the actual backend integration for final testing? I'll deploy to the dev server then.

I hope your code is working for both sql v1 and v2 at the same time still?

src/app/features/reporting/sql-report/sql-report.service.spec.ts
```diff
@@ -262,4 +263,38 @@ describe("SqlReportService", () => {
...
+      {
+        key: "Female",
+        value: 4,
+        level: 1,
+      },
+    ];
+
+    const result = service.getCsvforV2(mockSqlData);
+
+    const expectedCsv = `Name,,Value
```
sleidig, line 292: why do we expect an empty column between these?

Abhinegi2: In the getCsv method, `deepestLevel + 1` adds an empty column before "Value", which may be to ensure the alignment of value in the CSV output or "Value" column consistently(that's why we expected an empty column in the test cases). Still, I’m not sure why it's needed. Tom might have some idea about it.

Abhinegi2:
I also reverted the mocks now for testing we can deploy this on dev site.

src/app/features/reporting/sql-report/sql-report.service.spec.ts
sleidig: ah, I think this was done to give people a visual kind of indentation also in the downloaded file :+1: 

```suggestion
    // extra empty column to allow for indentation of name at child levels
    const expectedCsv = `Name,,Value
```

src/app/features/reporting/sql-report/sql-report.service.spec.ts
```diff
@@ -262,4 +263,38 @@ describe("SqlReportService", () => {
       },
     ]);
   });
+  it("should convert report into CSV string for version 1 SQL", () => {
+    const mockSqlData: SqlReportRow[] = [
+      {
+        key: "Total students",
```
sleidig, line 269: I believe this is the v2 format, not v1?

And maybe indeed we should rename service.getCsvforV2 if that method is only handling that format ...

src/app/features/reporting/reporting/reporting.component.spec.ts
```diff
@@ -328,4 +328,23 @@ describe("ReportingComponent", () => {
         expect(component.errorDetails).not.toBeNull();
       });
   });
+
+  it("should return raw data for version 1 SQL reports", () => {
+    const mockData = [
+      { key: "Total students", value: 6 },
+      { key: "Students gender", value: 10 },
```
sleidig, line 335: isn't this the format of v2 reports?

I would like to have two tests actually - one for v1 data and one for v2 data

src/app/features/reporting/sql-report/sql-report.service.spec.ts
```diff
@@ -262,4 +263,39 @@ describe("SqlReportService", () => {
       },
     ]);
   });
+  it("should convert report into CSV string for version 1 SQL", () => {
```
sleidig, line 266:
```suggestion

  it("should convert report into CSV string for version 1 SQL", () => {
```

sleidig (commented):
Great :smiley: 
Works as expected on https://dev.aam-digital.net/report

sleidig (approved):

Issue #2654, opened 2024-11-11
"Report calculated at" should display in local timezone


sleidig:
**Describe the bug**
Currently the Reports View loads previous calculations and displays the time the result was calculated on:

> This report was calculated at: 11.11.2024 14:46

However, this time is not displayed in the local time of the user at the moment. This must reflect local time.

### To reproduce:
1. Visit the "Reports" section (in a system where server-side reports are enabled): https://dev.aam-digital.net/report
2. Execute a server-side (`mode: "sql"`) report (e.g. in dev system select "Beneficiary (Individuals) [SQL]" report)
3. See the description text above the results once the calculation is finished and compare the time (it is not in current time-zone)

Abhinegi2:
@sleidig  after investigating, I found that "Report calculated at" comes from the `endDate` value, retrieved from the backend's report API. To display "Report calculated at" in the local timezone, I’m taking the following approach:  

1. Create a `timezone-pipe` component to convert datetime to the local timezone.  
2. Integrate it with the `ReportingComponent`.  
3. Add test cases to ensure correct timezone conversion.  

also, let me know if you have any alternative approaches or suggestions.

sleidig:
Can you check if there are such pipes already available? (possibly Date pipe has some way to support it)
Seems like a good approach overall.

(tests probably sufficient in the pipe without the need to re-test in reporting-component as logic is only within new/existing pipe's implementation)

PR #2864, opened 2025-02-18
 display report calculation time in local timezone

Abhinegi2:
closes: #2654

### DIFF
--- a/src/app/features/reporting/reporting/reporting.component.html
+++ b/src/app/features/reporting/reporting/reporting.component.html
@@ -19,7 +19,7 @@
 
 <div style="color: gray; padding: 10px" *ngIf="reportCalculation" i18n>
   This report was calculated at:
-  {{ reportCalculation?.endDate | date: "dd.MM.yyyy HH:mm" }} <br />
+  {{ localTime | date: "dd.MM.yyyy HH:mm" }} <br />
   Click again on "Calculate" to re-calculate the report including any changes of
   records since then.
 </div>
--- a/src/app/features/reporting/reporting/reporting.component.ts
+++ b/src/app/features/reporting/reporting/reporting.component.ts
@@ -139,7 +139,7 @@ export class ReportingComponent {
           ),
         );
         if (this.reportCalculation?.endDate) {
-          // Convert the UTC to local timezone
+          // Convert the UTC to local timezone (as the date string doesn't include timezone information (e.g. ending with "Z") we have to handle this explicitly
           this.localTime = moment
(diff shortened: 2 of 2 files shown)


Abhinegi2:
@sleidig it's my initial approach to use the local timezone, can you also please deploy this on the dev site for testing?

src/app/features/reporting/reporting/reporting.component.html
```diff
@@ -19,7 +19,7 @@
 
 <div style="color: gray; padding: 10px" *ngIf="reportCalculation" i18n>
   This report was calculated at:
-  {{ reportCalculation?.endDate | date: "dd.MM.yyyy HH:mm" }} <br />
+  {{ localTime | date: "dd.MM.yyyy HH:mm" }} <br />
```
sleidig, line 22: You solution seems to work well. Doing a quick research if we can avoid the extra property, I noticed the following:

```JavaScript
new Date("2025-02-18T08:05:45.909")
// = Tue Feb 18 2025 08:05:45 GMT+0100 (Central European Standard Time)
new Date("2025-02-18T08:05:45.909" + "Z")
// = Tue Feb 18 2025 09:05:45 GMT+0100 (Central European Standard Time)
```

i.e. probably our backend should write the date including "Z" to indicate UTC timezone.

Maybe we can just add that in `fetchReportCalculation()` service as a workaround, without having to edit the component here?

e.g. 
```
this.http.get<ReportCalculation>(...)
  .pipe(map(res => {
    if(res?.endDate && !res.endDate.endsWith("Z")) {
      res.endDate = res.endDate + "Z";
    }
    return res;
  })`

**Review by sleidig — COMMENTED** [review]
deployed to dev. Works as expected :+1:

File: src/app/features/reporting/reporting/reporting.component.ts [hunk 2]
```diff
@@ -137,7 +138,13 @@ export class ReportingComponent {
             reportData.calculation.id,
           ),
         );
-
+        if (this.reportCalculation?.endDate) {
+          // Convert the UTC to local timezone
```
sleidig, line 142:
```suggestion
          // Convert the UTC to local timezone (as the date string doesn't include timezone information (e.g. ending with "Z") we have to handle this explicitly
```

src/app/features/reporting/reporting/reporting.component.html
sleidig: nvermind. The proposed solution is working and not too complex anyway :sweat_smile:

sleidig (approved):

Issue #2867, opened 2025-02-19
Import records and link them as members/participants directly during import


sleidig:
_As a event coordinator I want to import a list of participants from spreadsheet (csv) so that the system creates both the profile of these people and also the link to my selected Event record in order to seamlessly add event participants without extra and confusing manual steps of adding them to an event later on._

Assuming a configuration similar to the School Enrollment ("School History") in our demo setup:
- a spreadsheet of children is imported as students of a certain school: During import I, as a user, select one existing School to which the students should be added.
- the Child records should be created from the imported data (already possible)
- a "School Enrollment" record (ChildSchoolRelation entity) should be created for each imported Child, linking to that selected School
  - some columns should be mappable to the relationship record (e.g. "class" of the "School Enrollment" record) in the same process where the other columns are mapped to core Child profile fields also.

-----
Additional considerations:
- This functionality could be offered to users from the Details View of an entity (e.g. from the School Details page, to import students linked to that school)

sleidig:
We have some logic like this already implemented. There is hard-coded support for creating ChildSchoolRelation records to link to a School, particularly. However, we will need to generalize this here

sleidig:
To test this functionality the following use case applies:

1. Import Children (e.g. with attached sample or any csv) [Students.csv](https://github.com/user-attachments/files/18885250/Students.csv)
2. In Import Step 2 (Select Import Type): Use the "Advanced Import Actions" section to add two actions:
   1. Link to "Recurring Activity" (select an existing activity)
   2. Link to "School" (select an existing school)
3. Run the import. Expecting:
   1. In the selected "Recurring Activity" (Attendance -> Manage Activities -> open the selected activity), all imported children should be already added to the "Participants" field of that activity
   2. In the selected "School", all imported children should already be added to the "Students" list of that school
4. Go to "Import" and in the first step use the "Undo import" button for the "Previous Import" listed on the right side
5. Expect to see all changes to get reverted:
   1. In the selected "Recurring Activity", the previously imported participants have disappeared (and those originally added outside of the import are still there)
   2. In the selected School, the imported students have disappeared
   3. In the children list, the imported students have disappeared

![Image](https://github.com/user-attachments/assets/b7a8ba28-29a4-4b23-8188-66181303836b)

![Image](https://github.com/user-attachments/assets/091b5377-9ebe-4bff-9dd2-67bfd22f1194)

PR #2870, opened 2025-02-20
Refactor Additional Import Actions service

sleidig:
see #2867


Abhinegi2:
Functionality is working as expected, and the code looks simple. 

However, I found two bugs. This may not be directly related to this pr/fix  -
1. When toggling to expert mode, the "Import as" dropdown displays empty options.
image
2. The advanced import actions fields should remain disabled until a child entity(relation entity) is selected else for another entity it does not show any option. 
image

sleidig:
Thanks, that's good observations! Will fix these issues together with the new functionality

Issue #2876, opened 2025-02-24
Ordering of dropdown not displayed correctly


sleidig:
If enabled, our dropdown field allows users to reorder the multiple selected elements. This is not correctly displayed in the dropdown list anymore:

- In closed field, showing the values in the expected, manually set order:
![Image](https://github.com/user-attachments/assets/9a628aa3-44e0-4aae-b174-85fc6e345fb2)
- In the dropdown list, order does not match. Dragging does reorder in the background but is not visible to the user until closing the dropdown list, which is very confusing:
![Image](https://github.com/user-attachments/assets/a8bff884-2779-4f3b-9632-a034f4c146d1)

Abhinegi2:
@george-neha Could you please test this bug fix on the deployed:(https://pr-2877.aam-digital.net/)?  

Steps to verify:  
1. Navigate to **Children → Edit Data Structure → General Setting**.  
2. Select **Generated Title of Record**.  
3. Check if you can reorder the options and confirm that the position changes are reflected in the field and the options list.

PR #2877, opened 2025-02-25
Ordering of dropdown not displayed correctly 

Abhinegi2:
closes: #2876

### DIFF
--- a/src/app/core/common-components/basic-autocomplete/basic-autocomplete.component.ts
+++ b/src/app/core/common-components/basic-autocomplete/basic-autocomplete.component.ts
@@ -133,6 +133,7 @@ export class BasicAutocompleteComponent<O, V = O>
   );
   autocompleteFilterFunction: (option: O) => boolean;
   @Output() autocompleteFilterChange = new EventEmitter<(o: O) => boolean>();
+
   /** whether the "add new" option is logically allowed in the current context (e.g. not creating a duplicate) */
   showAddOption = false;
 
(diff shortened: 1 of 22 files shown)


src/app/core/common-components/basic-autocomplete/basic-autocomplete.component.ts
```diff
@@ -256,6 +256,10 @@ export class BasicAutocompleteComponent<O, V = O>
         event.previousIndex,
         event.currentIndex,
       );
+      const reorderedOriginalOptions = this.autocompleteOptions.map(
+        (o) => o.initial,
+      );
+      this.reorderdOptions.emit(reorderedOriginalOptions);
```
sleidig, line 262: Shouldn't a reordering also trigger a valueChange event (I guess it probably does, calling this.onChange() below)? And if so, shouldn't that event be enough for parent components to handle reorderings?

For me, logically there is not really a difference between the number or the order of selected items changing. It always results in a different array representing the current value?

sleidig (commented):
Works as expected :100: 

I do wonder why we need a separate output explicitly for reordering, however?

src/app/core/common-components/basic-autocomplete/basic-autocomplete.component.ts
Abhinegi2: Yes, I checked, and the onChange() call does emit the updated value (reflecting the new order) to the parent component.
However, I found that while the form value (toStringAttributes) updates correctly, the parent's toStringAttributesOptions array wasn't reflecting the reordered sequence. This was causing the dropdown reordering issue. I’ve now added a method to update the sequence—please check it out.

src/app/core/common-components/basic-autocomplete/basic-autocomplete.component.ts
```diff
@@ -133,7 +133,6 @@ export class BasicAutocompleteComponent<O, V = O>
   );
   autocompleteFilterFunction: (option: O) => boolean;
   @Output() autocompleteFilterChange = new EventEmitter<(o: O) => boolean>();
```
sleidig, line 135:
```suggestion
  @Output() autocompleteFilterChange = new EventEmitter<(o: O) => boolean>();

```

sleidig (approved):
Thanks for finding this even simpler solution :+1:

PR #2883, opened 2025-02-27
fix: bulk delete for special loaders correctly handles deleted entities

sleidig:


Abhinegi2 (approved):
Thanks, working as expected!

Issue #2882, opened 2025-02-27
Fields seem to cause errors if the "Field ID" includes a "."?


sleidig:
Certain Field IDs cause errors.

**To Reproduce**

1. Edit data structure of an entity type
2. Add a new field, change the "Field ID" manually and add a "." character
4. Save and open the details view to see the form field
5. In the console errors are shown:

> dynamic-component.directive.ts:69 TypeError: Cannot read properties of null (reading 'parent')
    at _.ngOnInit (edit-component.ts:70:36)
    at dynamic-component.directive.ts:69:25
    at Generator.next (<anonymous>)

> TypeError: Cannot read properties of null (reading '_rawValidators')
    at template (edit-text.component.html:5:21)

...

## ToDo:

- Are there other special characters that also cause such errors?
   - Check what happens if the user uses Arabic or other characters in the Field ID manually
- Check what the system generates as ID when filling the label with other scripts like Hindi or Arabic characters
- Validate the input (using our regex validator function?) of that field

george-neha:
1. for some of the special characters like . & @ %, the system generates an error like the one mentioned above.
2. if unique character . is added in the field ID, then that field is always editable. 

<img width="964" alt="Image" src="https://github.com/user-attachments/assets/3d6b6940-fac8-4861-a069-8da13b9c5038" />

3. Field ID is blank if other scrips like Hindi, Urdu, etc are used. The field is populated only if the characters are in the English script. 
<img width="897" alt="Image" src="https://github.com/user-attachments/assets/38dd36d2-694e-4a85-a981-d11364e1f7e1" />

sleidig:
Thanks @george-neha , that's super helpful analysis 🙏 

I think we should only allow alphanumeric characters and _ in IDs, to avoid any possible issues (i.e. probably a regex like this `^[a-zA-Z0-9_]*$`). And update the `AdminEntityFieldComponent` with its "Field ID" form control in two ways:
1. update the generation of IDs (`generateIdFromLabel`) to replace any other special character with an underscore
2. add a validator to that form control to mark any value that includes other characters as invalid (e.g. when the user directly changes the field's value)

PR #2884, opened 2025-02-28
Fields seem to cause errors if the "Field ID" includes a "."?

Abhinegi2:
closes: #2882

### DIFF
--- a/src/app/core/admin/admin-entity-details/admin-entity-field/admin-entity-field.component.html
+++ b/src/app/core/admin/admin-entity-details/admin-entity-field/admin-entity-field.component.html
@@ -67,6 +67,12 @@
                 icon="lock"
                 matSuffix
               ></fa-icon>
+              <mat-error *ngIf="fieldIdForm.hasError('pattern')">
+                <span i18n>
+                  Invalid ID pattern. Only letters, numbers, and underscores are
+                  allowed.</span
+                >
+              </mat-error>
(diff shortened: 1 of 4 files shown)


Abhinegi2:
@sleidig, Could you please review the current approach? `generateIdFromLabel` is now replacing special characters with underscores, which affects the previous test cases. 
Do the updated test results align with our current approach, or should we adjust the logic/test?

sleidig:
Oh, I didn't realize we are already removing any special characters @Abhinegi2 :see_no_evil: 
Hmm, I guess either one is fine ... actually let's stick with the existing logic then. But I would like to allow "_" as one special character to make longer IDs more readable.

I wonder how we got the IDs including "." at the client. maybe only because they entered IDs manually. Do we display a useful error for the form field yet, when the validator fails?

Abhinegi2:
> Oh, I didn't realize we are already removing any special characters @Abhinegi2 🙈 Hmm, I guess either one is fine ... actually let's stick with the existing logic then. But I would like to allow "_" as one special character to make longer IDs more readable.
> 
> I wonder how we got the IDs including "." at the client. maybe only because they entered IDs manually. Do we display a useful error for the form field yet, when the validator fails?

Currently, no useful error message is displayed. I’m planning to add one that says: "Only letters, numbers, and underscores are allowed."
What do you think about this error message for the validation?

sleidig (approved):

PR #2885, opened 2025-02-28
fix import issues

sleidig:


Abhinegi2 (commented):
The import as dropdown is now working as expected, and the Advanced option is disabled until an entity type is selected.

I wanted to confirm if we're only addressing the first issue in this branch (https://github.com/Aam-Digital/ndb-core/pull/2870#issuecomment-2671200624), as I'm still encountering the second issue.

sleidig:
> The import as dropdown is now working as expected, and the Advanced option is disabled until an entity type is selected.
> 
> I wanted to confirm if we're only addressing the first issue in this branch ([#2870 (comment)](https://github.com/Aam-Digital/ndb-core/pull/2870#issuecomment-2671200624)), as I'm still encountering the second issue.

Yes, I think the selection of additional actions is tightly connected to the logic changes of what additional actions we have (done in the next PR). The "Additional Actions" are disabled until at least some entity is selected now, however.

Abhinegi2 (approved):

Issue #2862, opened 2025-02-17
Updating records through import from spreadsheet


sleidig:
Our Import Module currently always creates new records. It is not possible to update fields of existing records from an external spreadsheet via file import. However
- sometimes users make corrections in their spreadsheet and want to import the update again.
- sometimes a few items in an imported spreadsheet may already be in the database and it would be preferred to update the existing data in Aam Digital (or possibly also skip without changes) instead of creating a new (possibly duplicate) record as part of the import

The details of the requirements and a feasible logic for this needs to still be discussed and planned however. Questions are:
- How can the system determine which record in Aam a row in the spreadsheet matches?
  - What if there is no (complete) match? What if there are multiple matches?
    - ... this further extension could be postponed also. When we currently import a link to an existing other record, the system simply takes the first, exact match. This could also be an initial solution here
- How to let the user configure between updating/creating?
- How to clearly show what will happen (update vs create) in the preview (and other places?) to the user?
- ... ?

sleidig:
For detecting existing records:

1. Selected "ID fields" are all mapped to columns in the imported file:
   1. Identifying existing by all "ID fields"
   2. Preview:
      - Sorts records, showing those getting updated first (to make sure the user sees that some are getting updated rather than created
      - (?) highlights rows for records getting updated with a different background color
      - shows an icon/button for each record getting updated, explaining why this is highlighted
2. Some selected "ID field" is not part of the imported file:
   1. Show warning on Column Mapping step that some ID field is not mapped (e.g. if ID fields are name + project number but project number is not part of the imported spreadsheet)
   2. Identifying existing by (partial) matching for only those ID fields that are mapped
   3. In Preview: give option to user to switch between create/update for each row

PR #2899, opened 2025-03-10
Imports that update existing records

sleidig:
closes: #2862


sleidig:
@george-neha , can you test this?

For simplicity I have not added manual options for the user to selected between update/create/skip yet, as I first would like to get a core version finalized

george-neha:
I added a student with an existing record(Jyotsana Pilla) in the file I imported. However, when I undid the import, this existing student was deleted from the list. Should only the updated data be revised, or should the record no longer exist?

sleidig:
> I added a student with an existing record(Jyotsana Pilla) in the file I imported. However, when I undid the import, this existing student was deleted from the list. Should only the updated data be revised, or should the record no longer exist?

Thank you for discovering this, @george-neha ! :pray: 

I have improved the undo. Could you try again, trying to break it (maybe also with other field types like category, date, etc.)?

src/app/core/import/import-review-data/import-review-data.component.ts
```diff
@@ -83,9 +117,10 @@ export class ImportReviewDataComponent implements OnChanges {
           data: {
             entitiesToImport: this.mappedEntities,
             importSettings: {
-              columnMapping: this.columnMapping,
               entityType: this.entityType,
+              columnMapping: this.columnMapping,
               additionalActions: this.additionalActions,
+              idFields: this.idFields,
```
Abhinegi2, line 123: Any specific reason behind naming this **idFields**? 
would it make sense to use something like **matchingIdFields** instead since these fields are used for record matching?

sleidig: good point. I am renaming this for clarity

Abhinegi2: Just checking— did you push the updated codebase with the renamed variable, or are you planning to rename it later? I can't see the changes yet.

Abhinegi2 (approved):

src/app/core/import/import-review-data/import-review-data.component.ts
sleidig: thanks :see_no_evil:   I pushed but didn't commit my changes

sleidig:
@george-neha , did you also test to update some records by matching name + dateOfBirth? (I just want to be sure these date fields work for matching; And that scenario may be very common and realistic for users also)

george-neha:
@sleidig Yeah, I tested multiple scenarios, and they are working as per expectation.
1. Name +DOB+ gender matching with the data in the children's profile> review screen indicates updating profile
2. Only Name matching and DOB+ gender not matching with the data in the children's profile> review screen indicates creating new profile

Issue #2857, opened 2025-02-14
Merge records: Multiple selected from from a list


george-neha:
In the notes/tasks entity, multiple children/teams involved/groups can be tagged to a note/task. While merging such records, the children/team involved/groups should be available as specified in the preview column. The user should also be able to choose additional values from the list.

Scenario 1:
Record 1 has child A, B, C
Record 2 has child B, C, D, and E. 

The user should be able to manually select the tagged children/team involved/groups to move to the merged record.

Merged Record: Child A, B, C, D, E

Scenario 2:

Record 1 has child A, B, C
Record 2 has child B, C, D, and E. 
Merged Record: Child A, B, C, F and G

Abhinegi2:
@george-neha, I have a few questions:  

1. Since we already implemented multi-select for the `entityType` field in #2855, I believe both scenarios mentioned above are covered there. Could you please verify and test this?  

2. For child entities that have related entity tabs like Education, Attendance, and Notes,  I don’t think we’ve addressed how the merge functionality will work for them in this issue and also in our first version. Do we already have a sub-issue for that? If not, how do we plan to handle their representation in the merge preview dialog?  

Let me know your thoughts, and we can connect to discuss this further.

george-neha:
During the multi-selection of beneficiaries in the notes entity, when a value is de-selected from records A or B, the other selected values are deleted. 

https://github.com/user-attachments/assets/84975f1e-c311-4d17-9683-f25b2becca86

sleidig:
when manually changing a field in the preview column, the record A / B radiobutton or checkbox should probably get unselected automatically

sleidig:
Shouldn't fields without / empty values be hidden? For some dataTypes this doesn't seem to work yet (e.g. expecting Notes, RelatedRecords, Groups to be hidden):

![Image](https://github.com/user-attachments/assets/71a5e58c-98f8-401a-8e2d-77dc6a5fc022)

Abhinegi2:
@george-neha, could you test the following in the deployed version: https://pr-2901.aam-digital.net/?
1. When manually changing a field in the preview column, the Record A/B radio button or checkbox should automatically get unselected (currently, only checkboxes are being unselected; previously selected values are still in preview).
2. Fields with an array datatype that are empty or contain no values should be hidden (e.g., related records, groups).

Also, let me know if you have any other suggestions/thoughts regarding these.

PR #2901, opened 2025-03-11
 Merge records: Multiple selected from from a list

Abhinegi2:
closes:  #2857

### Visible/Frontend Changes

- [x]
- [ ]

- [x]
- [ ]

### DIFF
--- a/src/app/features/de-duplication/bulk-merge-records/bulk-merge-records.component.ts
+++ b/src/app/features/de-duplication/bulk-merge-records/bulk-merge-records.component.ts
@@ -85,7 +85,10 @@ export class BulkMergeRecordsComponent<E extends Entity> implements OnInit {
   private initFieldsToMerge(): void {
     this.entityConstructor.schema.forEach((field, key) => {
       const hasValue = this.entitiesToMerge.some(
-        (entity) => entity[key] !== undefined && entity[key] !== null,
+        (entity) =>
+          entity[key] !== undefined &&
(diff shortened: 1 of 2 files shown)


Abhinegi2:
@sleidig After our discussion, I made some changes:  

1. In the merge preview, removing an option(which is available in any Record A and B) will now unselect the related checkbox.  
2. For string and long-text fields, updating any value will also unselect the related checkbox.  
there are still some cleanups pending, but could you quickly check if this aligns with what we want?

george-neha:
I added a new value in the string and long-text fields, but when I de-select a value from record A/B, the new value is also deleted.  @Abhinegi2

src/app/features/de-duplication/bulk-merge-records/bulk-merge-records.component.ts
```diff
@@ -140,6 +151,97 @@ export class BulkMergeRecordsComponent<E extends Entity> implements OnInit {
     control.patchValue(value);
   }
 
+  private meregePreviewFieldChanges(
```
sleidig, line 154:
```suggestion
  private mergePreviewFieldChanges(
```

sleidig (commented):
Seems to work well but this component is also becoming a really complex piece of code. I am getting scared of making future changes to this ... Let's briefly review together in a call if there are any opportunities to reduce the cascade of methods needed when updating a value.

sleidig (approved):

Issue #2858, opened 2025-02-14
Merge: Other entities linking to either one of the two merged records


george-neha:
When two records are merged, the details captured in other entities should not be lost or deleted; they should be available in the merged record. 

Other entities linking to either one of the two merged records (e.g., Notes related to two records of a Child that are getting merged) must be updated to link to the single merged record. 
Details captured in other entities (all the records listed within the details tabs captured) should be available in a list in the merged record to ensure no data loss. 

**Scenario ** : Child Record A has four linked Notes. Child record B has five notes that are linked. 
In the merged record, nine notes should be listed to ensure no data loss. 

In scenarios where two-child records are merged for the other entity-related tabs(Education, Notes & tasks, Educational Materials, Observations) to ensure no data loss, all the details captured for both records are available in a list in the current design.

sleidig:
Edge case: What if record A + B are both linked, so that after merging them the same record would be linked twice (e.g. twice on the attendance list of an event)?

- ID of the removed merged record gets replaced by the ID of the remaining record
- if in array fields the same ID is included twice now, then we remove the second, duplicate ID
- in case this has a different attendance status, we still delete that second entry and log a warning
   - --> 2nd version: proper UI for asking user about such cases

Abhinegi2:
After reviewing both recommended methods, I found the following:  

1. **`findFieldsReferencingEntityType`** provides details about related entity types, such as:  

   ```
   Aser: { additional: "Child", dataType: "entity", entityReferenceRole: "composite", id: "child" }  
   EducationalMaterial: { additional: "Child", dataType: "entity", entityReferenceRole: "composite", id: "child" }  
   ChildSchoolRelation: [{…}]
   ```  
   However, it does not return the related entity IDs or other necessary details.  

2. **`CascadingActionResult`** (from `entityActionService`) provides the information we need—a list of related entities for the selected rows. For example:  

   ```
   Note {  
     _id: 'Note:35e63d4b-d0bd-45f7-8c8f-07d9292d9991',  
     children: ['Child:1'],  
     childrenAttendance: EventAttendanceMap(0),  
     authors: ['User:Devak'],  
     relatedEntities: []  
   }
   ```  
   Since this approach includes entity IDs, it is more suitable for our use case.  

### Proposed Approach:  
1. First, write test cases to verify that after merging, the related entities of Record B are added to the related entities of Record A.  
[...]

PR #2904, opened 2025-03-12
Merge: Other entities linking to either one of the two merged records 

Abhinegi2:
closes: #2858

Edge case: What if record A + B are both linked, so that after merging them, the same record would be linked twice (e.g. twice on the attendance list of an event)?

- [x] ID of the removed merged record gets replaced by the ID of the remaining record
- [x] if in array fields the same ID is included twice now, then we remove the second, duplicate ID
- [x] In case this has a different attendance status, we still delete that second entry and log a warning


src/app/features/de-duplication/bulk-merge-service.spec.ts
```diff
@@ -46,4 +47,23 @@ describe("BulkMergeService", () => {
...
+    // Todo: this should be done in the service after updating executeMerge method
+    noteA.relatedEntities.push(recordB.getId());
+    await entityMapper.save(noteA);
+
+    const updatedNoteA = entityMapper.get(
+      Note.ENTITY_TYPE,
+      noteA.getId(),
+    ) as Note;
+    console.log(updatedNoteA);
+    expect(updatedNoteA.relatedEntities).toContain(recordB.getId());
```
sleidig, line 67: An additional test case can be: when `multiRelated = [recordA.getId(), recordB.getId()]` then only one time the merged ID should be included (not including it twice)

Abhinegi2:
Thanks for the suggestions! and also added an additional test case for multi-merge. now, I'll start working on the development

sleidig:
I have merge https://github.com/Aam-Digital/ndb-core/pull/2916 . Maybe you can switch to the EntityRelationsService in your implementation instead of the (somewhat complicated) cascading action extension, @Abhinegi2 . Then this should be good to go, I think.

sleidig (approved):
Looks great!

Clean code and functionality seems to work well. Thanks, @Abhinegi2

Issue #2911, opened 2025-03-18
Public Form "Text Blocks" lose markdown


sleidig:
**To Reproduce**

1. Go to the Admin Screen for a Public Form (e.g. "Edit Data Structures" -> Public Forms -> click on a Public Form to edit in dialog)
2. Switch to the "Configure Fields" tab and "edit" it
3. Add a new Text Block
4. Save
5. Edit the fields tab again
6. Edit the existing Text Block and add markdown (e.g. for bold: "**foo**")
7. Save the dialog for "Edit Field", where you entered the text block text
8. ... Markdown formatting is now showing as expected in the preview
9. Save the "Configure Fields" tab
10. (!) Markdown formatting now disappears

![Image](https://github.com/user-attachments/assets/39412829-8cfb-4d9e-ba27-374a02508c92

Abhinegi2:
After investigating -  
I found that the issue is not just the loss of markdown formatting. Once a text block is initially saved, any updates to that text field are not reflected in both the public form and the details view.

**to reproduce** - 
create a new text field and save it, then try to update its value, the updated value appears correctly after saving the "Edit Text Field" dialog. However, once you save the entire details view, the updated value is lost.

Abhinegi2:
@george-neha Could you test the fixes for the mentioned issues at (https://pr-2914.aam-digital.net/child)?

PR #2914, opened 2025-03-19
Public Form "Text Blocks" lose markdown 

Abhinegi2:
closes: #2911


sleidig (approved):
Thanks for the good analysis and quick fix :+1:

PR #2916, opened 2025-03-19
refactor: extract EntityRelationsService with functions to find linki…

sleidig:
…ng records

- tests still failing


sleidig:
> After the functional review- I found that the delete and anonymize actions are working fine except for schools.
> 
> * When anonymizing a school, it is not being anonymized properly—I can still see it in the list. However, the anonymized action works correctly for other entities.

Thanks for the thorough testing, @Abhinegi2 ! :blush: 
I checked this and found that this is actually intended behavior: The "School" entity type is not marked to have any personal information (PII), therefore anonymization does not apply (in the entity details it's hidden then). I have now fixed and also hidden that action from bulk actions in this context.

Abhinegi2 (approved):
Looks good to me!

PR #2917, opened 2025-03-20
refactor: bulkMergeComponent

Abhinegi2:
closes: #XX

### Visible/Frontend Changes

- [x]
- [ ]

- [x]
- [ ]

### DIFF
--- a/src/app/features/de-duplication/bulk-merge-records/bulk-merge-records.component.html
+++ b/src/app/features/de-duplication/bulk-merge-records/bulk-merge-records.component.html
@@ -11,42 +11,23 @@
           <strong>{{ field.label }}</strong>
         </div>
 
-        @for (
-          entityToMerge of entitiesToMerge;
-          track entityToMerge;
-          let i = $index
-        ) {
-          <div class="record-cell">
-            @if (field.allowsMultiValueMerge) {
-              <mat-checkbox
-                [checked]="existingFieldSelected[field.id][i]"
-                (change)="selectExistingValue(field.id, i, $event.checked)"
-                [disabled]="isFieldDisabled[field.id][i]"
-              >
-                <app-entity-field-view
-                  [field]="field"
-                  [entity]="entityToMerge"
-                ></app-entity-field-view>
-              </mat-checkbox>
-            } @else {
-              <mat-radio-group>
-                <mat-radio-button
-                  [value]="entityToMerge[field.id]"
-                  (click)="selectExistingValue(field.id, i)"
-                  [checked]="existingFieldSelected[field.id][i]"
-                  [disabled]="isFieldDisabled[field.id][i]"
-                >
-                  <app-entity-field-view
-                    [field]="field"
-                    [entity]="entityToMerge"
-                  ></app-entity-field-view>
-                </mat-radio-button>
-              </mat-radio-group>
-            }
-          </div>
-        }
+        <app-merge-fields
+          [field]="field"
+          [entities]="entitiesToMerge"
+          [existingSelected]="existingFieldSelected[field.id]"
+          [isDisabled]="isFieldDisabled[field.id]"
+          [mergedControl]="mergeForm.formGroup.get(field.id)"
+          [fieldEntities]="fieldValues[field.id]"
+          (selectedValue)="
+            selectExistingValue(
+              $event.fieldId,
+              $event.entityIndex,
+              $event.checked
+            )
+          "
+        >
+        </app-merge-fields>
 
-        <!-- Merge Preview -->
         <div class="preview-cell">
           <app-entity-field-edit
             [field]="field"
--- a/src/app/features/de-duplication/bulk-merge-records/bulk-merge-records.component.ts
+++ b/src/app/features/de-duplication/bulk-merge-records/bulk-merge-records.component.ts
@@ -21,9 +21,9 @@ import {
 import { AbstractControl, ReactiveFormsModule } from "@angular/forms";
 import { MatError } from "@angular/material/form-field";
 import { MatCheckboxModule } from "@angular/material/checkbox";
-import { th } from "@faker-js/faker";
+import { MergeFieldsComponent } from "../merge-field/merge-field.component";
 
-type MergeField = FormFieldConfig & { allowsMultiValueMerge: boolean };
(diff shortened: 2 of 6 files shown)


src/app/features/de-duplication/bulk-merge-records/bulk-merge-records.component.html
```diff
@@ -11,40 +11,15 @@
           <strong>{{ field.label }}</strong>
```
sleidig, line 11: Why not put the field label also inside the sub-component?

src/app/features/de-duplication/bulk-merge-records/bulk-merge-records.component.html
```diff
@@ -11,40 +11,15 @@
...
-            }
-          </div>
-        }
+        <app-merge-fields
+          [field]="field"
+          [entities]="entitiesToMerge"
+          [existingSelected]="existingFieldSelected[field.id]"
+          [isDisabled]="isFieldDisabled[field.id]"
+          [control]="mergeForm.formGroup.get(field.id)"
+          (valueChanged)="updateMergePreviewValue(field.id, $event)"
```
sleidig, line 20: I think existingSelected, isDisabled and valueChanged should be possible to handle (create and update) entirely inside the sub-component. Did you try that?

`existingFieldSelected` and `isFieldDisabled` as a status for all fields is probably not needed and each app-merge-fields component can internally track these just for its own field without inputs?

src/app/features/de-duplication/bulk-merge-records/bulk-merge-records.component.ts
```diff
@@ -117,53 +112,12 @@ export class BulkMergeRecordsComponent<E extends Entity> implements OnInit {
...
-      ];
     }
-    return currentValue;
-  }
-
-  private areValuesIdentical(a: any, b: any): boolean {
-    return JSON.stringify(a) === JSON.stringify(b);
   }
 
   private allowsMultiValueMerge(field?: FormFieldConfig): boolean {
```
sleidig, line 123: This can probably also be moved into the merge-fields component, as it only relates to a specific field?

src/app/features/de-duplication/bulk-merge-records/bulk-merge-records.component.ts
```diff
@@ -57,7 +50,6 @@ export class BulkMergeRecordsComponent<E extends Entity> implements OnInit {
    */
   existingFieldSelected: Record<string, boolean[]> = {};
```
sleidig, line 51: not used here anymore, I think?

src/app/features/de-duplication/bulk-merge-records/merge-fields/merge-fields.component.ts
```diff
@@ -0,0 +1,176 @@
...
+    EntityFieldEditComponent,
+  ],
+  templateUrl: "./merge-fields.component.html",
+  styleUrls: ["./merge-fields.component.scss"],
+})
+export class MergeFieldsComponent implements OnInit {
+  @Input() field!: FormFieldConfig;
+  @Input() entities!: Entity[];
+  @Input() control!: AbstractControl;
+  @Output() valueChanged = new EventEmitter<any>();
```
sleidig, line 28: I think this is not used anymore, as we only rely on the control?

src/app/features/de-duplication/bulk-merge-records/merge-fields/merge-fields.component.ts
```diff
@@ -0,0 +1,176 @@
...
+
+  existingSelected: boolean[] = [];
+  isDisabled: boolean[] = [];
+  allowsMultiValue: boolean = false;
+
+  ngOnInit() {
+    this.initializeFieldState();
+    this.control?.valueChanges.subscribe((newValue) => {
+      this.updateSelectedStatus(newValue);
+      const control = this.control.get(this.field.id);
```
sleidig, line 38: isn't `this.control` already the sub-control pointing to this field's value? Why do we need to get a child control of it?

src/app/features/de-duplication/bulk-merge-records/bulk-merge-records.component.html
```diff
@@ -7,44 +7,12 @@
       <div class="header" i18n>Merge Preview</div>
 
       @for (field of fieldsToMerge; track field.id) {
-        <div class="label-cell">
-          <strong>{{ field.label }}</strong>
-        </div>
```
sleidig, line 12: sorry, I now realize why you had this here rather than inside: the  block with the form field is also at this top-level (in my mind I was expecting it inside the merge-fields component).

In this case, actually it does feel right to have the label here in the template :see_no_evil:

sleidig (changes requested):
The UI is somewhat broken now, as the layout is separated and all blocks in the sub-component are displayed below each other.
Having record A + B listed below instead of side by side is actually not bad (especially if we expect to have even more than 2 in the future or work on small screens). But at least the label has to be separate column and the header doesn't fit anymore currently.

Abhinegi2:
> The UI is somewhat broken now, as the layout is separated and all blocks in the sub-component are displayed below each other. Having record A + B listed below instead of side by side is actually not bad (especially if we expect to have even more than 2 in the future or work on small screens). But at least the label has to be separate column and the header doesn't fit anymore currently.

I’ve already fixed the layout issue but forgot to push the changes. For now, I’m keeping the old layout, but we can explore the suggested layout changes in the future if needed

src/app/features/de-duplication/bulk-merge-records/merge-fields/merge-fields.component.ts
```diff
@@ -0,0 +1,172 @@
...
+
+  existingSelected: boolean[] = [];
+  isDisabled: boolean[] = [];
+  allowsMultiValue: boolean = false;
+
+  ngOnInit() {
+    this.initializeFieldState();
+    this.control?.valueChanges.subscribe((newValue) => {
+      this.updateSelectedStatus(newValue);
+      this.control?.patchValue(newValue, { emitEvent: false });
```
sleidig, line 37: does this really make sense? We are waiting for the control value to change, and then patch that same new value to the same control?

sleidig (approved):
Works well :+1:

Issue #2742, opened 2024-12-19
support two-dimensional drag&drop for admin views


sleidig:
see new feature available in Angular 19:
https://material.angular.io/cdk/drag-drop/overview#list-wrapping

The Admin UI (editing list filters, reordering field groups of details view) should support wrapped drag&drop.

sleidig:
blocked by #2792

Abhinegi2:
@george-neha Can you please test the following on the:(https://pr-2921.aam-digital.net/)  

1. Try dragging and dropping field groups in the details view (`Edit Data Structure → Details View`).  
2. Try dragging and dropping filters in the list view (`Edit Data Structure → List View`).  

Let me know if you encounter any issues.

PR #2921, opened 2025-03-24
 support two-dimensional drag&drop for admin views 

Abhinegi2:
closes: #2742


sleidig (approved):
Seems to work well! :heart_eyes: 

But why do we actually not need the `CdkDropListGroup` anymore while we get more functionality, @Abhinegi2 ? :sweat_smile:

Issue #2595, opened 2024-10-02
"Duplicate" should also be available as Entity Action from EntityDetails menu


sleidig:
Currently "cloning" is only available from "bulk actions" in the list view. This should also be possible for a single record from its details view.

sleidig:
see https://github.com/Aam-Digital/ndb-core/blob/6c0449d1cca9a0aff7ddbaa139e3790b53afb68c/src/app/core/entity-details/entity-actions-menu/entity-actions-menu.service.ts#L16

PR #2924, opened 2025-03-26
"Duplicate" should also be available as Entity Action from EntityDetails menu 

Abhinegi2:
closes: #2595

### Visible/Frontend Changes

- I added a duplicate action in the details view.
- Also added a success message while duplicating records

### DIFF
--- a/src/app/core/entity-list/duplicate-records/duplicate-records.service.ts
+++ b/src/app/core/entity-list/duplicate-records/duplicate-records.service.ts
@@ -14,7 +14,7 @@ export class DuplicateRecordService {
     private alertService: AlertService,
   ) {}
 
-  async duplicateRecord(sourceData: Entity | Entity[]) {
+  async duplicateRecord(sourceData: Entity | Entity[]): Promise<boolean> {
     const entities = Array.isArray(sourceData) ? sourceData : [sourceData];
     const duplicateData = this.clone(entities);
 
--- a/src/app/core/entity-list/duplicate-records/duplicate-records.service.ts
+++ b/src/app/core/entity-list/duplicate-records/duplicate-records.service.ts
@@ -58,7 +58,7 @@ export class DuplicateRecordService {
     } else {
       return $localize`:Entity action confirmation message:${
         sourceData[0].getConstructor().label
-      } "${sourceData.toString()}"  duplicated Successfully`;
+      } "${sourceData.toString()}" duplicated Successfully`;
     }
   }
 }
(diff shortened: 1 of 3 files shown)


src/app/core/entity-list/duplicate-records/duplicate-records.service.ts
```diff
@@ -42,4 +49,16 @@ export class DuplicateRecordService {
...
+
+  generateSuccessMessage(sourceData: Entity[]): string {
+    if (sourceData.length > 1) {
+      return $localize`:Entity action confirmation message:${sourceData.length} ${
+        sourceData[0].getConstructor().labelPlural
+      } duplicated Successfully`;
+    } else {
+      return $localize`:Entity action confirmation message:${
+        sourceData[0].getConstructor().label
+      } "${sourceData.toString()}" duplicated Successfully`;
```
sleidig, line 61:
```suggestion
      } duplicated successfully`;
    } else {
      return $localize`:Entity action confirmation message:${
        sourceData[0].getConstructor().label
      } "${sourceData.toString()}" duplicated successfully`;
```

src/app/core/entity-list/duplicate-records/duplicate-records.service.ts
```diff
@@ -10,11 +11,17 @@ export class DuplicateRecordService {
   constructor(
     private entitymapperservice: EntityMapperService,
     private entityService: EntitySchemaService,
+    private alertService: AlertService,
   ) {}
 
-  async duplicateRecord(sourceData: Entity[]) {
-    const duplicateData = this.clone(sourceData);
-    return await this.entitymapperservice.saveAll(duplicateData);
+  async duplicateRecord(sourceData: Entity | Entity[]) {
```
sleidig, line 17:
```suggestion
  async duplicateRecord(sourceData: Entity | Entity[]): Promise<boolean> {
```

src/app/core/entity-list/duplicate-records/duplicate-records.service.ts
```diff
@@ -10,11 +11,17 @@ export class DuplicateRecordService {
...
+    const entities = Array.isArray(sourceData) ? sourceData : [sourceData];
+    const duplicateData = this.clone(entities);
+
+    await this.entitymapperservice.saveAll(duplicateData);
+
```
sleidig, line 22: Maybe we can navigate to the newly cloned entity? (if duplicateRecord() is called with a special navigateTo parameter that we only set to true for single clone)

sleidig (commented):
Seems to work well and looks good. Thanks for thinking of the confirmation alert also!

Let's try to add navigation for further convenience

sleidig (approved):

Issue #2509, opened 2024-08-14
Notes should be sorted by order of creation if on the same day


sleidig:
**Describe the bug**
If you create multiple notes for the same day, the most recently created record is not showing on top of the list (we only sort by date, without any time information). This is very confusion to users.

**To Reproduce**

1. Go to "Notes"
2. Add a new note for the same date of the last note in the list
3. Find your new note (it's probably not at the top of the list where you would expect it)

**Expected behavior**
Ideally the notes of the same day should be sorted "sensibly". Can we use the "created.at" timestamp as a secondary sorting criteria for those that have the same date, maybe?

sleidig:
Yes, underlying all the lists is `EntitiesTableComponent`. I think we have some special logic to transform columns into sortable values already in src/app/core/common-components/entities-table/table-sort/table-sort.ts, maybe that is a good place to look into

Abhinegi2:
@george-neha could you please test the above-mentioned issue here-  https://pr-2925.aam-digital.net/?

PR #2925, opened 2025-03-26
 Notes should be sorted by order of creation if on the same day 

Abhinegi2:
closes: #2509

### Visible/Frontend Changes

- [x]
- [ ]

- [x]
- [ ]

### DIFF
--- a/src/app/core/common-components/entities-table/table-sort/table-sort.ts
+++ b/src/app/core/common-components/entities-table/table-sort/table-sort.ts
@@ -23,7 +23,17 @@ export function tableSort<OBJECT extends Entity, PROPERTY extends keyof OBJECT>(
   data.sort((objA, objB) => {
     const valueA = getComparableValue(objA.record, active);
     const valueB = getComparableValue(objB.record, active);
-    return compareValues(valueA, valueB);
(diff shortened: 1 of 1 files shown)


Abhinegi2:
@sleidig, I just wanted to confirm that I've also fixed the age column sorting. Do we still need the `disableSortingHeaderForAdvancedFields` method to disable sorting for specific fields, or can I remove it?

src/app/core/common-components/entities-table/entities-table.component.ts
```diff
@@ -453,6 +453,10 @@ export class EntitiesTableComponent<T extends Entity>
    */
   private disableSortingHeaderForAdvancedFields(c: FormFieldConfig) {
     // if no dataType is defined, these are dynamic, display-only components
+    if (c.viewComponent === "DisplayAge") {
+      return;
+    }
+
```
sleidig, line 459:
```suggestion
    if (c.viewComponent === "DisplayAge") {
      // we have implemented support for age specifically
      return;
    }

    // if no dataType is defined, these are dynamic, display-only components
```

sleidig (approved):
Works well :smiley: 

I just generalized the age detection a bit.

Issue #2908, opened 2025-03-17
Show some "Not Found" page when opening entity details of an ID that doesn't exist


sleidig:
More explicitly handle if the user opens the details view for an ID that does not exist.
Currently, the Entity Details view opens empty. However, this is not very clear user feedback and may cause errors in components that get loaded and depend on an entity object being available.

To replicate the problem:
1. open the app with an entity ID that does not exist (e.g. in demo mode: https://demo.aam-digital.com/child/000 )
2. the tabs of the details view are showing a loading bar endlessly
3. --> instead, catch this situation and show a message

Saransh1524:
Hi, @sleidig  I would like to take up this issue. Could you please assign it to me?

sleidig:
Hi @Saransh1524 , welcome to the Aam Digital community 😃   Thanks for offering to contribute!

I have slightly clarified the description of this issue with an example for you. Assigning this here to you to explore and work on a solution.

Please do get back with questions or if you are stuck. Happy to help. Our codebase is fairly complex so it may be challenging to find the right spot sometimes.

Saransh1524:
hi @sleidig, thankyou for assigning the issue. I'll start exploring the codebase and working on a solution. I'll surely reach out to you if i have any questions. Thankyou!!

Saransh1524:
Hi @sleidig,

I have implemented a separate "Entity Not Found" page to handle cases where an entity ID does not exist. When an invalid entity ID is encountered, the user is redirected to this page with the message: "The requested entity ID does not exist. Please check the ID and try again."

Please review the changes and let me know if there are any further adjustments needed. Once confirmed, I will proceed with generating a pull request.

Thanks!

sleidig:
Can you please open a Pull Request here toward our repository for us to review and discuss your implementation?
https://github.com/Aam-Digital/ndb-core/pulls

Saransh1524:
Hi @sleidig i have opened a pull request #2929. Please check and let me know about any further changes that needs to be made
Thanks!

PR #2929, opened 2025-03-27
fix 2908

Saransh1524:
Reference to issue 2908

closes: #2908 

### Visible/Frontend Changes
{8C97A241-4FEC-4684-84BC-465651ABDC30}

This function now navigates to 404 if the entity is not found or an error occurs.
{209E5237-AB0E-4027-B31F-BC2F08939CD9}

- [x] Ensure the details view does not enter an endless loading state when an invalid entity ID is accessed.
- [x] Add error handling for invalid entity IDs in the backend to ensure components depending on the entity object do not crash.


src/app/app.routing.ts
```diff
@@ -56,7 +56,7 @@ export const allRoutes: Routes = [
   },
   { path: "login", component: LoginComponent },
   { path: "404", component: NotFoundComponent },
-
+  { path: 'entity-not-found', component: EntityNotFoundComponent },
```
sleidig, line 59: As you can see above (line 58) we already have a `NotFoundComponent`. I think we can reuse the same UI instead of introducing a similar component additionally. Probably you can just route to /404 to even reuse the same existing paths, @Saransh1524

src/app/app.module.ts
```diff
@@ -95,6 +95,7 @@ import { NotificationService } from "./features/notification/notification.servic
 import { AngularFireModule } from "@angular/fire/compat";
 import { FirebaseConfiguration } from "./features/notification/notification-config.interface";
 
+
```
sleidig, line 98:
```suggestion
```

revert to minimize code changes

src/app/core/entity-details/abstract-entity-details/abstract-entity-details.component.ts
```diff
@@ -60,18 +60,29 @@ export abstract class AbstractEntityDetailsComponent implements OnChanges {
...
-      if (this.ability.cannot("create", this.entityConstructor)) {
-        this.router.navigate([""]);
-        return;
+    try {
+      if (this.id === 'new') {
+        if (this.ability.cannot('create', this.entityConstructor)) {
+          this.router.navigate(['']);
+          return;
+        }
+        this.entity = new this.entityConstructor();
```
sleidig, line 69: Following best practices, let's only wrap the smallest possible part of the code into a try-catch. I think the part for "new" entities is not related to your issue. It will be enough to just try-catch the `entityMapperService.load` call?

sleidig (changes requested):
Thanks, @Saransh1524 ! :smiley: 

This seems to work well. I just have a few small suggestions to align the code even further with our project

Saransh1524:
Hi @sleidig ,

Thank you for your feedback. The changes you mentioned are very helpful, and I appreciate the detailed suggestions. I will work on implementing these updates and will update the pull request soon.

Saransh1524:
Hi @sleidig,

I wanted to discuss the changes I am planning to implement based on the requirements. Here is my approach:

1. I will use a try-catch block only for the entityMapperService.load call.

2. Instead of creating a new UI for the NotFoundComponent, I will reuse the existing UI.  And make it route to 404. This will resolve the infinite loding bar problem.

3. The message displayed by the NotFoundComponent will be dynamic and change according to the situation. To achieve this, I plan to pass a query parameter reason with a value indicating that the entity was not found while routing to the 404 page.
The NotFoundComponent will be modified to check for the query parameter. If the parameter exists, the message will change to "The entity you are looking for does not exist." The message will be dynamic inside the not-found.component.html, changing based on the situation.

4. The route for entityNotFound will be removed along with the component files, and the NotFoundComponent UI will be reused.
I believe this approach will address the issues and pass the necessary checks. Please share your feedback on this implementation plan.

Thank you.

sleidig:
Hi @Saransh1524 , this sounds good to me :+1: 

As there already is a routing for /404, your step (2) will be very simple.

Actually, I am not sure a different text for "entity not found" compared to the "normal" not found is adding much information: Users never type an ID directly (and often do not even understand the concept, as it is somewhat technical). Probably the simple "this page does not exist or you do not have permission", which is the existing `NotFoundComponent` is enough

Saransh1524:
I've reverted the unnecessary changes and applied the requested updates. Let me know if further modifications are needed! @sleidig

src/app/app.routing.ts
```diff
@@ -56,7 +56,6 @@ export const allRoutes: Routes = [
   },
   { path: "login", component: LoginComponent },
   { path: "404", component: NotFoundComponent },
-
   {
```
sleidig, line 59:
```suggestion

  {
```

src/app/core/entity-details/abstract-entity-details/abstract-entity-details.component.ts
```diff
@@ -57,21 +57,31 @@ export abstract class AbstractEntityDetailsComponent implements OnChanges {
...
-
   protected async loadEntity() {
     this.isLoading = true;
-    if (this.id === "new") {
-      if (this.ability.cannot("create", this.entityConstructor)) {
-        this.router.navigate([""]);
+  
+    if (this.id === 'new') {
+      if (this.ability.cannot('create', this.entityConstructor)) {
+        this.router.navigate(['']);
```
sleidig, line 65: Sorry to bother you with another minor styling point: We use [eslint](https://eslint.org/) and prettier to ensure the code is looking consistent, making it easier to read. (For example, this ensures all strings use " instead of ' quotes).

You can run the code checks with `npm run lint`. Often it can even fix things automatically with `npm run lint --fix`.

I also find the GitHub PR tab "Files changed" useful: You can have a look at all lines of code you changed. And if you notice some changes that are just changing the formatting and are not strictly necessary, you can go back and revert these.
... These are minor things but we require them so that our big open source code base remains clean and readable.

 Could you format your code according to this, please @Saransh1524 ?

src/app/core/entity-details/abstract-entity-details/abstract-entity-details.component.ts
sleidig: Thanks for adjusting this :+1:

sleidig (changes requested):
Thanks for the changes! Just need to pass "linting" now.

src/app/core/entity-details/abstract-entity-details/abstract-entity-details.component.ts
Saransh1524: sure, i will make the changes

Saransh1524: @sleidig i made the linting changes requested by you but  the command (npm run lint) shows a lot of errors related to Delete ␍. I am actually not sure what that is. can you help me with this

Saransh1524:
@sleidig Sorry for the late changes!.  I fixed the linting issues and pushed them, could you please check if everything looks good and works fine?

.github/workflows/on-official-release-push.yml
```diff
@@ -1,6 +1,5 @@
 name: Semantic Release
-on:
-  workflow_dispatch
+on: workflow_dispatch
```
sleidig, line 2:
```suggestion
on:
  workflow_dispatch
```

src/app/core/config/dynamic-routing/empty/application-loading.component.scss
```diff
@@ -4,5 +4,5 @@
   place-items: center;
 
   text-align: center;
-  margin: 2em
+  margin: 2em;
```
sleidig, line 7:
```suggestion
  margin: 2em
```

sleidig (approved):
Sorry for the slow feedback @Saransh1524 , we have had a very busy week.

Thank you for cleaning this up and figuring out the linting!

I am undoing a few unrelated formatting changes, to keep the final commit simpler.

Your code looks good now and the functionality is working as expected. Nice job! :smiley:

Issue #2905, opened 2025-03-12
Creating a new "related" note from a details view wrongly links it twice


sleidig:
**Describe the bug**

1. Precondition: config for `entity:Note` has the "relatedEntities" field enabled to link to some other record type (e.g. `"relatedEntities": { "label": "Related Records", "dataType": "entity", "additional": "RecurringActivity", "isArray": true }`)
   - dev.aam-digital.net is set up like this currently 
3. Open the Details view of a child
4. Go to "Notes" tab in the details view and click the add button to create a new note that is linked to this child
5. Now the Child is selected both in the "Children" and the "Related Entities" fields:![Image](https://github.com/user-attachments/assets/ac1bea91-6e2f-45c0-8e12-765bd9f054ed)

It should only link in the "children" field. The "relatedEntities" field doesn't even allow this entity type in its selection, actually.

(-> probably to be fixed in `RelatedNotesComponent`)

Abhinegi2:
@sleidig  I think we need to deploy this fix on the dev site for testing, as related entity fields are not visible in demo mode when linking/creating a note to a child.

PR #2933, opened 2025-03-31
Creating a new "related" note from a details view wrongly links it twice

Abhinegi2:
closes:  #2905

### Visible/Frontend Changes

- [x]
- [ ]

- [x]
- [ ]


src/app/child-dev-project/notes/notes-related-to-entity/notes-related-to-entity.component.ts
```diff
@@ -90,12 +90,14 @@ export class NotesRelatedToEntityComponent
...
       }
 
-      for (const e of [
-        this.entity.getId(),
-        ...this.getIndirectlyRelatedEntityIds(this.entity),
-      ]) {
-        if (!newNote.relatedEntities.includes(e)) {
-          newNote.relatedEntities.push(e);
+      if (this.entity.getType() !== "Child") {
+        for (const e of [
```
Abhinegi2, line 94: This was adding all entity types, including Child, to the `relatedEntities` array. To prevent that, I'm now excluding the Child entity, and it is working fine.

src/app/child-dev-project/notes/notes-related-to-entity/notes-related-to-entity.component.ts
```diff
@@ -90,12 +90,14 @@ export class NotesRelatedToEntityComponent
...
+          this.entity.getId(),
```
Abhinegi2, line 95: This was adding all entity types, including Child, to the relatedEntities array. To prevent that, I'm now excluding the Child entity, and it is working fine.

sleidig: I investigated this a bit further myself, as I couldn't make sense why this is happening. Adjusted the unit tests to cover the issue also.

I think your workaround is not sufficient, unfortunately. This bug can also occur with different entity types. We probably have to somehow check if this.entity.getId() is already linked into the new note somewhere and only if not, include it in this for loop

Abhinegi2: I’ve added a helper method to include `this.entity.getId()` when it’s not linked with another entity. The new tests are now passing, but one of the previous tests is failing. The test is as follows:

```
entity = createEntityOfType("User");
component.entity = entity;
component.filter = undefined;
component.property = undefined;
await component.ngOnInit();
note = component.createNewRecordFactory()();
expect(note.relatedEntities).toEqual([entity.getId()]);
```

Previously, in the above test the `note` looked like this:
```
authors: ['User:7766f008-83d9-4a91-9857-b473be6565ca']
children: []
relatedEntities: ['User:7766f008-83d9-4a91-9857-b473be6565ca']
schools: []
```

After adding the helper method, it now looks like this:

```
authors: ['User:8a8d36fd-bf98-4f06-b09b-0f8fc8c37dfa']
children: []
relatedEntities: []
schools: []
```

**Can you confirm whether this test code block is still valid in any scenario?** (because now we are not adding those entity id which are already mapped to another column)

sleidig: oh thanks, that's an interesting point! Yes, I guess the test should be changed. Otherwise there also it is linked twice. And as "User" entity, it should be linked as "authors".

src/app/child-dev-project/notes/notes-related-to-entity/notes-related-to-entity.component.spec.ts
```diff
@@ -65,14 +66,15 @@ describe("NotesRelatedToEntityComponent", () => {
...
+    expect(note.relatedEntities).not.toContain(entity.getId());
 
     entity = createEntityOfType("User");
     component.entity = entity;
     component.filter = undefined;
     component.property = undefined;
     await component.ngOnInit();
     note = component.createNewRecordFactory()();
-    expect(note.relatedEntities).toEqual([entity.getId()]);
+    expect(note.relatedEntities).not.toContain(entity.getId());
```
sleidig, line 77:
```suggestion
    expect(note.authors).toContain(entity.getId());
    expect(note.relatedEntities).not.toContain(entity.getId());
```

src/app/child-dev-project/notes/notes-related-to-entity/notes-related-to-entity.component.ts
```diff
@@ -103,6 +103,18 @@ export class NotesRelatedToEntityComponent
     };
   }
 
+  /**
+   * check if an entityId is already referenced in any array properties of the given note
+   */
+  private isAlreadyLinked(newNote: any, id: string): boolean {
+    for (const key in newNote) {
+      if (Array.isArray(newNote[key]) && newNote[key].includes(id)) {
```
sleidig: What if the other field, like "author", would only allow a single id, not an array?

sleidig (approved):

PR #2934, opened 2025-03-31
fix(core): keep PouchDatabase changesFeed on resets to keep interconn…

### LINKED ISSUE #2890: Unable to Create New Entity Type as Demo Admin in demo mode(only)
### Describe the bug
In demo mode/demo-site, when logged in as demo-admin, attempting to create a new entity type fails, and an error appears in the console. However, the same process works correctly when performed by a normal user.

### To Reproduce

1. Log in as demoadmin in the demo mode.
2. Navigate to Administer Entity Types -> Click Add New Entity Type.
3. Enter "test" as the entity name.
4. Observe that the entity is not created, and the following console error appears:
```
Error in .on("change", function): TypeError: Cannot read properties of undefined (reading 'next')
    at Yt.<anonymous> (pouch-database.ts:221:59)
```
5. Log out and log in as a normal user.
6. Go to https://demo.aam-digital.com/admin/entity.
7. Add a new Entity.

The entity is successfully created now.

### Expected behavior
The new entity type should be created successfully for demoadmin.

### LINKED ISSUE #2932: General Settings update not getting saved in Demo Mode
**Describe the bug**
In [demo mode](https://demo.aam-digital.com/), attempting to update general settings for any entity (e.g., Label, Label Plural) displays a success alert ("Configuration Updated"). However, the changes are not actually applied.

**To Reproduce**
**Logged in as demo admin**

1. Navigate to the 'Child' entity.
2. Click on 'Edit Data Structure' → 'General Settings'.
3. Modify the Label and save the changes.
4. Verify the update—it does not persist, and the changes are not reflected.

showing this error in the console - 

Image
Image

**Expected behavior**
The updated Label should be saved and reflected in the system.

**It seems like the issue only occurred when we logged in as demo-admin**

sleidig:
…ections

otherwise demo mode faces some problems after logouts

fixes #2932, fixes #2890

### DIFF
--- a/src/app/core/database/pouchdb/pouch-database.ts
+++ b/src/app/core/database/pouchdb/pouch-database.ts
@@ -252,8 +252,7 @@ export class PouchDatabase extends Database {
    */
   async reset() {
     this.pouchDB = undefined;
-    // keep this.changesFeed because some services are already subscribe to this reference and will find updates if we recreate a new subject
-
+    // keep this.changesFeed because some services are already subscribed to this reference
     this.databaseInitialized = new Subject();
   }
 
--- a/src/app/core/database/pouchdb/pouch-database.ts
+++ b/src/app/core/database/pouchdb/pouch-database.ts
@@ -202,8 +202,8 @@ export class PouchDatabase extends Database {
   }
 
   /**
-   * Listen to changes to documents which have an _id with the given prefix
-   * @param prefix for which document changes are emitted
+   * Listen to changes to documents in the database.
+   * Use rxjs operators to filter for specific prefixes etc. if needed.
    * @returns observable which emits the filtered changes
    */
   changes(): Observable<any> {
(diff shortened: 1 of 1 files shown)


Abhinegi2 (approved):
The functionality is working fine. 

The issue was that we were setting `this.changesFeed = undefined` after ensuring it was completed?

sleidig:
> The functionality is working fine.
> 
> The issue was that we were setting `this.changesFeed = undefined` after ensuring it was completed?

Yes, @Abhinegi2 .

During startup of the system, some central services, like ConfigService subscribe to the "changes" feed of the database. When we reset that subject to undefined and then recreated it when logging in as demo-admin, then the ConfigService was still remaining subscribed to the old observable of the first database initialization (because we don't reload the page with all services completely in demo mode). So now, there was (a) a call to undefined leading to the error message and (b) even after fixing that, the ConfigService didn't listen to the correct changes feed anymore and didn't notice the new updates.

PR #2941, opened 2025-04-07
Refactor/simplify account backend

sleidig:
closes: https://github.com/Aam-Digital/account-backend/issues/26

after this, the [account-backend API](https://github.com/Aam-Digital/account-backend) can be deprecated and archived. User management requests are made directly to the Keycloak API. The requests and process of steps with Keycloak are moved from the existing implementation of that backend.

--> deployed to https://dev.aam-digital.net for testing

### TODOs
- [x] error handling (e.g. show error if email already exists)

### TODO in server configs:
- [x] CORS :warning: 
   - This is currently facing CORS issues. We have to add `access-control-allow-origin` headers (maybe for all aam-digital subdomains, [something like this stackoverflow?](https://stackoverflow.com/questions/1653308/access-control-allow-origin-multiple-origin-domains)). @tomwwinter ?
   - --> https://github.com/Aam-Digital/ndb-setup/pull/67
- [ ] update default realm (ndb-setup) and current realms:
   - [x] add "(realm-management) manage-users" and "(realm-management) view-realm" roles as "associated roles to "account_manager role
   - [ ] Client Scope > Mappers: "client roles" (to add `"realm-management": { "roles": [] }`) to token
   (and this Client Scope added to the "app" Client)


Abhinegi2:
I have tested the functionality and it seems that account creation, role changes, and email updates are working as expected. However, I have two doubts:
1. After sending the invitation email, the message says, "After verifying your email, you can open the Application with the link below. Open Aam Digital." The URL in the email currently points to .com, but it should be https://dev.aam-digital.net/.
2. By default, I have the **user_app** role, but I’m still able to update users. Just wanted to confirm—shouldn't this action be restricted to users with the **account_manager** role?

sleidig:
Thanks for the thorough testing.

> 1. After sending the invitation email, the message says, "After verifying your email, you can open the Application with the link below. Open Aam Digital." The URL in the email currently points to .com, but it should be https://dev.aam-digital.net/.(basically, the link should reflect the domain where the account was created)

This is a limitation of our dev system. Can be disregarded here.

> 2. By default, I have the **user_app** role, but I’m still able to update users. Just wanted to confirm—shouldn't this action be restricted to users with the **account_manager** role?

The permissions to create a "User" entity and to add an actual account to log in are given separately. The "account_manager" role only guards the power of creating the account. This may be a bit confusing however. Should consider maybe linking this :thinking:

Issue #2694, opened 2024-12-02
Automated Status-Updates based on linked entity


sleidig:
_As a user I want the system to keep related status of different records automatically up-to-date so that I don't forget and end up with inconsistent, outdated data._

## Scenario
- entity types: Mentor, Mentee, Mentorship
- Mentorship contains fields linking to a Mentor and a Mentee
- Mentorship "status" field with different dropdown options (e.g. "active", "finished")
- Mentee and Mentor "status" fields with different dropdown options (e.g. "open for mentorship", "in mentorship", "alumni")

## Goal
For users, the status of the Mentorship and its linked Mentee and Mentor are logically related. Updating the status of the Mentorship should automatically update the related Mentor's and Mentee's status as well:

1. When I create a new Mentorship, I want the linked Mentor's and Mentee's status to change to "in mentorship".
2.  When I change the Mentorship's status to "finished", I want the linked Mentor's and Mentee's status to change to "alumni".

### Edge cases
3. When the Mentor has a second Mentorship that is still active, I actually want to keep the Mentor's status as "in mentorship" even when "finishing" one of the Mentorships.
4. When I have talked to the Mentor before and she is willing to take up a new Mentorship again, I may want to change the Mentor's status to "open for mentorship" instead of "alumni".
5. When I have previously set the Mentor's status manually back to "open for mentorship" for some exceptional reason, I may want the Mentor's status to stay in that status instead of being changed to "alumni" when "finishing" the Mentorship.

--> User needs to be aware of the automated updates and have control over them to possible overrule the automation.

-----
# Suggested Functionality

## Semi-automatic status updates
After saving the Mentorship with a new status, the following process gets triggered automatically:
1. A dialog popup appears, listing affected linked status:
   1. "You have updated the _status_ of this record. The _status_ is linked to the following records' _status_. Do you want to update the following?" 
   2. within the dialog the affected linked records are listed
   3.  in our example:
       - Mentor _Jane Doe_ [current status: "in mentorship"] [new status: "alumni"]
       - Mentee _John Doe_ [current status: "open to mentorship"] [new status: "open to mentorship"]
   4. the "new status" field is an editable form field that allows to manually select any status from dropdown
   5. the "new status" is pre-filled with the default automation mapping (see configuration below)
   6. dialog buttons "Update related fields", "Do not update" let the user confirm the process
2. The system automatically edits and saves the affected records (Mentee, Mentor) according to the dialog result
[...]

PR #2942, opened 2025-04-08
Feat/automate update related field

Abhinegi2:
closes: #2694

### Todo

- [x] update `entityformcomponent` do not use `automatedservice` directly?
- [x] Optimize `automatedservice`?
- [x] Update automateddialogcomponent to use newly saved rules ("when"->" then")


george-neha:
Few suggestions: 
1. The new icon can be renamed "Automated rule configurator."
2. From a UI perspective, maybe the font size for "Configure mapping" can be reduced? Right now, the font size and the button size are mismatched.
3. The "select related field" field is not intuitive in the configure mapping screen. As a user, I'm trying to capture the configuration for the dropdown values directly.  How about highlighting the field in bold/ or a different colour scheme, or an underlying text urging the user to select the field before configuring the dropdown values?
4. Text suggestion for configure mapping screen- "You can create an automation rule for this field. Once set up, any changes will automatically update the related field, ensuring data consistency and reducing manual work. Please select the field and configure the mapping carefully to avoid unintended updates."
@Abhinegi2

src/app/core/entity/schema/default-value-config.ts
```diff
@@ -41,5 +41,6 @@ export type DefaultValueMode =
 export interface AutomatedConfigRule {
   relatedEntity: string;
   relatedField: string;
+  mappedProperty: string;
   automatedMapping: { [key: string]: string };
 }
```
sleidig, line 46:
```suggestion
/**
 * Special default value rule to continuously updates the value based a field of another related entity.
 */
export interface AutomatedConfigRule {
  /**
   * ID of the related entity that triggers updates of this value
   */
  relatedEntityId: string;

  /**
   * field ID of the relatedEntity that holds the reference to this field's entity.
   * 
   * e.g. if a child's field should be updated based on a school's value then
   * this is the field on the school entity of type "entity", holding the child's entity for the relevant relation
   */  
  relatedReferenceField: string;

  /**
   * field ID of the relatedEntity which triggers updates of this value
   */
  relatedTriggerField: string;

  /**
   * Key-Value map defining "when-then" rules to update the value, where
   * keys are values of the relatedEntity's relatedTriggerField and
   * mapped values are the new values of this field that get updated automatically
   */
  automatedMapping: { [key: string]: string };
}
```

src/app/core/entity/schema/default-value-config.ts
```diff
@@ -41,5 +41,6 @@ export type DefaultValueMode =
 export interface AutomatedConfigRule {
   relatedEntity: string;
   relatedField: string;
+  mappedProperty: string;
```
sleidig, line 44: show dropdown in admin UI to allow user to select (if multiple fields link to the entity type)

src/app/features/automated-status-update/automated-status-update.component.ts
```diff
@@ -1,9 +1,82 @@
...
+  entityForm: EntityForm<Entity>;
+  constructor(
+    @Inject(MAT_DIALOG_DATA) public data: { entities: AffectedEntity[] },
+    private dialogRef: MatDialogRef<AutomatedStatusUpdateComponent>,
+    private entityFormService: EntityFormService,
+  ) {}
+
+  async ngOnInit(): Promise<void> {
+    await Promise.all(
+      this.data.entities.map(async (entity) => {
```
sleidig, line 52: this has only very light calculations, I would prefer a linear loop to reduce the nesting and complexity.

(i.e. instead of `Promise.all(this.data.entities.map(...))` doing a `for(let entity of this.data.entities) { ... }` with simple awaits within the for loop)

src/app/core/common-components/entity-form/entity-form.service.ts
```diff
@@ -267,7 +267,6 @@ export class EntityFormService {
     const form: EntityFormGroup<T> = entityForm.formGroup;
 
     this.checkFormValidity(form);
-
     const updatedEntity = entity.copy() as T;
```
sleidig, line 270:
```suggestion

    const updatedEntity = entity.copy() as T;
```

undoing change to simplify git changelog

src/app/core/entity/schema/default-value-config.ts
```diff
@@ -38,8 +38,32 @@ export type DefaultValueMode =
...
+/**
+ * Special default value rule to continuously updates the value based a field of another related entity.
+ */
 export interface AutomatedConfigRule {
-  relatedEntity: string;
-  relatedField: string;
+  /**
+   * ID of the related entity that triggers updates of this value
+   */
+  relatedEntityId: string;
```
sleidig, line 48: setting up a sample config, I now believe this is wrongly labeled:
Actually this holds the "relatedEntityType" (e.g. "School") not a specific ID (like "School:123"). After all, the exact relationships are dynamically loaded from the data.

sleidig (approved):
this part works as expected now :partying_face:

Issue #769, opened 2021-04-14
Hide sections without permissions


TheSlimvReal:
After finishing #768 some entities will not be visible to some user groups. If for example access to the `HealthCheck` subentity is denied, the user will not be able to load the health check data. The sections connected with this data should also be hidden from the user. This should happen automatically based on the users permissions, without further configuration settings.

---
### Implementation steps:
- [x] extend the EntityDetailsComponent to support for an optional config property `permittedUserRoles` (an array of user-role strings) and only display those tabs for which a user has permission.

similar implementation already exists for Dashboard Widgets:
https://github.com/Aam-Digital/ndb-core/blob/e6b40960babd7d8b89fd4bca554de63e24c2a52c/src/app/core/dashboard/dashboard/dashboard.component.ts#L69-L77

TheSlimvReal:
Once the replication backend is finished, [CASL](https://casl.js.org/v5/en/) should also be used in the frontend to enforce visual feedback on permissions

TheSlimvReal:
This is actually still not done, the release was wrongly linked to this issue

dvdaruri:
Hi, I have created a local branch to solve this issue and am ready to push to origin so I can create a pull request. I am getting the error: remote: Permission to Aam-Digital/ndb-core.git denied to dvdaruri. May I get access so I can push to create a branch?

sleidig:
Hi @dvdaruri , 
thanks for contributing to Aam Digital!

As a new external contributor, you will have to create a PR from a "fork" of the repository. This guide is describing the process fairly well: https://codex.so/fork-and-pull-en

Sorry for the extra steps. Do let me know if you run into any problems or questions! If you end up contributing regularly, we can also add you to our core repository to push directly.

PR #2946, opened 2025-04-15
Completing task: Hide sections without permissions # 769

dvdaruri:
closes: #769 

This code mimics the code provided for the task regarding dashboard user roles and duplicates it. The appropriate sections should now be hidden from users when utilized.

### DIFF
--- a/src/app/core/entity-details/entity-details/entity-details.component.ts
+++ b/src/app/core/entity-details/entity-details/entity-details.component.ts
@@ -23,14 +23,61 @@ import { AblePurePipe } from "@casl/angular";
 /**
  * This component can be used to display an entity in more detail.
  * It groups subcomponents in panels.
-	@@ -62,6 +67,16 @@ export class EntityDetailsComponent
+ * Any component that is registered (has the `DynamicComponent` decorator) can be used as a subcomponent.
+ * The subcomponents will be provided with the Entity object and the creating new status, as well as its static config.
+ */
+@RouteTarget("EntityDetails")
+@UntilDestroy()
+@Component({
+  selector: "app-entity-details",
+  templateUrl: "./entity-details.component.html",
+  styleUrls: ["./entity-details.component.scss"],
+  imports: [
+    AblePurePipe,
+    MatButtonModule,
+    MatMenuModule,
+    FontAwesomeModule,
+    Angulartics2OnModule,
+    MatTabsModule,
+    TabStateModule,
+    MatTooltipModule,
+    MatProgressBarModule,
+    NgIf,
+    NgForOf,
+    ViewTitleComponent,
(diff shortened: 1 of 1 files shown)


src/app/core/entity-details/entity-details/entity-details.component.ts
```diff
@@ -19,7 +19,12 @@ import { RouteTarget } from "../../../route-target";
 import { AbstractEntityDetailsComponent } from "../abstract-entity-details/abstract-entity-details.component";
 import { ViewActionsComponent } from "../../common-components/view-actions/view-actions.component";
 import { AblePurePipe } from "@casl/angular";
-
+import { EntityAbility } from "../../permissions/ability/entity-ability";
+import { Router } from "@angular/router";
+import { EntityRegistry } from "../../entity/database-entity.decorator";
+import { UnsavedChangesService } from "../form/unsaved-changes.service";
+import { EntityMapperService } from "../../entity/entity-mapper/entity-mapper.service";
+import { SessionSubject } from "../../session/auth/session-info";
```
sleidig, line 27:
```suggestion
import { SessionSubject } from "../../session/auth/session-info";

```

sleidig (changes requested):
Sorry for the slow review @dvdaruri .

This is a good starting point but at the moment you are not calling the new method yet so it does not have any effect in the app.

For testing, I would suggest to change the `config-fix.ts` (which is the demo configuration that is automatically loaded locally) like this:
```typescript
// find the following section in "view:child/:id"
      panels: [
        // ... some other tabs are here also
        {
          title: $localize`:Panel title:Attendance`,
          components: [
            {
              title: "",
              component: "GroupedChildAttendance"
            }
          ],
// add the following line:
          permittedUserRoles: ["admin_app"]
        },
```
The goal then is that when opening the details view of a specific child, the "Attendance" tab will not be visible with the normal "demo" account. (While, if you log out and log in with the "demo-admin" account, it should then be visible)


-----
You probably still need to look at two aspects in your code:
1. understand the exact structure of the config and find where the "permittedUserRoles" property is added to the config (I would like format above). 
   - Our dynamic component system is automatically setting the config objects as `@Input` properties to the component. For example `this.panels` is holding the array of panels that you find in the config-fix.ts. Your new "permittedUserRoles" should be found within that "panels" config then. (Maybe use breakpoints or console.log statements to debug and understand the structure)
2. Filter the `panels` array being displayed by this component, based on your method that checks the role.

sleidig:
Hi @dvdaruri , thanks again for your contribution so far! Are you still planning to continue working on this PR or should we hand this over to someone else to try and complete it?

dvdaruri:
Hi you can hand it over, sorry I have been busy with work.

Issue #2821, opened 2025-01-29
Keep sorting when navigating back to list (like filters)


sleidig:
We persist filters when jumping to a details page and then navigating back to the list already (by holding them as part of the URL query params (e.g. "demo.aam-digital.com/child?center=alipore"). We should consider the same for the column by which the list is sorted.

sleidig:
This issue hasn't been describe in a lot of detail, sorry about that @anandwetzel. We are in particular looking at the main lists (Children, Schools, etc. in the main menu of the demo config), implemented through the `EntityListComponent`.

The additional, embedded lists implemented with `RelatedEntitiesComponent` (and the components extending that) could be skipped for now also.

To persist selected filters as url params the `FilterComponent` is already doing something similar. Maybe that can give some inspiration. (And maybe things need to be tested to ensure both filter and sort params are working properly in combination)

PR #2953, opened 2025-04-17
Sorting #2821

anandwetzel:
closes: #2821

### Visible/Frontend Changes
- Keep sorting when navigating back to the list (based on column)

- Use URL query params to persist sorting


sleidig (commented):
Hi @anandwetzel , thanks for contributing to Aam Digital! :smiley: 

This looks like a good approach. You still seem to be missing a way to call your `onSortChange` method. (With our inheritance of different component this is a bit convoluted ...). You could for example subscribe to your own sortChange event for that (see my comment / code suggestion). That makes it successfully write the sort to the url for me.

Could you move this from the `ChildSchoolOverviewComponent` (which is a very special sub-component used only in one place) to the `EntityListComponent`? (see [my comment in the issue](https://github.com/Aam-Digital/ndb-core/issues/2821#issuecomment-2830718993) also)

sleidig:
Hi @anandwetzel , thanks again for your contribution so far! Are you still planning to continue working on this PR or should we hand this over to someone else to try and complete it?

Issue #2880, opened 2025-02-27
Long field labels cannot be read fully


sleidig:
If a form field has a longer text as its label, the label is simply cut off at some point and not fully readable. We should offer better support for these cases.

Options (to be discussed) could be:
- Tooltip over the label to show the full text
- Prevent users to configure labels that are longer than can be displayed (set max-length in Admin UI)
- Show labels with line break, if necessary
- ... other options?

-----
To reproduce:
1. Open the details view of an entity, e.g. a specific school's profile
2. "Edit data structure" (through the three dot menu at the top right)
3. Edit a field (from the "Details View & Fields" section of the Admin UI) by hovering over a field and clicking the "Edit Field" button
4.  Change the "Label" for that field to a longer text
5. Save the changes and see that the label is not fully shown in the form:

![Image](https://github.com/user-attachments/assets/04684cd4-4410-47be-af38-818d72f34c22)

PR #2957, opened 2025-04-23
Long field labels cannot be read fully

Abhinegi2:
closes:#2880


Abhinegi2:
- I’ve currently added a hover effect on the label so that the full text appears on hover, like this:

image

- [ ] I’m also working on adding ellipsis (...) for truncated labels.

sleidig:
This looks good! :+1: 

Is a max-length still necessary if longer labels are displayed like this?

sleidig (changes requested):
It seems the tooltip location is at the end (?) of the unshortened label ... sometimes far from the actually visible label and the current cursor location?

image

sleidig:
> Updated the tooltip position as above. Please check the ss and let me know if it makes sense now?
> 

Did this improve the display for you? I do not see any difference yet in my browser (chromium on linux).

I think the issue is the span (and also the <mat-label> but not the <label>) still having the full width of the long text and the tooltip is therefore displaying at the middle if this. If shorting the css width is not feasible, tooltip position "before" would probably give a reasonable result also

sleidig (approved):

Issue #2835, opened 2025-02-04
Default values of filters should support placeholders like "$current_user"


sleidig:
_As a admin I want to filter lists to match only records that link the current logged in user, to make the interface as simple as possible for users._

Entity lists can display filters above the table to let users narrow down the results. These are configured in the overall system config (Config:CONFIG_ENTITY doc in database / config-fix.ts demo config). We can already define a default value for a filter, to pre-select and pre-filter automatically (e.g. `filters: [{ "id": "authors", "default": "User:ff8b8bf8-ca1b-4ea4-82ff-4491d1c4ee06" }]`). This system should also suppport placeholders, like we use them for entity field defaultValue definitions.

e.g. support a `FilterConfig` definition like this:
```
{
  "id": "authors",
  "default": "$current_user"
}
```

(this should support all "dynamic" default values, i.e. also "$now")

**Proposed Solutions & Alternatives**
Reuse the implementation of the `DynamicPlaceholderValueService` and integrate this into the `FilterGeneratorService`.
Write unit tests first and develop this following the TDD approach.

PR #2958, opened 2025-04-23
Added placeholder support for default filter values

apkoch:
closes: #2835 

### Visible/Frontend Changes

- Filters now support placeholders as default values

- Integrated DynamicPlaceholderValueService placeholder fill implementation into FilterGeneratorService
- Added unit tests adapted from DynamicPlaceholderValueService tests


src/app/core/filter/filter-generator/filter-generator.service.spec.ts
```diff
@@ -20,12 +19,14 @@ import { ConfigurableEnumFilter } from "../filters/configurableEnumFilter";
...
+  EntitySchemaField,
+  PLACEHOLDERS,
+} from "../../entity/schema/entity-schema-field";
+import { CurrentUserSubject } from "app/core/session/current-user-subject";
 
 describe("FilterGeneratorService", () => {
-  let service: FilterGeneratorService;
-  let filterService: FilterService;
-
+  let service: FilterGeneratorService, filterService: FilterService;
```
sleidig, line 29:
```suggestion
  let service: FilterGeneratorService;
  let filterService: FilterService;
  
```

Let's keep this a bit more verbose and clearly separated to make it easier to read and consistent across our code base.

src/app/core/filter/filter-generator/filter-generator.service.spec.ts
```diff
@@ -138,19 +129,15 @@ describe("FilterGeneratorService", () => {
...
     await TestBed.inject(EntityMapperService).saveAll([school1, school2]);
     const csr1 = new ChildSchoolRelation();
     csr1.schoolId = school1.getId();
     const csr2 = new ChildSchoolRelation();
     csr2.schoolId = school2.getId();
-    const csr3 = new ChildSchoolRelation();
-    csr3.schoolId = school2.getId();
-    const csr4 = new ChildSchoolRelation();
-    csr4.schoolId = school1.getId();
+    const [csr3, csr4] = [csr2, csr1];
```
sleidig, line 140: I am not sure this is equivalent to the previous code? Doesn't this just assign the object references now instead of a new object / deep copy?

sleidig (commented):
Thanks for contributing to Aam Digital, @apkoch ! :smiley: 

Works as expected :100:  
I love that you also wrote a unit test!

The date filter is a special component and doesn't seem to work with this approach for PLACEHOLDERS.NOW at the moment. But that's okay considering it's different logic. The important use case was for the current user.

If you feel ambitious, you could also do a small refactoring to avoid duplicating code from `DynamicPlaceholderValueService`: We could add a new method in that service that just returns the value for a PLACEHOLDERS string. Then reuse that method both in the FilterGeneratorService and the DynamicPlaceholderValueService. That way we are future-proofing this for any additional placeholder that may be added.


-----
Appreciate you trying to simplify the code! To make the reviewing easier (and keep consistent code style across our code base) we prefer to avoid changes to code that's not directly related to the exact issue and test case, however. Could you undo the unrelated changes - especially where now some commented code is left over? (sorry to be nit-picky about such a small point :blush: )

apkoch:
Apologies, I made those changes mainly to comply with codeclimate and have reverted them in my latest commit. I'd love to refactor this to future proof for additional placeholders and attempt to get the date filter working in the future! Unfortunately I won't be able to start these tasks until next week, just so that you're aware.

sleidig:
Thanks, @apkoch . Highly appreciate it! Feel free to pick this up next week and comment if you should get stuck or have any questions.

src/app/core/default-values/dynamic-placeholder-value.service.ts
```diff
@@ -52,4 +53,12 @@ export class DynamicPlaceholderValueService extends DefaultValueStrategy {
         break;
     }
   }
+
+  getDefaultValueString(filterConfig: FilterConfig): string {
```
sleidig, line 57: I would to suggest the following, simplified method signature:

```suggestion
  /**
   * Get the current value for a given placeholder.
   * @param placeholder The placeholder string to resolve.
   * @returns The value for the given placeholder or `undefined` if not found.
   */
  getPlaceholderValue(placeholder: string): string | undefined {
```


Can you also refactor the `setDefaultValue` above to use this new function? We can replace the switch-statement with a call to this new `getPlaceholderValue` method. (Or maybe actually move the switch statement into the new method, so that it can also resolve other placeholders than CURRENT_USER).

src/app/core/filter/filter-generator/filter-generator.service.ts
```diff
@@ -33,6 +36,8 @@ export class FilterGeneratorService {
     private entityMapperService: EntityMapperService,
     private filterService: FilterService,
     private schemaService: EntitySchemaService,
+    private currentUser: CurrentUserSubject,
```
sleidig, line 39: don't need `currentUser` here directly anymore :blush:

sleidig (commented):
Thanks for implementing the feedback and improving this further, @apkoch ! Just added a few more thoughts for your approach.

sleidig:
Hi @apkoch , thanks again for your contribution so far! I think this is almost done with a final polishing.
Are you still up to implementing these last suggested cleanups or should someone from our team finalize it?

apkoch:
So sorry about the delay, I plan on implementing these suggested cleanups in the next few days!

sleidig:
> So sorry about the delay, I plan on implementing these suggested cleanups in the next few days!

Thanks, @apkoch ! Don't hesitate to ping back with thoughts and questions (if you like, I can also invite you to our community Slack). We have refactored some of the related code recently, moving the `DynamicPlaceholderService`, so you'll have to resolve some merge conflicts, unfortunately.

apkoch:
Got it! I would appreciate the invite!
I was having trouble finding the proper form control subclass for a filter to pass into the existing `setDefaultValue` method in `DynamicPlaceholderService`, should I write an additional override for this method to handle filters specifically?

sleidig:
@apkoch , do join us here :slightly_smiling_face: : https://join.slack.com/t/aam-digital/shared_invite/zt-34ms2fhjq-6wZ1~8VcSFKLx5DFDs6qug

sleidig:
> I was having trouble finding the proper form control subclass for a filter to pass into the existing `setDefaultValue` method in `DynamicPlaceholderService`, should I write an additional override for this method to handle filters specifically?

Yeah, the way this services takes a FormControl and only sets a value on that instead of just returning us a replaced value makes it quite complex ...

That's why I suggested to split the existing `setDefaultValue` method and you had created the `getDefaultValueString` (it somehow got lost when resolving the merge conflicts, I think? :grimacing: )
Then here in the filters you shouldn't have to worry about any form controls

src/app/core/filter/filter-generator/filter-generator.service.ts
```diff
@@ -33,6 +35,8 @@ export class FilterGeneratorService {
...
```
sleidig, line 38:
```suggestion
```

src/app/core/filter/filter-generator/filter-generator.service.ts
```diff
@@ -22,6 +22,8 @@ import { DateFilter } from "../filters/dateFilter";
 import { BooleanFilter } from "../filters/booleanFilter";
 import { ConfigurableEnumFilter } from "../filters/configurableEnumFilter";
 import { EntityFilter } from "../filters/entityFilter";
+import { CurrentUserSubject } from "app/core/session/current-user-subject";
```
sleidig, line 25:
```suggestion
```

src/app/core/default-values/x-dynamic-placeholder/dynamic-placeholder-value.service.ts
```diff
@@ -70,4 +54,27 @@ export class DynamicPlaceholderValueService extends DefaultValueStrategy {
...
+  getPlaceholderValue(placeholder: string): string | Date | undefined {
+    switch (placeholder) {
+      case PLACEHOLDERS.CURRENT_USER:
+        let userId = this.currentUser.value?.getId();
+        return userId;
+      case PLACEHOLDERS.NOW:
+        let date = new Date();
+        return date;
+      default:
+        Logging.warn(
```
sleidig, line 72:
```suggestion
        Logging.debug(
```

turns out this happens quite a lot now. So let's tone this down to "debug" log level only (otherwise it'll show up on our central monitoring platform)

sleidig (approved):
Awesome! This works well and looks very clean now.

Thanks for your work and the follow-ups upon the review, @apkoch ! :smiley:

Issue #2278, opened 2024-03-07
Allow search/filtering of fields when editing filters or columns using EntityFieldsMenuComponent 


sleidig:
Users can add/remove filters and columns in the list view by toggling any of the fields of the entity type (see #2177). The `EntityFieldsMenuComponent` is implemented as a "display component" (without business logic) to provide a simple UI dropdown with the available fields for this. For complex entity types there can be dozens of fields in the list, making it difficult to find the relevant field. To improve UX, this component should offer a way to search/filter the fields to help narrow down the fields and find the relevant ones to select.

maybe use https://primeng.org/multiselect ?

or use our basic-autocomplete component in multi-select mode?

PR #2961, opened 2025-04-25
Allow search/filtering of fields when editing filters or columns using EntityFieldsMenuComponent  

Abhinegi2:
closes: #2278

### Visible/Frontend Changes

- [x]
- [ ]

- [x]
- [ ]


Abhinegi2:
While working on this, I had two options in mind since we are using our **basicAutocompleteComponent**, which allows us to display options as plain text or chips, so attaching screenshots for both approaches below.

Let me know which one you think looks better:
1. Chip view(we can also use the full width of field)
image

2. text view
image

src/app/core/common-components/entity-fields-menu/entity-fields-menu.component.html
```diff
@@ -1,27 +1,9 @@
...
-    ></fa-icon>
-
-    <app-entity-field-label
-      [field]="field"
-      [entityType]="entityType"
-    ></app-entity-field-label>
-  </button>
-</mat-menu>
+<mat-form-field>
+  <app-basic-autocomplete
```
sleidig, line 2: isn't this basically the `EntityFieldSelectComponent` now?

src/app/core/common-components/entity-fields-menu/entity-fields-menu.component.html
```diff
@@ -1,27 +1,9 @@
...
-      [icon]="activeFields.includes(field.id) ? 'toggle-on' : 'toggle-off'"
-    ></fa-icon>
-
-    <app-entity-field-label
-      [field]="field"
-      [entityType]="entityType"
-    ></app-entity-field-label>
-  </button>
-</mat-menu>
+<mat-form-field>
```
sleidig, line 1: let's also display this form a bit wider in the admin view for easier reading of field names

sleidig:
> While working on this, I had two options in mind since we are using our **basicAutocompleteComponent**, which allows us to display options as plain text or chips, so attaching screenshots for both approaches below.

Good thoughts, @Abhinegi2 . I would like to consider the chip layout in the future. To keep it simple now, let's finalize the current text-based implementation though. That seems to work very well already.

src/app/core/common-components/entity-fields-menu/entity-fields-menu.component.html
Abhinegi2: Yes, it's the same. I'm still using EntityFieldsMenuComponent for now because we might want to customize it in the future.
But what do you think, should we use EntityFieldSelectComponent instead? If so, we can remove this.

src/app/core/common-components/entity-fields-menu/entity-fields-menu.component.html
Abhinegi2: I tried both full width and 30% width. Here are the results—let me know which one looks better: 
1. full width -
image

2. 30% width -
image

src/app/core/common-components/entity-fields-menu/entity-fields-menu.component.html
sleidig: Having a basis to customize again is a fair point. How about using <app-entity-field-select> within the template of this component for now?

src/app/core/common-components/entity-fields-menu/entity-fields-menu.component.html
Abhinegi2: initially, I thought the full-width looked better. But later I realized that if there are only 2–3 selected items, it appears too large on wider screens. **so I also prefer the 30% width now.**

sleidig (changes requested):
When removing a column through the "x" in the preview (outside of the select field), the select field still shows this as selected when clicking into it again.

Abhinegi2:
> When removing a column through the "x" in the preview (outside of the select field), the select field still shows this as selected when clicking into it again.

updated, please have a look now

sleidig (approved):
works now. Thanks!

re-assigning variables within the template feels a little messy but any else would be much more extensive, I guess. And this component needs further UX tuning in the future anyway :wink:

Issue #2400, opened 2024-06-06
"Clear all filters" button for lists


sleidig:
To quickly reset all selected filters, users have request a button to clear all existing filters at once.

george-neha:
To clarify, if I filter or search for a specific name or age, then the "Clear All Filters" option does not apply to them. 

<img width="964" alt="Image" src="https://github.com/user-attachments/assets/1355cdd9-5792-4951-8022-b97590694f2b" />

PR #2965, opened 2025-05-01
"Clear all filters" button for lists 

Abhinegi2:
closes: #2400

### DIFF
--- a/src/app/core/entity-list/entity-list/entity-list.component.html
+++ b/src/app/core/entity-list/entity-list/entity-list.component.html
@@ -30,6 +30,8 @@
       [entities]="allEntities"
       [useUrlQueryParams]="true"
       [(filterObj)]="filterObj"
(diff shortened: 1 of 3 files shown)


src/app/core/entity-list/entity-list/entity-list.component.html
```diff
@@ -30,6 +30,8 @@
       [entities]="allEntities"
       [useUrlQueryParams]="true"
       [(filterObj)]="filterObj"
+      [filterString]="filterString"
+      (filterStringChange)="filterString = ''; applyFilter('')"
```
sleidig, line 34: I understand it's only resetting the string currently, but let's make it emit a new string rather than just void. That seems more intuitive to me (and may be useful in the future.

sleidig (approved):
Nice work, thanks! :1st_place_medal:

Issue #2963, opened 2025-04-30
autocomplete dropdown now has two scrollbars


sleidig:
![Image](https://github.com/user-attachments/assets/f99c1be8-6aa4-4efa-b9c0-49a2f2076528)

sleidig:
Maybe introduced with https://github.com/Aam-Digital/ndb-core/pull/2960 ?

Abhinegi2:
@george-neha, as discussed could you also test this issue here-Deployed to https://pr-2968.aam-digital.net/

PR #2968, opened 2025-05-02
autocomplete dropdown now has two scrollbars 

Abhinegi2:
closes: #2963

### DIFF
--- a/src/app/core/common-components/basic-autocomplete/basic-autocomplete.component.ts
+++ b/src/app/core/common-components/basic-autocomplete/basic-autocomplete.component.ts
@@ -136,7 +136,10 @@ export class BasicAutocompleteComponent<O, V = O>
   /** whether the "add new" option is logically allowed in the current context (e.g. not creating a duplicate) */
   showAddOption = false;
 
-  // maximum height of the autocomplete panel
+  /**
+   * maximum height of the autocomplete panel.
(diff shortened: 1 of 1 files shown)


src/app/core/common-components/basic-autocomplete/basic-autocomplete.component.html
```diff
@@ -32,6 +32,8 @@
   (optionSelected)="select($event.option.value)"
   autoActiveFirstOption
   [hideSingleSelectionIndicator]="true"
+  [class.custom-panel]="true"
+  [style.max-height.px]="maxPanelHeight"
```
sleidig, line 36: Have you tried finding a purely css property to get this working? Having layout calculations in the typescript code adds a lot of complexity

sleidig (approved):
works :+1:

src/app/core/common-components/basic-autocomplete/basic-autocomplete.component.html
Abhinegi2: Yes, I tried using only CSS. It works fine with just a few options (like 5–8). But when there are many options (like 10 or more), it shows two scrollbars. 
This also happens on small screens, even with fewer options. Because of that, I decided to handle it with typeScript code.

Issue #2380, opened 2024-05-08
Autocomplete / search to narrow down dropdown options in filters


sleidig:
_In the filter components above the list views, I can select options to be filtered from suggestions. If there are many values, this is unwieldy and I would like to narrow down the dropdown options like in our auto-complete fields._

When filtering values for a list, the system offers a dropdown with all options. This can be quite long. Users should be able to type for autocomplete / narrowing down the displayed dropdown options (similar to how entity-select fields allow that in forms already).

![Image](https://github.com/user-attachments/assets/d9771c9a-33ae-4194-b45c-2c640f90e22f)

UX conditions:
- actually filter logic must not change (as the user clicks on one or more elements to activate them as filter criteria)
- filter value component should allow typing to narrow down the dropdown options offered
- upon leaving the focus from the filter field, the searched text should be cleared and the UI must display the same as it is currently (i.e. the actually selected filter value(s))
- ideally this should work for any type of filter (filter for a "string" dataType field as well as for "entity" or other dataType)

Ideally no code should be written for this afresh. Either the existing BasicAutocompleteComponent should be reused (or possibly a library that provides this kind of function)

sleidig:
`AppListFilter` directly uses a `MatSelect` at the moment. Probably we can use out `BasicAutocompleteComponent` internally there even without changing its Inputs and Outputs.

george-neha:
A different issue, where the dropdown value are showing blank entries on clearing the filter was identified while testing this: https://github.com/Aam-Digital/ndb-core/issues/2983

PR #2969, opened 2025-05-05
Autocomplete / search to narrow down dropdown options in filters 

Abhinegi2:
closes: #2380, [#2983](https://github.com/Aam-Digital/ndb-core/issues/2983)

### Visible/Frontend Changes

- [x]
- [ ]

- [x]
- [ ]


src/app/core/filter/filter/filter.component.spec.ts
```diff
@@ -152,37 +150,4 @@ describe("FilterComponent", () => {
     expect(component.filterSelections[0].name).toBe("category");
     expect(component.filterSelections[0].selectedOptionValues).toBeEmpty();
   });
-
```
Abhinegi2, line 155: As we are using basicautocompletecomponent now, we not need this anymore?

Abhinegi2:
@sleidig I found the issue with the dropdown position while clicking Select All and Clear for the school filter — the dropdown is misaligned, and two scrollbars are showing.

While debugging, I noticed this is happening because we expect each option to have a max height of 48px in the BasicAutoComplete component. However, some school names are quite long, which causes the option height to exceed the expected value. This increases the height of the parent panel, leading to layout issues like:
- Misplaced dropdown(click on select all and then try to scroll)
- Duplicate scroll bars

To fix this, I’m thinking we should avoid wrapping long option text and instead allow horizontal scrolling inside the dropdown.

1. Right now, it's like this -
image

2. After horizontal scrolling, it would look like this-
image?

Also, I am not sure how it affects users, so if you have another approach, let me know.

Abhinegi2:
I’ve added the ellipsis effect on the mat-option, and it’s working as expected.
I also attempted to make the dropdown options wider than the field, but it was more complex and time-consuming than anticipated.
Would you like me to handle this in a separate PR/Issue?

sleidig:
> I’ve added the ellipsis effect on the mat-option, and it’s working as expected. I also attempted to make the dropdown options wider than the field, but it was more complex and time-consuming than anticipated. Would you like me to handle this in a separate PR/Issue?

yes, maybe create an issue for that then and we decide when to prioritize this later

sleidig:
> > I’ve added the ellipsis effect on the mat-option, and it’s working as expected. I also attempted to make the dropdown options wider than the field, but it was more complex and time-consuming than anticipated. Would you like me to handle this in a separate PR/Issue?
> 
> yes, maybe create an issue for that then and we decide when to prioritize this later

see https://github.com/Aam-Digital/ndb-core/issues/2981

src/app/core/filter/list-filter/list-filter.component.html
```diff
@@ -1,30 +1,17 @@
...
-        [appBorderHighlight]="option.color"
-        style="padding-top: 0.5em; padding-bottom: 0.5em"
-      >
-        {{ option.label }}
-      </mat-option>
-    }
-  </mat-select>
+    <ng-container autocompleteHeader>
+      <div class="flex-row">
+        <button mat-flat-button i18n (click)="selectAll()">Select All</button>
```
sleidig, line 12: the `class="util-button"` did have a purpose. Without it, the hover background color looks a bit off right now

src/app/core/filter/list-filter/list-filter.component.ts
```diff
@@ -15,28 +15,37 @@ import { MatButtonModule } from "@angular/material/button";
...
-      (option) => option.key,
-    );
-    this.filterConfig.selectedOptionChange.emit(
-      this.filterConfig.selectedOptionValues,
-    );
+    const allValues = this.filterConfig.options.map((opt) => opt.key);
+    this.autocompleteControl.setValue(allValues);
   }
 
   deselectAll() {
```
sleidig, line 48: if I do a "Clear" for an individual filter here, the "Clear all" button still remains visible in the overall list although everything is cleared. (Maybe not directly related to this PR but still a tiny glitch we should fix)

src/app/core/filter/filter/filter.component.spec.ts
sleidig: yes, I think the autocomplete-component only replaces the UI part but we still rely on other code to find and create the available options and to create the correct filter object, which is checked in the expect() statements here. Let's keep it

sleidig (changes requested):
Seems to work well. This is a huge UX improvement :smiley:

src/app/core/filter/list-filter/list-filter.component.ts
Abhinegi2: Ah, thanks for noticing this , i have fixed this now

sleidig (changes requested):
I stumble over a UX issue on final testing:

When I type to narrow down filter options and then leave the focus (even using "clear" of the filter or "clear all" button), the filter-text is still there (but "invisible" in the background). Opening the filter again still shows pre-filtered options instead of all available ones in the dropdown:
Screenshot From 2025-05-14 22-04-38
Screenshot From 2025-05-14 22-04-45

Abhinegi2:
> I stumble over a UX issue on final testing:
> 
> When I type to narrow down filter options and then leave the focus (even using "clear" of the filter or "clear all" button), 

Ah, I see the issue now. After debugging, I found that this bug/behavior was introduced earlier when `multi` was set to `true`. At that time, we wanted to retain the filtered text even after selecting options ( #2337), which is why it's not getting cleared now. I’ll update the code to handle this scenario appropriately (for both clear and clear all buttons).

sleidig:
> > I stumble over a UX issue on final testing:
> > When I type to narrow down filter options and then leave the focus (even using "clear" of the filter or "clear all" button),
> 
> Ah, I see the issue now. After debugging, I found that this bug/behavior was introduced earlier when `multi` was set to `true`. At that time, we wanted to retain the filtered text even after selecting options ( #2337), which is why it's not getting cleared now. I’ll update the code to handle this scenario appropriately (for both clear and clear all buttons).

I think actually, the filter autocomplete text should reset when leaving that filter's focus also, not only when a clear button is used

Abhinegi2:
> I think actually, the filter autocomplete text should reset when leaving that filter's focus also, not only when a clear button is used

I’ve made some updates:  now, when the filter input loses focus, the autocomplete text resets. Could you have a look now?

src/app/core/common-components/basic-autocomplete/basic-autocomplete.component.ts
```diff
@@ -428,6 +431,7 @@ export class BasicAutocompleteComponent<O, V = O>
         this.select(undefined);
       }
       this.isInSearchMode.set(false);
+      this.retainSearchValue = "";
```
sleidig, line 434: This fixes the issue with search values being kept too long. We did implement something upon an explicit user request to _not_ reset the search during multi-select (so that you can select multiple things under the same search term). But for now, this function here is much better then the confusing previous behavior. So let's go with this :+1:

sleidig (approved):
Nice :+1:

PR #2974, opened 2025-05-07
test with native gh runners on arm64

sleidig:
see https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/


sleidig:
Seems to work well with native GitHub Action Runners now! :heart_eyes:

Issue #2977, opened 2025-05-08
auto-update: existing trigger-mapping lost when opening edit dialog again


sleidig:
**To Reproduce**

1. Go to Admin UI to edit a field's default value in "updated from referencing entity" mode
2. Click on the "Configure Automation" to open the dialog
3. Select some value mappings but not all
4. Save the dialog
5. Click on "configure automation" again
6. fill some mapping value that is still empty
7. save and open a third time
8. --> only the mapping selected on the second turn has been selected and saved (previous values are always lost on saving the dialog)

PR #2978, opened 2025-05-08
auto-update: existing trigger-mapping lost when opening edit dialog again 

Abhinegi2:
closes: #2977

### DIFF
--- a/src/app/features/automated-status-update/automated-field-mapping/automated-field-mapping.component.ts
+++ b/src/app/features/automated-status-update/automated-field-mapping/automated-field-mapping.component.ts
@@ -176,12 +176,17 @@ export class AutomatedFieldMappingComponent implements OnInit {
     });
     if (this.isInvalid) return;
 
-    const formattedMappings: { [key: string]: any } = {};
-    Object.entries(this.selectedMappings).forEach(([key, value]) => {
-      formattedMappings[key] = this.schemaService.valueToDatabaseFormat(
-        value,
-        this.targetFieldConfig,
-      );
+    const formattedMappings: { [key: string]: any } = {
+      ...this.selectedMappings,
+    };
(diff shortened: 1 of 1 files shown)


src/app/features/automated-status-update/automated-field-mapping/automated-field-mapping.component.ts
```diff
@@ -176,12 +176,17 @@ export class AutomatedFieldMappingComponent implements OnInit {
...
     if (this.isInvalid) return;
 
-    const formattedMappings: { [key: string]: any } = {};
-    Object.entries(this.selectedMappings).forEach(([key, value]) => {
-      formattedMappings[key] = this.schemaService.valueToDatabaseFormat(
-        value,
-        this.targetFieldConfig,
-      );
+    const formattedMappings: { [key: string]: any } = {
+      ...this.selectedMappings,
```
sleidig, line 180: as the forms are initialized with the values from selectedMappings, this shouldn't really be necessary?

src/app/features/automated-status-update/automated-field-mapping/automated-field-mapping.component.ts
```diff
@@ -176,12 +176,17 @@ export class AutomatedFieldMappingComponent implements OnInit {
...
+    };
+    Object.entries(this.mappingForms).forEach(([key, mappingForm]) => {
+      const value = mappingForm.formGroup.get(this.targetFieldConfig.id)?.value;
+      if (value) {
```
sleidig, line 184: I thought this might prevent it from saving a deleted mapped value. That doesn't seem to be a problem ... But do we really need the check here or can we just always use whatever value is in the form control currently?

sleidig (approved):
Thanks! Looks good

Issue #2881, opened 2025-02-27
Default Value selection should use component for the relevant datatype


sleidig:
When selecting a default value for a field or a pre-filled value for a Public Form, we currently offer a simple text input to the user. However, when the relevant field is a dropdown or other type, then this is difficult to handle. Instead, we should dynamically show the user a dropdown select for the current field (or a date select for a date field, etc.)

![Image](https://github.com/user-attachments/assets/f13f8ab5-aa0d-4edd-a638-4e6b5568852d)

sleidig:
This may also apply to "Edit Field" when setting a "static" default value for an Entity Field (not only when doing that for public forms)

PR #2984, opened 2025-05-09
Default Value selection should use component for the relevant datatype

Abhinegi2:
closes:  #2881


src/app/core/admin/admin-entity-details/admin-entity-field/admin-entity-field.component.ts
```diff
@@ -281,6 +283,14 @@ export class AdminEntityFieldComponent implements OnInit {
...
     const fieldId = this.fieldIdForm.getRawValue();
 
+    // If a default value is configured, convert it to the database format
+    const defaultValueConfig = this.entitySchemaField?.defaultValue?.config;
+    if (defaultValueConfig?.value) {
+      defaultValueConfig.value = this.entitySchemaService.valueToDatabaseFormat(
+        defaultValueConfig.value,
+        this.entitySchemaField,
+      );
+    }
```
sleidig, line 293: why does this have to happen here? (I would see such a defaultValue specific task as the responsibility for the admin-default-value component)

And shouldn't the defaultValue stored in the entity schema by definition already be in database-format anyway?

src/app/core/common-components/entity-field-edit/entity-field-edit.component.ts
```diff
@@ -67,8 +71,18 @@ export class EntityFieldEditComponent<T extends Entity = Entity>
...
 
   private updateField() {
+    if (!this.field) {
+      this._field = undefined;
+      return;
+    }
+
+    this._field = toFormFieldConfig(this.field);
+
     if (!this.entity?.getConstructor()) {
```
sleidig, line 81: might get easier to read now with an if(...constructor) ... else ...

sleidig (commented):
I little UI polishing may be nice:
image

can we get the inner form-field line to align with the buttons (or maybe even better: don't show any line or box at all)

The "label" is also unnecessary, as we are editing that specific field. If possible, I would hide that as well.

src/app/features/public-form/edit-prefilled-values/edit-prefilled-values.component.ts
```diff
@@ -110,11 +118,24 @@ export class EditPrefilledValuesComponent
...
-      ({ field, defaultValue, hideFromForm }) => ({
-        id: field,
-        defaultValue,
-        hideFromForm: hideFromForm ?? true,
-      }),
+      ({ field, defaultValue, hideFromForm }) => {
+        const transformedValue = this.entitySchemaService.valueToDatabaseFormat(
+          defaultValue.config?.value,
+          this.getSchemaField(field),
+        );
```
sleidig, line 125: I think we should try to move this into the admin-default-value-static component. The defaultValues should always be stored in database format, so that component could make sure it only sets/emits it in the correctly transformed format to the outside already

Abhinegi2:
> I little UI polishing may be nice: ![image](https://private-user-images.githubusercontent.com/1682541/443264518-4f65b969-6389-4092-9a31-0ad3ea07a1a5.png?

 I’ve made some UI refinements. Could you please have a look now? here - https://pr-2984.aam-digital.net/

src/app/core/admin/admin-entity-details/admin-entity-field/admin-entity-field.component.ts
Abhinegi2: Sure, I will move this to the admin-default-value component

While trying to save the updated field schema(defaultvalueconfig), right now it's not already in database format, that's why I am doing this -
image

sleidig: oh, I see. Are we saving it in this format to the schema? Or do we transform it somewhere else along the way?

I thing it would be best to keep this as an internal implementation detail of admin-default-static and just expose the database-format to the outside

sleidig:
> > I little UI polishing may be nice: ![image](https://private-user-images.githubusercontent.com/1682541/443264518-4f65b969-6389-4092-9a31-0ad3ea07a1a5.png?
> 
> I’ve made some UI refinements. Could you please have a look now? here - https://pr-2984.aam-digital.net/

Nice! Looks very clean now :slightly_smiling_face:

src/app/core/admin/admin-entity-details/admin-entity-field/admin-entity-field.component.ts
Abhinegi2: I think we have been saving this format to the schema as, is until now (probably because it only supported simple text, so there was no need for transformation).
But now I’ve added the logic inside the default-value-component itself to transform the value before emitting/saving it.

src/app/core/common-components/entity-field-edit/entity-field-edit.component.ts
```diff
@@ -67,14 +76,23 @@ export class EntityFieldEditComponent<T extends Entity = Entity>
...
+      this._field = this.entityFormService.extendFormFieldConfig(
+        this.field,
+        this.entity.getConstructor(),
+      );
+    } else {
+      // add editComponent (because we cannot rely on the entity's schema yet for a new field)
+      this._field.editComponent =
+        this._field.editComponent ??
+        this.entitySchemaService.getComponent(this._field, "edit");
+    }
```
sleidig, line 96:
```suggestion
    if (this.entity?.getConstructor()) {
      this._field = this.entityFormService.extendFormFieldConfig(
        this.field,
        this.entity.getConstructor(),
      );
    } else {
      this._field = toFormFieldConfig(this.field);
      // add editComponent (because we cannot rely on the entity's schema yet for a new field)
      this._field.editComponent =
        this._field.editComponent ??
        this.entitySchemaService.getComponent(this._field, "edit");
    }
```

sleidig (approved):
Good to go now :+1: 

I did some small final polishing myself.

PR #2987, opened 2025-05-12
Fix e2e tests and add CI

geigerzaehler:
Fix end-to-end tests and a job for the tests to CI.

I chose not to add the test script to the Dockerfile and instead run it on the test runner directly. It looks like caching with docker does not work and preparing the cache build images is quite slow.

I discovered that the tests were slow because the DB views [are prebuilt](https://github.com/Aam-Digital/ndb-core/blob/1f2f0a6e9efd9573b035789bc1c662c5b5a98286/src/app/core/database/pouchdb/pouch-database.ts#L313) whenever the app loads. This would take up to 30s seconds on every app load. It seems this is not strictly necessary for the app to work so I commented the code out temporarily.

~~We need to decide whether we want to proceed with this approach and how to configure this behavior for tests or develop another solution to the tests being slow.~~

We’ve decided to only pre-build the views in production. I’ve also removed the corresponding assertions from the unit tests.

I didn’t touch the actual tests to much to keep the PR small. I’ll update them in a follow-up PR and apply best practices for writing tests.


doc/compodoc_sources/how-to-guides/end-to-end-tests.md
```diff
@@ -0,0 +1,43 @@
+# End-to-end tests
+
+End-to-end tests run the app in the browser using browser-local database. We use
+[Playwright][] as the testing framework.
+
+Before run the tests you need to start a local development server with `npm
+start`. You can use the `playwright` CLI to run the tests from the terminal or
```
sleidig, line 7:
```suggestion
Before you run the tests you need to start a local development server with 
`npm start`. You can use the `playwright` CLI to run the tests from the terminal or
```

e2e/tests/attendance-tests.spec.ts
```diff
@@ -103,33 +98,14 @@ test.describe("Attendance Module", () => {
...
-      const borderColor = await card.evaluate(
-        (el) => getComputedStyle(el).borderLeftColor,
-      );
-
-      if (i === 0) {
-        expect(borderColor).toBe("rgb(76, 175, 80)"); // Green
-      } else {
-        expect(borderColor).toBe("rgb(255, 152, 0)"); // Orange
-      }
-    }
```
sleidig, line 127: did you intentionally remove the color checks?

(I guess a better way for this will be the visual regression testing soon ...)

package.json
```diff
@@ -9,8 +9,6 @@
     "test": "ng test",
     "test-ci": "ng test --code-coverage --karma-config=build/karma-ci.conf.js ",
     "lint": "ng lint ndb-core",
-    "e2e": "playwright test",
-    "e2e:debug": "playwright test --project chromium --headed",
```
sleidig, line 13: can we keep these? I kinda like to have all the actions listed out here ...

src/app/core/database/pouchdb/pouch-database.ts
```diff
@@ -310,7 +311,9 @@ export class PouchDatabase extends Database {
     }
 
     await this.put(designDoc, true);
-    await this.prebuildViewsOfDesignDoc(designDoc);
+    if (environment.production) {
```
sleidig, line 314:
```suggestion

    // for faster initial loading we disable prebuilding views in development
    // TODO: check if this should be completely removed, also for production systems
    if (environment.production) {
```

sleidig (approved):
Looks great! Nice cleanups and performance tuning.

Thanks a lot, @geigerzaehler ! :heart_eyes:

geigerzaehler:
> do you have any idea how we can get rid of these waiting CI jobs `test (linux/arm64, self-hosted, Linux, ARM64)`?

Yes, you can remove these requirements in the repository settings under the branch protection rules for master.

Issue #2473, opened 2024-07-24
Auto-select and/or auto-create values for enum fields during import


sleidig:
When importing data that includes a column that maps to a dropdown field ("configurable-enum") the user currently has to select the target dropdown value for every value that occurs in the imported data (after clicking on the the "Configure Value mapping" button). This can be a very tedious process if there are many dropdown options.

![Image](https://github.com/user-attachments/assets/24d9d634-6bf9-44ef-83fd-610a1f50ac7a)

--> Automatically preselect/map the imported values to enum values if they match the id or label of an enum item

PR #2992, opened 2025-05-13
Auto-select and/or auto-create values for enum fields during import 

Abhinegi2:
closes: #2473

### Visible/Frontend Changes

- [x] Added smart selection during import by matching label or ID
- [x] do not auto-select where someone manually updated options


src/app/core/basic-datatypes/discrete/discrete-import-config/discrete-import-config.component.ts
```diff
@@ -54,23 +56,30 @@ export class DiscreteImportConfigComponent implements OnInit {
...
 
   private getFormValues(additional: any) {
-    if (!additional) {
-      additional = {};
-    }
+    additional = additional || {};
+
+    const enumEntity = this.configurableEnumService.getEnum(
+      this.schema?.additional,
+    );
```
sleidig, line 63: we probably shouldn't do this if schema.additional is undefined. Looking at the code, this might create a new enum entity. And we do use the same import-config component for boolean and enums, so this may be a valid scenario to not have an `additional`

src/app/core/basic-datatypes/discrete/discrete-import-config/discrete-import-config.component.spec.ts
```diff
@@ -69,7 +69,11 @@ describe("DiscreteImportConfigComponent", () => {
 
     expect(component.form.getRawValue()).toEqual({
       male: genders.find((e) => e.id === "M"),
-      female: null,
+      female: {
```
sleidig, line 72:
```suggestion
      // unmapped values will be imported as "invalid option" to not lose data:
      female: {
```

sleidig (approved):

PR #2993, opened 2025-05-14
Simplify existing e2e test case

geigerzaehler:
Simplify an existing e2e test case to make them more readable and robust.

I converted one flexible test with a more complex and dynamic structure to simple recipe of instructions and expectations. I also tried to only use accessibility-based locators since they are more readable and more robust. Only one test was changed, the other changes are mostly white-space changes. Probably a good idea to review this with white-space changes hidden.

There are a couple of minor changes to the application code to improve the accessibility tree of the app so that we can use accessibility-based locators.

I’ve also switched to mocking the datetime for all tests. This means that demo data is generated with respect to the mocked datetime.


e2e/tests/dashboard-tests.spec.ts
```diff
@@ -1,35 +1,15 @@
...
-    */
-    await expect(page.getByText("Quick Actions")).toBeVisible();
-
-    // Verify "Record Attendance" button in Quick Actions
-    const recordAttendanceButton = page.getByRole("cell", {
-      name: "Record Attendance",
-    });
-    await expect(recordAttendanceButton).toBeVisible();
-
-    await recordAttendanceButton.click();
```
geigerzaehler, line 20: If we want to click on something it’s usually not necessary to expect visibility before. The `.click()` call will wait for the element to show up

e2e/tests/dashboard-tests.spec.ts
```diff
@@ -1,35 +1,15 @@
 import { expect, test } from "#e2e/fixtures.ts";
 
-test.describe("Dashboard Module", () => {
```
geigerzaehler, line 3: No need to wrap everything in a test file. This flattens the structure a bit and makes it more readable.

e2e/tests/check-browser-title.spec.ts
```diff
@@ -1,5 +0,0 @@
-import { expect, test } from "#e2e/fixtures.ts";
-
-test("has title", async ({ page }) => {
```
geigerzaehler, line 3: This test is too expensive for what it does. A single test case requires about 5s of setup. It’s much easier to assert this in other tests

e2e/tests/attendance-tests.spec.ts
```diff
@@ -1,360 +1,278 @@
...
-    /*
-      Verify the date field displays the current date
-    */
-    const fixedDate = new Date(FIXED_DATE);
-    const formattedDateRegex = new RegExp(
-      `${fixedDate.getMonth() + 1}/${fixedDate.getDate()}/${fixedDate.getFullYear()}|` +
-        `${(fixedDate.getMonth() + 1).toString().padStart(2, "0")}/${fixedDate.getDate().toString().padStart(2, "0")}/${fixedDate.getFullYear()}`,
-    );
-    const dateField = page.getByLabel("Date");
-    await expect(dateField).toHaveValue(formattedDateRegex);
```
geigerzaehler, line 23: We set the date to a fixed constant in the test setup so we know it and can just assert it directly:
```typescript
wait expect(dateField).toHaveValue("1/23/2025");
```
It’s totally okay to hard-code the values for now.

e2e/tests/attendance-tests.spec.ts
```diff
@@ -1,360 +1,278 @@
...
+    page.getByRole("heading", { name: "School Class 2B" }),
+  ).toBeVisible();
 
-    /*
-      Verify list of classes is displayed
-    */
-    await page.waitForSelector("mat-card-header");
-    const classList = page.locator("mat-card-header");
-    const count = await classList.count();
-    expect(count).toBeGreaterThan(0);
```
geigerzaehler, line 38: We’re testing an artifact of the implementation (the class list) here. I removed this because we should focus on things that are directly visible to the user and I don’t know what this is in this case.

e2e/tests/attendance-tests.spec.ts
```diff
@@ -1,360 +1,278 @@
...
-
-    const pageIndicator = page.locator("text=/\\d+\\s\\/\\s\\d+/"); // "1 / 5" pattern
-    // Extract total participants from "x / y" format
-    if (await pageIndicator.isVisible()) {
-      const text = await pageIndicator.textContent();
-      const match = text?.match(/(\d+)\s\/\s(\d+)/); // Extract current/total numbers
-      if (match) {
-        totalParticipants = parseInt(match[2]); // Total participants
-      }
-    }
```
geigerzaehler, line 56: In general we should avoid making the test code dependent of the fixtures in the database. Since we setup the tests in a specific state we know the participant count and can assert it directly:

```typescript
await expect(page.getByText("1 / 6")).toBeVisible();
```

e2e/tests/attendance-tests.spec.ts
```diff
@@ -1,360 +1,278 @@
...
-    const ReviewDetailsBtn = page.getByRole("button", {
-      name: "Review Details",
-    });
-
-    // Loop to mark attendance dynamically until the "Review Details" button is visible
-    while (!(await ReviewDetailsBtn.isVisible())) {
-      const randomIndex = Math.floor(Math.random() * 4); // Random index for attendance buttons
-      const attendanceButton = page.getByRole("paragraph").nth(randomIndex);
-      await attendanceButton.click();
-      await page.waitForTimeout(1000); // Wait for the next participant to load
```
geigerzaehler, line 66: We should try to avoid this fixed time-outs at all costs because they are guaranteed to slow down the tests. (With 6 attendees we’re just waiting here for 6s). It’s usually possible to rewrite this as an expectation that is only waited for until it succeeds.

To work around the loading issue I used
```typescript
await expect(page.getByRole("button", { name: "Absent" })).toHaveCount(1
````

File: e2e/tests/attendance-tests.spec.ts [hunk 8]
```diff
@@ -1,360 +1,278 @@
...
-        "app-attendance-status-select > .mat-mdc-form-field > .mat-mdc-text-field-wrapper > .mat-mdc-form-field-flex",
-      )
-      .first();
-    await EditStatusBtn.click();
-
-    const statusOptions = page.getByText("Present Absent Late Excused");
-    const randomIndex = Math.floor(
-      Math.random() * (await statusOptions.count()),
-    );
-    const selectedStatus = statusOptions.nth(randomIndex);
```
geigerzaehler, line 86: Using Math.random makes the tests non-deterministic which makes them hard to follow and debug because each test run is different. Instead, just [chose an option](https://xkcd.com/221/) when you run it.

doc/compodoc_sources/how-to-guides/end-to-end-tests.md
```diff
@@ -3,15 +3,19 @@
...
+Before you run the tests you need to start a local development server with `npm
+start`. You can use the `playwright` CLI to run the tests from the terminal or
 start the [Playwright UI].
 
 ```bash
 npx playwright test
 npx playwright test --ui
 ```
 
+To debug test runs your can use Playwright’s [Trace Viewer]. On CI traces for
```
sleidig, line 15:
```suggestion
To debug test runs you can use Playwright’s [Trace Viewer]. On CI traces for
```

Issue #2915, opened 2025-03-19
UX issues in Notifications


george-neha:
When I click on a notification or a hyperlink from the notifications list, the drawer remains open. Ideally, the drawer should close automatically once I make a selection. Currently, I have to click elsewhere on the screen to close it.

<img width="964" alt="Image" src="https://github.com/user-attachments/assets/d343f695-e436-42e1-a0d3-9cbc6eb9668b" />

sleidig:
icon and label are not properly aligned with longer (translated) texts:
![Image](https://github.com/user-attachments/assets/c23d9983-8d7f-4e34-aee0-5ff6532dcc3d)

Abhinegi2:
While working on alignment, I found that the issue occurs when the text is long and there's no spacing between the text. That’s why the layout breaks.

- I made some updates to add spacing, and here’s how it looks with spacing
   ![Image](https://github.com/user-attachments/assets/f17aa128-2f8c-4bd3-9a33-4245ca878c6e)


- However, if there’s no spacing, it still appears like this
  ![Image](https://github.com/user-attachments/assets/d515cdea-596b-4a71-849f-e94383d94b7a)


**I wanted your suggestion on the second point** (Would truncation, wrapping help to display long text properly?)

sleidig:
Both don't look great 😅 

Let's maybe make the menu width grow (if possible) or otherwise truncate the text to keep it one line, @Abhinegi2 ?

Abhinegi2:
@george-neha could you test the above-mentioned things here? - https://pr-2996.aam-digital.net/

PR #2996, opened 2025-05-15
UX issues in Notifications 

Abhinegi2:
closes: #2915

### Visible/Frontend Changes

- [x]
- [ ]

- [x]
- [ ]


src/app/features/notification/notification.component.html
```diff
@@ -92,8 +98,8 @@ <h2 class="notification-title" i18n>Notifications</h2>
     (click)="closeOnlySubmenu(settingsMenuTrigger, $event)"
   >
     <button mat-menu-item (click)="markAllRead()">
-      <fa-icon icon="check"></fa-icon>
-      <span class="menu-option" i18n>Mark all as read</span>
+      <fa-icon icon="check" class="standard-icon-with-text"></fa-icon>
```
sleidig, line 101: let's additionally change the overflow (a class "flex-row" around these might help). At the moment, if a translation is even longer than the German one, we still have the line break

sleidig (commented):
Code looks good :-)

@abhinegi2, can you remove the test code, preparing for merge?

src/app/features/notification/notification.component.html
Abhinegi2: For longer translation, I am thinking like this (or also can show full text onhover) -
image

What's your thought?

sleidig: This looks good in my opinion. Let's go with this for now - should only be a safeguard for exceptional translations anyway

sleidig (approved):

PR #2997, opened 2025-05-15
feat: Migrate eslint configuration

geigerzaehler:
Migrate the eslint configuration to the [new configuration format][1].

I was working on setting up linting and type-checking for the test code and I thought that this would be useful.

[1]: https://eslint.org/docs/latest/use/configure/migration-guide


sleidig (approved):
cool, didn't even know this exists :sweat_smile: 

Looks good to me (and linting still seems to pass) :100:

Issue #2428, opened 2024-06-20
Import: show entity reference fields in preview with icon / more clearly visualized


sleidig:
The import data preview table does not make it very clear if a value is an entity reference, because that is just displayed as a plain toString text also. That makes it hard to distinguish whether a name string was imported or the lookup to an entity in the DB (or if the referenced entity has an empty toStringAttribute, the user may see an empty slot despite the successful lookup).

sleidig:
This is actually not true or not true anymore:

![Image](https://github.com/user-attachments/assets/c36343e2-fc43-4338-9dee-afe7e152fc30)

It may be necessary that the referenced entity type has an icon configured to clearly recognize this, however. Maybe the entity-block component by default should display some fallback icon if no custom icon is defined?

sleidig:
@george-neha , do you have thoughts about the UI of this aspect?

Abhinegi2:
@sleidig, how can I export the school-enrollment CSV? I need to test the mentioned behavior.

PR #2999, opened 2025-05-15
Import: show entity reference fields in preview with icon / more clearly visualized 

Abhinegi2:
closes: #2428

also fixed below error during import(step3) - 
RuntimeError: NG0100: ExpressionChangedAfterItHasBeenCheckedError: Expression has changed after it was checked. Previous value: 'false'. Current value: 'true'. Expression location: ImportComponent component. Find more at https://angular.dev/errors/NG0100
    at throwErrorIfNoChangesMode (core.mjs:11396:11)
    at bindingUpdated (core.mjs:20555:17)
    at Module.ɵɵproperty (core.mjs:26066:9)
    at ImportComponent_Template (import.component.html:97:13)


sleidig (commented):
code looks good :+1: 


@Abhinegi2 , @george-neha  what was your opinions on the choice of icons? :sweat_smile: 
Seeing it in action, the filled diamond icon feels rather heavy (especially for the fact that this actually is a placeholder for not having an icon). An alternative could be an empty circle maybe?
image

george-neha:
I wanted a common icon like a diamond that would work for both Schools/organizations and participants/students. I also thought having an icon that is filled would be in sync with the icons that we have at present throughout the system.

This is just a thought: an empty circle can be confused with a radio button.

sleidig (approved):

PR #3000, opened 2025-05-15
feat: Better handling address

sadaf895:
closes: #XX

### Visible/Frontend Changes

- [x]
- [ ]

- [x]
- [ ]


package.json
```diff
@@ -112,6 +112,7 @@
     "angular-eslint": "^19.4.0",
     "babel-loader": "^10.0.0",
     "eslint": "^9.24.0",
+    "eslint-config-prettier": "^10.1.5",
```
sleidig, line 115: I think this is not related to the issues in this PR. Please remove it here (and if it's an important change, open a separate PR to add eslint-config-prettier). What is this needed for?

src/app/child-dev-project/attendance/attendance-details/attendance-details.component.html
```diff
@@ -47,6 +47,8 @@ <h1 mat-dialog-title>
         type="number"
         [value]="entity.countEventsPresent(forChild)"
         readonly
+        title="Number of days the child was present"
```
sleidig, line 50: any reason why you added these here while working on the location field? This looks familiar but probably we removed it for a reason (for example, these fields are read-only, so telling users to "enter number of .." doesn't make sense). I think we can skip adding this

src/app/features/location/address-search/address-search.component.ts
```diff
@@ -88,10 +90,17 @@ export class AddressSearchComponent implements OnInit {
...
+    private location: GeoService,
+    private dialog: MatDialog,
+  ) {}
 
   ngOnInit() {
     this.initSearchPipeline();
+    // Keep a local array for isInputInOptions
+    this.filteredOptions.subscribe((options) => {
+      this.filteredOptionsArray = options;
+    });
```
sleidig, line 103: I think we could simplify this by using BehaviorSubject instead of Subject as the type of `filteredOptions`: https://www.learnrxjs.io/learn-rxjs/subjects/behaviorsubject

it then has a property `filteredOptions.value` without having to create this ourselves

src/app/features/location/address-search/address-search.component.ts
```diff
@@ -162,4 +172,11 @@ export class AddressSearchComponent implements OnInit {
       }),
     );
   }
+  isInputInOptions(input: string): boolean {
```
sleidig, line 175:
```suggestion

  isInputInOptions(input: string): boolean {
```

src/app/features/location/no-location-dailog/no-location-dailog.component.html
```diff
@@ -0,0 +1,7 @@
+<h2 mat-dialog-title i18n>No mapped location</h2>
+<mat-dialog-content i18n>
+  There is no mapped location for the entered address.
```
sleidig, line 3: We can use our `ConfirmationDialogService` for this, so that we don't have to create a new component for a simple confirmation message:

https://github.com/Aam-Digital/ndb-core/blob/16b2c9d0b7da1021256b143944f216d90835d7cb/src/app/core/common-components/confirmation-dialog/confirmation-dialog.service.ts#L11-L27

src/app/features/location/geo.service.ts
```diff
@@ -69,14 +69,28 @@ export class GeoService {
   private reformatDisplayName(result: OpenStreetMapsSearchResult): GeoResult {
     const addr = result?.address;
     if (addr) {
-      const city = addr.city ?? addr.town;
+      const city = addr.city ?? addr.town ?? "";
+      let street = addr.road ? addr.road : "";
```
sleidig, line 73: a shorter way of writing this:
```suggestion
      let street = addr.road ?? "";
```

(see https://angular.love/double-question-marks-typescript-3-7-nullish-coalescing)

src/app/features/location/geo.service.ts
```diff
@@ -69,14 +69,28 @@ export class GeoService {
...
 
       result.display_name = [
         addr.amenity ?? addr.office,
-        addr.road ? addr.road + " " + addr.house_number : undefined,
-        addr.postcode ? addr.postcode + " " + city : city,
+        street || undefined,
+        postcodeCity || undefined,
       ]
-        .filter((x) => !!x)
+        .filter((x) => !!x && x !== "undefined")
```
sleidig, line 93: nice solution overall!

Could you write a unit test in geo.service.spec.ts also? I would suggest we call `service.reformatDisplayName(testResult)` for different values of a testResult variable and check if the resulting string is nice and as expected. The reformatDisplayName is currently private, so the test file will not find it. But you can remove the "private" here from the method also, because this method could be useful even outside of the class.

sleidig (changes requested):
Nice work! I added a few remarks to simplify the code still.

src/app/features/location/no-location-dailog/no-location-dailog.component.html
sadaf895: @sleidig Ah, Yeah I was about to do this as now I am working on the new location issue there I have found that ConfirmationDialogService is there I am using it in my new PR will do it in this PR also.

src/app/features/location/address-search/address-search.component.ts
```diff
@@ -34,7 +34,7 @@ import { MatIconButton } from "@angular/material/button";
 import { FaIconComponent } from "@fortawesome/angular-fontawesome";
 import { GeoLocation } from "../geo-location";
 import { HttpErrorResponse } from "@angular/common/http";
-
+import { ConfirmationDialogService } from "../../../core/common-components/confirmation-dialog/confirmation-dialog.service";
```
sleidig, line 37:
```suggestion
import { ConfirmationDialogService } from "../../../core/common-components/confirmation-dialog/confirmation-dialog.service";

```

src/app/features/location/address-search/address-search.component.ts
```diff
@@ -125,12 +128,23 @@ export class AddressSearchComponent implements OnInit {
...
   }
 
-  selectLocation(selected: GeoResult | string | undefined) {
+  async selectLocation(selected: GeoResult | string | undefined) {
     let result: GeoLocation;
     if (typeof selected === "object") {
       result = { geoLookup: selected };
     } else if (typeof selected === "string") {
-      // special case to set address text from search without mapped location (when no result was found)
+      // Show confirmation dialog instead of custom dialog
```
sleidig, line 136:
```suggestion
```

src/app/features/location/address-search/address-search.component.ts
```diff
@@ -125,12 +128,23 @@ export class AddressSearchComponent implements OnInit {
...
+      await this.confirmationDialog.getConfirmation(
+        $localize`No mapped location`,
+        $localize`There is no mapped location for the entered address. You can still save the address as free text.`,
+        [
+          {
+            text: $localize`OK`,
+            dialogResult: true,
+            click: () => {},
+          },
+        ],
```
sleidig, line 146: you can simply import `OkButton` from confirmation-dialog.component.ts
```suggestion
        OkButton,
```

src/app/features/location/geo.service.ts
```diff
@@ -66,18 +66,35 @@ export class GeoService {
       );
   }
 
-  private reformatDisplayName(result: OpenStreetMapsSearchResult): GeoResult {
+  private getCity(addr: any): string {
```
sleidig, line 69: I think you can use the type `OpenStreetMapsSearchResult` instead of `any` for the addr to these methods?

sleidig (changes requested):
one test is still failing:
https://github.com/Aam-Digital/ndb-core/actions/runs/15752343588/job/44399914133?pr=3000#step:8:3367

very nice simplifications with the adress methods and good test cases :+1:

PR #3001, opened 2025-05-15
feat: Lint and type-check e2e test code

geigerzaehler:
Lint and type-check e2e test code. I’ve created a separate tsconfig file for the e2e code so that we cna start with best practices like setting strict mode.

I also created a new `code-quality` step in the workflows and moved the lint step there from the test step. This means we can run the steps in parallel and get quicker feedback.

I opted to run `eslint` instead of `ng lint ndb-core` because the former is more flexible and faster.

### DIFF
--- a/.github/workflows/on-master-push.yml
+++ b/.github/workflows/on-master-push.yml
@@ -26,6 +26,19 @@ jobs:
           destination_dir: documentation
           force_orphan: true
 
+  code-qualty:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: lts/*
+          cache: npm
+
+      - run: npm install --ci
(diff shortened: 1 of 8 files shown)


sleidig (approved):
Looks good to me. I do like the way we move checks out of the Dockerfile into Github Actions. Let's try that with the unit tests also some time :blush: 

How about extracting the code-quality and test steps into one joint workflow that runs both on master, tags and pull_requests so that we do not have to duplicate that?

geigerzaehler:
> How about extracting the code-quality and test steps into one joint workflow that runs both on master, tags and pull_requests so that we do not have to duplicate that?

I think that makes a lot of sense but I’d like to tackle it in a follow-up. I need to familiarize myself better with the workflows because modularizing workflows can be tricky. Simplifying the unit test job will probably make this easier as well so I’ll start with that.

PR #3002, opened 2025-05-16
Don’t run unit test in docker container in CI

geigerzaehler:
- [ ] Verify that code coverage works
- [ ] Update repository status checks to include `test-unit` and remove `test`

PR #3003, opened 2025-05-16
Fix lint warnings and fail on warnings on CI

geigerzaehler:
Make CI fail when there are lint warnings. For this to pass we fix all the existing lint warnings.

### DIFF
--- a/.github/workflows/pull-request-update-or-push-tag.yml
+++ b/.github/workflows/pull-request-update-or-push-tag.yml
@@ -20,7 +20,7 @@ jobs:
           cache: npm
 
       - run: npm install --ci
-      - run: npx eslint
+      - run: npx eslint --max-warnings=0
       - run: npx tsc -p e2e
 
   prepare-code-coverage:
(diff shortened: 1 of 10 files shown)


sleidig (approved):
this is a nice solution, marking the exceptions explicitly one by one :+1:

sleidig:
@geigerzaehler looks like we messed up some detail in the other PR #3002 :

> [Invalid workflow file: .github/workflows/on-master-push.yml#L108](https://github.com/Aam-Digital/ndb-core/actions/runs/15067898694/workflow)
The workflow is not valid. .github/workflows/on-master-push.yml (Line: 108, Col: 9): Job 'pre-release' depends on unknown job 'test'.

[source](https://github.com/Aam-Digital/ndb-core/actions/runs/15067898694)

Do you wanna fix this here? Otherwise I'll look into it later

geigerzaehler:
> Do you wanna fix this here? Otherwise I'll look into it later

I’m on it

PR #3005, opened 2025-05-16
refactor: Extract QA CI workflow

geigerzaehler:
Extract QA into a separate workflow and call it on master, tag and PR builds.

- [x] Update names of required status checks for main branch (@sleidig)

### DIFF
--- a/.github/workflows/on-master-push.yml
+++ b/.github/workflows/on-master-push.yml
@@ -6,6 +6,9 @@ on:
       - master
 
 jobs:
+  qa:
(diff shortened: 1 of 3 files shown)


.github/workflows/on-master-push.yml
```diff
@@ -6,6 +6,10 @@ on:
       - master
 
 jobs:
+  qa:
+    uses: ./.github/workflows/_qa.yaml
```
sleidig, line 10: nice. Always wondered whether something like this is possible

sleidig (approved):

PR #3007, opened 2025-05-17
refactor: Build webapp files once outside of docker

geigerzaehler:


sleidig (approved):
Once again a really nice simplification. Thanks, @geigerzaehler !

PR #3011, opened 2025-05-19
refactor: Prepare for new Angular build system

geigerzaehler:
This is the first set of fixes for issues that popped up when I tried to change the Angular build architect
1. Convert CommonJS modules in the codebase to ES modules
2. Hardcode dynamic import identifiers

### DIFF
--- a/src/bootstrap-i18n.ts
+++ b/src/bootstrap-i18n.ts
@@ -4,7 +4,7 @@ import {
 } from "./app/core/language/language-statics";
 import { loadTranslations } from "@angular/localize";
 import { registerLocaleData } from "@angular/common";
-import * as parseXliffToJson from "./app/utils/parse-xliff-to-js";
+import parseXliffToJson from "./app/utils/parse-xliff-to-js";
 
 /**
  * Load translation files and apply them to angular localize system.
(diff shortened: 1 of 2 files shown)


src/bootstrap-i18n.ts
```diff
@@ -32,9 +32,16 @@ export async function initLanguage(locale?: string): Promise<void> {
   $localize.locale = locale;
   // This is needed for locale-aware components & pipes to work.
   // Add the required locales to `webpackInclude` to keep the bundle size small
-  const localeModule = await import(
-    /* webpackInclude: /(fr|de|it)\.mjs/ */
-    `../node_modules/@angular/common/locales/${locale}`
-  );
-  registerLocaleData(localeModule.default);
+  let localeModule;
+  if (locale == "de") {
```
sleidig, line 36: if possible, can we build the locales generically, maybe with the following enum (to limit the number of code changes when adding a new locale in the future)?

https://github.com/Aam-Digital/ndb-core/blob/7afebe91d198266cb715d52ebf0b8abd68f03cf0/src/app/core/language/languages.ts#L8-L13

geigerzaehler: > if possible, can we build the locales generically

I don’t quite understand what you mean, could you elaborate?

The main thing about the code change is that all the imports are hard-coded so they can be statically analyzed. That means there is always a code change required when adding a locale. But this is intentional, because we’re telling the builder to also build the new file.

sleidig: ah, so the exact locales to need to be hard-coded here? Okay, that's fine then.

geigerzaehler: Yeah. The parameter to `import()` must be a string literal so that the builder can analyze it at build time. It cannot be runtime dependent.

Issue #2348, opened 2024-04-18
Fields should have a max width


sleidig:
If only one fieldGroup / column is displayed in a form, the fields fill the whole width. On large screens, this is not ideal:
![image](https://github.com/Aam-Digital/ndb-core/assets/1682541/36ccef0f-5fe3-476f-9bd4-93c040f34131)

we should set a max width there somewhere

Abhinegi2:
Currently working on this issue and, after reviewing the code, two possible approaches come to mind:

1. We can apply a max-width by targeting the `entity-field-edit` component. However, it seems we can't directly set the width there due to how it's being used (like we are using appdynamiccomponent there ). Instead, we could apply a max-width style in the SCSS of the parent components wherever `entity-field-edit` is used.
2. We can apply max-width styles directly within the SCSS files of the edit components themselves (e.g., edit-text, edit-date, etc.). A cleaner approach might be to define a global class for max-width styling, which can then be reused across these components. This way, we won’t need to redefine styles every time we use these components in the future.

So the 2nd approach looks good to me, also wanted to know your thoughts on this.

PR #3013, opened 2025-05-20
Fields should have a max width

Abhinegi2:
closes: #2348

### DIFF
--- a/src/app/core/common-components/entity-form/entity-form/entity-form.component.scss
+++ b/src/app/core/common-components/entity-form/entity-form/entity-form.component.scss
@@ -16,6 +16,9 @@
      of the columns-wrapper class */
   mat-form-field {
     width: 100%;
+    @media screen and (min-width: 1440px) {
+      width: 60%;
+    }
   }
 
   /* We align the photo (and only tht photo) to the center of the cell if there is one.


Abhinegi2:
@george-neha 

**To reproduce the issue:** In[ demo mode](https://demo.aam-digital.com/admin/entity/School?mode=details), go to the School section and try removing all fields except one, then save it. If you now open any school details or try creating a new school, you’ll notice that the remaining field takes up the full width of the screen.

**Problem/Issue:** This behavior is not ideal on large screens, when there's only one fieldGroup or column, it shouldn't use the full width.

**Fix:**
here: https://pr-3013.aam-digital.net/
Now, on wider screens (e.g., above 1440px), the single field no longer visible across the full screen. You can test this by inspecting the form and resizing the screen.

src/app/core/common-components/entity-form/entity-form/entity-form.component.scss
```diff
@@ -16,6 +16,9 @@
      of the columns-wrapper class */
   mat-form-field {
     width: 100%;
+    @media screen and (min-width: 1440px) {
+      width: 60%;
+    }
```
sleidig, line 21: is there a special reason why you decided to use `@media` rather than a simple `max-width: 864px` (= 60% * 1440)?

sleidig (approved):

src/app/core/common-components/entity-form/entity-form/entity-form.component.scss
Abhinegi2: not a specific reason, but the thought I had was- if we directly use max-width: 864px, and the screen is wider than 1440px (which is rare), there would be a lot of gap between the form fields and the save button due to the fixed width.

That’s why I preferred using a media query with a 60% width to make it responsive for larger screens.

PR #3017, opened 2025-05-21
Public Form: n.getId is not a function

Abhinegi2:
closes: #3015


sleidig (approved):
Thanks for the good debugging. I move your fix in a slightly more generic variant to the config-service.

Tested on affected customer system :+1:

Issue #2976, opened 2025-05-07
user docs for automated record updates feature


sleidig:
The newly released feature (automated updates of values in one record based on rules when a linked record is edited) could use a short user guide for our Support Center.

george-neha:
@sleidig Can you please review the draft i have created- https://docs.google.com/document/d/1ElhTTwAVHtd-mPRY28vlSrAw8KaSViitQfDHCPWJFvA/edit?usp=sharing

sleidig:
Adapted the standard demo setup here (including a pre-configured status automation rule):
https://pr-3029.aam-digital.net/

Please have a look and share your thoughts and suggestions, @george-neha

PR #3029, opened 2025-05-26
fix(demo): add default config for auto-update rule

sleidig:
- see #2976

added some additional field and config to standard demo to demonstrate the new feature


george-neha:
Thank you for adding the field in the demo env. Have updated the user document accordingly.

Issue #3027, opened 2025-05-23
Public form become blank on editing the label.


george-neha:
**Describe the bug**
On editing the label of "Public forms", the entire screen becomes blank.

**To Reproduce**

1. Navigate to " Admin"
2. Click on "Manage Public forms"
2. Click on Three dot menu on top right corner to Edit data structure
3. Click on "General Settings"
4. Edited the field "Label" and "Label(plural). 
5. Click on "Save"
6. The list of records is not longer available and the page becomes blank.

<img width="964" alt="Image" src="https://github.com/user-attachments/assets/92fbf04e-045a-4767-8cf3-f2f128e9fa16" />

Abhinegi2:
@george-neha I've implemented a quick fix for this issue, so you can proceed with your testing now. As we discussed, we're considering restricting user access to update settings(Edit data structure) for these types of entities. I'm currently working on that, but in the meantime, feel free to continue your work.

PR #3030, opened 2025-05-27
 Public form become blank on editing the label. 

Abhinegi2:
closes: #3027


sleidig (approved):
Nice simple solution. Thanks!

Issue #2989, opened 2025-05-13
Demo-Flow: UI to choose use case


sleidig:
Directly at startup of the app, show a view to users where they can choose from different "use cases". Each use case contains a different config doc to set up a specific demo scenario. This will replace the different manually set up demo systems currently [linked on our website](https://aam-digital.com/).

Mockup:
![Image](https://github.com/user-attachments/assets/3c94818a-7004-49be-8d24-2beaeaaa5838)
[Figma](https://www.figma.com/design/8kEk9KOKSyyMzWkR3JWqV0/Aam-Digital---Designs?node-id=1703-6784&t=lK64nZF2Cci7HEC5-0)

-----
## Implementation steps

- [ ] new "assistant" component that adds a button to the toolbar and opens a kind of side pane. This component should become the container for various different virtual assistance modes.
- [ ] new "choose-use-case" component that offers a UI to select different setups and see a description for each
- [ ] new service to load available use-cases from the assets folder (?). Probably the existing format of "baseConfigs" [used in ndb-setup](https://github.com/Aam-Digital/ndb-setup/tree/master/baseConfigs/default) could be adopted here
- [ ] adapt the bootstrap process of the app in AppComponent. 
   - [ ] This is currently showing a loading screen until config is available. Needs to be changed to allow us to show the assistant with the choose-use-case view instead.
   - [ ] If there is a config, we should however continue "normally" as before. And if there is a remote sync connection, we should make sure to first sync and load any existing config from the server.

george-neha:
Suggestion for the text: Welcome to Aam Digital!

Your path to customizing your case management needs. You can explore some of our pre-set up scenarios below

Aam Digital is fully customizable and can be quickly adapted to your needs and data structures. It is not limited to these scenarios. Through Aam Digital, you can track, measure, and manage your project.

george-neha:
In the current design the grey-ing out the options in the dropdown is giving the illusion to the user that these options are non clickable and also adding to the cognitive load of reading instructions below. Also in the future we offer the demo in more languages this might not be optimal way to showcase this.

<img width="459" alt="Image" src="https://github.com/user-attachments/assets/85f590ef-f5e0-4b49-af8b-2e0058cc4b31" />

 Instead I would suggest all the dropdown options, can these be in the same font color, and instead show a tick sign beside the option selected by the user.

[...]

PR #3031, opened 2025-05-27
create initial prototype of setup-service

sleidig:
- closes: #2989

todo-

- [x] add i18n (check if npm packages are available or extend own script to extract markers)
- [x] update hint text (neha will share)
- [x] Update democomponent for small screens.
- [x] Add/Update some basic test cases
- [x] only show "choose use case" if no config is there yet
- [x] fix problems with other base-configs
- [x] add German (de) and French (fr) json files for MarkdownContent_help, Config_SetupWizard
- [x] ? switch locale based on base-config
- [x] queryParam to start with automatically loading a use-case already 
- [x] Update demo initialization logic for sync mode to handle 404 screen issue (occurs after clearing site data).
- [x] On clicking "Initialize System," verify if the selected use case matches the current local storage language. If not, update the local storage language and query parameters accordingly.
- [x] Implement Neha’s suggestion on handling use case options and localization settings.
- [x] Use use case ID instead of name/label in query parameters for consistency and reliability.
- [x] fix not being able to sign out anymore with demo_mode (e.g. for switching to demo-admin)
- [x] in mock mode, logging out and logging in as demo-admin re-routes to 404 page after login
- [x] toolbar height on mobile screen is different -> ensure all styles are relative to this
- [x] scrolling in assistant view is not working properly

### DIFF
--- a/.releaserc
+++ b/.releaserc
@@ -1,6 +1,7 @@
 {
   "tagFormat": "${version}",
   "branches": [
+    "official-release",
     {"name": "master", "prerelease": true}
   ],
   "plugins": [


src/app/core/admin/setup-wizard/setup-wizard-config.ts
```diff
@@ -17,88 +17,3 @@ export interface SetupWizardStep {
   text: string;
   actions?: MenuItem[];
 }
-
-export const defaultSetupWizardConfig: SetupWizardConfig = {
```
sleidig, line 21: TODO: switch demo to German / French and copy the texts into json files for assets under education_de + education_fr

src/app/core/setup/base-config.ts
```diff
@@ -0,0 +1,31 @@
...
+   * (can contain Markdown formatting).
+   */
+  description: string;
+
+  /**
+   * Names of json files that should be imported during the setup process.
+   * Files have to be located in the `assets/base-configs/${id}/` folder
+   * and match
+   */
+  entitiesToImport: string[];
```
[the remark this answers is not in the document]
sleidig: resolved by allowing the files to be either arrays of entity docs or a single entity doc. This will allow us to flexibly compose use cases sharing different pieces of config.

src/app/core/setup/demo-assistant-button/demo-assistant-button.component.scss
```diff
@@ -0,0 +1,19 @@
+@use "variables/breakpoints";
+
+.demo-assistant-btn {
+  @media screen and (max-width: breakpoints.$sm) {
+    min-height: 56px;
+  }
+  min-height: 65px;
```
sleidig, line 7: is it possible to work with relative heights (100%?) here instead or was this the only way to get it to work?

src/app/core/ui/ui/ui.component.html
```diff
@@ -52,14 +53,16 @@
...
+    <app-demo-assistant-button></app-demo-assistant-button>
   </div>
 </mat-toolbar>
 
 <!-- MAIN NAVIGATION + CONTENT -->
 <mat-sidenav-container (backdropClick)="closeSidenavOnMobile()" autosize>
   <mat-sidenav
     #sideNav
-    *ngIf="isLoggedIn()"
+    *ngIf="isLoggedIn | async"
```
sleidig, line 65: somehow this is not hiding it during choose-use-case anymore for me right now ...

src/app/core/setup/demo-assistant-button/demo-assistant-button.component.ts
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  /**
+   * Opens a dialog to assist the user in setting up a demo environment.
+   * Depending on the user's login state, it opens either a context-aware dialog (if logged in)
+   * or a demo assistance dialog (if not logged in). The dialog guides the user through selecting
+   * a use case and initializing the system with the corresponding demo data.
+   * @returns A promise that resolves with the dialog result when the dialog is closed.
+   */
+  async openDemoSetupDialog() {
```
sleidig, line 46: assistant should close when clicking anywhere outside the panel (if not locked, like initial setup)

src/app/core/setup/choose-use-case/choose-use-case.component.html
```diff
@@ -0,0 +1,26 @@
...
+      <mat-select
+        [(value)]="selectedUseCase"
+        (selectionChange)="onSelectionChange()"
+      >
+        @for (useCase of demoUseCases; track useCase) {
+          <mat-option
+            [value]="useCase"
+            [class.greyed-out]="useCase.locale !== locale"
+          >
+            {{ useCase.name }}
```
sleidig, line 14: seeing this now, I think we have to somehow also show the language and explain the greyed out entries ...

src/app/core/setup/choose-use-case/choose-use-case.component.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  constructor(@Inject(WINDOW_TOKEN) private window: Window) {
+    this.locale =
+      this.window.localStorage.getItem(LANGUAGE_LOCAL_STORAGE_KEY) ||
+      DEFAULT_LANGUAGE;
+  }
+
+  onSelectionChange() {
+    if (this.selectedUseCase?.locale) {
+      this.window.localStorage.setItem(
```
sleidig, line 48: this doesn't seem to work as intended yet: language is not change directly (only when manually re-loading the page). In the language switcher we also reload the page to change the language ... maybe that's necessary.

Either way, we should implement an option to pass the selected base-config through url param, so maybe this is an option to solve this?

src/app/core/ui/ui/ui.component.html
Abhinegi2: I found the issue. Earlier, when things were working, we mistakenly imported merge from hammerjs. At that time, the detectConfigReadyState() method wasn’t returning anything, so the navigation remained hidden until the setup demo process was completed, and it started returning true.

now, however, we’ve correctly switched to using merge from rxjs. In this setup, detectConfigReadyState() returns true as either a config is present or the sync is complete (which indicates that there’s no existing config on the server). Because of this, the navigation is shown immediately after the system initializes, even when the user is just choosing a use case.
That’s why the navigation isn’t being hidden anymore.

src/app/core/setup/demo-assistant-button/demo-assistant-button.component.ts
```diff
@@ -0,0 +1,81 @@
...
+  private readonly dialog = inject(MatDialog);
+
+  private isDialogOpen: boolean = false;
+
+  assistantEnabled: boolean;
+
+  ngOnInit(): void {
+    this.assistantEnabled = environment.demo_mode;
+
+    if (!this.configService.hasConfig()) {
```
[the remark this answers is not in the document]
sleidig: Thanks for the analysis, I agree with you: first option, combining the conditions, seems the better appraoch to me.

src/app/app.component.ts
```diff
@@ -16,69 +16,54 @@
...
+    private loginState: LoginStateSubject,
+    private demoDataInitializer: DemoDataInitializerService,
+    private setupService: SetupService,
   ) {
-    this.detectConfigReadyState();
+    this.configReady$ = this.loginState.pipe(
+      // if logged out, we don't wait for config and treat this separately
+      map((loginState) => loginState !== LoginState.LOGGED_IN),
+      // immediately switch the state based on loginState but then take time for config readiness
+      mergeMap((loggedOut) => {
```
sleidig, line 55: @Abhinegi2 I decided to move the full-screen mode of admin views to the UI component for better UX and simpler logic here. Do have a look whether my code makes sense

sleidig:
e2e tests are still failing. Could you have a look at them, @Abhinegi2 ?

Abhinegi2:
> e2e tests are still failing. Could you have a look at them, @Abhinegi2 ?

sure i will check it

src/app/core/setup/system-init-assistant/choose-use-case/choose-use-case.component.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  @Output() selectionChanged = new EventEmitter<BaseConfig>();
+
+  selectedUseCase: BaseConfig;
+  locale: string;
+
+  constructor(@Inject(WINDOW_TOKEN) private window: Window) {
+    this.locale =
+      this.window.localStorage.getItem(LANGUAGE_LOCAL_STORAGE_KEY) ||
+      DEFAULT_LANGUAGE;
```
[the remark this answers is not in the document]
Abhinegi2: We currently have a `currentRegionCode` method in the LanguageService, but it seems unused (only used in test files). 
I’m replacing/updating it with a getCurrentLocale method, which returns the locale (e.g., ‘en-US’). This updated method will be used in the ChooseUseCaseComponent.

e2e/fixtures.ts
```diff
@@ -39,10 +36,14 @@ export const test = base.extend<{ forEachTest: void }>({
 });
 
 async function initSystemWithBaseConfig(page: Page) {
```
sleidig, line 38: @geigerzaehler , could you briefly have a look at this e2e init change?

We are completely reworking the UX of the (demo) system initialization, so the user now first has to interact with a system-init-assistant view before anything else is set up and available.

src/app/core/demo-data/demo-data-initializer.service.ts
```diff
@@ -88,6 +89,8 @@ export class DemoDataInitializerService {
   }
 
   private syncDatabaseOnUserChange() {
+    if (this.databaseSynced) return;
```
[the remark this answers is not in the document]
sleidig: I can't see any use of the run() method except in tests. Is this really still called in the app?

src/styles/variables/_sizes.scss
```diff
@@ -24,3 +24,8 @@ $margin-main-view-top: 16px;
 $margin-main-view-bottom: 64px;
 
 $form-group-min-width: 250px;
+
+$toolbar-height-mobile: 56px;
+$toolbar-height-desktop: 64px;
+
+$toolbar-height: clamp(#{$toolbar-height-mobile}, 10vw, #{$toolbar-height-desktop});
```
sleidig, line 31: I don't understand how/why this works but it seems to work :sweat_smile:

PR #3036, opened 2025-05-30
Stabilize faker date generation

geigerzaehler:
Make use of `faker.defaultRefDate` to ensure that date generation with faker is
stable across test runs.

Unfortunately, this does not fully address flaky tests. It seems like the
ordering of notes in the dashboard is not stable which points to some
indeterminism in the database.


sleidig:
Might be that using uuid functions is based on certain randomness outside of the fixed seeds

sleidig (approved):
Thanks! :-)

PR #3039, opened 2025-06-01
refactor: Type-check app code on CI

geigerzaehler:
I think it’s a good idea to type-check all app code separately. This surfaced on
type issue in one of the Storybook stories.

### DIFF
--- a/.github/workflows/_qa.yaml
+++ b/.github/workflows/_qa.yaml
@@ -24,6 +24,7 @@ jobs:
 
       - run: npm install --ci
       - run: npx eslint --max-warnings=0
+      - run: npx tsc -p .
       - run: npx tsc -p e2e
       - name: actionlint
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/_qa.yaml
```diff
@@ -24,6 +24,7 @@ jobs:
 
       - run: npm install --ci
       - run: npx eslint --max-warnings=0
+      - run: npx tsc -p .
       - run: npx tsc -p e2e
```
sleidig, line 28: is the first line automatically including the second, more narrow?

sleidig (approved):

.github/workflows/_qa.yaml
geigerzaehler: The project at the root excludes the e2e folder

https://github.com/Aam-Digital/ndb-core/blob/8def237856a75a92bbd7bb6b6999190748858c0e/tsconfig.json#L52-L54

This is necessary (I think) for the app code to build properly. But we still want to type-check the e2e test code and this is what the second line is for.

sleidig: Ah, thanks for the explanation :+1:

Issue #2991, opened 2025-05-13
Demo-Flow: Welcome and further actions


sleidig:
After selecting a use case (#2989) and having demo data generated (#2990) the user should see a simple intro with important hints. From there, the user can close the assistant and start exploring the demo.

When the user opens the assistant again by clicking on the toolbar, the system should provide some basic explanations and (in the demo mode) some actions, like resetting the demo or registering for an account.

Mockups:
![Image](https://github.com/user-attachments/assets/79d1cd81-8500-4f0f-b555-30cf55e4a49b)
![Image](https://github.com/user-attachments/assets/4276531e-9151-484d-9141-0803869578ab)

-----
## Implementation Steps

- [x] new component for (context-aware?) hints
- [ ] load different actions from a database entity (where we can configure multiple items with title and (external) link)

PR #3040, opened 2025-06-02
Demo-Flow: Welcome and further actions 

Abhinegi2:
closes: #2991

### Visible/Frontend Changes

- [x]
- [ ]

- [x]
- [ ]

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -22,6 +22,7 @@ import { ConfigService } from "./core/config/config.service";
 import { LoginStateSubject } from "./core/session/session-type";
 import { LoginState } from "./core/session/session-states/login-state.enum";
 import { SetupService } from "./core/setup/setup.service";
+import { environment } from "environments/environment";
 
 /**
  * Component as the main entry point for the app.
@@ -51,7 +52,11 @@ export class AppComponent {
     protected loginState: LoginStateSubject,
     private setupService: SetupService,
   ) {
-    this.setupService.openDemoSetupDialog();
+    if (environment.demo_mode) {
+      // If we are in demo mode, we open the setup dialog immediately
+      // to allow the user to select a base config.
+      this.setupService.openDemoSetupDialog();
+    }
 
     this.detectConfigReadyState();
     this.detectConfigMode();
--- a/src/app/core/setup/context-aware-dialog/context-aware-dialog.component.ts
+++ b/src/app/core/setup/context-aware-dialog/context-aware-dialog.component.ts
@@ -1,0 +1,30 @@
+import { Component } from "@angular/core";
+import { MatButtonModule } from "@angular/material/button";
+import { MatDialogActions, MatDialogContent } from "@angular/material/dialog";
+import { DemoAssistanceDialogComponent } from "../demo-assistance-dialog/demo-assistance-dialog.component";
+
+// Define a shared interface for dialog components
+export interface DemoDialogComponentBase {
+  // Add any shared methods/properties here
+}
+
+@Component({
+  selector: "app-context-aware-dialog",
+  imports: [MatDialogContent, MatDialogActions, MatButtonModule],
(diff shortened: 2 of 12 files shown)


sleidig:
> When using the system with demo_mode and a session local, the following error occurs after selecting a use case: Error initializing demo data: conflict: Document update conflict (unable to resolve) ID: {"data": ...}

Whatever docs are already in the database (e.g. the Config entity, but maybe also ConfigurableEnum entities etc.) should be kept and not overwritten. "local" mode should function like "synced" only without the remote server database as a counterpart, so if the system is set up already, the setup actions have to be skipped automatically

src/app/core/setup/context-aware-dialog/context-aware-dialog.component.ts
```diff
@@ -0,0 +1,30 @@
+import { Component } from "@angular/core";
+import { MatButtonModule } from "@angular/material/button";
+import { MatDialogActions, MatDialogContent } from "@angular/material/dialog";
+import { DemoAssistanceDialogComponent } from "../demo-assistance-dialog/demo-assistance-dialog.component";
+
+// Define a shared interface for dialog components
+export interface DemoDialogComponentBase {
+  // Add any shared methods/properties here
+}
```
sleidig, line 9: good idea. can we already outline some basic properties?

src/app/core/setup/demo-assistance-dialog/demo-assistance-dialog.component.html
```diff
@@ -1,34 +1,65 @@
...
+      </app-choose-use-case>
+    </div>
+
+    <p class="footer-note" i18n>
+      Aam Digital is fully customizable and can be quickly adapted to your needs
+      and data structures. It is not limited to these scenarios. Through Aam
+      Digital, you can track, measure, and manage your project.
+    </p>
+  } @else {
+    <p>
```
sleidig, line 28: i18n (below also)

src/app/core/setup/demo-assistance-dialog/demo-assistance-dialog.component.ts
```diff
@@ -34,9 +35,26 @@ export class DemoAssistanceDialogComponent implements OnInit {
...
+            this.demoInitialized = true;
+            this.generatingData = false;
+          })
+          .catch((error) => {
+            console.error("Error initializing demo data:", error);
+            this.generatingData = true;
+          });
+      } else {
+        this.generatingData = false;
+      }
```
sleidig, line 53: For me, it would be easier to read with a structure like this:

```
loading = true;
await init();
loading = false;
```

sleidig (approved):
Looks pretty good :+1: 

@Abhinegi2 , as this overlaps with some of the base PR, maybe combine them soon and then work on remaining improvements in one branch?

src/app/app.component.ts
```diff
@@ -51,7 +52,11 @@ export class AppComponent {
     protected loginState: LoginStateSubject,
     private setupService: SetupService,
   ) {
-    this.setupService.openDemoSetupDialog();
+    if (environment.demo_mode) {
```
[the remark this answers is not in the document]
Abhinegi2: Yes, we need to check if there’s any existing config, and if not, we should trigger the demo flow. 
I’m also facing a similar issue in local mode, so it might sync up on this.

 For now, we could merge this branch into the base and handle the config check there?

src/app/core/setup/context-aware-dialog/context-aware-dialog.component.ts
sleidig: if it's not used anywhere yet, I would tend to remove it for now until we need it. (We sometimes ended up with forgotten duplicate structures :sweat_smile: )

sleidig (approved):
Go ahead and merge this into the other feature-branch, @Abhinegi2 :+1:

PR #3043, opened 2025-06-02
fix(user management): fix errors loading account page for records wit…

sleidig:
fix errors loading account page for records without existing account


Abhinegi2 (approved):
Thanks, working as expected!

PR #3045, opened 2025-06-02
Allow e2e tests to use dedicated database fixtures

### LINKED ISSUE #2995: E2E: Database fixtures
Currently our E2E tests use the same demo data that's used for development, which causes several problems:

- Changes to demo data for development can accidentally break tests
- Demo data generation is slow and contains more data than needed for tests
- We can't create specific test data for individual test cases

Proposed solution: Create a test fixture system to populate the database with specific test data.

We can build on the existing `DemoDataGenerator` code and extend it for test-specific needs.

Technical challenge: Need to create an interface between browser-based database and test code running on the host.

geigerzaehler:
- closes #2995

Add a minimal mechanism that allows us to build database fixtures specific to end-to-end test cases and load them in the app.

I tried to keep the scope small so we can discuss the approach without implementing everything already.

Follow-ups for #2995

- [ ] Allow for more control how data is generated. E.g. ensure that there’s a task that is due.
- [ ] Generate tasks that are due and notes that need a follow-up and test for them
- [ ] Use mechanism in remaining test cases
- [ ] Refactor generator code so that it is less coupled to Angular DI and its APIs are simpler to use


geigerzaehler:
I was able to get rid of EntityMapper :tada:

geigerzaehler:
I’m still waiting for feedback but I think we should move forward with this for know and adapt along the way.

I noticed that now the `DemoDataGenerator` base class requires `EntityRegistry` to be injected. This means that I can no longer instantiate these classes as I was doing [in this PR initially](https://github.com/Aam-Digital/ndb-core/pull/3045/files#diff-8a8a908cf39d2b2dfd04537c9e26f503889e65761ab5be41847fcb58a4a83416R9).

This is not a huge problem but I’d like to adjust our plan a bit and create an intermediate PR that moves the data generation logic from class methods to static functions so that we can use the functions in tests. For example

```typescript
@Injectable()
export class DemoUserGeneratorService extends DemoDataGenerator<Entity> {
  static DEFAULT_USERNAME = TEST_USER;
  static ADMIN_USERNAME = "demo-admin";

  static provider() {
    return [
      { provide: DemoUserGeneratorService, useClass: DemoUserGeneratorService },
    ];
  }

  generateEntities = generateUsers;
}

export function generateUsers(): Entity[] {
  /// ...
}
```

geigerzaehler:
This is ready for review again

e2e/fixtures.ts
```diff
@@ -49,3 +64,63 @@ export async function argosScreenshot(
...
+        new EventAttendanceMapDatatype(entitySchemaService),
+        new LongTextDatatype(),
+      ] satisfies DefaultDatatype[];
+    },
+  });
+
+  const config = new Config(Config.CONFIG_KEY, defaultJsonConfig);
+  const entityConfigService = new EntityConfigService(null, entityRegistry);
+  entityConfigService.setupEntities(
+    Object.entries(config.data.data).flatMap(([id, config]) => {
```
sleidig, line 114:
```suggestion
  const entityConfigService = new EntityConfigService(null, entityRegistry);
  entityConfigService.setupEntities(
    Object.entries(defaultJsonConfig.data).flatMap(([id, config]) => {
```

Could possibly simplify it like this?

src/app/core/demo-data/demo-data-initializer.service.ts
```diff
@@ -69,7 +69,12 @@ export class DemoDataInitializerService {
       DemoDataGeneratingProgressDialogComponent,
     );
 
-    await this.demoDataService.publishDemoData();
+    if (window["ndbDemoData"]) {
```
sleidig, line 72: the "ndb" prefix (assuming you picked this from the ndb-core repo name?) is super old and a bit out of context nowadays. This was the working name initially but now we are only using "aam".
Maybe I would rename this to just "demoData" here - or "e2eDemoData"?

tsconfig.json
```diff
@@ -35,6 +35,7 @@
       "node"
     ],
     "paths": {
+      "#src/*": ["./*"],
```
sleidig, line 38: did we resolve the problems that in new code IDE's use this for imports but then the references don't work properly?

src/app/core/demo-data/demo-data-initializer.service.ts
```diff
@@ -69,7 +69,12 @@ export class DemoDataInitializerService {
...
+      const db = this.dbResolver.getDatabase();
+      await db.putAll(window["ndbDemoData"]);
```
sleidig, line 74: if we are saving the demo data within the Angular app, couldn't we do the `entities.map((e) => entitySchemaService.transformEntityToDatabaseFormat(e));` only here and save the hassle to init that in the e2e fixture?

I know we talked about why this didn't work - but I forgot or am not sure if those problems still apply in the final architecture here now :blush:

tsconfig.json
geigerzaehler: This now works with the angular builder. So if an app source file uses this path the builder resolves it properly. But please check that it works with your IDE as expected.

src/app/core/demo-data/demo-data-initializer.service.ts
geigerzaehler: Unfortunately this doesn’t work since `transformEntityToDatabaseFormat(entity)` expects entity to be an instance of `Entity` so that `entity.getSchema()` is defined. But `window.demoData` is only raw JSON. We can only send JSON from the the test runner to the browser.

src/app/core/demo-data/demo-data-initializer.service.ts
geigerzaehler: Makes sense, fixed in https://github.com/Aam-Digital/ndb-core/pull/3045/commits/93552204d956d80413e2df23b8d9adc081728969

PR #3051, opened 2025-06-03
form field label is cut off with "..." but no tooltip is shown for entity array fields (?)

Abhinegi2:
closes:  #3050


Abhinegi2:
@george-neha, could you please test the fix for the tooltip issue on multiselect entity-type fields?

Steps to test:
1. Create a new field linked to another entity and enable the multi-select option.
(For example: Add a new field with a long label in the School entity and link it to the Child entity with multi-select enabled.)
2. Open the details of any School entry and select multiple Child entities in that field.
When the field label gets truncated (shown as "..."), hover over it tooltip should now show the full label.

sleidig (approved):

PR #3052, opened 2025-06-03
fix: Make viewport scroll container

### LINKED ISSUE #3046: Take screenshots of full page to check for visual regressions
It’s a good practice to take screenshots of the full page to check for visual regressions so that you don’t miss something that is outside of the viewport. Although we use `fullPage: true` for the screenshots

https://github.com/Aam-Digital/ndb-core/blob/b52b166744ca0e5ec930ba73a7faff1a8d4ba250/e2e/fixtures.ts#L44

the screenshots [are not](https://app.argos-ci.com/sleidig/ndb-core/builds/73/164772602) of the full page.

The reason for this behavior is that in the app not the view port is scrolled but the content of `<mat-sidenav-content>`. To fix this we need to change the layout styles of the app so that the document element (`<html>`) is not restricted in size and scrolls freely.

geigerzaehler:
Fixes #3046

### DIFF
--- a/src/app/core/ui/primary-action/primary-action.component.scss
+++ b/src/app/core/ui/primary-action/primary-action.component.scss
@@ -3,7 +3,7 @@
 }
 
 :host {
-  position: absolute;
+  position: fixed;
   bottom: 10px;
   right: 10px;
   z-index: 999;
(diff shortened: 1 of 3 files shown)


geigerzaehler:
As you can [see](https://app.argos-ci.com/sleidig/ndb-core/builds/84/165123098) the screenshots are a lot more useful now.

sleidig (changes requested):
The previous css looks weird but I think we were struggling with some points to get the ng-content filling the space properly ...

At the moment the changes still have a side effect, making the primary action scroll along:
image

PR #3054, opened 2025-06-03
refactor: Use CSS animation for roll call

geigerzaehler:
This allows us to remove workarounds from tests and makes the tests about 3s faster.


sleidig (approved):
Looks good :+1:

PR #3055, opened 2025-06-05
Hide sections without permissions #769

Abhinegi2:
closes: #769

### DIFF
--- a/src/app/core/entity-details/entity-details/entity-details.component.spec.ts
+++ b/src/app/core/entity-details/entity-details/entity-details.component.spec.ts
@@ -97,6 +97,36 @@ describe("EntityDetailsComponent", () => {
       }),
     );
   }));
+
+  it("filters out panels not permitted for the current user role", fakeAsync(() => {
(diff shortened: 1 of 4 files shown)


src/app/core/entity-details/entity-details/entity-details.component.spec.ts
```diff
@@ -97,6 +97,36 @@ describe("EntityDetailsComponent", () => {
       }),
     );
   }));
+
+  it("filters out panels not permitted for the current user role", fakeAsync(() => {
+    const testChild = new TestEntity("Role-Test");
+    testChild["_rev"] = "1";
```
sleidig, line 103: setting _rev should not be necessary?

sleidig (approved):
Looks and works well! :100: 

(after reverting the config change done for testing the e2e tests should also pass again, I think)

Abhinegi2:
> Looks and works well! 💯
> 
> (after reverting the config change done for testing the e2e tests should also pass again, I think)

yes e2e tests are working fine now

PR #3056, opened 2025-06-05
fix(core): init logging only after loading config into environment

sleidig:


sleidig:
> Looks good! just one question out of curiosity, why was the previous approach changed? Was there any specific issue it caused?

The key change is that we only init the sentry logging _after_ loading the config.json. Currently, we cannot overwrite the sentry dsn or the production true/false for the logging because it is initalized before. This means all dev and pr systems log errors marked as "production" environment.

Issue #2541, opened 2024-08-27
EditEntity (entity select field) may discard selected values if the user is missing permissions on them


sleidig:
**To Reproduce**
assume a user who has access only to children with a certain category.

1. Edit a Note that links to children, including some the current user cannot access
2. Trying to close the form causes a warning about unsaved changes - even without enabling edit mode
3. Editing and saving the note will discard the children from the entity-select field that this user cannot access

**Expected behavior**
We currently remove refs that are not found. Instead, we should probably keep them and display some placeholder chip in the select fields saying "not found or not accessible".

TODO:

- [x] Hide unrelated entities (i.e., entities the current user does not have access to) and store their reference instead of removing them entirely.
- [x] Display a warning indicating:
  *"Some records could not be found, accessed, or appear to be corrupted."*

Abhinegi2:
@sleidig ,

I tried to replicate the reported issue but haven’t been able to reproduce it yet. Here’s what I did:
1.	I created a new role called self-only in Keycloak.
2.	Then, I updated the config:permission document for the Child entity with the following condition:
```
{
  "subject": "Child",
  "action": ["create", "read", "update"],
  "conditions": {
    "created.by": {
      "$eq": "${user._id}"
    }
  }
}
```

3.	However, after applying this, I ran into a problem — our Keycloak server is blocking localhost, so I couldn’t create a new user with the self-only role via the Users screen for local testing.
4.	To work around this, I hardcoded the condition temporarily like so:
```
"created.by": {
  "$eq": "User:demo"
}
```

5.	As expected, this prevents me from creating or updating a Child unless the created.by field matches User:demo.
6.	Then I went to the Notes module and tried editing the children array. but, it still allowed saving a reference to a child I do not have access to (i.e., a Child created by another user), and the note was saved with that inaccessible child still included in the array.

Let me know if I might be missing a step in replicating the original issue.

sleidig:
Hmm, I am able to reproduce the problem as described. Using the https://github.com/Aam-Digital/aam-services/tree/main/docs/developer local setup 

with permission rule:
```
{
  "_id": "Config:Permissions",
  "data": {
    "default": [
      {
        "subject": [
          "Config",
          "SiteSettings",
          "PublicFormConfig",
          "ConfigurableEnum"
        ],
        "action": "manage"
      }
    ],
    "admin_app": [
      {
        "subject": "all",
        "action": "manage"
      }
    ],
    "user_app": [
      {
[...]

PR #3065, opened 2025-06-11
EditEntity (entity select field) may discard selected values if the user is missing permissions on them 

Abhinegi2:
closes: #2541

### DIFF
--- a/src/app/core/common-components/basic-autocomplete/basic-autocomplete.component.html
+++ b/src/app/core/common-components/basic-autocomplete/basic-autocomplete.component.html
@@ -126,7 +126,6 @@
     <ng-container>
       <mat-chip-row
         *ngFor="let item of _selectedOptions"
-        [hidden]="item.isHidden"
         [editable]="!_disabled"
         class="chip"
       >
--- a/src/app/core/common-components/basic-autocomplete/basic-autocomplete.component.html
+++ b/src/app/core/common-components/basic-autocomplete/basic-autocomplete.component.html
@@ -126,6 +126,7 @@
     <ng-container>
       <mat-chip-row
         *ngFor="let item of _selectedOptions"
+        [hidden]="item.isHidden"
         [editable]="!_disabled"
         class="chip"
       >
--- a/src/app/core/common-components/basic-autocomplete/basic-autocomplete.component.ts
+++ b/src/app/core/common-components/basic-autocomplete/basic-autocomplete.component.ts
@@ -59,6 +59,7 @@ interface SelectableOption<O, V> {
   asString: string;
   asValue: V;
   selected: boolean;
+  isHidden: boolean;
 }
 
 export const BASIC_AUTOCOMPLETE_COMPONENT_IMPORTS = [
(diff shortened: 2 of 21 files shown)


src/app/core/common-components/entity-select/entity-select.component.ts
```diff
@@ -265,6 +284,15 @@ export class EntitySelectComponent<
...
+   * Combine the user’s selected IDs with any hidden ones we’re preserving.
+   */
+  private combineWithHiddenEntities(selected: string[]): string[] {
+    return Array.from(
+      new Set([...selected, ...Array.from(this.hiddenEntities)]),
+    );
+  }
 }
 
 function isMulti(
```
Abhinegi2, line 298: I think we can remove this as now we are not using it anymore!

src/app/core/common-components/basic-autocomplete/basic-autocomplete.component.ts
```diff
@@ -420,6 +423,7 @@ export class BasicAutocompleteComponent<O, V = O>
       asValue: this.valueMapper(opt),
       asString: this.optionToString(opt),
       selected: false,
+      isHidden: (opt as any)?.isHidden ?? false,
```
sleidig, line 426: would this also work?

```suggestion
      isHidden: (opt as SelectableOption)?.isHidden ?? false,
```

src/app/core/common-components/basic-autocomplete/basic-autocomplete.component.html
```diff
@@ -126,6 +126,7 @@
     <ng-container>
       <mat-chip-row
         *ngFor="let item of _selectedOptions"
+        [hidden]="item.isHidden"
```
sleidig, line 129: didn't we already filter these out when creating _selectedOptions? Is this still needed here now?

src/app/core/common-components/entity-select/entity-select.component.html
```diff
@@ -16,11 +16,13 @@
       [createOption]="disableCreateNew ? null : createNewEntity"
     >
       <ng-template let-item>
-        <app-entity-block
-          style="margin: auto"
-          [entity]="item"
-          [linkDisabled]="form.enabled"
-        ></app-entity-block>
+        @if (!item.isHidden) {
```
sleidig, line 20: couldn't / shouldn't we do this condition within the auto-complete component instead of just here?

src/app/core/filter/filter/filter.component.ts
```diff
@@ -102,9 +102,12 @@ export class FilterComponent<T extends Entity = Entity> implements OnChanges {
 
   filterOptionSelected(filter: Filter<T>, selectedOptions: string[]) {
     filter.selectedOptionValues = selectedOptions;
-    this.hasActiveFilters = this.filterSelections.some(
-      (f) => f.selectedOptionValues?.length > 0,
-    );
+    setTimeout(() => {
```
sleidig, line 105: I guess this had change detection issues? Any idea why/what exactly? (we should probably add code comments to such workarounds for documentation)

sleidig (commented):
works great, as expected. Thanks, @Abhinegi2 !

Just some small questions about potential fine-tuning.

sleidig (approved):

Issue #3066, opened 2025-06-12
Stabilize test screenshots of dashboard


geigerzaehler:
The e2e test screenshot of the dashboard changes from test run to test run (see [this Argos build](https://app.argos-ci.com/sleidig/ndb-core/builds/101/167347197) for example). We want to make the screenshot deterministic.

It looks like the biggest change comes from a different order of the items in the dashboard panels “Notes needing follow-up”, “Absences this week”, “Absences last week” and “Late last week”. We should be able to fix this by setting the order in the DB query for the items.

In addition it looks like there are also imperceptible single pixel differences. We probably can get rid of them by adjusting the sensitivity of the comparison with the [`threshold` parameter](https://argos-ci.com/docs/playwright#argosscreenshotpage-name-options).

sleidig:
Seems pretty stable now. Thanks a ton, @geigerzaehler ! 😍

PR #3067, opened 2025-06-12
Remove indeterminism from attendance and note demo data

geigerzaehler:
This should stabilize dashboard appearance. See #3066

### DIFF
--- a/src/app/core/entity/model/entity.ts
+++ b/src/app/core/entity/model/entity.ts
@@ -230,7 +230,7 @@ export class Entity {
   }
 
   /** actual id without prefix */
-  private get entityId(): string {
+  get entityId(): string {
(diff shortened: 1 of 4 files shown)


src/app/core/entity/model/entity.ts
```diff
@@ -230,15 +230,15 @@ export class Entity {
...
+  get entityId(): string {
     return Entity.extractEntityIdFromId(this._id);
   }
 
   /**
    * Set id without prefix.
    * @param newEntityId The new id without prefix.
    */
-  private set entityId(newEntityId: string) {
+  set entityId(newEntityId: string) {
```
sleidig, line 241:
```suggestion
  private set entityId(newEntityId: string) {
```

PR #3071, opened 2025-06-13
Properly limit width of autocomplete dropdown browsers other than Webkit

geigerzaehler:
Before it looked like this on Firefox

Screenshot 2025-06-13 at 10-46-08 Aam Digital - Demo

Now it’s fixed and looks the same across all browsers

Screenshot 2025-06-13 at 10-47-13 Aam Digital - Demo


sleidig (approved):
Thanks!

Issue #3076, opened 2025-06-16
Capture address details and Change the pinned location on the map


george-neha:
Scenario 4: As a user, I should be able to manually move the pin on the map without losing on the address details that i have captured.
When entering an address in the search bar, the possible/ suggested locations are available in a listed drop-down. I should be able to move the pinned location on the map. If the address captured does not match with the new pinned location, a dialog box should appear: " Address details captured does not match with the location on the map. Do you want to continue or edit the address?”

Dialog buttons: Continue/Edit Address
Continue> The address details captured should be saved along with the new the pinned location on the map.

PR #3080, opened 2025-06-17
feat: Allow manual pin move with address mismatch confirmation dialog

### LINKED ISSUE #2944: Editing the address details
**Edit Address details**

**Scenario :** I should be able to edit the child's address details. As a user, I should be able to edit the full address or parts. The newly captured address details should override the previously captured ones.
In scenarios where the user clicks on on one of suggested/possible locations listed, then the functionality should be as scenario 1 in https://github.com/Aam-Digital/ndb-core/issues/2943.

In scenarios where:
User edits the address details, but does not click one of the suggested/possible locations. Then the following should be triggered on save: “Do you want to update the location pinned on the map?”
If the user clicks no, then the address details are captured without changing the location on the map.
Example : 
Old Location:  Impact Hub, Rollbergstraße 28A, 12053 Berlin
User edits:  Impact Hub, 2nd floor, cabin 3 Rollbergstraße 28A, 12053 Berlin
New Location:  Impact Hub, 2nd floor, cabin 3 Rollbergstraße 28A, 12053 Berlin

Example : 
Old Location:  Impact Hub, Rollbergstraße 28A, 12053 Berlin
User edits: Impact Hub Stuttgart
New Location: Impact Hub, Reuterstrasse 17A, 12053 Stuttgart

### LINKED ISSUE #3074: Capturing additional details in the address field
As a user, I should be able to capture the address details given by the child in the Aam Digital System so that I can visualise them, if available, on a map without losing any additional data.

Scenario 2: As a user, I should be able to capture additional details like the house number, flat number, building name etc. so that these details are available in the system.
A second line should contain additional details captured that do not match the data pulled from the map API.

Example 2:
User input: “Impact Hub, 2nd floor, cabin 3 Berlin”
Address Displayed: Impact Hub, Rollbergstraße 28A, 12053 Berlin.
2nd floor, cabin 3
Example 3: User input: “Reuterstrasse 17A, Berlin”
Address displayed: “Reuterstrasse, 12053 Berlin.
17A

When you enter an address with additional details (like the house number, flat number, building name, etc.) in the search bar, the possible/ suggested locations should be available in a drop-down menu.
The user should be able to click on the suggested location from the list.
The additional details captured do not match the API; they should be available in a second line to avoid losing this information.
The selected location should be displayed along with the pinned map in the UI.

sadaf895:
- closes: #3076
- closes #2944
- closes #3074

### DIFF
--- a/src/app/features/location/map-popup/map-popup.component.ts
+++ b/src/app/features/location/map-popup/map-popup.component.ts
@@ -10,7 +10,7 @@ import { BehaviorSubject, firstValueFrom, Observable, Subject } from "rxjs";
 import { MapComponent } from "../map/map.component";
 import { AsyncPipe } from "@angular/common";
 import { MatButtonModule } from "@angular/material/button";
-import { GeoService, GeoResult } from "../geo.service";
+import { GeoResult, GeoService } from "../geo.service";
 import { AddressEditComponent } from "../address-edit/address-edit.component";
 import { ConfirmationDialogService } from "../../../core/common-components/confirmation-dialog/confirmation-dialog.service";
 import { LocationProperties } from "../map/map-properties-popup/map-properties-popup.component";
@@ -178,7 +178,7 @@ export class MapPopupComponent {
         },
       ],
     );
-    return result === "continue" || result === "update" ? result : undefined;
+    return result as "continue" | "update" | undefined;
   }
 
   private async handleConfirmationResult(


george-neha:
In the current design for the map, as a user i have to scroll up/down after i select a pin on the map, which gets very confusing. Would it be possible to spilt the screen left-right? with the form details on the left and the map on the right? @sadaf895 @sleidig

sleidig (approved):
Good work, @sadaf895 . The code is quite easy to understand with the well-named internal methods.

Seems to work really well, even with additional details in the search. Great tuning, @sadaf895 and @george-neha  :pray:

sleidig (approved):

Issue #2552, opened 2024-08-30
Import: lookup entity reference fields using multiple columns of an import


sleidig:
_When importing data that includes a link to existing records, I may need multiple columns to accurately match the right record to be linked in the import, e.g. using both firstname and lastname fields (if these are stored separately)._

**Objective**
We currently are able to import "entity" dataType fields by looking up one column from the import and selecting an existing entity matching that value. Extend this to use a combination of multiple fields that all (?) need to match, for more accurately linking.

---
Implementation Steps:
- [x] extend the `EntityDatatype.importMapFunction` to support this (and write unit tests for it first)
- [x] define an interface what "config" data is passed into the importMapFunction as "additional" (as this now has to contain more than just the ID of one matching field)
- [x] extend the `EntityImportConfigComponent` to offer a UI for configuring multiple fields

PR #3082, opened 2025-06-18
Import entity-refs with multiple columns as conditions

sleidig:
- closes: #2552

Allows to import an entity-ref (entity type field that links to an existing record) using multiple columns of the imported file as combined criteria.

see docs freshly extended for this here: https://chatwoot.help/hc/aam-digital/articles/1730371253-importing-data-from-a-spreadsheet-csv-file#linking-imported-records-with-individual-parent-entities

### DIFF
--- a/src/app/core/basic-datatypes/entity/entity-import-config/entity-import-config.component.html
+++ b/src/app/core/basic-datatypes/entity/entity-import-config/entity-import-config.component.html
@@ -2,19 +2,26 @@
   <h3 i18n="Label for property selection">
     Select matching {{ entity.label }} property
   </h3>
+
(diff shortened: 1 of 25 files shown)


Abhinegi2 (approved):
The functionality is working as expected! 👍 

I also followed the Chatwoot steps mentioned above and watched the video (https://youtu.be/IYyArbZCp6Y). 
However, video seems a bit outdated for example, we’ve already added features like smart selection on the configure field and a few other improvements. So we might consider updating these videos in the future.(it just a thought)

george-neha:
An updated video with our newly released features seems like a great idea. I can look into it. 😊

Issue #3096, opened 2025-06-26
Improve filters (especially "completed tasks") to automatically include archived


sleidig:
Some special filters like the Todo List view's "completed" tasks only make sense if archived records are included (Todo entities are automatically archived when the user completes them). Currently, the user sees an empty list when selecting that filter and has to additionally toggle the "include archived records" below the table to see results.

Some special logic should detect such filters and automatically activate the "include archived records" when such a filter option is selected, so to save time and avoid confusion.

![Image](https://github.com/user-attachments/assets/f07932db-db2a-4453-86b8-0a23e655e6ff)

PR #3099, opened 2025-06-26
Remove custom Todo components and use standard Entity code

### LINKED ISSUE #2511: Refactor TodoList and related filters + list so that this can use EntityList without custom component
Currently we have a custom `TodoListComponent` that adds some custom filter and overwrites that the todo details are opened as a special component in the popup. With a few more generalizations of the code we should be able to remove this custom component and instead just use EntityList here as well.

### LINKED ISSUE #2918: Todo Details cannot be navigated to (404 because they are always popups currently)
for Todo entities the entity block link (e.g. in Notification Event items) is broken:
This leads to a 404 page, because TodoDetailsComponent is designed to always show in a popup.

sleidig:
- closes #2918 
- closes #2511


src/app/core/entity-list/EntityListConfig.ts
```diff
@@ -100,10 +102,6 @@ export interface DateRangeFilterConfigOption {
   label: string;
 }
 
-export interface PrebuiltFilterConfig<T> extends BasicFilterConfig {
```
sleidig, line 103: this interface was copied twice in the file

src/app/features/todos/add-default-todo-views.ts
```diff
@@ -0,0 +1,155 @@
...
+      key: "completed",
+      label: $localize`:Filter-option for todos:Completed`,
+      filter: { isCompleted: true, isActive: { $exists: true } },
+    },
+    {
+      key: "open",
+      label: $localize`:Filter-option for todos:All Open`,
+      filter: { isCompleted: false },
+    },
+    { key: "", label: $localize`Any`, filter: {} },
```
Abhinegi2, line 151: Currently, when selecting the "Any" option, the selected value is not displayed as expected in the dropdown-

image

```suggestion
    { key: "any", label: $localize`Any`, filter: {} },
```

Abhinegi2 (changes requested):
After marking a task as completed, it is no longer visible when applying the "Task due: Completed" filter in the list view. (It's not showing anything.)

1. Mark any ToDo/task as **completed**.
2. Apply the **"Completed"** filter (list view).
3. Notice that no results are shown, even though completed tasks exist.

sleidig:
> After marking a task as completed, it is no longer visible when applying the "Task due: Completed" filter in the list view. (It's not showing anything.)
> 
> 1. Mark any ToDo/task as **completed**.
> 2. Apply the **"Completed"** filter (list view).
> 3. Notice that no results are shown, even though completed tasks exist.

Thanks for highlighting this. I had created a separate issue for better dealing with this special case (the problem is that we need to additionally "include archived records") --> https://github.com/Aam-Digital/ndb-core/issues/3096

But I have now included inactive by default in this list again, to make this less of a problem and get back the current behavior

PR #3104, opened 2025-06-26
refactor: Extract demo data generation into static functions

geigerzaehler:
- see #2995 

This change will allow us to generate demo data without depending on DI.

### DIFF
--- a/src/app/child-dev-project/children/demo-data-generators/demo-child-generator.service.ts
+++ b/src/app/child-dev-project/children/demo-data-generators/demo-child-generator.service.ts
@@ -10,6 +10,7 @@ import { genders } from "../model/genders";
 import { calculateAge } from "../../../utils/utils";
 import { DateWithAge } from "../../../core/basic-datatypes/date-with-age/dateWithAge";
 import { createEntityOfType } from "../../../core/demo-data/create-entity-of-type";
+import { range } from "lodash-es";
 
 export class DemoChildConfig {
   count: number;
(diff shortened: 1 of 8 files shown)


src/app/child-dev-project/children/demo-data-generators/demo-child-generator.service.ts
```diff
@@ -31,36 +32,48 @@ export class DemoChildGenerator extends DemoDataGenerator<Entity> {
...
+function generateChildren(params: { count: number }): Entity[] {
+  return range(params.count).map((i) => generateChild(String(i + 1)));
+}
 
-    if (faker.number.int(100) > 90) {
-      DemoChildGenerator.makeChildDropout(child);
-    }
-    return child;
-  }
+/** @deprecated */
```
sleidig, line 48:
```suggestion
/** @deprecated */
```
what should be used instead or why should it not be used? Let's add this to the "deprecated" comment

sleidig (approved):
Looks good to me :+1:

src/app/child-dev-project/children/demo-data-generators/demo-child-generator.service.ts
geigerzaehler: Addressed in https://github.com/Aam-Digital/ndb-core/pull/3104/commits/c13f7e7274ee2b6ae83dc1b49461d11e0d58d84f

Issue #3078, opened 2025-06-17
use date-range picker for reports view


sleidig:
Our Reports view (https://test.aam-digital.net/report) let's users select a date range for many reports. We currently use the native Angular Material component for this. With our date-range-filter component we have a more advanced version of this field that allows us to define a few pre-defined ranges that users can easily select. We should use this for the reporting period selection as well.

- [ ] use our custom date-range-filter component (or a sub-component of it) in the Report component
- [ ] offer the following ranges as default:
   - current month
   - last month
   - current quarter
   - last quarter
   - current year
   - last year
- [ ] allow the view:report config to overwrite which default ranges are shown to users

PR #3105, opened 2025-06-26
feat : allows few pre-defined date ranges to report view

sadaf895:
closes: #3078

### DIFF
--- a/src/app/core/basic-datatypes/date/date-range-filter/date-range-filter-panel/date-range-filter-panel.component.ts
+++ b/src/app/core/basic-datatypes/date/date-range-filter/date-range-filter-panel/date-range-filter-panel.component.ts
@@ -30,17 +30,38 @@ export const defaultDateFilters: DateRangeFilterConfigOption[] = [
   },
   {
     startOffsets: [{ amount: -1, unit: "weeks" }],
-    label: $localize`:Filter label:Since last week`,
+    endOffsets: [{ amount: -1, unit: "weeks" }],
+    label: $localize`:Filter label:Last week`,
   },
   {
     startOffsets: [{ amount: 0, unit: "months" }],
     endOffsets: [{ amount: 0, unit: "months" }],
-    label: $localize`:Filter label:This month`,
+    label: $localize`:Filter label:Current month`,
   },
   {
     startOffsets: [{ amount: -1, unit: "months" }],
     endOffsets: [{ amount: -1, unit: "months" }],
     label: $localize`:Filter label:Last month`,
+  },
+  {
+    startOffsets: [{ amount: 0, unit: "quarters" }],
+    endOffsets: [{ amount: 0, unit: "quarters" }],
+    label: $localize`:Filter label:Current quarter`,
+  },
+  {
+    startOffsets: [{ amount: -1, unit: "quarters" }],
+    endOffsets: [{ amount: -1, unit: "quarters" }],
+    label: $localize`:Filter label:Last quarter`,
+  },
+  {
+    startOffsets: [{ amount: 0, unit: "years" }],
+    endOffsets: [{ amount: 0, unit: "years" }],
+    label: $localize`:Filter label:Current year`,
+  },
+  {
+    startOffsets: [{ amount: -1, unit: "years" }],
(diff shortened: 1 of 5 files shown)


src/app/core/basic-datatypes/date/date-range-filter/date-range-filter-panel/date-range-filter-panel.component.ts
```diff
@@ -30,18 +30,39 @@ export const defaultDateFilters: DateRangeFilterConfigOption[] = [
   },
   {
     startOffsets: [{ amount: -1, unit: "weeks" }],
```
sleidig, line 32: The default options here have been designed for filtering notes. If we add quarters and others, that list is becoming very long and overwhelming. Instead, I would like to define different default ranges when we use this for the report view

i.e. probably in the `SelectReportComponent` we can create a different object of defaultReportDateFilters, similar to this one but with different ranges (as listed in the issue) and pass that in its template there

sadaf895:
@sleidig Please have a look if it will be good to go then I'll undo the changes for the config file if required.

src/app/core/basic-datatypes/date/date-range-filter/date-range-filter.component.ts
```diff
@@ -20,6 +27,12 @@ export class DateRangeFilterComponent<T extends Entity> implements OnChanges {
 
   @Input() filterConfig: DateFilter<T>;
 
+  /**
+   * Emits an array with two elements, the selected from and to dates as string
+   */
+  @Output() selectedOptionChange = new EventEmitter<string[]>();
```
sleidig, line 33: - this doesn't seem to be used/emitted now (and does almost the same thing as dateRangeChange)?

src/app/features/reporting/reporting/reporting.component.ts
```diff
@@ -53,17 +55,35 @@ export class ReportingComponent {
   data: any[];
   exportableData: any;
 
+  dateRangeOptions: any[] = [];
```
sleidig, line 58:
```suggestion
  dateRangeOptions: DateRangeFilterConfigOption[] = [];
```

src/app/features/reporting/reporting/select-report/select-report.component.ts
```diff
@@ -76,9 +124,21 @@ export class SelectReportComponent implements OnChanges {
...
+      this.toDate = range.end;
+    }
+    this.dataChanged.emit();
+  }
+
+  onDateRangeChange(event: { from: Date; to: Date }) {
+    this.fromDate = event.from;
+    this.toDate = event.to;
     this.dataChanged.emit();
   }
```
sleidig, line 143: what's the difference between these two so that we need separate methods? Couldn't we just use the same method? Also, we are now listening to two outputs, probably one of the outputs and one of the methods here is enough?

sleidig (changes requested):

src/app/features/reporting/reporting/select-report/select-report.component.ts
sadaf895: @sleidig I'll looked into the changes tomorrow

src/app/core/basic-datatypes/date/date-range-filter/date-range-filter.component.ts
```diff
@@ -36,9 +44,7 @@ export class DateRangeFilterComponent<T extends Entity> implements OnChanges {
     ) {
       this.fromDate = range.start;
       this.toDate = range.end;
-      this.filterConfig.selectedOptionChange.emit(
-        this.filterConfig.selectedOptionValues,
-      );
```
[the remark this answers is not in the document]
sleidig: Despite you marking this as resolved, I don't see changes to the code and still have the functional error. Did you commit and push the changes, @sadaf895 

1. test Notes List, selecting "Today" as date filter. Currently nothing happens
2. Revert back the removed calls to `this.filterConfig.selectedOptionChange.emit` here (keep them in addition to the new `dateRangeChange` output)

sleidig (approved):
Good work, @sadaf895 . Thanks!

PR #3107, opened 2025-06-26
refactor: Remove that doesn’t run meaningful assertions

geigerzaehler:
The removed test should check that dashboard includes widgets with attendance summary. When you run the tests, you can see that no attendance summary is visible because of the default fixtures don’t include the necessary records. The tests never catch this because of their dynamic nature. (`daysCount` is 0)

There are other issues with the tests (locators) and we don’t have a good way to generate the necessary fixtures yet (see #2995). So instead of replacing the test I opted to remove it for now.


geigerzaehler:
Ignore this. The tests don’t run any meaningful assertions but my analysis was wrong: The problem is that we don’t way to the widgets to load there content. Still this is caused by the dynamic nature of the tests. I’m going to replace this with screenshot tests.

PR #3108, opened 2025-06-26
fix: Stable sorting for groups in Entity Count dashboard widget

### LINKED ISSUE #1068: Children Count Dashboard widget should keep to order of enum
At the moment, the order of the statistics groups (e.g. by center in this screenshot) are order seemingly random: Which center is listed first depends on the center assigned to the first participant in the database. To make this logical and configurable, the order should rather depend on the order of the configurabe-enum items (if that property is an enum and not just a string).

image

geigerzaehler:
Fixes #1068

### DIFF
--- a/src/app/features/dashboard-widgets/entity-count-dashboard-widget/entity-count-dashboard/entity-count-dashboard.component.ts
+++ b/src/app/features/dashboard-widgets/entity-count-dashboard-widget/entity-count-dashboard/entity-count-dashboard.component.ts
@@ -1,5 +1,6 @@
 import { Component, Input, OnInit } from "@angular/core";
 import { Router } from "@angular/router";
+import { sortBy } from "lodash-es";
 
 import { DynamicComponent } from "../../../../core/config/dynamic-components/dynamic-component.decorator";
 import { EntityMapperService } from "../../../../core/entity/entity-mapper/entity-mapper.service";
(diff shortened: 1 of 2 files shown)


src/app/features/dashboard-widgets/entity-count-dashboard-widget/entity-count-dashboard/entity-count-dashboard.component.ts
```diff
@@ -161,19 +162,22 @@ export class EntityCountDashboardComponent
...
+            ? groupByType.additional
+            : undefined;
+        return {
+          label: label,
+          value: entities.length,
+          id: group?.["id"] || label,
+          groupedByEntity: groupedByEntity,
+        };
+      }),
+      (group) => group.id,
```
sleidig, line 179: My expectation for this, functionality wise, would be:
1. I edit the configurable-enum options (by editing a Child's details and using the tool icon of the "center" dropdown field), re-ordering the items in the dropdown edit dialog to make "Alipore" the last item in the dropdown list
2. I then go to the dashboard and see the items as disaggregations in that order, i.e. "Alipore" is listed last.

That's not the case yet.

I am not sure how the order of dropdown items is managed, but they support drag&drop when editing, so that component probably can give some context

geigerzaehler: Ah, I understand. I tried to implement this, PTAL.

PR #3113, opened 2025-07-01
Extract drag and drop from the admin-entity-list into a reusable component

Abhinegi2:
- Extract the drag-and-drop logic from the AdminEntityList component into its own (new) component for reusability.

### DIFF
--- a/src/app/core/admin/admin-entity-details/admin-entity-panel-component/admin-entity-panel-component.component.ts
+++ b/src/app/core/admin/admin-entity-details/admin-entity-panel-component/admin-entity-panel-component.component.ts
@@ -57,7 +57,6 @@ export class AdminEntityPanelComponentComponent implements OnInit {
   }
 
   private initializeFields(): void {
-    console.log(this.config);
     if (!this.targetEntityType) return;
     const targetEntitySchemaFields = Array.from(
       this.targetEntityType.schema.keys(),
--- a/src/app/core/config/config.service.ts
+++ b/src/app/core/config/config.service.ts
@@ -479,7 +479,6 @@ const migrateChildSchoolOverviewComponent: ConfigMigration = (
     deprecatedComponents.includes(configPart.component)
   ) {
     return {
-      title: "School History",
       component: "RelatedEntities",
       config: {
         entityType: "ChildSchoolRelation",
--- a/src/app/core/entity-details/related-entity-config.ts
+++ b/src/app/core/entity-details/related-entity-config.ts
@@ -23,4 +23,10 @@ export interface RelatedEntityConfig {
    * (Optional) A custom loader method to fetch related entities.
    */
   loaderMethod?: string;
+
+  /**
+   * If set to `true`, the component will assume that there can be only one active related entity at a time
+   * (e.g., only one currently attended school per child).
+   */
+  single?: boolean;
 }
--- a/src/app/utils/related-entities-default-config.ts
+++ b/src/app/utils/related-entities-default-config.ts
@@ -8,7 +8,7 @@ import { RelatedEntityConfig } from "../core/entity-details/related-entity-confi
 
 export const RELATED_ENTITIES_DEFAULT_CONFIGS: Record<
   string,
-  { entityType: string; columns?: FormFieldConfig[]; single?: boolean }
+  { entityType: string; columns?: FormFieldConfig[] }
 > = {
   NotesRelatedToEntity: {
     entityType: "Note",
@@ -33,10 +33,6 @@ export const RELATED_ENTITIES_DEFAULT_CONFIGS: Record<
       { id: "completed", hideFromForm: true },
     ],
   },
-  ChildSchoolOverview: {
-    entityType: "ChildSchoolRelation",
-    single: true,
-  },
 };
 
 export const RELATED_ENTITY_OVERRIDES: Record<string, RelatedEntityConfig> = {
@@ -60,5 +56,10 @@ export const RELATED_ENTITY_OVERRIDES: Record<string, RelatedEntityConfig> = {
   RecurringActivity: {
     entityType: "RecurringActivity",
     component: "RelatedEntities",
+  },
+  ChildSchoolOverview: {
+    entityType: "ChildSchoolRelation",
+    component: "RelatedEntities",
+    single: true,
   },
 };


src/app/core/admin/admin-entity-details/admin-entity-panel-component/admin-entity-panel-component.component.html
```diff
@@ -8,15 +8,33 @@
...
+      <ng-template #fieldTemplate let-field let-remove="remove">
+        <div class="drop-item column-preview flex-row gap-small">
+          <fa-icon icon="grip-vertical" class="drag-handle"></fa-icon>
+          <app-entity-field-label
+            [field]="field"
+            [entityType]="targetEntityType"
+          ></app-entity-field-label>
+          <fa-icon icon="times" (click)="remove(field)"></fa-icon>
+        </div>
+      </ng-template>
```
Abhinegi2, line 28: Maybe we can move this logic part in the ReorderableListComponent to avoid repeating

src/app/core/admin/admin-entity-details/admin-entity-panel-component/admin-entity-panel-component.component.scss
```diff
@@ -1,4 +1,27 @@
...
+}
+
+.remove-icon {
+  cursor: pointer;
+  margin: auto;
+}
+.drop-item:has(.remove-icon:hover) {
+  color: rgb(255, 0, 0);
+  background-color: rgba(255, 0, 0, 0.1);
+}
```
Abhinegi2, line 27: It's a duplicate!

src/app/core/admin/admin-entity-details/admin-entity-panel-component/admin-entity-panel-component.component.ts
```diff
@@ -5,12 +5,24 @@ import { EntityConstructor } from "../../../entity/model/entity";
...
+    EntityFieldsMenuComponent,
+    ReorderableListComponent,
+    EntityFieldLabelComponent,
+    FontAwesomeModule,
+  ],
   templateUrl: "./admin-entity-panel-component.component.html",
-  styleUrl: "./admin-entity-panel-component.component.scss",
+  styleUrls: [
+    "./admin-entity-panel-component.component.scss",
+    "../../admin-entity-list/admin-entity-list.component.scss",
```
sleidig, line 24: for which styling do we need this?

src/app/utils/reorderable-list/reorderable-list.component.html
```diff
@@ -0,0 +1,19 @@
...
+>
+  <div class="flex-row gap-regular flex-wrap gap-large column">
+    @for (item of items; track item) {
+      <div cdkDrag class="drop-item">
+        <ng-container
+          *ngTemplateOutlet="
+            itemTemplate;
+            context: { $implicit: item, remove: remove.bind(this) }
+          "
+        ></ng-container>
```
sleidig, line 15: shall we make the template optional and build in a default template already? Or do/should these look very different for the two admin-components? (if not, we may even skip the option to input a template completely for now and only add that later if needed)

src/app/core/admin/admin-entity-list/admin-entity-list.component.html
```diff
@@ -61,31 +59,31 @@
...
+            <fa-icon
+              icon="times"
+              class="remove-icon"
+              (click)="remove(col)"
+            ></fa-icon>
+          </div>
+        </ng-template>
 
         <div class="table-content-preview">
           <span i18n>Drag & drop table column headers. Add columns:</span>
```
sleidig, line 89: I am wondering where we should cut the components :thinking: 

Is there a use case to only use the drag-and-drop-columns without an option to add columns? Or should we put the the entity-fields-menu inside that also, to have a rather large component that can administer columns completely with an [(columns)] input+output?

Abhinegi2: I tried this out and found a few issues:

In the AdminEntityList, we were using drag-and-drop for both filters and columns. However, filters are of type string[], while columns are of type ColumnConfig[]. Because of this difference, we need to handle drag-and-drop, as well as field addition/removal, separately for each.

also, I noticed that the current drag-and-drop implementation (in our new component) is not working correctly for columns in the list view.


just wanted to let you know

src/app/core/admin/admin-entity-details/admin-entity-panel-component/admin-entity-panel-component.component.html
```diff
@@ -6,25 +6,25 @@
...
-        [availableFields]="allFields"
-        (activeFieldsChange)="updateFields(activeFields)"
-      ></app-entity-fields-menu>
-    </div>
+    <app-reorderable-list
+      [items]="activeFields"
+      [availableItems]="allFields"
+      [entityType]="targetEntityType"
+      [orientation]="'mixed'"
+      [fieldLabel]="'Columns shown:'"
```
sleidig, line 25:
```suggestion
      orientation="mixed"
      fieldLabel="Columns shown:"
      i18n-fieldLabel
```

sleidig: can simplify this in the other html template also

src/app/utils/reorderable-list/reorderable-list.component.ts
```diff
@@ -0,0 +1,61 @@
...
+  ],
+  templateUrl: "./reorderable-list.component.html",
+  styleUrl: "./reorderable-list.component.scss",
+})
+export class ReorderableListComponent {
+  @Input() items: any[] = [];
+  @Input() availableItems: (string | ColumnConfig)[] = [];
+  @Input() entityType: EntityConstructor;
+  @Input() orientation: "vertical" | "horizontal" | "mixed" = "vertical";
+  @Input() dropLockAxis: "x" | "y" | null = null;
```
sleidig, line 37: looks like currently we are always using this with "mixed" orientation. Do we already see a use case where this will need to be changed?

Otherwise, let's remove them from the "public interface" (i.e. not make them `@Input()` to keep things simple for now.

src/app/utils/reorderable-list/reorderable-list.component.html
```diff
@@ -0,0 +1,61 @@
...
+              class="remove-icon"
+              (click)="remove(item)"
+            ></fa-icon>
+          </div>
+        }
+      </div>
+    }
+  </div>
+</div>
+
```
sleidig, line 50: I think this line (see mouse cursor) should be between filter and column editing and not within each (now the two modes to edit filters/columns look visually more separated within than towards each other)

image

sleidig (changes requested):
Seems to work well and code looks pretty good also :+1:

src/app/utils/reorderable-list/reorderable-list.component.html
Abhinegi2: Removed the bottom line from filters (as it wasn't present earlier for filters). The line will now only be visible for columns or the default template.

sleidig: oh, I guess this line was supposed to look like the bottom line of a "table" there :sweat_smile: 

Looks good now :+1:

sleidig (approved):
Looks very good now :+1:

PR #3121, opened 2025-07-02
fix(core): remove favicon from site-settings

### LINKED ISSUE #1995: PWA app install icons should reflect the configured favIcon
Design of the platform is now widely configurable (see #1949), including the favicon. However, to support a fully "white-labeled" platform, the app icon (after installing as PWA) and splash screens also need to reflect the custom favIcon.

This needs to reflect in both
- index.html
- manifest.webmanifest
- images need to be available without permissions?

### LINKED ISSUE #2254: SiteSettings favIcon loads before user is authorized
**Describe the bug**
If the SiteSettings entity is not having "public" read permissions (accessible without login), then loading the favIcon fails with a 401 error.

**To Reproduce**
Steps to reproduce the behavior (needs actual couchdb and permissions):
1. Upload a favicon through .../admin --> Site Settings
2. reload the page (the problem only occurs if the site settings are loaded cached from localStorage)
3. see in console and default favIcon

---
Probably we need to delay the loading of the favIcon until login is complete?

sleidig:
because this was unreliable.
Instead, assets folder can be overwritten on servers using docker volumes

- fixes #2254
- closes #1995


Abhinegi2 (approved):
Looks good!

PR #3125, opened 2025-07-02
feat(adminUI): auto-create menu item on new entity type creation

### LINKED ISSUE #3118: when creating a new entity type in Admin UI, also add menu item
We already allow to create a new entity type in the Admin UI through the `AdminEntityTypesComponent`. While the UI is not very polished yet, it serves as a useful tool for internal use and advanced users. When creating a new type, we currently add config for views and the entity config.

Also create a new menu item `{ "entityType": "MY_NEW_TYPE" }` (using the EntityMenuItem format) there to automate a very common step that admins need to take.

sadaf895:
- closes #3118


sadaf895:
@sleidig Can you please review the code once. I guess the requirement of this task is implemented if I am not wrong.

src/app/core/admin/admin-entity-types/admin-entity-types.component.ts
```diff
@@ -117,6 +117,20 @@ export class AdminEntityTypesComponent implements OnInit {
...
+      data.navigationMenu = { items: [] };
+    }
+    const menuItems = data.navigationMenu.items;
+
+    // Avoid duplicate menu items
+    if (!menuItems.some((item: any) => item.entityType === id)) {
+      menuItems.push({
+        entityType: id,
+      });
+    }
```
sleidig, line 132: maybe put this into its own private method, so that this function itself is not too long.

sleidig (approved):
Looks good and works as expected :+1:

src/app/core/admin/admin-entity-types/admin-entity-types.component.ts
sadaf895: OK , Thanks!

Issue #3115, opened 2025-07-01
menu remains hidden while syncing


sleidig:
When logging in again after a longer time, so that there are a lot of changes to sync, the system shows the dashboard but hides the menu (although an old, valid config doc is available) until the sync is completed. That is unnecessary and can block users for a long time when there is a lot of data to sync

PR #3127, opened 2025-07-04
menu remains hidden while syncing 

Abhinegi2:
closes: #3115

### DIFF
--- a/src/app/core/ui/ui/ui.component.ts
+++ b/src/app/core/ui/ui/ui.component.ts
@@ -50,6 +50,7 @@ import { BehaviorSubject } from "rxjs";
 import { LoginStateSubject } from "../../session/session-type";
 import { toSignal } from "@angular/core/rxjs-interop";
 import { LoginState } from "../../session/session-states/login-state.enum";
+import { ConfigService } from "../../config/config.service";
 
 /**
  * The main user interface component as root element for the app structure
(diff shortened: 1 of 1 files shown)


src/app/core/ui/ui/ui.component.ts
```diff
@@ -127,9 +129,13 @@ export class UiComponent {
...
     );
 
-    this.setupService
-      .waitForConfigReady(true)
-      .then((ready) => this.configReady$.next(ready));
+    if (this.configService.hasConfig()) {
+      this.configReady$.next(true);
+    } else {
+      this.setupService
+        .waitForConfigReady(true)
```
sleidig, line 136: I actually thought waitForConfigReady(false) would do this also, but that doesn't seem to hide the menu on initial init. Did you understand what our param in waitForConfigReady currently does? (let's add that to the JsDoc there)

Abhinegi2: I think what’s happening is this:
We’re using waitForConfigReady(false), which checks if either:
- Config is ready (i.e., configUpdates emits a non-undefined config), or
- Sync is completed (syncState === COMPLETED)
But configUpdates$ is based on: `configUpdates = this.entityUpdated.pipe(shareReplay(1))` 
If no config has been emitted yet, then configUpdates is undefined,
Maybe that's why, then in this case it's waiting till sync is COMPLETED

sleidig (approved):

Issue #119, opened 2018-06-01
user customizable Dashboard widgets


sleidig:
Allow users to reorder/remove/add widgets blocks on the Dashboard dynamically.

codingfabi:
maybe keep in mind to implement a Drag and Drop feature, so users can organize there widgets as they wand to

PR #3132, opened 2025-07-08
feat(Admin UI): customize Dashboard widgets

sadaf895:
closes: #119

-----
custom settings components for individual widgets:
- [x] Entity Count
- [x] Shortcuts 
- [x] Notes needing follow up
- [x] tasks due
- [x] children with / without recent reports
- [x] attendance week (absences this week)
- [x] Progress
   - probably doesn't need any settings, we can generate a unique id (--> uuid() function) when the component gets created and set it as config without asking the user (still need to do this when add widget functionality will be implement)
- [x] Birthdays

### DIFF
--- a/src/app/core/dashboard/dashboard-widget-registry.service.ts
+++ b/src/app/core/dashboard/dashboard-widget-registry.service.ts
@@ -13,8 +13,7 @@ export class DashboardWidgetRegistryService {
 
   /**
    * Register a new widget and its details
-   * @param widgetName The name of the widget
-   * @param settingsComponent The settings component associated with the widget
+   * @param widgetDefinition The widget to register
    */
   register(widgetDefinition: DashboardWidgetDefinition) {
     this.dashboardWidgets.push(widgetDefinition);
@@ -62,9 +61,11 @@ export interface DashboardWidgetDefinition {
    */
   settingsComponent: string;
 
-   * Some default configuration settings for the widget. This configuration is used
-   * as the initial settings when a widget is first added to a dashboard. Users can
-   * later override these settings with their own preferences.
+  /**
+   * Default configuration settings for the widget
+   * used as the initial settings when a widget is first added.
+   *
+   * Users can later override these settings with their own preferences.
    */
   defaultConfig: any;
 }


src/app/features/dashboard-widgets/shortcut-dashboard-widget/shortcut-dashboard-settings.component/shortcut-dashboard-settings.component.ts
```diff
@@ -20,42 +20,31 @@ import { AdminIconComponent } from "#src/app/admin-icon-input/admin-icon-input.c
...
-    console.log('Shortcuts input:', this.shortcuts);
-    console.log('Subtitle input:', this.subtitle);
-    console.log('Explanation input:', this.explanation);
-    
     this.localConfig = {
-      shortcuts: this.shortcuts ? [...this.shortcuts.map(s => ({ ...s }))] : [],
-      subtitle: this.subtitle || "Quick Actions",
-      explanation: this.explanation || "Shortcuts to quickly navigate to common actions"
+      shortcuts: this.shortcuts
+        ? [...this.shortcuts.map((s) => ({ ...s }))]
```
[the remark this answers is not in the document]
sadaf895: We need this mapping if we want to avoid mutating the original input array and its objects. If not, we can remove it.

src/app/core/admin/admin-dashboard/admin-dashboard.component.html
```diff
@@ -0,0 +1,109 @@
...
+  insights you need.
+</div>
+
+<div class="overall-container flex-row" [class.disabled]="isDisabled">
+  <!-- WIDGET PREVIEW AREA -->
+  <div
+    cdkDropList
+    [cdkDropListData]="dashboardConfig.widgets"
+    (cdkDropListDropped)="drop($event)"
+    cdkDropListOrientation="mixed"
```
sleidig, line 28: drag&drop of a widget between two rows of widgets suddenly creates a whole empty row --> needs optimization in the drag preview (unfortunately I can't take screenshots while dragging something ... 🙄 )

@sadaf895 , can you connect with @Abhinegi2 and optimize the drag&drop preview stylings together in a pair programming session? (please bring back the discussion if you should find it needs any larger restructuring, though)

src/app/core/admin/admin-entity-details/widget-component-select/widget-component-select.component.ts
```diff
@@ -29,17 +30,111 @@ import { MatTooltipModule } from "@angular/material/tooltip";
...
+        value: {
+          component: "AttendanceWeekDashboard",
+          config: { daysOffset: 0, periodLabel: "last week" },
+        },
+      },
+      {
+        label: $localize`Progress`,
+        value: {
+          component: "ProgressDashboard",
+          config: { dashboardConfigId: "1" },
```
sleidig, line 127: TODO: needs to generate a new ID, otherwise the progress states will always be exactly the same between multiple widgets of this type

[--> Pair Programming Session]

src/app/core/admin/admin-dashboard/admin-dashboard.component.html
sleidig: maybe we can even share/import the styling (e.g. dotted lines as containers) between the drag&drop layout in admin of entity-forms and here for consistency?

sleidig: for drag&drop of form fields, we also let the user drag it from anywhere in the field (although we also show the drag-handle icon). For consistency and easier use, I would suggest we allow the same for dashboard widgets (i.e. let users drag wherever they click on the widget)

src/app/core/admin/admin-dashboard/admin-dashboard.component.html
```diff
@@ -0,0 +1,109 @@
...
+  <div class="flex-row gap-small save-buttons">
+    <button mat-raised-button color="accent" (click)="save()" i18n>Save</button>
+    <button mat-stroked-button (click)="cancel()" i18n>Cancel</button>
+  </div>
+</div>
+
+<div class="hint-banner" i18n>
+  You can customize the widgets on your dashboard to highlight the information
+  that is most useful to you. You can personalize the dashboard to show the key
+  insights you need.
```
sleidig, line 19: I feel like the two sentences here say more or less the same? Suggesting a little adjustment:


```suggestion
  Drag and drop the widgets to reorder them. You can configure details of each widget by clicking on them.
  Add additional widgets or remove widgets to fully customize the dashboard to your needs.
  The dashboard layout is shared by all users, your changes will be visible to everyone.
```

src/app/core/admin/admin-dashboard/admin-dashboard.component.html
```diff
@@ -0,0 +1,109 @@
...
+        </mat-card>
+      </div>
+    }
+
+    <!-- DROP AREA: ADD NEW WIDGET -->
+    <mat-card class="widget-preview-card drop-area-hint">
+      <mat-card-content class="add-widget-content">
+        <span i18n="Header of drag and drop box">
+          click here to add new widget
+        </span>
```
sleidig, line 87:
```suggestion
```

strictly speaking you have to click on the button and not exactly here 😉 
I'd remove it to avoid any confusion

src/app/core/admin/admin-dashboard/admin-dashboard.component.html
```diff
@@ -0,0 +1,109 @@
...
+        <fa-icon
+          icon="question-circle"
+          matTooltip="Add widgets to customize your dashboard. You can drag and drop to reorder them."
+          i18n-matTooltip
+        ></fa-icon>
```
sleidig, line 92:
```suggestion
```

I have added a very similar explanation to the top (where it'll be more easily visible). So I'd remove that here

src/app/core/admin/admin-dashboard/admin-dashboard.component.ts
```diff
@@ -0,0 +1,209 @@
...
+  templateUrl: "./admin-dashboard.component.html",
+  styleUrls: [
+    "./admin-dashboard.component.scss",
+    "../../dashboard/dashboard/dashboard.component.scss",
+    "../admin-entity/admin-entity-styles.scss",
+  ],
+})
+export class AdminDashboardComponent implements OnInit {
+  @Input() dashboardViewId: string;
+  @Input() isDisabled: boolean = false;
```
sleidig, line 52: what does this do? Do we ever allow people to access the admin view but then disable it?

src/app/core/admin/admin-dashboard/admin-dashboard.component.ts
```diff
@@ -0,0 +1,209 @@
...
+  dashboardConfig: DashboardConfig;
+  private originalDashboardConfig: DashboardConfig;
+
+  private readonly configService = inject(ConfigService);
+  private readonly dialog = inject(MatDialog);
+  private readonly location = inject(Location);
+  private readonly snackBar = inject(MatSnackBar);
+
+  ngOnInit() {
+    this.loadDashboardViewConfig();
```
sleidig, line 63: if you move the code of the implementation for `loadDashboardViewConfig` immediately after this method, it is slightly easier to find & read for someone trying to understand the class

src/app/core/admin/admin-dashboard/admin-dashboard.component.ts
```diff
@@ -0,0 +1,209 @@
...
+        event.previousIndex,
+        event.currentIndex,
+      );
+    } else {
+      // Handle drop from available widgets
+      const widgetType = event.previousContainer.data[event.previousIndex];
+      const newWidget: DynamicComponentConfig = {
+        component: widgetType.component,
+        config: { ...widgetType.defaultConfig },
+      };
```
sleidig, line 79: when does this case happen? I think we only have one simple drag&drop between existing widgets in our dashboard admin view

src/app/core/admin/admin-dashboard/admin-dashboard.component.ts
```diff
@@ -0,0 +1,209 @@
...
+      widgetConfig,
+      settingsComponent,
+    );
+    if (updatedConfig) {
+      this.dashboardConfig.widgets[idx] = updatedConfig;
+      this.dashboardConfig.widgets = [...this.dashboardConfig.widgets];
+    }
+  }
+
+  getSettingsComponentforWidget(widgetConfig: DynamicComponentConfig): any {
```
sleidig, line 99: we can try to design a more generic way to register dashboard components, so that this hard-coded list is not necessary 

[--> Pair Programming Session]

src/app/core/admin/admin-dashboard/admin-dashboard.component.ts
```diff
@@ -0,0 +1,209 @@
...
+        return "BirthdayDashboardSettings";
+
+      default:
+        return null;
+    }
+  }
+
+  private prettifyWidgetTitle(componentName: string): string {
+    return componentName.replace(/([a-z])([A-Z])/g, "$1 $2");
+  }
```
sleidig, line 132: I think this can be handled in `AdminWidgetDialogComponent` if we don't need the title anywhere else here

src/app/core/admin/admin-dashboard/admin-dashboard.component.ts
```diff
@@ -0,0 +1,209 @@
...
+    const selectedWidget = await firstValueFrom(dialogRef.afterClosed());
+    if (selectedWidget) {
+      this.dashboardConfig.widgets.push(selectedWidget);
+    }
+  }
+
+  cancel() {
+    this.dashboardConfig = JSON.parse(
+      JSON.stringify(this.originalDashboardConfig),
+    );
```
sleidig, line 172: can you test if we need this? I think we navigate away and thereby discard the whole component instance, so it should not make a difference if we update a class property (`this.dashboardConfig`) or not.

(As we don't save the config to the DB, changes should automatically be lost anyway)

sleidig: If we can remove it here, I think we don't need `this.originalDashboardConfig` at all and can delete that property completely.

src/app/core/admin/admin-entity-details/widget-component-select/widget-component-select.component.ts
```diff
@@ -29,17 +30,111 @@ import { MatTooltipModule } from "@angular/material/tooltip";
...
+          config: { daysOffset: 7, periodLabel: "this week" },
+        },
+      },
+      {
+        label: $localize`Attendance (Last Week)`,
+        value: {
+          component: "AttendanceWeekDashboard",
+          config: { daysOffset: 0, periodLabel: "last week" },
+        },
+      },
```
sleidig, line 122: This is the same widget with different settings. We can offer it only once here and let the user configure the widget as needed later.

```suggestion
      {
        label: $localize`Attendance (recent absences)`,
        value: {
          component: "AttendanceWeekDashboard",
          config: { daysOffset: 7, periodLabel: "this week" },
        },
      },
```

src/app/core/admin/admin-entity-details/widget-component-select/widget-component-select.component.ts
```diff
@@ -89,3 +184,9 @@ export interface WidgetOption {
    */
   disabled?: string;
 }
+
+export interface DashboardWidgetOption {
+  label: string;
+  value: DynamicComponentConfig;
+  disabled?: string;
+}
```
sleidig, line 192: I think this is structurally exactly the same as `WidgetOption`? Let's use the existing interface then

sleidig (changes requested):
notes for finishing touches for the UI (let's do some of these together in a pair programming session):
- [ ] background color of admin view (title is currently colored but rest is white
- [ ] padding for drag handles and "x" buttons
- [ ] possibly make the actual widgets less opaque, so that the UI elements for editing become more easily visible
- [x] disable any actions of widgets in admin view (e.g. prevent navigation to a shortcut item when clicking on it in admin mode)


I really like the tooltips on the "x" and drag-handle elements and the overall clarity of the UI 😃

sleidig (approved):
Looks good now, I think :-)

src/app/features/dashboard-widgets/birthday-dashboard-widget/birthday-dashboard-settings.component/birthday-dashboard-settings.component.ts
```diff
@@ -0,0 +1,129 @@
...
+    this.updateLocalConfig();
+    this.emitConfigChange();
+  }
+
+  updateLocalConfig() {
+    this.localConfig = {
+      threshold: this.formControl.value?.threshold ?? 32,
+      entities: {
+        [this.selectedEntityType]: this.selectedProperty,
+      },
```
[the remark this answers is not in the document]
sleidig: moved to separate issue [#3257](https://github.com/Aam-Digital/ndb-core/issues/3257)

Issue #3101, opened 2025-06-26
Admin UI: Option to add new related-entities component


Abhinegi2:
Currently, the Details View only supports adding new sections for default field groups. There is no option to add related entities as sections within the details view.

<img width="1507" alt="Image" src="https://github.com/user-attachments/assets/bc7749c1-accf-4177-ad99-da65de1a273b" />

Proposed solution:
Extend the Details View so that when a user clicks "Add Section", they are presented with two options:

1. Add a standard (default) section( current functionality).
2. Add a Related Entity section( new functionality).

To support this, clicking "Add Section" should open a dialog for the user to choose between these two options. Based on the selection, either the existing flow or a new UI for configuring related entities should be shown.

PR #3134, opened 2025-07-08
Admin UI: Option to add new related-entities component 

Abhinegi2:
closes: #3101


src/app/core/admin/admin-entity-details/admin-entity-details/admin-entity-details.component.ts
```diff
@@ -44,15 +46,38 @@ export class AdminEntityDetailsComponent {
...
+          });
+        } else if (selectedSection === "related-form") {
+          panel.components.push({
+            title: $localize`:Default title:New Related Section`,
+            component: "RelatedEntities",
+            config: {
+              entityType: this.entityConstructor.ENTITY_TYPE, //we can not use empty string
+            },
+          });
+        }
```
sleidig, line 80: should we move this part into the dialog component, so that it returns a fully usable config object? If we get more and more different components, that may be more scalable

src/app/core/admin/admin-entity-details/admin-entity-details/admin-entity-details.component.ts
```diff
@@ -44,15 +46,38 @@ export class AdminEntityDetailsComponent {
...
+            title: $localize`:Default title:New Section`,
+            component: "Form", // TODO: make this configurable
+            config: { fieldGroups: [] },
+          });
+        } else if (selectedSection === "related-form") {
+          panel.components.push({
+            title: $localize`:Default title:New Related Section`,
+            component: "RelatedEntities",
+            config: {
+              entityType: this.entityConstructor.ENTITY_TYPE, //we can not use empty string
```
sleidig, line 77: This has a bit weird behavior, especially when the entity-type doesn't have any valid other types to be listed as related entities. In that case it would probably be best to disable the option for users in the dialog already. And otherwise we could pre-select the first valid related type rather than itself.

src/app/core/admin/admin-entity-details/admin-entity-details/admin-entity-details.component.ts
```diff
@@ -44,15 +46,38 @@ export class AdminEntityDetailsComponent {
...
-      title: $localize`:Default title:New Section`,
-      component: "Form", // TODO: make this configurable
-      config: { fieldGroups: [] },
-    });
+    this.dialog
+      .open(EntityComponentSelectComponent, {
+        height: "20vh",
+      })
+      .afterClosed()
+      .subscribe((selectedSection) => {
```
sleidig, line 61: Seeing the code and testing this a bit the following idea for a more scalable/extendible architecture comes to my mind (let me know what you think, @Abhinegi2 ):

- here, when opening the dialog, pass along some context that the dialog can use to figure out which options to offer, maybe pass in even the whole view config (EntityDetailsConfig here). That would give the dialog component the entityType as well as any other sections that are already there to give a very smart suggestion for new components (or for now, simply use the entityType of the EntityDetailsConfig to filter if any RelatedEntities exist for that type)
   - or maybe, to start simple, we just pass in the parent entity type for now
- Return a whole config object from the dialog (maybe with some stuff pre-selected, like the best guess for the related-entity-type). That way, as the number of different components that can be added grows, we can delegate the logic to create a good default/empty config for itself to the components
   - or as an alternative: Design the Admin components (e.g. AdminRelatedEntitiesComponent - or what's its actual name...?) so that they work without receiving any existing config object and inside itself automatically set some default config where necessary. This may be even more elegant, then the selection dialog doesn't need to worry about that

src/app/core/admin/admin-entity-details/entity-component-select-component/entity-component-select-component.ts
```diff
@@ -0,0 +1,31 @@
...
+    MatFormFieldModule,
+    MatSelectModule,
+    DialogCloseComponent,
+  ],
+  templateUrl: "./entity-component-select-component.html",
+  styleUrl: "./entity-component-select-component.scss",
+})
+export class EntityComponentSelectComponent {
+  options = [
+    { value: "default-form", label: $localize`Default Section` },
```
sleidig, line 20:
```suggestion
    { value: "entity-form", label: $localize`Form Fields Section` },
```

sleidig (commented):
Works well already. A few further thoughts, looking ahead to make this the basis of more complex workflows

sleidig (commented):
Looks good to me :-)

src/app/core/admin/admin-entity-details/entity-component-select-component/entity-component-select-component.ts
```diff
@@ -0,0 +1,70 @@
...
+        },
+      },
+      {
+        label: $localize`Related-Entity Section`,
+        value: {
+          title: $localize`:Default title:New Related Section`,
+          component: "RelatedEntities",
+          config: {},
+        },
+      },
```
sleidig, line 51: if no getEntityTypesReferencingType exists, this option should be disabled with a tooltip to the user explaining "Add a field with type that links to this record to enable this section."

sleidig (changes requested):

src/app/core/admin/admin-entity-details/entity-component-select-component/entity-component-select-component.html
```diff
@@ -0,0 +1,20 @@
+<h2 mat-dialog-title i18n>Select Section Type</h2>
+<app-dialog-close mat-dialog-close></app-dialog-close>
+
+<mat-dialog-content>
+  <p i18n>
+    Choose which kind of section to add: a Default Section” for editing fields,
+    or a “Related-Entity Section” for displaying linked records.
```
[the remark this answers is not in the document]
george-neha: I agree that widget makes sense, since we are finally using the sub-components to help users customise the application to their needs. 
Since we are also looking at customising the dashboard widget, do you think maybe specifying the widget like "Dashboard Widget"/"Admin Widget" be helpful?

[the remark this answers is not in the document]
sleidig: :+1:  thanks Neha.

I think because this screen is only shown when clicking a button inside either the Details or Dashboard admin views, out of context it is pretty clear what widget is added. So we can probably keep implementation simple and just talk about "widget" generically.

sleidig (approved):
Looks good to me!

PR #3137, opened 2025-07-09
 Sometimes Admin (Edit Field?) has undefined variables #3136

Abhinegi2:
closes: #3136


sleidig (approved):
Thanks a lot for identifying the cause and working out a fix, @Abhinegi2 !

I have adjusted it a bit further because I observed that the field type we showed for selecting a static default value was only changing after saving the dialog

PR #3155, opened 2025-07-16
refactor: update entity handling to use ColumnConfig type across comp…

Abhinegi2:
- [x] Expend EntityFieldMenu component to emit full column config instead of just a string.
- [x] Added some additional columns in the available columns dropdown.
- [x] Updated Admin list and Matching view to work around the ColumnConfig instead of just string


src/app/core/admin/admin-matching-entities/admin-matching-entities.component.ts
```diff
@@ -184,8 +181,11 @@ export class AdminMatchingEntitiesComponent implements OnInit {
     };
 
     this.configService.saveConfig(fullConfig).then(() => {
+      console.log(
+        fullConfig["appConfig:matching-entities"],
+        "updated matching config",
+      );
```
Abhinegi2, line 187:
```suggestion
```

src/app/core/common-components/entity-fields-menu/entity-fields-menu.component.spec.ts
```diff
@@ -29,4 +30,25 @@ describe("EntityFieldsMenuComponent", () => {
...
+        label: "Custom Name",
+      },
+      {
+        id: "age",
+        label: "Age",
+      },
+    ];
+    expect(component._availableFields.length).toBe(2);
+
+    const name = component._availableFields.find((f) => f.id === "name")!;
```
sleidig, line 48: what does the `!` at the end of the statement do?

src/app/core/common-components/entity-fields-menu/entity-fields-menu.component.ts
```diff
@@ -28,38 +29,67 @@ import { EntityFieldSelectComponent } from "app/core/entity/entity-field-select/
...
+      });
+
+      this.activeFieldsChange.emit(mappedFields);
+    });
+  }
+
   ngOnChanges(changes: SimpleChanges): void {
     if (changes.activeFields) {
-      this.selectedFieldsControl.setValue(this.activeFields, {
+      const selectField = this.activeFields?.map((field) =>
```
sleidig, line 87: maybe better named `selectedFields`? (as this is an array)

src/app/core/admin/admin-list-manager/admin-list-manager.component.ts
```diff
@@ -36,41 +36,56 @@ import { MatSelectModule } from "@angular/material/select";
...
-  @Input() items: (string | ColumnConfig)[] = [];
+  @Input() items: ColumnConfig[] = [];
   @Input() entityType: EntityConstructor;
   @Input() fieldLabel: string;
   @Input() templateType: "default" | "filter" = "default";
-  @Input() activeFields: (string | FormFieldConfig)[] = [];
+  @Input() activeFields: ColumnConfig[] = [];
+  @Input() onlyIDs: boolean = false;
+  @Output() itemsChange = new EventEmitter<ColumnConfig[]>();
+  @Output() idsChanges = new EventEmitter<string[]>();
```
sleidig, line 46: if we have an output for itemsChange and one for idsChange (which seems a smart appraoch, actually), then I think the `onlyIDs` input as a toggle is not needed, is it?

src/app/core/common-components/entity-fields-menu/entity-fields-menu.component.spec.ts
Abhinegi2: Previously i was using the testEntity and trying to find some fields, and that time, sometimes coming back as undefined. So I used the non-null assertion (!) to avoid errors. 
But now, using/defining a custom availableFields array, I forgot to remove those. Let me remove those, as the fields will no longer be undefined or null.

sleidig: Does it really help in such cases? I have not used the `!` before but the StackOverflow sounded like it is just a TypeScript assertion (i.e. during static code checking not at runtime if something is undefined)?

Abhinegi2: It did not help me. I still got the error 

I also used this first time by going through this - https://www.geeksforgeeks.org/typescript/typescript-non-null-assertion-operator-postfix-type/

src/app/core/admin/admin-list-manager/admin-list-manager.component.ts
Abhinegi2: We need something to differentiate between what to emit, for eg, the full config or just the IDs, so in that way, it’s helpful

sleidig (commented):
code looks very good now.

functional testing:
- [x] the "_id" column can't be deleted from the columns drag&drop currently
- [x] "[name of record]" looks super stupid :see_no_evil:  ... how should we call this for users? Maybe just "Name (record preview)" or just "Name"?

sleidig (approved):
Looks all good now :+1: 

-----
... almost. The double scrollbars here make it very ugly to use. Feel free to look into it after merging, if you prefer:
image

PR #3156, opened 2025-07-17
Fix: automated updates also from popup forms

### LINKED ISSUE #2980: auto-update: make update-dialog trigger from RowDetails (Matching review dialog) also

sleidig:
- closes #2980


src/app/features/automated-status-update/automated-status-update-config-service.ts
```diff
@@ -134,20 +134,38 @@ export class AutomatedStatusUpdateConfigService {
...
 
     if (affectedEntities.length > 0) {
       await this.confirmAndSaveAffectedEntities(affectedEntities);
     }
   }
 
+  private getChangedFields(
+    newEntity: Entity,
+    originalEntity: Entity,
+  ): [string, any][] {
```
Abhinegi2, line 155:
```suggestion
  ): [string, any][] {
```
I was thinking we could use a more specific type, but I guess we’re using any because the value can be a string, an object, or maybe an address config?

Abhinegi2 (approved):
Tested on the dev, and it seemsworking fine for matching entities now (also thanks for detailed description),
Code also looks good and understandable.

PR #3165, opened 2025-07-23
chore: Use test-specific db fixtures for attendance test

geigerzaehler:
Use test-specific fixtures instead of demo data for one attendance tests.


geigerzaehler:
@sleidig Could you have another look at this? I believe the qlty checks are false positives

e2e/fixtures.ts
```diff
@@ -79,22 +99,47 @@ export async function loadApp(page: Page, entities?: Entity[]) {
...
-    new DateOnlyDatatype(),
-    new LongTextDatatype(),
-
-    // the following datatypes have dependencies through Angular DI - need workarounds in case they are needed for test data
-    //new ConfigurableEnumDatatype(null),
-    //new LocationDatatype(null),
-    //new EntityDatatype(null, null, null),
-    //new EventAttendanceMapDatatype(entitySchemaService),
-  ]);
+  const injector = Injector.create({
```
sleidig, line 102: ah, nice. This looks easier than I would've thought :slightly_smiling_face: 

Then we could also undo the split between EntitySchemaTransformer and EntitySchemaService that I previously added to disentangle the part requiring DI. I think having those as two classes is only confusing, if it is not absolutely necessary

sleidig (commented):
Looks good to me. Thanks for working this out (again).

Shall we undo the EntitySchemaService refactoring here still?

sleidig:
> @sleidig Could you have another look at this? I believe the qlty checks are false positives

sorry for the slow reaction. I was out for some days. Agreed, most of these look invalid. I have already deactivated the active PR comments from the tool for new PRs for now.

sleidig (approved):

Issue #3146, opened 2025-07-11
date range filter is still showing date text after clearing (all) filters


sleidig:
when using "clear all" filters button, the filtering is removed but the date range filter still shows the text of the previously selected dates

PR #3169, opened 2025-07-28
date range filter is still showing date text after clearing (all) filters 

Abhinegi2:
closes: [#3146](https://github.com/Aam-Digital/ndb-core/issues/3146)


george-neha:
In the Notes/Tasks entity, after I initially clear the filter, when I select a different date range and/or also select a different filter like category, the button "clear all filters is not available.

image
image

The clear all button is available only if I enter a value in the "Filter"field.

Abhinegi2:
@george-neha Thanks for highlighting this issue. I seem to have missed it during some cleanups, but it's now fixed. Could you please check it now?

sleidig (approved):

Issue #2422, opened 2024-06-19
replace custom groupBy with built-in JavaScript groupBy function


sleidig:
We have our own small utility function to group data: https://github.com/Aam-Digital/ndb-core/blob/27952f4f5cf0b9113ac6c86f6d9773877042390a/src/app/utils/utils.ts#L40-L65

A native groupBy is available now:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/groupBy

--> check if we can replace our custom code with this method and if so, remove our utils function.

PR #3175, opened 2025-07-29
refactor: Use native `Map.groupBy` where possible

geigerzaehler:
See #2422

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -27,7 +27,7 @@
     "module": "ES2022",
     "resolveJsonModule": true,
     "lib": [
-      "ES2022",
+      "ES2024",
       "dom"
     ],
     "types": [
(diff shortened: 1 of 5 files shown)


tsconfig.json
```diff
@@ -27,7 +27,7 @@
     "module": "ES2022",
     "resolveJsonModule": true,
     "lib": [
-      "ES2022",
+      "ES2024",
```
sleidig, line 30: do you know if there are / we need polyfills for any important backward compatibility on older browsers?

sleidig (approved):
:pray:

tsconfig.json
geigerzaehler: Yes, we do need polyfills for backwards compatibility. I don’t know what your browser policy is and how it is implemented. But I’m happy to investigate.

sleidig: I don't know either what our browser policy is :laughing: 

But we should support somewhat older browsers. We are catering to a diverse, often low-tech user base.
According to user analytics we had people using Chrome Mobile 119, Firefox 133, Edge 131 or Safari 16.0 this year.

sleidig: @geigerzaehler , do we still need to add the polyfill here before this can be merged?

sleidig: agreed. Let's postpone this for now. I'll close the PR to have a up-to-date overview on GitHub but let's keep the branch for reference.

Issue #3171, opened 2025-07-29
Public Form field group title field loses focus on every key press


sleidig:
**Describe the bug**

**To Reproduce**

1. "Configure Fields" of a Public Form
2. Type a group title (as shown below)
3. The field loses focus after every character and you need to click into the field again to continue typing 

<img width="1012" height="414" alt="Image" src="https://github.com/user-attachments/assets/63ee58d0-c902-41dd-b87b-0ed72aab119e" />

This bug doesn't seem to occur when editing the group titles in the Entity Details.

Abhinegi2:
@george-neha, could you test the fixes for the above-mentioned issue that have been deployed here- https://pr-3177.aam-digital.net/

PR #3177, opened 2025-07-30
fix: prevent unnecessary value change events in updateValue method

Abhinegi2:
closes: #3171

### DIFF
--- a/src/app/features/public-form/edit-public-form-columns/edit-public-form-columns.component.ts
+++ b/src/app/features/public-form/edit-public-form-columns/edit-public-form-columns.component.ts
@@ -44,7 +44,9 @@ export class EditPublicFormColumnsComponent
 
   updateValue(newConfig: FormConfig) {
     // setTimeout needed for change detection of disabling tabs
-    setTimeout(() => this.formControl.setValue(newConfig.fieldGroups));
+    setTimeout(() =>
+      this.formControl.setValue(newConfig.fieldGroups, { emitEvent: false }),
+    );
     this.formControl.markAsDirty();
   }
 }


src/app/features/public-form/edit-public-form-columns/edit-public-form-columns.component.ts
```diff
@@ -44,7 +44,9 @@ export class EditPublicFormColumnsComponent
 
   updateValue(newConfig: FormConfig) {
     // setTimeout needed for change detection of disabling tabs
-    setTimeout(() => this.formControl.setValue(newConfig.fieldGroups));
+    setTimeout(() =>
+      this.formControl.setValue(newConfig.fieldGroups, { emitEvent: false }),
```
sleidig, line 48: will this not make it impossible for entity forms or other parent components to subscribe to changes of this, if it should be necessary?

Abhinegi2: I debugged this again and found the issue: every time any value changes in the form (even typing a single letter in the title), we update this.formConfig, which is passed as an @Input to admin-entity-form. This re-triggers change detection, causing the input to re-render and focus being lost

src/app/features/public-form/edit-public-form-columns/edit-public-form-columns.component.ts
```diff
@@ -36,9 +36,15 @@ export class EditPublicFormColumnsComponent
       this.formConfig = {
         fieldGroups: publicFormConfig.columns,
       };
-      this.formControl.valueChanges.subscribe(
-        (v) => (this.formConfig = { fieldGroups: v }),
-      );
+      const originalFormConfig = JSON.parse(JSON.stringify(this.formConfig));
```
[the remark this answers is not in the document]
sleidig: @Abhinegi2 , but why would a page reload be necessary? Why is `EntityFormService.resetForm` triggered by `FormComponent`'s cancel button not working here?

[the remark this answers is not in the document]
Abhinegi2: In EditPublicFormColumnsComponent, we are using a custom formConfig to show the columns as 
```
      this.formConfig = {
        fieldGroups: publicFormConfig.columns,
      };
```
 When resetForm() is called, it only resets the form field values, not the custom configuration, so that's why it's working in other components where we are directly passing original fields/config

src/app/features/public-form/edit-public-form-columns/edit-public-form-columns.component.ts
```diff
@@ -36,9 +36,17 @@ export class EditPublicFormColumnsComponent
...
+      // Manually reset the config to reflect the cancelled changes immediately in the UI,
+      // otherwise a full page reload would be needed to see the cancelled changes.
```
sleidig, line 41:
```suggestion
      // also resetting the local config object only internal to this component
```

Issue #3026, opened 2025-05-23
save and display filename of import in the Import History


sleidig:
To make it easier to distinguish between multiple previous imports (and which one the user may want to undo), we could save the original filename and display it there:

![Image](https://github.com/user-attachments/assets/bccf5667-c6be-453c-9be8-0842708acd89)

1. save the filename in the `ImportMetadata` entity after the import
2. display the filename in the Import History entries to give users additional context information
3. if `additionalActions` are defined on the ImportMetadata entity, also display information about the record the imported data was linked to on the Import History entry

PR #3181, opened 2025-07-31
feat: display filename of import in Import History

sadaf895:
closes: #3026


sadaf895:
@sleidig Ready for review now

src/app/core/import/import-history/import-history.component.scss
```diff
@@ -1,3 +1,13 @@
 .no-results {
   font-style: italic;
 }
+
+.date-filename {
```
Abhinegi2, line 5:
```suggestion
.file-info {
```

src/app/core/import/import-history/import-history.component.scss
```diff
@@ -1,3 +1,13 @@
...
+  white-space: nowrap;
+  display: inline-block;
+}
+
+.filename {
+  margin-left: 0.5em;
```
Abhinegi2, line 11: You can use margin-left-small or something?

sleidig:
> 2. We can simplify the card by removing the date from the top and showing it like this (file icon is optional):
> 
> 3. Or if we still want to keep the date visible at the top

Good point. Maybe indeed the file name is more informative and we keep the date only for sorting and in the second row. I would skip file icon but maybe add the file extension (.csv) for clarification.

sleidig:
nice, looks good to me

src/app/core/import/import-history/import-history.component.html
```diff
@@ -5,7 +5,10 @@ <h3 i18n>Previous Imports</h3>
     <mat-expansion-panel [expanded]="i === 0">
       <mat-expansion-panel-header>
         <mat-panel-title>
-          {{ item.date | date }}
+          <span class="file-info">
+            <fa-icon [icon]="'file-alt'" class="margin-right-small"></fa-icon>
+            {{ item.filename || item.config?.filename }}
```
Abhinegi2, line 10: In what scenario would item.filename be empty, so that we are using item.config?.filename as fall back?

src/app/core/import/import-history/import-history.component.html
```diff
@@ -22,6 +25,24 @@ <h3 i18n>Previous Imports</h3>
         ></app-entity-block>
         at {{ item.date | date: "medium" }}
       </div>
+
+      @if (item.config.additionalActions?.length > 0) {
+        <div class="margin-top-regular">
+          <strong i18n>Additional Actions:</strong>
```
Abhinegi2, line 31:
```suggestion
          <span i18n>Additional Actions:</span>
```

Issue #3143, opened 2025-07-10
Switching the language in Site Settings is not changing the UI language


sleidig:
The app language is not easily changeable anymore.

**To Reproduce**

1. Go to Admin > Site Settings
2. Edit and change the "Default Language"
3. Save. The language is still the old one
4. Reload. The language is still the old one
5. Reset the app (from (i) icon bottom left). The language is still the old one

**Expected behavior**
The language should switch to the new one immediately after saving (?). If necessary, automatically reloading the app.

Abhinegi2:
> While after initial setup we usually have config in one language that cannot easily be switched, the app language should still be changeable, I think.

tried to replicate the issue and noticed that after switching the language to German, some parts were translated (see ss).

<img width="1789" height="1252" alt="Image" src="https://github.com/user-attachments/assets/38c6817a-750a-4ee5-91dd-fe31692e4ebf" />

@sleidig  Are we expecting the language switch to update the entire interface, similar to when the app is first set up (including the translated side menu and other sections)?

PR #3187, opened 2025-08-04
 Switching the language in Site Settings is not changing the UI language 

Abhinegi2:
closes: #3143

### DIFF
--- a/src/app/core/language/language-select/language-select.component.spec.ts
+++ b/src/app/core/language/language-select/language-select.component.spec.ts
@@ -46,7 +46,7 @@ describe("LanguageSelectComponent", () => {
     expect(component).toBeTruthy();
   });
 
-  it("should write selected language to local storage and reload page", () => {
+  it("should switch locale", () => {
     component.changeLocale("de");
     expect(mockLanguageService.switchLocale).toHaveBeenCalledWith("de");
   });
--- a/src/app/core/language/language-select/language-select.component.spec.ts
+++ b/src/app/core/language/language-select/language-select.component.spec.ts
@@ -2,8 +2,6 @@ import { ComponentFixture, TestBed } from "@angular/core/testing";
 
 import { LanguageSelectComponent } from "./language-select.component";
 import { RouterTestingModule } from "@angular/router/testing";
-import { LANGUAGE_LOCAL_STORAGE_KEY } from "../language-statics";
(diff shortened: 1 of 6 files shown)


Abhinegi2:
@sleidig  now using the EntityMapperService to update the language in localStorage. Review the current approach.

But now, the E2E tests for "Translated and localized app versions (i18n)" are currently failing.
i tried to fix those but did not find the issue

src/app/core/site-settings/site-settings.service.ts
```diff
@@ -58,6 +61,26 @@ export class SiteSettingsService extends LatestEntityLoader<SiteSettings> {
...
+        const updatedLanguage =
+          updatedSiteSettings?.entity?.defaultLanguage?.id;
+        const currentLanguage = localStorage.getItem(
+          LANGUAGE_LOCAL_STORAGE_KEY,
+        );
+        if (updatedLanguage !== currentLanguage) {
+          // Override the language in localStorage and reload the app
+          localStorage.setItem(LANGUAGE_LOCAL_STORAGE_KEY, updatedLanguage);
+          this.window.location.reload();
+        }
```
sleidig, line 84: could you extract this into a method in LanguageService and reuse the code for this as well as language-select component?

sleidig (commented):
fixed id check, now it seems to work well for me.

src/app/core/site-settings/site-settings.service.ts
Abhinegi2: I tried that, but it’s creating a circular dependency. If I add a new method in LanguageService and call it here, it causes a loop because LanguageService also uses SiteSettingsService. But I’ll check again to see if there’s a way to make it work.

src/app/core/language/language-select/language-select.component.spec.ts
```diff
@@ -44,17 +42,12 @@ describe("LanguageSelectComponent", () => {
...
 
-  afterEach(() => {
-    window.localStorage.removeItem(LANGUAGE_LOCAL_STORAGE_KEY);
-  });
-
   it("should create", () => {
     expect(component).toBeTruthy();
   });
 
   it("should write selected language to local storage and reload page", () => {
```
sleidig, line 49:
```suggestion
  it("should switch locale", () => {
```

Issue #2396, opened 2024-06-06
improve UX of configurable-enum fields


sleidig:
- on dashboard group by: display a label (e.g. "not defined") for the group without value instead of just having that empty
- on dashboard group by: display color of label (if configured)
- on edit field: display color (if configured), in the field for selected value and in the dropdown for all options.

PR #3197, opened 2025-08-05
enhancement: UX of configurable-enum fields

sadaf895:
closes: #2396

### DIFF
--- a/src/app/features/dashboard-widgets/entity-count-dashboard-widget/entity-count-dashboard/entity-count-dashboard.component.html
+++ b/src/app/features/dashboard-widgets/entity-count-dashboard-widget/entity-count-dashboard/entity-count-dashboard.component.html
@@ -67,7 +67,10 @@
       <ng-container matColumnDef="label">
         <td *matCellDef="let group">
           @if (!group.groupedByEntity) {
-            <span>{{ group.label }}</span>
+            <span
+              [ngClass]="{ 'not-defined-label': group.label === 'not defined' }"
+              >{{ group.label }}</span
+            >
           } @else {
             <app-entity-block [entityId]="group.id"></app-entity-block>
           }
--- a/src/app/features/dashboard-widgets/entity-count-dashboard-widget/entity-count-dashboard/entity-count-dashboard.component.ts
+++ b/src/app/features/dashboard-widgets/entity-count-dashboard-widget/entity-count-dashboard/entity-count-dashboard.component.ts
@@ -23,6 +23,7 @@ import { MatIconButton } from "@angular/material/button";
 import { EntityFieldLabelComponent } from "../../../../core/common-components/entity-field-label/entity-field-label.component";
 import { ConfigurableEnumValue } from "app/core/basic-datatypes/configurable-enum/configurable-enum.types";
 import { ConfigurableEnumService } from "app/core/basic-datatypes/configurable-enum/configurable-enum.service";
+import { CommonModule } from "@angular/common";
 
 /**
  * Configuration (stored in Config document in the DB) for the dashboard widget.
@@ -67,6 +68,7 @@ interface GroupCountRow {
     MatTooltipModule,
     MatIconButton,
     EntityFieldLabelComponent,
+    CommonModule,
   ],
 })
 export class EntityCountDashboardComponent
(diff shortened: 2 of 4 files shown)


sadaf895:
@sleidig I am mentioning the all improvements done in this PR.

1.  Display "not-defined" label for empty or undefined strings
2.  Display "invalid options" 
3.  On dashboard group by: display color of label (if configured)
4.  Make the color bubbles always the same width / fill the space (and those items that don't have a color are put into a transparent bubble so that all texts align)

src/app/features/dashboard-widgets/entity-count-dashboard-widget/entity-count-dashboard/entity-count-dashboard.component.html
```diff
@@ -67,7 +67,20 @@
       <ng-container matColumnDef="label">
         <td *matCellDef="let group">
           @if (!group.groupedByEntity) {
-            <span>{{ group.label }}</span>
+            @if (group.label === undefined && group.isInvalidOption) {
+              <span class="not-defined-label" style="color: #d9534f"
```
sleidig, line 71:
```suggestion
              <span class="not-defined-label" style="color: #d9534f" i18n
```

src/app/features/dashboard-widgets/entity-count-dashboard-widget/entity-count-dashboard/entity-count-dashboard.component.html
```diff
@@ -67,7 +67,20 @@
       <ng-container matColumnDef="label">
         <td *matCellDef="let group">
           @if (!group.groupedByEntity) {
-            <span>{{ group.label }}</span>
+            @if (group.label === undefined && group.isInvalidOption) {
```
sleidig, line 70: why is label undefined important here? isn't isInvalidOption enough as condition?

src/app/features/dashboard-widgets/entity-count-dashboard-widget/entity-count-dashboard/entity-count-dashboard.component.ts
```diff
@@ -160,27 +164,70 @@ export class EntityCountDashboardComponent
...
-      const groupCountSorted = enumValues
-        .map((enumValue) => groupCountsMap.get(enumValue.id))
+
+      // Mark and collect invalid options
+      const invalidOptions = groupCounts
+        .filter((g) => g.id && !validIds.has(g.id))
+        .map((g) => ({
+          ...g,
+          label: undefined,
+          isInvalidOption: true,
```
sleidig, line 208: now each different invalid option category is listed separately on the dashboard. Can you all combine them, so that there is only once a "[invalid option]" that summarizes them all?

image

src/app/features/dashboard-widgets/entity-count-dashboard-widget/entity-count-dashboard/entity-count-dashboard.component.ts
```diff
@@ -160,27 +164,70 @@ export class EntityCountDashboardComponent
     const groupedByEntity =
```
sleidig, line 164: The `calculateGroupCounts` method already was super long. Could you try to create private sub-methods - maybe one to get groupCounts of enums, one for groupCounts of others, one to handle the notDefined?

src/app/features/dashboard-widgets/entity-count-dashboard-widget/entity-count-dashboard/entity-count-dashboard.component.html
```diff
@@ -67,7 +67,20 @@
...
+              <span
+                class="colored"
+                [style.background-color]="group.color || 'transparent'"
+              >
+                {{ group.label }}
+              </span>
+            }
           } @else {
             <app-entity-block [entityId]="group.id"></app-entity-block>
           }
```
sleidig, line 86: maybe we can use DisplayConfigurableEnumComponent here to handle the coloring? Less duplicated code and ensured UI consistency.

sleidig (changes requested):

sleidig:
> @sleidig I have added the logic for displaying the color in dropdown options also in this PR only. I have created the different PR for this but I need to render the current code of DisplayConfigurableEnumComponent which is in this branch so that's why I have added the logic here only

That's okay, thanks for explaining it here. One alternative in such a case (if you have complex new things to add that should really be separate):
1. Create a new branch not from `master` but from this feature branch
2. for the new PR, you can select the feature branch as a "base" (instead of the default master)

image

Github then only shows your new changes from the additional branch. But when we merge the first feature, the PR automatically changes to base onto master. So we can seamlessly review two things separately that build onto each other

sleidig (approved):
Nice work! 🥇

Issue #3184, opened 2025-08-01
demo is missing "observations" enum


sleidig:
The demo setups are missing a enum for the "observations" data:

<img width="1813" height="728" alt="Image" src="https://github.com/user-attachments/assets/9b22c51d-2c74-48a8-8dbf-6d882740a40e" />

PR #3200, opened 2025-08-11
demo is missing "observations" enum 

Abhinegi2:
closes: #3184


sleidig (approved):
I have revisited this and implemented a more lightweight fix. Had you noticed the json files for the rating-answer enums that already existed?

PR #3214, opened 2025-08-13
Add Claude Code command for writing e2e tests

geigerzaehler:
Also add some basic project memory for Claude to make working with the
codebase easier.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,66 @@
+# Aam Digital - Project Context
+
+## Project Overview
+Aam Digital is a comprehensive case management software for social organizations, designed to improve effectiveness and transparency in work with beneficiaries. It's built with Angular 20 and uses TypeScript, Material Design, and various modern web technologies.
+
+## Architecture & Tech Stack
+- **Frontend**: Angular 20, TypeScript, Angular Material
+- **State Management**: RxJS, Entity system with PouchDB
+- **Authentication**: Keycloak integration
+- **Database**: PouchDB (CouchDB compatible) with offline-first approach
+- **Testing**: Jasmine, Karma, Playwright for E2E
+- **Documentation**: Compodoc for API docs
+- **Build & Deploy**: Angular CLI, Docker
+
+## Key Features
+- Entity-based data modeling system
+- Offline-first architecture with sync capabilities
+- Multi-language support (i18n) via POEditor
+- Flexible configuration system
+- Dashboard widgets and reporting
+- Import/export functionality
+- Permission-based access control
+- Demo data generation for testing
+
+## Project Structure
+- `src/app/core/` - Core system modules and services
+- `src/app/features/` - Feature-specific modules
+- `src/app/child-dev-project/` - Child development specific features
+- `e2e/` - End-to-end tests with Playwright
+- `doc/` - Documentation and API reference
+- `build/` - Build configuration and scripts
+
+## Development Guidelines
+- Uses ESLint for linting (`npm run lint`)
+- Prettier for code formatting (pre-commit hooks)
+- Comprehensive test suite with Karma and Playwright
+- TypeScript strict mode enabled
+- Material Design components preferred
+- Entity-based architecture for data models
+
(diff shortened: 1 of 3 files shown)


CLAUDE.md
```diff
@@ -0,0 +1,66 @@
+# Aam Digital - Project Context
```
sleidig, line 1: This is required for claude as context?

Can we put it into .claude subfolder? Or actually merge it with the public README for humans, as these details may be useful for them (and other AI models) as well?

doc/compodoc_sources/how-to-guides/end-to-end-tests.md
```diff
@@ -94,6 +94,18 @@ option](https://argos-ci.com/docs/playwright#argosscreenshotpage-name-options)
 in `argosScreenshot()`. Set a default threshold with `argos upload
 --threshold=...`
 
+## AI-Assisted Test Development
+
+You can streamline test development using Claude Code with the custom `/write-e2e-test` command, which provides comprehensive guidelines for writing Playwright tests.
```
sleidig, line 99: Could you link some external docs on how to install/use claude? Does it also work via GitHub Copilot, which me and some others in our team have set up already?

sleidig (commented):
:heart_eyes:

CLAUDE.md
geigerzaehler: > Can we put it into .claude subfolder?

Unfortunately, not. To be automatically loaded as context the location needs to be CLAUDE.md. (See [docs](https://docs.anthropic.com/en/docs/claude-code/memory#determine-memory-type).

> Or actually merge it with the public README for humans, as these details may be useful for them (and other AI models) as well?

CLAUDE.md serves as the entry point. We could [import](https://docs.anthropic.com/en/docs/claude-code/memory#claude-md-imports) README.md for more info. But the readme has a lot more tokens and will make sessions more expensive. I would suggest keeping the files separate. If there are details that are useful for humans as well we should add them to the appropriate file (readme, docs/, etc.).

If there are rules relevant for Claude I usually ask it to extract it from the user docs and put it into the project memory.

doc/compodoc_sources/how-to-guides/end-to-end-tests.md
geigerzaehler: > Could you link some external docs on how to install/use claude?

Done in f9504a92fe6cfa1a7be46595d0a464be04029a8c

> Does it also work via GitHub Copilot, which me and some others in our team have set up already?

I don’t know. But it shouldn’t be a big issue to point Copilot to the command file and follow the instructions as well.

doc/compodoc_sources/how-to-guides/end-to-end-tests.md
```diff
@@ -94,7 +94,24 @@ option](https://argos-ci.com/docs/playwright#argosscreenshotpage-name-options)
 in `argosScreenshot()`. Set a default threshold with `argos upload
 --threshold=...`
 
-[aria-busy]: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-busy
-[Playwright]: https://playwright.dev/
-[Playwright UI]: https://playwright.dev/docs/test-ui-mode
-[Trace Viewer]: https://playwright.dev/docs/trace-viewer
```
sleidig, line 100: aren't we losing any external links here now?

geigerzaehler: Deleted this inadvertently. It’s restored now

sleidig (approved):
Awesome. Looking forward to try this

doc/compodoc_sources/how-to-guides/end-to-end-tests.md
sleidig: I did a bit of research and while you can select the Claude "model" in GitHub Copilot, this is not the same as using Claude Code and cannot execute the claude slash commands.

There is a free tier for Claude Code though. I'll experiment with that.

PR #3215, opened 2025-08-14
Admin Public Form pre-defined value: RangeError Maximum call stack size exceeded 

Abhinegi2:
closes: #3172

### Visible/Frontend Changes

- [x]
- [ ]

- [x]
- [ ]

### DIFF
--- a/src/app/core/admin/admin-entity-details/admin-entity-field/admin-entity-field.component.ts
+++ b/src/app/core/admin/admin-entity-details/admin-entity-field/admin-entity-field.component.ts
@@ -1,4 +1,4 @@
-import { Component, OnInit, inject } from "@angular/core";
+import { Component, inject, OnInit } from "@angular/core";
 import { Entity, EntityConstructor } from "../../../entity/model/entity";
 import {
   MAT_DIALOG_DATA,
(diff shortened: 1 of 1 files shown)


src/app/core/common-components/basic-autocomplete/custom-form-control.directive.ts
```diff
@@ -144,6 +148,7 @@ export abstract class CustomFormControlDirective<T>
 
   /** @deprecated the this.value setter seems to already do the same? */
   writeValue(val: T): void {
+    // only update if the value actually changed to prevent loops (we are using this in basicautocomplete component and also in multi select it getting called twice and creating loop)
```
Abhinegi2, line 151:
```suggestion
```

src/app/core/default-values/x-static/admin-default-value-static/admin-default-value-static.component.ts
```diff
@@ -113,5 +113,11 @@ export class AdminDefaultValueStaticComponent
     }
 
     this.value = { value: newValue };
+
+    // update if the value actually changed
+    const newConfig = { value: newValue };
+    if (JSON.stringify(newConfig) !== JSON.stringify(this.value)) {
+      this.value = newConfig;
+    }
```
Abhinegi2, line 121:
```suggestion
```

src/app/core/admin/admin-entity-details/admin-entity-field/admin-entity-field.component.ts
```diff
@@ -175,7 +175,7 @@ export class AdminEntityFieldComponent implements OnInit {
     this.schemaFieldsForm.valueChanges.subscribe((v) => {
       if (JSON.stringify(v) === JSON.stringify(this.data.entitySchemaField))
         return;
-      this.data.entitySchemaField = { ...this.getUpdatedSchemaField(v) };
+      Object.assign(this.data.entitySchemaField, this.getUpdatedSchemaField(v));
```
sleidig, line 178: We added this originally because we want to follow the pattern that a component doesn't create side effects (e.g. change an object that is passed in) and only makes changes by emitting Output properties. That makes undo/cancel operations easier and gives us clearer control and less issues over change detection.

One alternative that might also help fix the losing-focus bug: Can we change the form control to [updateOn](https://v17.angular.io/api/forms/AbstractControlOptions#updateOn) "blur" - and would that avoid repeated updates even if we keep the re-creating a new object here?

Abhinegi2: Thanks, the blur seems to be working as expected (hadn’t used it before, so new thing for me)

I applied the blur event to the entire schemaFieldsForm, but during testing, I found that for new fields, the auto-generate ID function wasn’t getting triggered until we left the input. So, now added it specifically for default values.

Issue #3186, opened 2025-08-02
inherited value: multi-select (isArray) field should be able to inherit a single value from a single select field


sleidig:
The defaultValue "inherited-from-referenced-entity" mode can copy one value from a linked entity into the field. e.g. copying the group coordinator from a group entity to the "responsible coordinators" of a participant entity. This currently breaks if one of the fields is an array (multi-select) but the other is only a single value.

Setup:
- (parent) entity: School
   - has a field "assigned user" (data type "entity", no multi-select)
- (child) entity: Child
   - has a field "assigned school" (data type "entity")
   - has a field "responsible users" (data type "entity", allows multi-select / isArray)
   - "responsible users" field has a defaultValue rule to "inherit" the value from "assigned user" field of the record linked in the "assigned school" field

-----

1. create a new Child record and select a School (that has an "assigned user in its record)
- Current behavior (broken): The "responsible users" field is not getting filled correctly (after saving, the doc in the DB actually shows each letter of the ID as its own entry in the array - the string is interpreted as an array
- Expected behavior: The one user should be automatically added in the field. The user can manually select additional users.

-----
## Suggested approach:
- [x] write a unit test that fails, replicating the problem
- [x] adapt InheritedValueService to handle copying from single to isArray value

Abhinegi2:
@george-neha could you test the fixes for the above-mentioned issue here - https://pr-3220.aam-digital.net/

PR #3220, opened 2025-08-15
inherited value: multi-select (isArray) field should be able to inherit a single value from a single select field 

Abhinegi2:
closes: #3186


sleidig (approved):
Thanks!

PR #3240, opened 2025-08-20
 Link2Mail simple entity action 

### LINKED ISSUE #3159: Link2Mail simple entity action
_As a user I want to quickly write emails to contacts managed in Aam Digital without copying around email IDs._

Our system can be used for contact management, including storing of email ID(s) as fields in the record. As a first step, which serves as the basis for more advanced mail features including templates, users should be able to open their email client (e.g. Outlook) with one click, addressing the current entity's email.

In the entity details actions menu, add an option "Send E-Mail". This should execute a `href="mailto:joe@example.org"` action, opening the local email client prefilled with the email ID of the entity. (for extension, see #...).

-----
## Suggested implementation steps

1. implement an "email" datatype, so that admin users can define specific fields as email IDs
   1. create a new datatype, similar to URL datatype
   2. move the existing `validEmail` validator from `DynamicValidatorsService` to the new datatype
2. implement the entity action   1. create a service that returns email fields of an entity
   1. implement a function in the service that executes the mailto link action
   2. register entity action

Abhinegi2:
- closes #3159

### Remaining TODOs:

- [ ] unit tests
- [ ] e2e test?

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -98,6 +98,7 @@ import { KeycloakAdminService } from "./core/user/user-admin-service/keycloak-ad
 import { defaultValueStrategyProviders } from "./core/default-values/standard-default-value-strategies";
 import { DemoDataModule } from "./core/demo-data/demo-data.module";
 import { MAT_FORM_FIELD_DEFAULT_OPTIONS } from "@angular/material/form-field";
+import { EmailClientServicetModule } from "./features/email-client/email-client.module";
 
 /**
  * Main entry point of the application.
@@ -144,6 +145,7 @@ import { MAT_FORM_FIELD_DEFAULT_OPTIONS } from "@angular/material/form-field";
     PublicFormModule,
     SkillModule,
     DemoDataModule,
+    EmailClientServicetModule,
     // top level component
     UiComponent,
     // Global Angular Material modules
--- a/src/app/core/basic-datatypes/string/display-email/display-email.component.ts
+++ b/src/app/core/basic-datatypes/string/display-email/display-email.component.ts
@@ -5,13 +5,6 @@ import { DynamicComponent } from "../../../config/dynamic-components/dynamic-com
 @DynamicComponent("DisplayEmail")
 @Component({
   selector: "app-display-email",
-  template: `
-    @if (value) {
-      <a [href]="'mailto:' + value" class="clickable">{{ value }}</a>
-    } @else {
-      <span>-</span>
-    }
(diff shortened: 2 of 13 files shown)


Abhinegi2:
there are few issues-

- [ ] If users don't have a desktop email application (like Outlook, Thunderbird, Apple Mail) set up as your default mail handler, nothing will open.
- [ ] Chrome and Edge do not have a mail client handler set by default.

Abhinegi2:
> > for showing a warning or info message (e.g., if the email client did not open, suggesting the user set an email app as their default mail client)i wanted suggestion what the best UX(like tooltip or where?) Maybe you can take a look at the current approach once?
> 
> @Abhinegi2 , is there a way I can disconnect my email client or test the case when there is no default?

i also tried to disconnect my default email client but did not find any way

Abhinegi2:
I tested this on my colleague’s laptop, and nothing happened if no email client was connected.
However, on Sadaf’s and Neha’s laptops, an option appeared to select services like Gmail, Yahoo, etc. After selecting one, it prompted for login, but nothing happened afterward.

I also researched whether there’s a way to detect if users have an email client connected, but I haven’t found any solution so far.
If I don’t find anything, then yes, we’ll need to show some details in the dialog.

sleidig (commented):
Functionality seems quite good!

- [x] if no email type field is present, we shouldn't show the "Send Email" action in the menu
- [x] if an email field is there but empty, we should show an alert explaining "Please fill an email address for this record to use this functionality." (or something similar)
- [x] can you add basic unit tests (maybe for those edge cases above)? I think we sometimes inject the "window" object through Angular DI to be able to mock it for tests also.

sleidig (approved):

e2e/tests/attendance-tests.spec.ts
```diff
@@ -269,6 +269,11 @@ test("Assign a recurring activity to a user", async ({ page }) => {
   // When I assign myself
   await page.getByRole("button", { name: "Edit" }).click();
   await page.getByLabel("Assigned user(s)").click();
+
+  // wait to ensure the dropdown option is rendered and interactable, to avoid playwriting timeout issue
+  await page
+    .getByRole("option", { name: currentUser.name, exact: true })
+    .waitFor({ timeout: 10000 });
```
sleidig, line 276: @geigerzaehler , any ideas why this apparently started running into timeouts?
And is this the best way to solve it or can/should we add timeout to the click?

Issue #3160, opened 2025-07-22
Link2Mail: email templates for mail action


sleidig:
_As a user I want to use prepared email templates to be able to communicate faster and more consistently._

Extending the very basic functionality of #3159, users should have the option to prepare templates with fixed subject and body text that can be selected with the "Send E-Mail" action. The user should see a selection dialog, similar to the popup shown for "Generate File". A dropdown offers all available templates to choose from. After selecting one and confirming the button "Open in email client", the local email app (e.g. Outlook) opens prefilled with the TO address as well as subject and body text from the template.

<img width="899" height="299" alt="Image" src="https://github.com/user-attachments/assets/24330237-60dd-4866-a120-abeca2395933" />

-----
## Suggested implementation steps:

1. Extend the "sendMail" method of the service, called by the existing entity action, to include a subject and body text (e.g. see https://stackoverflow.com/a/4782084/1473411) (maybe start with a static text for testing)
2. Implement an Entity Type (isInternalEntity=true) for the Templates
   1. a similar architecture to the `template-export` module can probably be followed. It will also be best to keep all files in a feature folder similarly
   2. create and register views to list and edit (details) relevant fields of the template entity
   3. add a menu entry in the admin overview to this list view
3. Implement a simple dialog component to select available templates (and offer a link to create new templates) similar to the one for generating files.

Abhinegi2:
If someone wanted to use formatting (e.g., bold text or structuring the body of an email),
 I don’t think it’s possible with the current approach. ( mailto: seems to be limited to plain text bodies only)

sleidig:
Yes, that's my understanding also. The implementation with mailto links is still vastly easier than setting up some server-side SMTP service. So we'll accept this limitation in this approach for now.

PR #3245, opened 2025-08-25
Feat/email template

Abhinegi2:
- closes #[#3160](https://github.com/Aam-Digital/ndb-core/issues/3160)

### Remaining TODOs:

- [ ] unit tests
- [ ] e2e test?

### DIFF
--- a/src/app/features/email-client/email-client.service.ts
+++ b/src/app/features/email-client/email-client.service.ts
@@ -26,6 +26,7 @@ export class EmailClientService {
     ) as EntityConstructor<Entity>;
 
     let recipient: string | null = null;
+
     for (const [, field] of entityType.schema.entries()) {
       if (field.dataType === EmailDatatype.dataType) {
         const emailValue = entity[field.id];
--- a/src/app/features/email-client/email-client.service.ts
+++ b/src/app/features/email-client/email-client.service.ts
@@ -5,6 +5,7 @@ import { inject, Injectable } from "@angular/core";
 import { AlertService } from "#src/app/core/alerts/alert.service";
 import { MatDialog } from "@angular/material/dialog";
 import { EmailTemplateSelectionDialogComponent } from "../email-template-selection-dialog/email-template-selection-dialog.component";
+import { lastValueFrom } from "rxjs";
 
 @Injectable({
   providedIn: "root",
@@ -19,15 +20,12 @@ export class EmailClientService {
    *
    * If no default email client is available on the device / configured in the browser, then nothing will happen here.
    */
-  executeMailtoFromEntity(entity: Entity): boolean {
+  async executeMailtoFromEntity(entity: Entity): Promise<boolean> {
     const entityType = this.entityRegistry.get(
       entity.getType(),
     ) as EntityConstructor<Entity>;
 
-    this.dialog.open(EmailTemplateSelectionDialogComponent, { data: entity });
-
     let recipient: string | null = null;
-
     for (const [, field] of entityType.schema.entries()) {
       if (field.dataType === EmailDatatype.dataType) {
         const emailValue = entity[field.id];
@@ -43,7 +41,21 @@ export class EmailClientService {
       return false;
     }
 
-    const mailto = "mailto:" + encodeURIComponent(recipient);
+    const dialogRef = this.dialog.open(EmailTemplateSelectionDialogComponent);
+    const template = await lastValueFrom(dialogRef.afterClosed());
+
+    if (!template) return false;
+
+    const enc = (v: string) => encodeURIComponent(v);
+    const params: string[] = [];
+    const subject = template.subject?.toString().trim();
+    const body = template.body?.toString();
+
+    if (subject) params.push(`subject=${enc(subject)}`);
+    if (body) params.push(`body=${enc(body)}`);
+
+    const mailto = `mailto:${enc(recipient)}${params.length ? `?${params.join("&")}` : ""}`;
+
     window.location.href = mailto;
     return true;
   }
--- a/src/app/features/email-client/email-client.service.ts
+++ b/src/app/features/email-client/email-client.service.ts
@@ -3,6 +3,8 @@ import { EntityRegistry } from "#src/app/core/entity/database-entity.decorator";
 import { Entity, EntityConstructor } from "#src/app/core/entity/model/entity";
 import { inject, Injectable } from "@angular/core";
 import { AlertService } from "#src/app/core/alerts/alert.service";
+import { MatDialog } from "@angular/material/dialog";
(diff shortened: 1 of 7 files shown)


src/app/features/email-template-selection-dialog/email-template-selection-dialog.component.html
```diff
@@ -0,0 +1,50 @@
...
+    >
+  </p>
+  <p>
+    <span i18n>
+      Note: Nothing will happen unless you have a default email client set up.
+      Please configure one to proceed.
+    </span>
+  </p>
+
+  <app-entity-select
```
sleidig, line 26: My password manager tries to fill this (and overlaps our selection dropdown ...).

Maybe we can add `autocomplete="off"` to the input html in the entity-select component in general?
https://developer.mozilla.org/en-US/docs/Web/Security/Practical_implementation_guides/Turning_off_form_autocompletion

src/app/features/email-client/email-client.service.ts
```diff
@@ -38,7 +42,22 @@ export class EmailClientService {
...
+
+    const enc = (v: string) => encodeURIComponent(v);
+    const params: string[] = [];
+    const subject = template.subject?.toString().trim();
+    const body = template.body?.toString();
+
+    if (subject) params.push(`subject=${enc(subject)}`);
+    if (body) params.push(`body=${enc(body)}`);
+
+    const mailto = `mailto:${enc(recipient)}${params.length ? `?${params.join("&")}` : ""}`;
```
sleidig, line 60: currently my body text always has an empty line above it in the email app. Can we remove that linebreak?

src/app/features/email-template-selection-dialog/email-template-selection-dialog.component.html
```diff
@@ -0,0 +1,50 @@
...
+      }"
+      matDialogClose
+      i18n
+      >configure available Email templates</a
+    >
+  </p>
+  <p>
+    <span i18n>
+      Note: Nothing will happen unless you have a default email client set up.
+      Please configure one to proceed.
```
sleidig, line 22:
```suggestion
      Note: "Sending" mail will open the email application on your device.
      If nothing happens you may have to set up a default email client.
```

src/app/features/email-template-selection-dialog/email-template-selection-dialog.component.html
```diff
@@ -0,0 +1,50 @@
...
+    </span>
+  </p>
+
+  <app-entity-select
+    [entityType]="EmailTemplate.ENTITY_TYPE"
+    [multi]="false"
+    [additionalFilter]="filteredTemplate"
+    label="Email Template"
+    i18n-label
+    placeholder="Select a email template"
```
sleidig, line 32:
```suggestion
    placeholder="Select an email template"
```

src/app/features/email-template-selection-dialog/email-template-selection-dialog.component.html
```diff
@@ -0,0 +1,50 @@
...
+    i18n-placeholder
+    [form]="emailTemplateSelectionForm"
+    [disableCreateNew]="true"
```
sleidig, line 35: hmm, we could even allow this here? I hadn't thought about the inline "create new" function in this context yet.

src/app/features/email-template-selection-dialog/email-template-selection-dialog.component.ts
```diff
@@ -0,0 +1,88 @@
...
+      );
+    } else {
+      return (
+        !Array.isArray(e.availableForEntityTypes) ||
+        e.availableForEntityTypes.length === 0
+      );
+    }
+  };
+
+  selectedTemplate(template: EmailTemplate) {
```
sleidig, line 79: maybe calling this "confirmSelectedTemplate" (or something that indicates this is returning and closing the dialog) would make it a bit more clear

src/app/features/email-template-selection-dialog/email-template-selection-dialog.component.ts
```diff
@@ -0,0 +1,88 @@
...
+    this.emailTemplateSelectionForm.setValue(template);
+
+    const selectedTemplate = this.availableTemplates.find(
+      (template: EmailTemplate) =>
+        template.getId() === this.emailTemplateSelectionForm.value,
+    );
```
sleidig, line 85: why do we find the template here again instead of just return the one that is passed in here as an argument to the method?

src/app/features/email-template-selection-dialog/email-template-selection-dialog.component.ts
```diff
@@ -0,0 +1,88 @@
...
+    } else {
+      return (
+        !Array.isArray(e.availableForEntityTypes) ||
+        e.availableForEntityTypes.length === 0
+      );
+    }
+  };
+
+  selectedTemplate(template: EmailTemplate) {
+    this.emailTemplateSelectionForm.setValue(template);
```
sleidig, line 80: isn't setting the value to the form already done by the EntitySelectComponent internally?

src/app/features/email-template-selection-dialog/email-template-selection-dialog.component.html
Abhinegi2: I think then we should add this into the basicautocomplete component because it's been used in entity-select

src/app/features/email-template-selection-dialog/email-template-selection-dialog.component.ts
Abhinegi2: In the argument, we’re only receiving the selected template ID, not the full email template config. Since I already loaded all available templates earlier, I’m just looking up the matching one from that list.

src/app/features/email-template-selection-dialog/email-template-selection-dialog.component.html
Abhinegi2: I tried to add this, but still the password extension is overlapping. I found that Browsers often ignore both autocomplete="off" and autocomplete="false" on input fields, especially if the input name, id, or context resembles an email, password, or credential field. This triggers the browser’s password manager popup regardless of your intentions.

src/app/features/email-client/email-template.entity.ts
```diff
@@ -0,0 +1,51 @@
...
+  /**
+   * Optional entity types this template is available for.
+   */
+  @DatabaseField({
+    label: $localize`:EmailTemplate:Available for Entity Type(s)`,
+    editComponent: "EditEntityType",
+    viewComponent: "DisplayEntityType",
+    isArray: true,
+  })
+  availableForEntityTypes?: string;
```
sleidig: This actually is mis-typed. with `isArray` it must be `string[]`

sleidig (approved):
made a few adjustments. otherwise looks good :pray:

PR #3250, opened 2025-08-26
Fix race condition in `app-entity-select`

geigerzaehler:
Before this change it could happen that `this.availableOptions.next(newAvailableEntities)` was called out of order which resulted in only the entity from `this.form.value` being displayed and not those loaded by `loadAvailbleEntities`. We rewrite the component with signals and Observables to prevent this issue.

See https://github.com/Aam-Digital/ndb-core/pull/3240#discussion_r2300337858

### DIFF
--- a/src/app/core/common-components/entity-select/entity-select.component.spec.ts
+++ b/src/app/core/common-components/entity-select/entity-select.component.spec.ts
@@ -22,8 +22,8 @@ import { createEntityOfType } from "../../demo-data/create-entity-of-type";
 import { DatabaseEntity } from "../../entity/database-entity.decorator";
 
 describe("EntitySelectComponent", () => {
-  let component: EntitySelectComponent<any, any>;
-  let fixture: ComponentFixture<EntitySelectComponent<any, any>>;
+  let component: EntitySelectComponent<any>;
+  let fixture: ComponentFixture<EntitySelectComponent<any>>;
   let test1Entities: TestEntity[];
   let test2Entities: Test2Entity[];
   let formControl: FormControl;
@@ -54,49 +54,38 @@ describe("EntitySelectComponent", () => {
 
   beforeEach(() => {
     fixture = TestBed.createComponent(EntitySelectComponent);
+    fixture.componentRef.setInput("form", formControl);
+    fixture.autoDetectChanges();
+
     component = fixture.componentInstance;
-    component.form = formControl;
-    fixture.detectChanges();
   });
-
-  it("should create", () => {
-    expect(component).toBeTruthy();
-  });
-
-  it("eventually loads all entities of the given type when the entity-type is set", fakeAsync(() => {
-    component.entityType = TestEntity.ENTITY_TYPE;
-    fixture.detectChanges();
-    tick();
-    expect(component.allEntities.length).toBe(test1Entities.length);
-  }));
 
   it("should not be in loading-state when all data is received", fakeAsync(() => {
-    component.entityType = TestEntity.ENTITY_TYPE;
-    expect(component.loading.value).toBe(true);
+    fixture.componentRef.setInput("entityType", TestEntity.ENTITY_TYPE);
+    fixture.autoDetectChanges();
+    expect(component.loading()).toBe(true);
     tick();
-    expect(component.loading.value).toBe(false);
+    expect(component.loading()).toBe(false);
   }));
 
   it("should suggest all entities after an initial load", fakeAsync(() => {
-    component.entityType = TestEntity.ENTITY_TYPE;
+    fixture.componentRef.setInput("entityType", TestEntity.ENTITY_TYPE);
+    fixture.autoDetectChanges();
     tick();
-    fixture.detectChanges();
-    expect(component.availableOptions.value).toEqual(
+    expect(component.availableOptions()).toEqual(
       jasmine.arrayWithExactContents(test1Entities),
     );
   }));
 
   it("suggests all entities of multiple different types if configured", fakeAsync(() => {
-    component.entityType = [TestEntity.ENTITY_TYPE, Test2Entity.ENTITY_TYPE];
+    fixture.componentRef.setInput("entityType", [
+      TestEntity.ENTITY_TYPE,
+      Test2Entity.ENTITY_TYPE,
+    ]);
+    fixture.autoDetectChanges();
     tick();
-    fixture.detectChanges();
 
-    expect(component.allEntities).toEqual(
-      [...test1Entities, ...test2Entities].sort((a, b) =>
-        a.toString().localeCompare(b.toString()),
-      ),
-    );
-    expect(component.availableOptions.value).toEqual(
+    expect(component.availableOptions()).toEqual(
       [...test1Entities, ...test2Entities].sort((a, b) =>
         a.toString().localeCompare(b.toString()),
       ),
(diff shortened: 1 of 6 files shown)


src/app/core/common-components/entity-select/entity-select.component.spec.ts
```diff
@@ -14,16 +14,13 @@ import { MockedTestingModule } from "../../../utils/mocked-testing.module";
...
-import { FormDialogService } from "../../form-dialog/form-dialog.service";
-import { of } from "rxjs";
 import { TestEntity } from "../../../utils/test-utils/TestEntity";
 import { createEntityOfType } from "../../demo-data/create-entity-of-type";
 import { DatabaseEntity } from "../../entity/database-entity.decorator";
 
-describe("EntitySelectComponent", () => {
-  let component: EntitySelectComponent<any, any>;
-  let fixture: ComponentFixture<EntitySelectComponent<any, any>>;
+fdescribe("EntitySelectComponent", () => {
```
sleidig, line 21:
```suggestion
describe("EntitySelectComponent", () => {
```

src/app/core/common-components/entity-select/entity-select.component.ts
```diff
@@ -139,86 +132,73 @@ export class EntitySelectComponent<
...
-   * Edit form value (currently selected) and the given available Entities to be consistent:
-   * Entities that do not exist should be removed from the form value
-   * and availableEntities should contain all selected entities, even from other types.
-   * @private
-   */
-  private async alignAvailableAndSelectedEntities(availableEntities: E[]) {
-    if (this.form?.value === null || this.form?.value === undefined) {
-      return;
-    }
+  #allEntities: Observable<E[]> = toObservable(this.entityType)
```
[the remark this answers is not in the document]
geigerzaehler: The `#` is for [private properties](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements). I didn’t even think about it. I’ll change to `private allEntities` for consistency.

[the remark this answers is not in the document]
sleidig: oh, interesting. Learned something new today :blush:    

Yeah, for consistency maybe let's stick with TypeScript's `private` keyword only?

src/app/core/common-components/entity-select/entity-select.component.ts
```diff
@@ -139,86 +128,87 @@ export class EntitySelectComponent<
...
-  private async updateAvailableOptions() {
-    const includeInactive = (entity: E) =>
-      this.includeInactive || entity.isActive;
-    const includeSelected = (entity: E) =>
-      asArray(this.form.value).includes(entity.getId());
-
-    const newAvailableEntities = this.allEntities.filter(
-      (e) => includeInactive(e) || includeSelected(e),
-    );
+  private entitiesForType: Resource<E[]> = resource({
```
sleidig, line 131: I first thought this is only loading the entities of a single type (of possibly multiple types). Maybe naming it `allEntities` again or `entitiesOfTypes` could be a bit more clear in that regard?

src/app/core/common-components/entity-select/entity-select.component.ts
```diff
@@ -139,86 +128,87 @@ export class EntitySelectComponent<
...
+        asArray(params.values).includes(entity.getId());
+
+      const availableEntities = params.allEntities.filter(
+        (e) => params.includeInactive || e.isActive || includeSelected(e),
+      );
 
-      if (availableEntities.find((e) => id === e.getId())) {
-        // already available, nothing to do
-        continue;
+      if (params.values !== null && params.values !== undefined) {
```
sleidig, line 182: with its initialValue and map(asArray), could we assume that `values` is never undefined?

src/app/core/common-components/entity-select/entity-select.component.ts
```diff
@@ -139,86 +128,87 @@ export class EntitySelectComponent<
...
+
+  private availableOptionsResource: Resource<E[]> = resource({
+    defaultValue: [],
+    params: () => ({
+      allEntities: this.entitiesForType.value(),
+      values: this.values(),
+      includeInactive: this.includeInactive(),
+    }),
+    loader: async ({ params }) => {
+      const includeSelected = (entity: E) =>
```
sleidig, line 175:
```suggestion
      const isSelected = (entity: E) =>
```

src/app/core/common-components/entity-select/entity-select.component.ts
```diff
@@ -139,86 +128,87 @@ export class EntitySelectComponent<
...
+          const additionalEntity = await this.getEntity(id);
+          if (additionalEntity) {
+            availableEntities.push(additionalEntity);
+          } else {
+            this.hasInaccessibleEntities = true;
+            availableEntities.push({
+              getId: () => id,
+              isHidden: true,
+            } as unknown as E);
+          }
```
sleidig, line 201: maybe extract into private helper method to reduce complexity of this loader function?

src/app/core/common-components/entity-select/entity-select.component.ts
```diff
@@ -139,86 +128,87 @@ export class EntitySelectComponent<
...
-    }
+      return availableEntities;
+    },
+  });
 
-    if (this.form.value !== updatedValue) {
-      this.form.setValue(updatedValue);
-    }
-  }
+  availableOptions: Signal<E[]> = computed(() =>
```
sleidig, line 209:
```suggestion
  /**
   * The currently visible options for the user to select from, considering current values and filters.
   */
  availableOptions: Signal<E[]> = computed(() =>
```

sleidig (commented):
I love the new structure with signals, @geigerzaehler  :heart_eyes: 
Thanks a lot for bringing in this new approach.

Commented a few thoughts to further simplify / clarify. I didn't make the changes directly yet because I am not experienced with signals myself and want to make sure I am not missing some intentional points of you here.

src/app/core/common-components/entity-select/entity-select.component.ts
```diff
@@ -114,19 +108,14 @@ export class EntitySelectComponent<
...
-   */
-  loading = new BehaviorSubject(true);
-  allEntities: E[] = [];
-  availableOptions = new BehaviorSubject<E[]>([]);
-
   hasInaccessibleEntities: Boolean = false;
 
-  @Input() includeInactive: boolean = false;
-  currentlyMatchingInactive: number = 0;
+  includeInactive = signal<boolean>(false);
```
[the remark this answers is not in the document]
geigerzaehler: This property is never used as an input but it [set internally](https://github.com/Aam-Digital/ndb-core/pull/3250/files#diff-8f8a539da43699e627525cbc19341821204f6f9eea56d02e702a8a754ec563d6R232). I’m making the property private to emphasize that

src/app/core/common-components/entity-select/entity-select.component.ts
geigerzaehler: Some of the other changes reduced the complexity of the function. To me it doesn’t make sense to split up the function because it’s logic belongs together. And creating a separate function would just create some indirection. Let me know if you still would like it to be extracted.

geigerzaehler:
> Clicking on a single select entity field doesn't select anymore (e.g. editing a school in the child's school history entries)

I’m not sure I understand the problem. I’ve tried the following and it works

[Peek 2025-09-02 10-03.webm](https://github.com/user-attachments/assets/be63d745-2953-49c2-b66d-9bb876aee95e)

sleidig:
> > Clicking on a single select entity field doesn't select anymore (e.g. editing a school in the child's school history entries)
> 
> I’m not sure I understand the problem. I’ve tried the following and it works

Somehow, switching from an existing selected works, but not if you first remove the existing and then select a new one (or create a new record and try to fill the empty field :see_no_evil:

geigerzaehler:
> Somehow, switching from an existing selected works, but not if you first remove the existing and then select a new one (or create a new record and try to fill the empty field 🙈

Yes, this didn’t work. While reloading, the resource value would revert to the default value (`[]`) and this would mess up something in `app-basic-autocomplete`. I couldn’t figure the root cause but I’ve added a `resourceWithRetention` variant that is more in line with our expectations about resources. (There’s an [ongoing discussion](https://github.com/angular/angular/issues/58602) on Angular to provide something similar.)

PR #3252, opened 2025-08-27
feat: add "invalid" and "empty" values in filter dropdown

### LINKED ISSUE #3147: include "invalid options" in filter dropdowns?
- [x] User request: for data cleaning of invalid dropdown options (i.e. values in data that don't exist as dropdown options), it would be useful to also offer these in the list filter. That way, users can quickly select all the records with a certain invalid value and use the bulk actions to correct these.

- [x] On the other hand, we should make sure to keep the filter options simple and clean - so maybe "invalid" values should only be listed at the end, after all the valid options.

- [x] Similarly, it would be useful for data cleaning to be able to filter for "empty" values. (this should also allow to navigate to such filtered list when clicking from the entity-count-dashboard)

-  "Invalid option" and "not defined(empty)" text should be style differently 
     - [x]  text style might be in "italic"
     - [x]  text color for "invalid option"  can be red and "not defined" can be grey similar to what is in the entity count dashboard @Abhinegi2 this one is still need to be implemented        

Image

sadaf895:
- closes #3147

### DIFF
--- a/src/app/core/filter/filter-generator/filter-generator.service.ts
+++ b/src/app/core/filter/filter-generator/filter-generator.service.ts
@@ -80,7 +80,7 @@ export class FilterGeneratorService {
           )
           .map((invalidId) => ({
             key: `invalid:${invalidId}`,
-            label: `[Invalid: ${invalidId}]`,
+            label: $localize`:filter option:[Invalid: ${invalidId}]`,
             isInvalid: true,
             filter: { [filterConfig.id + ".id"]: invalidId } as DataFilter<T>,
           }));
@@ -93,7 +93,7 @@ export class FilterGeneratorService {
           ? [
               {
                 key: null,
-                label: "not defined",
+                label: $localize`:filter option:not defined`,
                 isEmpty: true,
                 filter: {
                   $or: [


src/app/core/filter/filters/configurableEnumFilter.ts
```diff
@@ -11,15 +10,17 @@ export class ConfigurableEnumFilter<
     label: string,
     enumValues: ConfigurableEnumValue[],
     singleSelectOnly: boolean = false,
+    InvalidOptions: FilterSelectionOption<T>[] = [],
```
Abhinegi2, line 13:
```suggestion
    invalidOptions: FilterSelectionOption<T>[] = [],
```

src/app/core/filter/filter-generator/filter-generator.service.ts
```diff
@@ -56,11 +57,61 @@ export class FilterGeneratorService {
...
+                key: "__empty__",
+                label: "<i>not defined</i>",
+                filter: {
+                  $or: [
+                    { [filterConfig.id]: undefined },
+                    { [filterConfig.id]: null },
+                    { [filterConfig.id + ".id"]: undefined },
+                    { [filterConfig.id + ".id"]: null },
+                    { [filterConfig.id + ".id"]: "" },
+                  ],
```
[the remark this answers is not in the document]
sadaf895: @Abhinegi2 
When filtering for "empty" values in a field, the data can be inconsistent depending on:
- How the field is stored (primitive vs. object)
- How missing/empty values are represented (undefined, null, or empty string)
- Whether the field is a direct property or a nested object with an [id] property
So , all these conditions are included to robustly match any record where the filter field is "empty", regardless of how that emptiness is represented in the data. This ensures users can reliably filter for missing or unset values, even if the data is inconsistent.

Abhinegi2:
while testing this I noticed a few things:

1. In the **list view filter**, invalid options are displayed in *italic*, but in the **list view and details view**, they are shown with normal text styling. Shouldn’t we make this consistent across all views? 
image

2. I deleted two centers from the dropdown, and in the filter it now shows `[Invalid:Id]`. This might be confusing, especially when the dropdown **ID and label are completely different**. For example, in the screenshot below I deleted *Alipore*, but the filter displays `Invalid:c1`. 
image

3. The following point still isn’t working- “Text color for invalid options should be red, and for not defined it should be grey (similar to the entity count dashboard).”

@sleidig and @sadaf895  What do you think about points 1 and 2?

src/app/core/filter/filter-generator/filter-generator.service.ts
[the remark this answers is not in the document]
Abhinegi2: @sadaf895 could we add a unit test for these conditions

src/app/core/common-components/basic-autocomplete/basic-autocomplete.component.ts
```diff
@@ -209,6 +209,11 @@ export class BasicAutocompleteComponent<O, V = O>
     });
   }
 
+  getPlainTextTooltip(htmlString: string): string {
+    // Remove HTML tags for tooltip display
+    return htmlString.replace(/<[^>]*>/g, "");
```
sleidig, line 214: let's not use html in strings and then parse it out again at other places. Why do we need this here? Let's better add explicit css classes based on conditions to the html templates to have clear control

src/app/core/common-components/basic-autocomplete/basic-autocomplete.component.html
```diff
@@ -116,8 +116,17 @@
...
               class="text-truncate"
-              matTooltip="{{ item.asString }}"
+              [matTooltip]="item.asString"
               matTooltipPosition="above"
+              [ngClass]="{
+                'special-option': item.isInvalid || item.isEmpty,
+              }"
+              [style.color]="
+                item.asString.includes('[Invalid:') ||
+                item.asString.includes('not defined')
```
sleidig, line 126: can't we use `item.isInvalid || item.isEmpty` for this also?

src/app/core/filter/filter-generator/filter-generator.service.ts
```diff
@@ -56,11 +57,65 @@ export class FilterGeneratorService {
...
+        // Add "empty" option if there are empty/undefined values
+        const hasEmpty = dataValues.some(
+          (v) => v === undefined || v === null || v === "",
+        );
+        const emptyOption = hasEmpty
+          ? [
+              {
+                key: "__empty__",
+                label: "not defined",
+                color: "#888",
```
sleidig, line 98: we wanted to use the same css class as for the enitty-count-dashboard, to ensure consistent styling (even if it should change in the future).

--> try to
1. create a second scss file and move the `not-defined-label` class (and one for invalid) to that
2. import it both in entity-count-dashboard and this component
3. use an `ngClass` to set the color conditionally for these special entries, instead of adding the color here

sadaf895:
@sleidig I hope everything is good to go now

src/app/core/filter/filter-generator/filter-generator.service.ts
```diff
@@ -56,11 +57,63 @@ export class FilterGeneratorService {
...
+
+        // Add "empty" option if there are empty/undefined values
+        const hasEmpty = dataValues.some(
+          (v) => v === undefined || v === null || v === "",
+        );
+        const emptyOption = hasEmpty
+          ? [
+              {
+                key: null,
+                label: "not defined",
```
sleidig, line 96:
```suggestion
                label: $localize`:filter option:not defined`,
```

src/app/core/filter/filter-generator/filter-generator.service.ts
```diff
@@ -56,11 +57,63 @@ export class FilterGeneratorService {
...
+
+        // Find invalid options (not in enum)
+        const invalidOptions = dataValues
+          .filter(
+            (v) =>
+              v !== undefined && v !== null && v !== "" && !validIds.has(v),
+          )
+          .map((invalidId) => ({
+            key: `invalid:${invalidId}`,
+            label: `[Invalid: ${invalidId}]`,
```
sleidig, line 83:
```suggestion
            label: $localize`:filter option:[Invalid: ${invalidId}]`,
```

Issue #3231, opened 2025-08-19
Pre-Match confirmation dialog "Save" button is disabled until some of the pre-filled fields are edited


sleidig:
<img width="1015" height="418" alt="Image" src="https://github.com/user-attachments/assets/50580ade-3c44-4fd8-84af-9275ddbdad67" />

Ideally, users can just confirm all the prefilled information when using the "matching" dialog. However, our form only enables the save button after something was changed.

sleidig:
@Abhinegi2 , apparently this is still a bug in the latest 3.64.0 version 🙄

PR #3253, opened 2025-08-27
Link2Mail: document sent email as Note

### LINKED ISSUE #3161: Link2Mail: document sent email as Note
_As a user I want to document my communication done using the "send email" action as easily as possible (and not forget to document it)._

After opening an email template in the local email app (#3160), the user should automatically see the popup form to create a new Note, prefilled with the subject and body of the email template as subject and text of the note. The user can then either simple save the Note or make changes (e.g. adding additional remarks or changing the text based on changes that were made in the email before sending). The user can also cancel to skip adding a Note for that email (or because the email was actually not sent - it won't be possible for us to know if it was sent in the external email app).

The Note should also have a category field pre-selected, based on a category field the user can select in the Template entity.

New Note generated with:
- Note Subject = email template subject
- Note Text = email template body
- Note related entities (linked records) = email TO receivers
- Note category (dropdown) = email template "category" field (a field in the template entity type that is just used for prefilling this, so that users can define how to categorize such notes)

Abhinegi2:
- closes #3161
- closes #ANOTHER_GITHUB_ISSUE_NUMBER

### Remaining TODOs:

- [ ] unit tests
- [ ] e2e test?


sleidig (commented):
It takes a (very short) moment for my Thunderbird email to open, while the Note in our app immediately shows up. Maybe for more clear UI we could add a short dialog saying:

> Opening email on your device ...
>
> If nothing is happening, please check your default email client [link to user guide]
>
> ["Continue" button]

and only continue to the note after a delay of 5 seconds (?) or when the user clicks "continue" explicitly.

What do you all think?


also, the new note has the same UX issue as this issue here: https://github.com/Aam-Digital/ndb-core/issues/3231

sleidig (approved):
Looks good now :-)

PR #3268, opened 2025-09-01
fix(bug) : show link dropdown options in admin menu details

### LINKED ISSUE #3267: Dropdown for Link Field Missing in Admin Menu Item Details After Refactor
**Describe the bug**
After refactoring the menu item form logic into the new MenuItemFormComponent, the dropdown for selecting a link (route) in the admin menu item details dialog is missing. Previously, this dropdown allowed admins to select from available routes loaded from the config. Now, the link field is always a free text input, even in the admin dialog.

**Expected behavior**
- In the admin menu item details dialog, the link field should be a dropdown populated with available routes from the config.
- In other usages (e.g., shortcut dashboard settings), the link field should remain a free text input.

sadaf895:
- closes #3267

### DIFF
--- a/src/app/menu-item-form/menu-item-form.component.ts
+++ b/src/app/menu-item-form/menu-item-form.component.ts
@@ -6,6 +6,7 @@ import { FormsModule } from "@angular/forms";
 import { AdminIconComponent } from "../admin-icon-input/admin-icon-input.component";
 import { FaIconComponent } from "@fortawesome/angular-fontawesome";
 import { MatTooltipModule } from "@angular/material/tooltip";
+import { MatSelectModule } from "@angular/material/select";
 
 @Component({
   selector: "app-menu-item-form",
@@ -17,6 +18,7 @@ import { MatTooltipModule } from "@angular/material/tooltip";
     AdminIconComponent,
     FaIconComponent,
     MatTooltipModule,
+    MatSelectModule,
   ],
   templateUrl: "./menu-item-form.component.html",
   styleUrls: ["./menu-item-form.component.scss"],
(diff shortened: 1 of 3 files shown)


src/app/menu-item-form/menu-item-form.component.ts
```diff
@@ -17,13 +18,16 @@ import { MatTooltipModule } from "@angular/material/tooltip";
...
     MatTooltipModule,
+    MatSelectModule,
   ],
   templateUrl: "./menu-item-form.component.html",
   styleUrls: ["./menu-item-form.component.scss"],
 })
 export class MenuItemFormComponent {
   @Input() item!: MenuItem;
   @Input() showLinkTooltip = false;
+  @Input() showLinkDropdown = false;
```
Abhinegi2, line 29:
```suggestion
  @Input() showLinkOptions = false;
```

Abhinegi2 (commented):
We don't need both showLinkDropdown and availableOptions. Instead, we can simply pass [linkOptions]="availableRoutes" to the child component. In the child, rather than checking (showLinkDropdown && linkOptions?.length), we can just check (linkOptions?.length).
@sadaf895

Abhinegi2 (approved):

sleidig (commented):
Works well, thanks!

Actually, it would maybe be nice to 
1. give the dropdown on the dashboard shortcut widgets also?
2. allow users to define a completely custom link for the main menu (in case it doesn't match a normal one), switching to the simple text input

How about adding a small icon right of the input/select that allows users to manually switch between the two modes?

sadaf895:
> Works well, thanks!
> 
> Actually, it would maybe be nice to
> 
> 1. give the dropdown on the dashboard shortcut widgets also?
> 2. allow users to define a completely custom link for the main menu (in case it doesn't match a normal one), switching to the simple text input
> 
> How about adding a small icon right of the input/select that allows users to manually switch between the two modes?

I agree we could do that. @sleidig Maybe we can create different issue as a new feature type and do it in a different PR?

sleidig:
> I agree we could do that. @sleidig Maybe we can create different issue as a new feature type and do it in a different PR?

can you create an issue for that and pick it up? :blush:

PR #3270, opened 2025-09-02
enhancement: feasible UI of array-type data in list view

### LINKED ISSUE #3222: Small UI Tweak: Spacing/formatting of array-type data in list view
When for a given field, there is more than 1 value listed/referenced in there, e.g. "Option 1", "Option 2" and "Option 3", then in a list view, this is given out as "Option 1 ,  Option 2 ,  Option 3". That is, the spacing between the options is not ", " but rather " ,  ", i.e. there is one space "too much" in front of the comma and one too much after the comma.

Image

This is also evident when looking at the created code (where it seems we enforced a space by adding `&nbsp;`, but then two other spaces were added).

 
Image

I observed this on https://toladata.aam-digital.com/, e.g. for the list views https://toladata.aam-digital.com/project (field "Topic Area(s)") and the list view https://toladata.aam-digital.com/user (field "Qualification").

Within the details view for the respective entity, the spacing looks fine.

Image

 

-----
To reproduce:
1. create an enum field that allows multiple values
2. select more than one element in that field
3. make sure that column is visible in the list view --> see the problem there

sadaf895:
- closes #3222


george-neha:
A few thoughts:

1. The size of the buttons be based on content length, but cap their maximum width so they look consistent without wasting space.

2.For extremely long values, we truncate the field  with "..." (e.g., [ Super Long Org Nam... ]) to ensure a single entry doesn’t dominate the screen. Example: [ Org A ] [ Super Long Org Nam... ] [ Another Org ]. 
Screenshot 2025-09-02 at 12 29 39 PM

3. To allow users to view the full value, we can provide a tooltip on hover (desktop), long-press (mobile) to reveal the complete text.

sleidig:
Thanks, @george-neha . I like your suggestions!

> 1. The size of the buttons be based on content length, but cap their maximum width so they look consistent without wasting space.

I am not completely sure I understand. We did, for a short time, have a shared minimum width without always filling the whole space. I think the fact that it still doesn't make it fully consistent has gotten us to the point of making all the same width of the longest entry.

> 2.For extremely long values, we truncate the field with "..." (e.g., [ Super Long Org Nam... ]) to ensure a single entry doesn’t dominate the screen. Example: [ Org A ] [ Super Long Org Nam... ] [ Another Org ]. 
> 3. To allow users to view the full value, we can provide a tooltip on hover (desktop), long-press (mobile) to reveal the complete text.

That's great :+1:   Let's define a max-width for the individual items and use an ellipsis for overflowing text (+tooltip for the full text).

We can also cap the max number of items (maybe >3?) and if more show "7 values" (+ tooltip listing them out)

george-neha:
1. For values that have a longer name, tooltip is not available for such values.
2. There is a mismatch with fields having the same width as the longest entry, in case some records have multiple values selected.
image

PR #3281, opened 2025-09-04
added migration + added code base entity if  missing

### LINKED ISSUE #3255: RecurringActivity fields cannot be saved to config doc in DB if it is not yet in the Config doc
**Describe the bug**
In a default initial setup, the "entity:RecurringActivity" object is not part of the config in the DB, because we simply rely on the class defined in the code base. Users can still open the Admin UI for that entity type and make changes to fields.

Such changes are currently not stored into the DB but discarded after reloading the page. We need to make sure to add an "entity:" block to the config if it is not already defined and store that to the DB.

Abhinegi2:
- closes #3255
- closes #ANOTHER_GITHUB_ISSUE_NUMBER

### Remaining TODOs:

- [x] unit tests
- [ ] e2e test?


src/app/core/admin/admin-entity.service.ts
```diff
@@ -82,12 +83,26 @@ export class AdminEntityService {
...
   }
+
+  private getDefaultEntityConfig(
+    entityTypeId: string,
+    name: string,
+  ): EntityConfig {
+    return {
+      label: name,
+    };
+  }
```
Abhinegi2, line 107:
```suggestion

```

PR #3298, opened 2025-09-12
 Admin UI: editing Notes structure saves a broken config 

### LINKED ISSUE #3291: Admin UI: editing Notes structure saves a broken config
**Describe the bug**
After using "Configure Data Structure" for the Note entity type, the Note details view (e.g. popup creating new notes) is broken due to an invalid config in the DB.

**To Reproduce**

1. Go to 'Notes'
2. Open the Admin UI (context menu -> "Configure Data Structure")
3. Save some changes (e.g. reordering columns in the Notes list)
4. Try to create a new Note --> Now only showing an empty window with loading indicator

**Additional context**
The NoteDetails view works different from other Entity Details, therefore the `view:note/:id` config has a different format. We have disabled the Admin UI for changing the details view of Notes. But somehow saving the config still adds some invalid config to that view.

-----

## Alternative Solution: Create an`AdminNoteDetailsComponent`
to edit the view config for that details view.

The `NoteDetailsComponent` accepts three config properties: `topForm`, `middleForm`, `bottomForm`.
Admins should be able to select (and reorder) entity fields for each of these.

- [x] Create an interface `NoteDetailsConfig`, similar to `EntityDetailsConfig` and let NoteDetailsComponent implement it (so that we have a type defined for these config `@Input` properties that can be configured in Admin UI)
- [x] Create a new `AdminNoteDetailsComponent` that allows to edit these options
- [x] Update `AdminEntityComponent` to use this new component as an alternative to `<app-admin-entity-details>` in the relevant context

Abhinegi2:
- closes #3291

### Remaining TODOs:

- [ ] unit tests
- [ ] e2e test?

## Summary by CodeRabbit

- **New Features**
  - Admin UI: Add configurable Note Details editor with draggable three-section form, hint banner, sensible defaults, live updates and accessibility ID.

- **Layout**
  - New compact "fields-only" mode and a full-width form option for improved presentation and responsive field sizing.

- **Refactor**
  - Admin entity view/save flow updated to support and sanitize Note Details configurations.

- **Tests**
  - Unit tests added for the Admin Note Details UI.


sleidig:
> I’ve updated the UI for the Admin Notes details. Please review and let me know your thoughts before I proceed to the next step. Right now, the changes are being saved, but somehow the updated values in the notes details are getting lost. I’ll fix that once the UI looks good. Also, let me know if you have any suggestions or thoughts I can't add SS here because it's too large

I think the UI is pretty good now :slightly_smiling_face: 
We could allow the individual fields to wrap (instead of being below each other) within each section, as that is the real behavior also. But if that makes the code much more complex, the current design works fairly well also.

Abhinegi2:
> > Just one question for clarification: In the current approach, I’m displaying 3 sections — Top, Middle, and Bottom forms. so that when saving the note details configuration, it’s stored in the same format (i.e. topForm, middleForm, bottomForm). I thought we have a new noteDetailsConfig and that the noteDetailsComponent is designed to work based on these 3 sections.

I experimented with UI, so something like this?
image

george-neha:
@Abhinegi2 In the Config data structure>Details Views and Fields, I added a new field in the section of Subject and Notes. I am unable to view more than two fields in this section. 
image
image

src/app/core/admin/admin-entity/admin-entity.component.ts
```diff
@@ -58,9 +61,10 @@ export class AdminEntityComponent implements OnInit {
   entityConstructor: EntityConstructor;
   private originalEntitySchemaFields: [string, EntitySchemaField][];
 
-  configDetailsView: DynamicComponentConfig<EntityDetailsConfig>;
+  configDetailsView: DynamicComponentConfig<any>;
```
sleidig, line 64: maybe we can keep this slightly more typed for now?

```suggestion
  configDetailsView: DynamicComponentConfig<EntityDetailsConfig | NoteDetailsConfig>;
```

src/app/core/admin/admin-entity/admin-entity.component.ts
```diff
@@ -140,11 +144,30 @@ export class AdminEntityComponent implements OnInit {
...
+      if (noteConfig?.topForm) cleanNoteConfig.topForm = noteConfig.topForm;
+      if (noteConfig?.middleForm)
+        cleanNoteConfig.middleForm = noteConfig.middleForm;
+      if (noteConfig?.bottomForm)
+        cleanNoteConfig.bottomForm = noteConfig.bottomForm;
+
+      detailsViewConfig = {
+        ...this.configDetailsView,
+        config: cleanNoteConfig,
+      };
```
sleidig, line 163: can we delegate the responsibility of delivering a valid NoteDetailsConfig to the AdminNoteDetailsComponent only? It would be awesome if we don't have to distinguish cases within the generic admin-entity component

sleidig (approved):
Works well and looks like clean code. Nice job! :+1:

src/app/core/admin/admin-entity/admin-entity.component.ts
sleidig: can you narrow down how the details view gets those wrong properties? My assumption and goal would be that editing the list view does not touch the details view config in any way

Abhinegi2: I found a few things:

- When updating the listConfig, no extra property is added if we only reorder one field.
- However, on the second time, if we reorder, an extra property "entity": { full entityObj } gets added.

After checking, I noticed this property isn’t being saved to the DB by the list view itself. The issue seems to occur when the AdminEntityComponent initializes. At that time, the `loadViewConfig` method is called, which is somehow adding the entity property into the note details config, and then on saving, it's getting added.

This might be happening due to : 
`const viewConfig: DynamicComponentConfig = this.configService.getConfig(viewId);`

also, the DynamicComponentConfig interface uses any for its config; maybe thats time its getting added(not sure 100%)

Fornow, I removed the extra handling from save method and added a check in the loadViewConfig method: if the component is noteDetails and has an entity property, then delete it.

PR #3366, opened 2025-10-06
refactor: bulk actions to use a dynamic service registering available actions

### LINKED ISSUE #2894: Refactor bulk actions to use a dynamic service registering available actions
Currently, every bulk action is hard-coded into the EntityListComponent's html template and component logic. This should be decoupled so that other feature modules can dynamically register a new "bulk action" from other pieces of code without having to edit the EntityListComponent.

Maybe can be implemented like this:
- Extend the `EntityAction` with a flag `availableFor: "all" | "bulk-only" | "individual-only"`
- Create a component "EntityBulkActionsComponent"  similar to `EntityActionsMenuComponent` that replaces the current code in EntityListComponent showing all the buttons for different entity actions
  - switch the UI to a dropdown of actions instead of displaying all buttons because we are adding more and more actions
- If necessary, create a service that does bulk executions (or add another executeBulk function to the `EntityAction` definitions
- Move the code executing different bulk actions from the EntityListComponent to other services or Module constructors, registering them at the EntityActionsMenuComponent

If there is two many differences between an action on a single entity record and the bulk actions, maybe a separate service besides EntityActionsMenuService is an alternative approach.

sadaf895:
- closes #2894

### PR Checklist

-----
### Remaining TODOs:
- [ ] unit tests

## Summary by CodeRabbit

* **New Features**
  * Bulk actions UI with autocomplete, Cancel and selection-aware behavior.
  * New bulk "link external profiles" action (dialog) gated by skill-feature check.
  * Service endpoint to query skill feature availability.

* **Improvements**
  * Actions can declare availability scope (all / bulk-only / individual-only) and accept single or multiple selections.
  * List header shows bulk actions only when rows selected; merge requires exactly two items.

* **Style**
  * Styling added for the bulk actions UI and overlay.

* **Bug Fixes**
  * Immutable updates for row selection.
  * Email send now shows a confirmation dialog when no recipients found.

* **Tests**
  * Tests updated to cover action availability and dialog-based email/error flows.

### DIFF
--- a/src/app/features/email-client/email-client.service.spec.ts
+++ b/src/app/features/email-client/email-client.service.spec.ts
@@ -6,7 +6,6 @@ import { AlertService } from "#src/app/core/alerts/alert.service";
 import { Entity } from "#src/app/core/entity/model/entity";
 import { EmailDatatype } from "#src/app/core/basic-datatypes/string/email.datatype";
 import { DatabaseField } from "#src/app/core/entity/database-field.decorator";
-import { MarkdownModule } from "ngx-markdown";
 import { MatDialog } from "@angular/material/dialog";
 import { of } from "rxjs";
 
@@ -27,7 +26,6 @@ describe("EmailClientService", () => {
     mockDialog = jasmine.createSpyObj("MatDialog", ["open"]);
 
     TestBed.configureTestingModule({
-      imports: [MarkdownModule.forRoot()],
       providers: [
         EmailClientService,
         { provide: EntityRegistry, useValue: mockRegistry },


src/app/core/entity-details/entity-actions-menu/entity-actions-menu.component.ts
```diff
@@ -66,9 +66,10 @@ export class EntityActionsMenuComponent implements OnChanges {
       this.actions = [];
       return;
     }
-    const allActions: EntityAction[] = this.entityActionsMenuService.getActions(
-      this.entity,
-    );
+    // Pre-filter out bulk-only actions for single-entity context
+    const allActions: EntityAction[] = this.entityActionsMenuService
```
sleidig: for consistency and easier reuse, I am moving this filter to a new method `getActionsForSingle()` in the service

src/app/core/entity/entity-actions/entity-actions.service.ts
```diff
@@ -71,6 +98,39 @@ export class EntityActionsService {
...
+        icon: "edit",
+        tooltip: $localize`:entity context menu tooltip:Edit multiple records at once.`,
+        availableFor: "bulk-only",
+        permission: "update",
+        execute: async (entity: Entity) => {
+          const entities = Array.isArray(entity) ? entity : [entity];
+          if (!entities.length) return false;
+          const entityType = entities[0].getConstructor();
+          if (!entityType) return false;
+          const entityEditService = this.injector.get(EntityEditService);
```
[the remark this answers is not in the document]
sadaf895: When I am injecting the service directly in the class then it is causing circular dependency error so that's why I have used injector.

src/app/core/entity-details/entity-actions-menu/entity-actions-menu.component.ts
```diff
@@ -66,9 +66,9 @@ export class EntityActionsMenuComponent implements OnChanges {
...
     }
-    const allActions: EntityAction[] = this.entityActionsMenuService.getActions(
-      this.entity,
-    );
+
+    const allActions: EntityAction[] =
+      this.entityActionsMenuService.getActionsForSingle(this.entity);
 
     // check each action’s `visible` property to hide actions not applicable to the current entity
     const visibleActions = (
```
sleidig: as I put the visible filtering into the service for the `getActionsForBulk` we should probably also move that here into the service method ...

Do you still want to do that, @sadaf895 ?

sleidig: maybe that even means we put the filtering into `getActions()` there, so that it can be used both for single and bulk action methods without duplicating it?

sleidig (changes requested):
Sorry, found a few more functionality issues 😊 :
- [x] the merge dialog doesn't close when merge/cancel is clicked (although in the background the list leaves the bulk-action mode as expected
  - [x] the same happens with the "Anonymize?" confirmation dialog
- [x] let's switch "You need to select exactly two records for merge" from an alert message to a confirmation dialog (with a single "OK" button), so that people are sure to notice it
  - [x] same for "Send Email" message
- [x] we used to show how many records are currently selected. Can we do this in the bulk-actions component again? (it's helpful to understand if only one page or the whole list across pagination is selected, for example)

The code looks great to me now, however 👍

src/app/features/email-client/email-client.service.spec.ts
```diff
@@ -6,6 +6,7 @@ import { AlertService } from "#src/app/core/alerts/alert.service";
 import { Entity } from "#src/app/core/entity/model/entity";
 import { EmailDatatype } from "#src/app/core/basic-datatypes/string/email.datatype";
 import { DatabaseField } from "#src/app/core/entity/database-field.decorator";
+import { MarkdownModule } from "ngx-markdown";
```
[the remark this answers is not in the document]
sleidig: @Abhinegi2 , is this indeed necessary?

sleidig (approved):
Functionality looks good now 🙏

src/app/features/email-client/email-client.service.ts
```diff
@@ -43,9 +43,22 @@ export class EmailClientService {
       this.getEmailsForEntities(entityList);
 
     if (!recipients.length) {
-      this.alertService.addWarning(
-        $localize`Please fill an email address for this record to use this functionality.`,
-      );
+      await this.dialog
```
sleidig: @Abhinegi2 , now that this is awaiting a dialog response, the test needs to mock the user reaction to close the dialog, otherwise it will never reach the `return false` during the test

PR #3428, opened 2025-11-04
 Admin UI should not store internal default fields' schema in config attributes 

### LINKED ISSUE #2926: Admin UI should not store internal default fields' schema in config attributes
When editing an Entity Type, the Admin UI stores all schema fields of that entity into the `Config:CONFIG_ENTITY` doc (as `"entity:<ENTITY_TYPE>": { "attributes": { ... } }`). At the moment, this also adds internal default fields there, that are defined in the code base in Entity class. This is bloating up the config doc (and maybe even causing issues if we update schema details in the code base).

All of these should not be added to the config doc automatically but only if the user explicitly adds/overwrites them (if that is even possible).

The following are examples:
```
        "_id": {
          "anonymize": "retain",
          "dataType": "string"
        },
        "_rev": {
          "anonymize": "retain",
          "dataType": "string"
        },
        "created": {
          "anonymize": "retain",
          "dataType": "update-metadata"
        },
        "updated": {
          "anonymize": "retain",
          "dataType": "update-metadata"
        },
        "inactive": {
          "anonymize": "retain",
          "dataType": "boolean"
        },
        "anonymized": {
          "anonymize": "retain",
          "dataType": "boolean"
        },
```

---
- [x] write unit test: when saving admin entity details, expect the saved config not to contain these fields
  - make sure this test is currently failing (before fixing this issue) to confirm that the test is implemented properly
- [x] implement some logic to filter out these fields

Possible approaches could be
- somehow flag these field in the Entity class and filter all fields that have that flag
- use a hard-coded list of fields that we don't save to config?
- ...

Abhinegi2:
- closes #2926

### PR Checklist

-----
### Remaining TODOs:
- [ ] unit tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Entity configuration saving no longer includes internal default fields (IDs, revisions, timestamps, inactive/anonymized flags).

* **New Features**
  * Added an "internal" flag for schema fields so admin forms, imports, field pickers, anonymization lists, and bulk-edit selectors omit system/internal fields.
  * Bulk edit now uses a dedicated field selector component for choosing fields.

* **Tests**
  * Added/updated tests to ensure internal fields are excluded and user-facing fields are persisted.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/app/core/entity/entity-actions/entity-bulk-edit/entity-bulk-edit.component.ts
+++ b/src/app/core/entity/entity-actions/entity-bulk-edit/entity-bulk-edit.component.ts
@@ -15,7 +15,6 @@ import {
 } from "@angular/material/dialog";
 import { MatFormFieldModule } from "@angular/material/form-field";
 import { MatInputModule } from "@angular/material/input";
-import { MatSelectModule } from "@angular/material/select";
 import { MatTooltipModule } from "@angular/material/tooltip";
 import { FontAwesomeModule } from "@fortawesome/angular-fontawesome";
 import { DialogCloseComponent } from "app/core/common-components/dialog-close/dialog-close.component";
@@ -36,7 +35,6 @@ import { EntityFieldSelectComponent } from "#src/app/core/entity/entity-field-se
     FontAwesomeModule,
     MatTooltipModule,
     MatFormFieldModule,
-    MatSelectModule,
     EntityFieldEditComponent,
     EntityFieldSelectComponent,
   ],
@@ -52,11 +50,6 @@ export class EntityBulkEditComponent<E extends Entity> implements OnInit {
 
   selectedFieldFormControl: FormControl;
   fieldValueForm: EntityForm<E>;
-
-  /**
-   * The available fields of the entity, from which the user can choose.
-   */
-  entityFields: Array<{ key: string; label: string; field: any }> = [];
 
   entityData: E;
   showValueForm: boolean = false;


src/app/core/admin/admin-entity/admin-entity.component.spec.ts
```diff
@@ -178,4 +178,35 @@ describe("AdminEntityComponent", () => {
...
+    // Verify internal fields are NOT saved
+    expect(savedAttributes["_id"]).toBeUndefined();
+    expect(savedAttributes["_rev"]).toBeUndefined();
+    expect(savedAttributes["created"]).toBeUndefined();
+    expect(savedAttributes["updated"]).toBeUndefined();
+    expect(savedAttributes["inactive"]).toBeUndefined();
+    expect(savedAttributes["anonymized"]).toBeUndefined();
+
+    // cleanup:
+    AdminTestEntity.schema.delete("userField");
```
Abhinegi2:
```suggestion
    AdminTestEntity.schema.delete(" name");
```

Abhinegi2:
> Reviewing this I realize that we are hiding these internal fields already in the admin form builder. What condition do we use there? No label? I think that condition should be the same.
> 
> To make it easier to follow, I now also think that a flag in EntitySchemaField (`isInternalField`, similar to the `isInternalEntity`) will be more scalable. What do you say, @Abhinegi2 ?

@sleidig Coderabbit suggested that we have some fields without labels configured, so I’m not entirely sure about using the “no label” approach.
note.ts:95 – childrenAttendance: user-defined field with anonymize: "retain" and no label
note.ts:152 – relatesTo: user-defined field with anonymize: "retain" and no label

but, I think we can add an isInternalField flag and use that instead. For now, we can apply this flag only to these fields: ['_id', '_rev', 'created', 'updated', 'inactive', 'anonymized']

PR #3435, opened 2025-11-07
Import "update existing" matching seems broken 

### LINKED ISSUE #3434: Import "update existing" matching seems broken
The matching to existing records is not working correctly:

if the column used to identify them is not mapped in step 3, all rows are updating some single record?

Image

Image

in dev system with [Children_no_projectnumber.csv](https://github.com/user-attachments/files/23393588/Children_no_projectnumber.csv)

also, the matching doesn't seem reliable (sometimes matching the wrong record in the system, especially if that has the identifying field undefined but maybe also other cases).

---

Expected:
- if the identifier field is not mapped, no row will match and all rows will create new records
- all problems matching to the correct records are solved 🙈

Abhinegi2:
- closes #3434

### PR Checklist

-----
### Remaining TODOs:
- [ ] unit tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated record matching logic for import updates to evaluate only mapped and selected identifier fields. Records missing mapped identifiers are no longer incorrectly matched.

* **Tests**
  * Updated test expectations to validate new matching behavior and verify proper handling of incomplete records.

### DIFF
--- a/src/app/core/import/update-existing/import-existing.service.spec.ts
+++ b/src/app/core/import/update-existing/import-existing.service.spec.ts
@@ -88,7 +88,7 @@ describe("ImportExistingService", () => {
 
     let expectedEntities: any[] = [
       {
-        _rev: existingRecords[0]["_rev"], // matched by name + dateOfBirth
+        _rev: existingRecords[0]["_rev"], // matched by name AND category (both have values)
         _id: existingRecords[0]["_id"],
         name: "A",
         category: genders[1].id,
@@ -101,14 +101,12 @@ describe("ImportExistingService", () => {
         category: genders[1].id,
       },
       {
-        _rev: existingRecords[1]["_rev"], // matched by name, entity missing dateOfBirth
-        _id: existingRecords[1]["_id"],
+        _id: jasmine.any(String), // not matched (existing B has no category, so can't match)
         name: "B",
         category: "O1",
       },
       {
-        _rev: existingRecords[2]["_rev"], // matched by name, imported row missing dateOfBirth
-        _id: existingRecords[2]["_id"],
+        _id: jasmine.any(String), // not matched (import row C has no category, so can't match)
         name: "C",
       },
       {
--- a/src/app/core/import/update-existing/import-existing.service.ts
+++ b/src/app/core/import/update-existing/import-existing.service.ts
@@ -114,13 +114,13 @@ export class ImportExistingService {
         );
         const rawImportValue = rawImportEntity[idField];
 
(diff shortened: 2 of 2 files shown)


sleidig (changes requested):
There are still problems with mappings:

- use the "no_projectnumber" csv
- map the "more fields can be configured" column to Project Number field (it does contain the project numbers actually)
- see the preview: various records are mapped to the same existing record wrongly

Image

sleidig:
> Root Cause: The partial matching logic `!importEntity.hasOwnProperty(idField) || !e.hasOwnProperty(idField) `allows any row with null/undefined Project Number to match any existing record that also has null/undefined Project Number, causing multiple import rows to update the same entity. I’ve removed that logic, but one test related to it is now failing; should we remove that test case as well since it was validating the previous (incorrect) behavior?

The dev system with this latest PR still behaves unexpected: How can it correctly map records where the project number is empty to update them, if I select "Project Number" as the identifier field?
image

Abhinegi2:
@sleidig  When "Project Number" was used as the identifier but the CSV had empty/null values, all rows were shown as updating existing records instead of creating new ones.
This happened because the old logic treated null == null as a valid match, so every row with an empty Project Number matched the same existing record.
The previous matching logic allowed records with missing identifier values (null/undefined) to be considered equal.

Fix:
I updated the logic to:
- Match only on fields that are mapped as identifiers
- Require non-null values for a valid match

I made this change after reviewing the Chatwoot import feature documentation; please check if this approach aligns with the intended behavior. (Sorry for the confusion; I wasn’t fully familiar with how the import feature worked before and edge cases)

Abhinegi2:
@sleidig the issue which we noticed during childschoolenrollment import for siblings record  is:

When entity reference fields (e.g., Name→Child, Last Name→Child) didn't find matching entities, rows were completely omitted if other fields were also null/empty. This happened because `hasMappedProperty = false`, causing the row to be filtered out.
Also created a separate PR for this -https://github.com/Aam-Digital/ndb-core/pull/3444, and now unmatched rows appear in the preview, allowing create new entities

sleidig:
> When entity reference fields (e.g., Name→Child, Last Name→Child) didn't find matching entities, rows were completely omitted if other fields were also null/empty. This happened because `hasMappedProperty = false`, causing the row to be filtered out. Also created a separate PR for this -#3444, and now unmatched rows appear in the preview, allowing create new entities

So this happened even when some other columns have been mapped for the omitted row, right @Abhinegi2 ?

sleidig (approved):
Looks ok and works for our client. We'll further improve on this with #3443

PR #3471, opened 2025-11-28
doc: add sample prompt for SQL query tasks with Copilot

sleidig:
Sample to easily generate SQL queries for our Reports using Copilot.

## Summary by CodeRabbit

* **Documentation**
  * Added a new subsection on using Copilot to generate SQL queries for reports, with sample prompts, context-extraction guidance, and examples showing both formatted and single-line SQL output.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "@angular/animations": "^20.3.14",
         "@angular/cdk": "^20.2.14",
-        "@angular/common": "^20.3.13",
+        "@angular/common": "^20.3.14",
         "@angular/compiler": "^20.3.14",
         "@angular/core": "^20.3.14",
         "@angular/fire": "^20.0.1",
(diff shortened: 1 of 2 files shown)


doc/compodoc_sources/how-to-guides/reports.md
```diff
@@ -16,6 +16,38 @@ SQS enables the possibility to create SQL based queries in reports.
...
 
+## Using Copilot to generate SQL Queries
+
+```json
+// copy the full Config:CONFIG_ENTITY doc from the database
+// and use the following code to extract the entities as context for Copilot
+entities = Object.entries(x.data).filter(([k, v]) => k.startsWith("entity:")).map(([k, v]) => v);
+```
+
+Use the following prompt sample:
```
Abhinegi2: maybe we can also add/mention that this is only for the v1, and later we can add a snippet for v2 as well?

PR #3605, opened 2026-01-22
feat: Import with hard-coded prefilled field

### LINKED ISSUE #3465: Import with hard-coded prefilled field to link to another record
_As a user I want to easily import items related to a certain parent item, regardless of whether the entity reference field is in the structure of my imported data or the related data's structure._

For example:
From a Child's profile, I would like to click "Import School Enrollments for this Child" and then use the Import Module's "additional" in step 2 to pre-select the childId:

Image

This is currently only available if the Child (not the imported School Enrollment) record hold the entity field, pointing to the imported records. In this example, the structure is the other way around.

(It is possible to import this reference in step 3 by mapping some column of the import data that holds a key to the child. But this requires more preparation from the user outside the system. And it is confusing for users why some linking can be done in step 2 and other only in step 3)

---
Very specifically as a test case:
The system should offer an Advanced Import Action when importing School Enrollment. This should let the user select one Child from the dropdown to which all records will be linked/prefilled.
(For this example, because School Enrollment entities also have a schoolId field, another Import Action would be available to do the same to linked to a fixed school, also)

sadaf895:
- closes #3465

### PR Checklist

-----
### Remaining TODOs:
- [ ] unit tests

### DIFF
--- a/src/app/core/import/additional-actions/import-additional.service.ts
+++ b/src/app/core/import/additional-actions/import-additional.service.ts
@@ -291,7 +291,7 @@ export class ImportAdditionalService {
     action: AdditonalDirectLinkAction,
   ) {
     const targetEntity = await this.entityMapper.load(
-      action.targetType,
+      action.targetEntityType,
       action.targetId,
     );
     const ids = entitiesToBeLinked.map((e) => e.getId());
@@ -314,7 +314,7 @@ export class ImportAdditionalService {
     action: AdditonalDirectLinkAction,
   ) {
     const targetEntity = await this.entityMapper.load(
-      action.targetType,
+      action.targetEntityType,
       action.targetId,
     );
     targetEntity[action.targetProperty] = targetEntity[


src/app/core/import/additional-actions/additional-import-action.ts
```diff
@@ -78,3 +79,26 @@ export interface AdditionalIndirectLinkAction extends AdditionalImportBaseAction
...
+
+  /**
+   * EntityType of the entity being referenced (the type of entity selected for prefill)
+   */
+  targetType: string;
+
+  /**
+   * ID of the entity to use as the prefilled value
+   */
+  targetId?: string;
```
Abhinegi2:
```suggestion
  targetEntityType: string;

  /**
   * ID of the entity to use as the prefilled value
   */
  targetEntityId?: string;
```

How about these names?

Abhinegi2 (changes requested):
code looks clear, just a few cleanups. Also, @sadaf895, could you please add some unit tests for the newly created component? 

and also be great to add some JSDoc for any methods that  are complex and not obvious

sleidig (changes requested):
From functional testing:

I think the UI should have the usual mat-form-field styles for the additional actions selection also
Image

Can we keep the styling from the inheritance UI please (i.e. bold for part of the text in each option, "field" clearly indicating what is what). Or is there a reason to do it differently now?
Image

Did we loose some of the special Actions (that are not the new, static prefill logic), like "Link imported Students to a Partner (through School Enrollments)"? That created new School Enrollment records for each imported row, which thereby linked the Student records to a fixed partner record.
Do we still offer all the "Additional Actions" that we previously had + new ones? Or are some of the previous options not available anymore?

I tried this with the following scenario but the linking did not happen:
1. Import "Note"
2. Additional Action: "Link to Student via children" + selected one existing student
3. Column Mappings: currently no indication that the "Children" field is getting prefilled ⚠️ 
4. Imported notes have an empty "children" field

PR #3886, opened 2026-04-23
 routing separated for feature modules from dynamic config 

### LINKED ISSUE #3884: routing separated for feature modules from dynamic config
We have a few mostly hard-wired feature modules that need to register view configs dynamically, to be reachable:
https://github.com/Aam-Digital/ndb-core/blob/8ab755e02720afa0b33e6893db86d94eafea88c9/src/app/core/config/config.service.ts#L159-L163

Some of these (like Import and Deduplication) do not really belong in this customized config part, in my opinion. They would be better off as hard-wired feature module routes, similar to the `/admin` routing.

With that, there is currently some danger of conflicting custom entities with their routes and those hard-wired routes, I think.

Therefore, the following suggestions:

1. Do fixed feature-module routing for routes that are not customized (--> import, deduplication, what else ...?)
2. put all config-based routes for entities behind a pre-fix (e.g. `/c/my-entity-route`), so that there is no risk of conflicting routes

Abhinegi2:
- closes #3884

### PR Checklist

-----
### Remaining TODOs:
- [ ] unit tests

## Summary by CodeRabbit

* **New Features**
  * Added Import and Deduplication navigation entries (auth-protected) with unsaved-change protection.

* **Refactor**
  * Entity links and navigation now resolve runtime-prefixed entity routes (e.g., /c/...), updating how menu items, dialogs, searches, widgets, and navigation compute routes.
  * Routing generation and migration logic updated to normalize and namespace entity view paths.

* **Chores**
  * Removed static Import view entries from base configs.

* **Tests**
  * Updated and added specs covering routing, migrations, menu/link rewriting, and deduplication guards.


Abhinegi2:
@coderabbitai review

src/app/core/entity/entity-config.service.ts
```diff
@@ -203,4 +204,39 @@ export class EntityConfigService {
       EntityConfigService.getListViewId(entityType),
     );
   }
+
+  /**
+   * Resolve the route to be used in runtime navigation.
+   * Entity routes configured in dynamic config are namespaced under `/c/...`.
+   */
+  getRuntimeRoute(entityType: EntityConstructor): string {
+    const fallbackRoute = entityType.route;
```
sleidig: Actually, what I was thinking initially was not a difference between entity paths that are in config and entity paths that are in code, but just making all entity paths prefixed will other special feature paths get their own root path/module.

@Abhinegi2 , do you think there is an additional benefit of the distinction here within EntityConfigService?

src/app/app.routing.ts
```diff
@@ -63,6 +65,24 @@ export const allRoutes: Routes = [
...
+  },
+  {
+    path: "review-duplicates",
+    component: RoutedViewComponent,
+    canActivate: [AuthGuard, EntityPermissionGuard],
+    canDeactivate: [() => inject(UnsavedChangesService).checkUnsavedChanges()],
+    data: {
+      component: "ReviewDuplicates",
+    },
+  },
```
sleidig, line 82: Let's make this a whole module routing for the feature. (might get additional views in the future)

```
  {
    path: "deduplication",
    canActivate: [AuthGuard],
    children: DeduplicationModule.routes,
  },
```

src/app/core/entity/entity-config.service.ts
Abhinegi2: i guess for now distinction doesn't add real value, so now using same prefixed for all entities path

src/app/core/entity/entity-config.service.ts
```diff
@@ -203,4 +207,20 @@ export class EntityConfigService {
       EntityConfigService.getListViewId(entityType),
     );
   }
+
+  /**
+   * Resolve the route to be used in runtime navigation.
+   * All entity routes are namespaced under `/c/...`.
+   */
+  getRuntimeRoute(entityType: EntityConstructor): string {
+    return `/${CONFIG_ENTITY_ROUTE_PREFIX}/${normalizeRoutePath(entityType.route)}`;
```
sleidig: if this is a fixed prefix now, prbably enough to export it as a plain function, so that we do not need to inject the service in so many places, @Abhinegi2 ?

src/app/app.routing.ts
sleidig: @Abhinegi2 , do we really need the redirect here? As this is a new, hardly used feature, maybe let's just make a clean cut to new structure?

```
{
    path: "review-duplicates",
    redirectTo: "deduplication/review-duplicates",
    pathMatch: "full",
  },
```

src/app/core/entity/entity-config.service.ts
Abhinegi2: hmm make sense, let me make these changes

src/app/app.routing.ts
Abhinegi2: done please check now

src/app/core/basic-datatypes/entity/entity-block/entity-block.component.ts
```diff
@@ -94,6 +94,9 @@ export class EntityBlockComponent {
       return;
     }
 
-    this.router.navigate([entity.getConstructor().route, entity.getId(true)]);
+    this.router.navigate([
+      getEntityRuntimeRoute(entity.getConstructor()),
+      entity.getId(true),
```
sleidig, line 99: Could simplify these calls a little bit by handling this inside the method? (i.e. implement two logics, if an instance is passed, build the whole path including ID in getEntityRuntimeRoute)


```suggestion
      getEntityRuntimeRoute(entity),
```

src/app/core/config/dynamic-routing/route-paths.ts
```diff
@@ -0,0 +1,98 @@
...
+ */
+export const CONFIG_ENTITY_ROUTE_PREFIX = "c";
+
+/**
+ * Fixed feature routes that must never be replaced by dynamic config.
+ */
+export const RESERVED_FIXED_FEATURE_ROUTE_PATHS = [
+  "import",
+  "deduplication",
+] as const;
```
sleidig: Is this necessary if all user-configured routes are hidden behind the `/c/` prefix anyway?

src/app/core/config/dynamic-routing/route-paths.ts
```diff
@@ -0,0 +1,98 @@
...
+const RESERVED_FIXED_FEATURE_ROUTE_SET = new Set<string>(
+  RESERVED_FIXED_FEATURE_ROUTE_PATHS,
+);
+
+interface RuntimeViewPathOptions {
+  /**
+   * When enabled, entity views are moved under `CONFIG_ENTITY_ROUTE_PREFIX`.
+   */
+  prefixEntityRoutes?: boolean;
+}
```
sleidig: let's always do this without an option, to keep code and config simple

src/app/core/config/dynamic-routing/route-paths.ts
```diff
@@ -0,0 +1,98 @@
...
+export interface RuntimeViewPath {
+  /**
+   * Final path to register in Angular router config.
+   */
+  path: string;
+  /**
+   * Optional legacy path (without prefix) for backward-compatible redirects.
+   */
+  legacyPath?: string;
+}
```
sleidig: maybe instead we can write a config migration that scans for links (I guess only navigation and dashboard can even have such routes defined in custom) and replaces them.

That allows us to keep the legacy handling out of any other part of the code base.

src/app/core/entity-list/entity-list/entity-list.component.html
sleidig: unrelated, but let's move this action above "Configure Data Structure", so that the Admin UI item is always at the bottom.

src/app/features/de-duplication/de-duplication-module.spec.ts
```diff
@@ -0,0 +1,53 @@
+import { TestBed } from "@angular/core/testing";
+import { DeDuplicationModule } from "./de-duplication-module";
+import { RoutedViewComponent } from "app/core/ui/routed-view/routed-view.component";
+import { UnsavedChangesService } from "app/core/entity-details/form/unsaved-changes.service";
+import { MockedTestingModule } from "app/utils/mocked-testing.module";
+
+describe("DeDuplicationModule routes", () => {
```
sleidig, line 6: where does the de-duplication list currently have anything that it saves?

src/app/core/basic-datatypes/entity/entity-block/entity-block.component.ts
sleidig: actually, much less (=never) reused than I though, so never mind

src/app/core/config/config.service.ts
```diff
@@ -657,3 +673,48 @@ const migrateNotesManagerComponent: ConfigMigration = (key, configPart) => {
...
+}
+
+function migrateNavMenuItems(items: any[]): any[] {
+  return items.map((item) => {
+    if (item.link) {
+      item.link = migrateNavMenuLink(item.link);
+    }
+    if (item.subMenu) {
+      item.subMenu = migrateNavMenuItems(item.subMenu);
+    }
```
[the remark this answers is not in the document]
Abhinegi2: I don't think we have those cases where subMenu is not an array

src/app/features/de-duplication/de-duplication-module.spec.ts
Abhinegi2: removed, its not needed now

Abhinegi2:
@sleidig could you check the changes now and let me know if everything looks good, or if you had something else in mind? We can connect to discuss further if needed.

src/app/core/config/dynamic-routing/route-paths.ts
```diff
@@ -0,0 +1,59 @@
...
+ */
+export function getRuntimePathFromViewConfig(
+  view: ViewConfig,
+): RuntimeViewPath {
+  const rawPath = getViewPathFromConfigId(view._id);
+  if (!isEntityViewConfig(view) || !rawPath) {
+    return { path: rawPath };
+  }
+
+  return { path: getEntityRuntimePath(rawPath) };
```
sleidig: any reason why not to simply return a string?

src/app/core/config/config.service.spec.ts
```diff
@@ -659,6 +659,98 @@ describe("ConfigService", () => {
...
+      updateSubject.next({ entity: config, type: "update" });
+      await vi.advanceTimersByTimeAsync(0);
+
+      expect(service.getConfig("view:import")).toBeUndefined();
+      expect(service.getConfig("view:review-duplicates")).toBeUndefined();
+    } finally {
+      vi.useRealTimers();
+    }
+  });
+
```
sleidig: migration removed, no need for a test to keep this up
```suggestion
```

src/app/core/config/config.service.ts
```diff
@@ -5,8 +5,6 @@ import { shareReplay } from "rxjs/operators";
 import { addDefaultNoteDetailsConfig } from "../../child-dev-project/notes/add-default-note-views";
 import { addDefaultTodoViews } from "../../features/todos/add-default-todo-views";
 import { migrateInheritedFieldConfig } from "../../features/inherited-field/inherited-field-config-migration";
-import { addDefaultImportViewConfig } from "../import/add-default-import-view";
-import { addDefaultReviewDuplicatesViewConfig } from "../../features/de-duplication/add-default-review-duplicates-view";
```
sleidig, line 9: did we delete these files?

src/app/core/config/config.service.ts
```diff
@@ -657,3 +673,48 @@ const migrateNotesManagerComponent: ConfigMigration = (key, configPart) => {
 
   return configPart;
 };
+
+const FIXED_ROUTE_PREFIXES = [
```
sleidig: we also sometimes have /help or /hilfe routes, also /matching or other custom stuff ...

Not trivial to migrate then, I guess ... 🤔 
We could potentially look up if the route in config points to an entity list/details component. But that's not really easy either.

in that case we could migrate to the menu-item format that defined `{ entityType: "Child" }` instead of giving the actual route

sleidig (changes requested):
Sorry .. one last concern regarding migrations 😢

src/app/core/config/config.service.spec.ts
sleidig: maybe related: /attendance route in menu now goes to `c/attendance` which is a 404

and extra challenge, we nested some entity lists under this feature module path:
`/c/attendance/recurring-activity`

-> /attendance is a feature, should not get prefixed
-> /attendance/recurring-activity should then :question: ... also not get prefixed at all? or get prefixed in between, like `/attendance/c/recurring-activity`?

maybe for nested routes we could/should build the path along the submenu relationships? But that will make things more complicated ...

src/app/core/config/config.service.ts
sleidig: configPart.hasKey("navigationMenu) (i.e. migrate while having the full config object with all parts, including view configs)

then check all menu entries and for each:
scan the relevant view:... entry in configPart
check if it is an EntityList or EntityDetails ... Todo / Note ... (maybe check if there is a config.entityType property?)
only then, migrate (replace with the `entityType` instead of `link` property)

sleidig: and we replace attendance/recurring-activity with just /recurring-activity route in our template configs

sleidig (approved):
Looks good to me. Thanks!

sleidig (changes requested):
Oh, one more gap:

Dashboard shortcuts widget needs to get the migration also: The "Add Child" is leading into nowhere, currently.

sleidig (approved):

PR #3894, opened 2026-04-28
Import: Limit/throttle address lookups and warn/prevent user from large imports 

### LINKED ISSUE #3637: Import: Limit/throttle address lookups and warn/prevent user from large imports
**Describe the bug**
The Nominatim address lookup API only allows a limited number of requests. During imports that include a `location` field, we put all imported records through the API to get initial coordinates. This can (a) slow down an import significantly and (b) result in crashing the import, when the API starts rejecting requests with a 502 Gateway Error.

--> Integrate something in the UI to prevent imports with location lookups that are too large for those limits.

Abhinegi2:
- closes #3637

### PR Checklist

-----
### Remaining TODOs:
- [ ] unit tests

## Release Notes

* **New Features**
  * Added warnings for large address imports that may be slow or fail.
  * Implemented request caching and sequential queuing for address lookups to improve performance.
  * Auto-skips address lookup when dataset size exceeds optimal thresholds.

* **Bug Fixes**
  * Improved error handling for address lookup failures; imports now continue with text-only results.

* **Style**
  * Added warning-themed styling to hint components.


src/app/features/location/location-import-config/location-import-config.component.html
```diff
@@ -1,4 +1,12 @@
 <mat-dialog-content>
+  @if (rowCount > LOOKUP_WARNING_THRESHOLD) {
+    <app-hint-box i18n>
```
sleidig: two grey boxes look a bit weird. Maybe the warning box should be a red / orange box?
(do we have any sample for that? Maybe we should add an input option to app-hint-box to switch between some modes ...)

Image

And maybe we rephrase the existing (second) box and make the visible checkbox text a bit longer, like "Skip address lookup and only import text", and then move the rest of the explanation to a (?) tooltip?

src/app/features/location/location-import-config/location-import-config.component.html
```diff
@@ -1,4 +1,12 @@
 <mat-dialog-content>
+  @if (rowCount > LOOKUP_WARNING_THRESHOLD) {
```
sleidig, line 2: The main overview doesn't indicate that skip is enabled, unless I click on "Configure value mapping" to open this dialog. It would be good if we can already show an ⚠️  icon or something immediately on the column mapping view.

sleidig: or is this only applied after opening the dialog anyway?

src/app/features/location/location-import-config/location-import-config.component.ts
```diff
@@ -37,11 +37,15 @@ export class LocationImportConfigComponent {
   private readonly dialog = inject<MatDialogRef<any>>(MatDialogRef);
 
   skipAddressLookup = new FormControl(false);
+  readonly rowCount = this.data.totalRowCount;
+  readonly LOOKUP_WARNING_THRESHOLD = 30;
```
sleidig:
```suggestion
  readonly LOOKUP_WARNING_THRESHOLD = 50;
```

src/app/features/location/geo.service.spec.ts
```diff
@@ -147,4 +147,25 @@ describe("GeoService", () => {
...
+
+    service.lookup(term).subscribe();
+    expect(mockHttp.get).toHaveBeenCalledTimes(1);
+
+    service.lookup(term).subscribe();
+    expect(mockHttp.get).toHaveBeenCalledTimes(1);
+  });
+
+  it("should return empty array when HTTP lookup fails instead of throwing", () => {
+    mockHttp.get.mockReturnValue(throwError(() => new Error("502")));
```
sleidig: are we still able to show an error/warning to users in the location edit dialogs of normal forms with this?

sleidig (approved):
Both code and functionality looks good! Great work, @Abhinegi2

src/app/features/location/location-import-config/location-import-config.component.ts
sadaf895: I have question here should the warning be based on unique address count instead of total rows? (The deduplication already happens in [uniqueValues] during mapping.)

sadaf895 (commented):
I just have one question @Abhinegi2 , otherwise looks good to me

src/app/features/location/location-import-config/location-import-config.component.ts
Abhinegi2: hmm,  technically accurate (caching means only unique values make API calls)
But confusing-  user uploads 100 rows and sees "only 15 unique addresses";  they don't know about caching, they just see a number that doesn't match their CSV

@sleidig what you think

sleidig: good question. If the delay and lookups are actually reduced, I think telling users that (unless it is very difficult to calculate at this point) is probably good. The idea of "different" or unique addresses shouldn't be extremely hard to understand

src/app/features/location/location-import-config/location-import-config.component.ts
```diff
@@ -37,11 +39,15 @@ export class LocationImportConfigComponent {
...
+  readonly LOOKUP_WARNING_THRESHOLD = 50;
 
   constructor() {
     const additional = this.data.col.additional as LocationImportConfig;
-    if (additional?.skipAddressLookup) {
+    if (additional?.skipAddressLookup !== undefined) {
       this.skipAddressLookup.setValue(additional.skipAddressLookup);
+    } else if (this.rowCount > this.LOOKUP_WARNING_THRESHOLD) {
+      this.skipAddressLookup.setValue(true);
     }
```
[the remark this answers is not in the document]
Abhinegi2: its intended

src/app/core/common-components/hint-box/hint-box.component.scss
```diff
@@ -7,3 +7,8 @@
   border-radius: 4px;
   @include mat-elevation.elevation(1);
 }
+
+:host([type="warning"]) .hint-banner {
```
sleidig, line 11: making a conditional class in the template (e.g. `[class.warning-mode]="type === 'warning'"`) would have been a bit more explicit and easy to follow, I think. But this works also 😊

src/app/core/import/import-column-mapping/edit-import-column-mapping/edit-import-column-mapping.component.ts
```diff
@@ -95,6 +96,23 @@ export class EditImportColumnMappingComponent implements OnChanges {
...
+      return false;
+    const uniqueAddressCount = new Set(
+      this.rawData.map((row) => row[this.columnMapping.column]),
+    ).size;
+    if (uniqueAddressCount <= 50) return false;
+    const additional = this.columnMapping.additional as {
+      skipAddressLookup?: boolean;
+    };
+    return additional?.skipAddressLookup !== true;
+  });
```
sleidig, line 114: we should really find a way to generalize / integrate this in a decoupled way. This component shouldn't hold any knowledge of datatype-specific logic and checks ...

sleidig (approved):

PR #3846, opened 2026-04-02
chore: trigger CodeRabbit review only via 'Ready for AI Review' label

Abhinegi2:
Replace default auto-review behaviour (fires on every push when PR is not draft) with a label-based trigger. CodeRabbit now only runs when the label 'Ready for AI Review' is present, works in draft mode too, and stops auto-reviewing on new commits once the label is removed.

## Summary by CodeRabbit

* **New Features**
  * Automated PR reviews now run for draft PRs and will add a "Final Review" label to trigger a visual/AI review.
  * Added an E2E-only test mode to run end-to-end checks independently of code-quality and unit tests.
  * New label-triggered review workflow runs visual review and screenshot uploads when "Final Review" is applied.

* **Documentation**
  * README and PR template updated with guidance on using labels to control automated reviews.

PR #3901, opened 2026-04-29
chore: gate E2E and CodeRabbit behind 'Final Review' label, split QA workflows

Abhinegi2:
Previously `_qa.yaml` bundled unit tests, code quality, and E2E together. This caused:
- **E2E running twice** — once on every push (via `pull-request-updated`) and again when the label was applied (via `label-triggered-review`), because nobody removed E2E from the normal flow when the label workflow was added
- **Confusing flags** — `skip-e2e`, `e2e-only`, `upload-e2e-screenshots` were needed to control what ran where, making the intent hard to follow


Abhinegi2:
In the previous PR #3846, `_qa.yaml` bundled unit tests, code quality, and E2E together. To control what ran where, we had to pass flags like `skip-e2e`, `e2e-only`, and `upload-e2e-screenshots` ,  which made the intent hard to follow.

This caused a concrete problem: when the `Ready for Review` label was applied, E2E ran **twice** — once from `pull-request-updated.yml` (normal push flow) and once from `label-triggered-review.yml` (label flow) — because nobody removed E2E from the normal flow when the label workflow was added.

By splitting into two focused reusable workflows:
- `_qa-unit.yaml` — unit tests + code quality only
- `_qa-e2e.yaml` — E2E + Argos only

Each caller now does exactly one thing with no flags needed:
- `pull-request-updated.yml` → `_qa-unit.yaml` (runs on every push)
- `label-triggered-review.yml` → `_qa-e2e.yaml` (runs only on `Final Review` label)

No duplicate E2E, no confusing flags, clear separation of responsibilities.

sleidig (approved):
Let's try this one :-)

PR #3919, opened 2026-05-05
Fix/login flow resilience phase2

### LINKED ISSUE #3916: Login sometimes endlessly loading
Users observe the "Checking login ..." to endlessly load without progress or exit.

## Analysis

> Login flow analysis — root causes & improvements
> I traced [login.component.ts](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html), [session-manager.service.ts](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html), and [keycloak-auth.service.ts](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html).
> 
> Likely root cause of "spinner spins forever"
> OnPush change detection vs. manual property mutation. [login.component.ts](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) uses [ChangeDetectionStrategy.OnPush](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) but updates view-bound fields from async callbacks without ever calling markForCheck() or routing them through signals / async pipe:
> 
> [ssoCheckDone = true](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is set inside [sessionManager.checkRemoteSession().then(...)](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (constructor, line ~118).
> [loginInProgress](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is set inside a manual [loginState.subscribe(...)](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (ngOnInit, line ~125).
> [enableOfflineLogin = true](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is set inside the [race(...).subscribe(...)](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (line ~134).
> [showOfflineSection = true](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is set in [tryLogin()](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) — that one is fine because it's an event handler.
> Initial render captures [loginInProgress = false → true](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (state is set to [IN_PROGRESS](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) in the constructor, the BehaviorSubject re-emits it synchronously inside the first CD pass). After that, every later state transition silently mutates the fields but the view is never re-checked. So when Keycloak rejects (e.g., 504 on the token endpoint, or check-sso iframe timeout), [loginState](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) becomes [LOGIN_FAILED](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [ssoCheckDone](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) becomes true in memory — the user just keeps staring at the indeterminate <mat-progress-bar> because Angular never re-renders.
> 
> This single bug explains the "stuck spinner" symptom regardless of the underlying network problem.
> 
> Other concrete issues
> No timeout around [keycloak.init()](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) / [updateToken()](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html). A 504 from Keycloak (or proxy in front of it) can leave fetches hanging until the browser gives up. [keycloak-js](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) only times out the silent-SSO iframe (~10 s) — it does not time out direct token-endpoint fetches. We should wrap both with a [Promise.race(..., timer(N))](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and treat timeout as [RemoteLoginNotAvailableError](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html).
> 
> 504 isn't recognised as "network unavailable". [keycloak-auth.service.ts](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) [NETWORK_ERROR_PATTERNS](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) only matches Failed to fetch / NetworkError / iframe timeout. A 504 from Keycloak typically surfaces as a non-OK response and is logged via [Logging.error("Keycloak init failed", err)](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) — Sentry noise plus user gets no graceful fallback. Add detection of 5xx responses (and AbortError/timeout) and map them to [RemoteLoginNotAvailableError](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html).
> 
> [OnTokenExpired](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) triggers a full Keycloak redirect. The handler calls [this.login()](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html). If the silent refresh fails, [login()](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) falls through to [keycloak.login(...)](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) which redirects away from the app, losing in-progress work. Should call [keycloak.updateToken(<minValidity>)](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (or [checkSession()](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)), and only escalate to a redirect on hard failure — and not at all if the user is mid-task.
> 
> Offline fallback is unreachable when SSO check hangs. The "Log in" button is hidden until [ssoCheckDone](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html), and the offline section is hidden until [showOfflineSection](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (only set by [!navigator.onLine](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) initially or by clicking the hidden "Log in" button). After 10 s, [enableOfflineLogin](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is flipped, but the section is still hidden. So an online-but-degraded user has no escape hatch even after 10 s. Either:
> 
> Show the "Log in" / "Continue offline" buttons after a short hard timeout regardless of [ssoCheckDone](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html), or
> Combine [enableOfflineLogin](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) and [showOfflineSection](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) so the 10 s fallback also reveals the section.
> [checkRemoteSession()](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) then-chain swallows nothing but races nothing either. If [databaseResolver.initDatabasesForSession(session)](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) hangs (e.g., remote DB handshake), [loginState](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) stays [IN_PROGRESS](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) permanently. Add a timeout / progress reporting around [initializeUser](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html), and put [ssoCheckDone = true](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) in a [finally](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)-equivalent rather than only in [.then(...)](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html).
> 
> [memoize(initKeycloak)](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) cache is only invalidated on failure. After a successful init, an expired/rotated session (e.g., user signed out elsewhere, or Keycloak realm reloaded) cannot re-init without a full reload. On a recoverable error path you may want to also clear the cache.
> 
> [reuseFirstAsync(checkSession)](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) + [reuseFirstAsync(login)](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) deduplicate but never time out. If the first call hangs, every subsequent call to [checkSession()](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) returns the same hung promise — including the auto-refresh [OnTokenExpired](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) path. Combine with #1 (timeout) so the dedup window is bounded.
> 
> [routeAfterLogin](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) uses [setTimeout(..., 0)](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) on every login. Harmless but unnecessary; [router.navigateByUrl](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) already runs asynchronously. Remove for simplicity.
> 
> [safeRedirectUrl](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) rejects the redirect silently. If the redirect fails validation it returns / with no log — fine for security, but worth a [Logging.debug(...)](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) when it triggers, otherwise this is invisible if users complain about being dropped on the wrong page after login.
> 
> [accessToken](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) on the Keycloak service is never refreshed proactively in PouchDB-sync paths. When the token expires mid-sync, requests get 401 and the only recovery is the [OnTokenExpired](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) listener calling [login()](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (see #3). For long-running sessions on flaky networks, this is the most common trigger of "user is stuck" reports. Prefer a single getValidToken() API that always calls [updateToken(30)](vscode-file://vscode-app/snap/code/237/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) first.

sleidig:
- second phase of https://github.com/Aam-Digital/ndb-core/issues/3916

## Release Notes

* **Bug Fixes**
  * Enhanced login retry logic to automatically recover from transient network failures and timeouts.
  * Session updates now include timeout protection for improved reliability.
  * Improved error detection for gateway and upstream service unavailability.

### DIFF
--- a/src/app/core/session/auth/keycloak/keycloak-auth.service.ts
+++ b/src/app/core/session/auth/keycloak/keycloak-auth.service.ts
@@ -70,11 +70,19 @@ export class KeycloakAuthService {
   checkSession = reuseFirstAsync(async (): Promise<SessionInfo | null> => {
     await this.initKeycloak();
 
-    await firstValueFrom(
-      defer(() => this.keycloak.updateToken()).pipe(
-        timeout({ each: KEYCLOAK_OPERATION_TIMEOUT_MS }),
-      ),
-    );
+    try {
+      await firstValueFrom(
+        defer(() => this.keycloak.updateToken()).pipe(
+          timeout({ each: KEYCLOAK_OPERATION_TIMEOUT_MS }),
+        ),
+      );
+    } catch (err) {
+      if (this.isOfflineOrUnavailableError(err)) {
+        Logging.debug("Keycloak updateToken failed (offline/unavailable)", err);
+        throw new RemoteLoginNotAvailableError(err);
+      }
+      throw err;
+    }
     const token = await this.keycloak.getToken();
     if (!token) {
       return null;
--- a/src/app/core/session/auth/keycloak/keycloak-auth.service.ts
+++ b/src/app/core/session/auth/keycloak/keycloak-auth.service.ts
@@ -23,12 +23,27 @@ const KEYCLOAK_OPERATION_TIMEOUT_MS = 15_000;
 /** Backoff schedule for explicit user-initiated login retries. */
 const LOGIN_RETRY_DELAYS_MS = [1_000, 3_000];
 
+/**
+ * Browser/keycloak-js error messages that indicate a transient network
+ * failure rather than a real authentication problem. Matched against both
+ * `err.message` and `err.toString()` since some thrown values (e.g. plain
+ * strings from keycloak-js) only expose the text via `toString()`.
+ */
+const NETWORK_ERROR_PATTERNS = [
+  "Failed to fetch",
+  "NetworkError",
+  "Load failed",
+  "Network request failed",
+  "Timeout when waiting for 3rd party check iframe message.",
+];
+
 function isRetryableNetworkError(err: any): boolean {
   if (!err) return false;
   if (err instanceof TimeoutError) return true;
   if (["AbortError", "TimeoutError"].includes(err?.name)) return true;
   if ([502, 503, 504].includes(err?.status)) return true;
-  return false;
+  const message = `${err?.message ?? ""} ${err?.toString?.() ?? ""}`;
+  return NETWORK_ERROR_PATTERNS.some((pattern) => message.includes(pattern));
 }
 
 /**
@@ -155,27 +170,11 @@ export class KeycloakAuthService {
     });
   });
 
-  private static readonly NETWORK_ERROR_PATTERNS = [
-    "Failed to fetch",
-    "NetworkError",
-    "Load failed",
-    "Network request failed",
-    "Timeout when waiting for 3rd party check iframe message.",
-  ];
-
   private isOfflineOrUnavailableError(err: any): boolean {
-    // All retryable network errors (5xx / timeout / abort) also count as
-    // "unavailable" so they map to RemoteLoginNotAvailableError instead of
-    // spamming Sentry.
+    // All retryable network errors (5xx / timeout / abort / fetch failures)
+    // also count as "unavailable" so they map to RemoteLoginNotAvailableError
+    // instead of spamming Sentry.
     if (isRetryableNetworkError(err)) {
-      return true;
-    }
-    const message = err?.message ?? "";
-    if (
-      KeycloakAuthService.NETWORK_ERROR_PATTERNS.some((pattern) =>
-        message.includes(pattern),
-      )
-    ) {
       return true;
     }
     return !navigator.onLine;
--- a/src/app/core/session/auth/keycloak/keycloak-auth.service.ts
+++ b/src/app/core/session/auth/keycloak/keycloak-auth.service.ts
@@ -10,6 +10,27 @@ import { KeycloakUserDto } from "../../../user/user-admin-service/keycloak-user-
 import { ActivatedRoute } from "@angular/router";
 import { ThirdPartyAuthenticationService } from "../../../../features/third-party-authentication/third-party-authentication.service";
 import { reuseFirstAsync } from "#src/app/utils/reuse-first-async";
+import { defer, firstValueFrom, throwError, timer, TimeoutError } from "rxjs";
+import { retry, timeout } from "rxjs/operators";
+
+/**
+ * Hard upper bound on individual Keycloak network operations
+ * (init, token refresh). Picked well above expected RTT but short enough
+ * that a hung proxy/upstream surfaces a clear failure to the user.
+ */
+const KEYCLOAK_OPERATION_TIMEOUT_MS = 15_000;
+
+/** Backoff schedule for explicit user-initiated login retries. */
+const LOGIN_RETRY_DELAYS_MS = [1_000, 3_000];
+
+function isRetryableNetworkError(err: any): boolean {
+  if (!err) return false;
+  if (err instanceof TimeoutError) return true;
+  if (err?.name === "AbortError" || err?.name === "TimeoutError") return true;
+  const status = err?.status;
+  if (status === 502 || status === 503 || status === 504) return true;
+  return false;
+}
 
 /**
  * Handles the remote session with keycloak
@@ -26,11 +47,20 @@ export class KeycloakAuthService {
   /**
    * Check for an existing SSO session without redirecting to Keycloak.
    * Returns the session info if already authenticated, or null if not.
+   *
+   * Intentionally does NOT retry on transient errors: this runs silently on
+   * page load and the user is already waiting on the spinner. Failing fast
+   * lets the login form render quickly; a retry happens implicitly when the
+   * user clicks "Log in".
    */
   checkSession = reuseFirstAsync(async (): Promise<SessionInfo | null> => {
     await this.initKeycloak();
 
-    await this.keycloak.updateToken();
+    await firstValueFrom(
+      defer(() => Promise.resolve(this.keycloak.updateToken())).pipe(
+        timeout({ each: KEYCLOAK_OPERATION_TIMEOUT_MS }),
+      ),
+    );
     const token = await this.keycloak.getToken();
     if (!token) {
       return null;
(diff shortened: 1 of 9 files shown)


src/app/core/session/auth/keycloak/keycloak-auth.service.ts
```diff
@@ -101,6 +167,18 @@ export class KeycloakAuthService {
...
+    if (err?.name === "AbortError" || err?.name === "TimeoutError") {
+      return true;
+    }
+    // Recognise transient gateway/upstream errors as "unavailable" so they
+    // surface as RemoteLoginNotAvailableError rather than spamming Sentry.
+    const status = err?.status;
+    if (status === 502 || status === 503 || status === 504) {
+      return true;
+    }
     const message = err?.message ?? "";
```
Abhinegi2: if (
  ["AbortError", "TimeoutError"].includes(err?.name) ||
  [502, 503, 504].includes(err?.status)
) {
  return true;
}

Maybe we can combine them?

Abhinegi2 (changes requested):

src/app/core/session/auth/keycloak/keycloak-auth.service.ts
sleidig: maybe also should be combined with "isRetryableNetworkError"?

Abhinegi2 (approved):

PR #3929, opened 2026-05-06
fix: MarkdownPage does not load entity due to OnPush change detection (#3922)

### LINKED ISSUE #3922: MarkdownPage doesn't seem to load entity
see https://dev.aam-digital.net/help

Abhinegi2:
Closes #3922

- Root cause: `MarkdownPageComponent` used `ChangeDetectionStrategy.OnPush` but stored loaded entity content in a plain class property. When `ngOnInit`'s async Promise resolved and set `this.markdownContent`, Angular's `OnPush` detector did not re-render the view — leaving "Content not found." visible even after a successful load.

## Release Notes

* **New Features**
  * Added Help menu item to the application's main navigation.
  * Help page located at `/help` displays markdown-formatted documentation and guidance.

* **Refactor**
  * Updated component state management implementation.

### DIFF
--- a/src/app/features/markdown-page/markdown-page/markdown-page.component.html
+++ b/src/app/features/markdown-page/markdown-page/markdown-page.component.html
@@ -1,6 +1,6 @@
 @if (markdownEntityId) {
-  @if (markdownContent) {
-    <div markdown [data]="markdownContent"></div>
+  @if (markdownContent()) {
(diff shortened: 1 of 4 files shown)


sleidig (approved):
Looks good to me (and works). Thanks!

We may want to add an e2e test with screenshot in addition to the unit test in the future. But the Markdown view is not widely used, so for now we are good, I think.

PR #3962, opened 2026-05-12
Fixes and sentry noise reduction

sleidig:
we still have a lot of sentry alerts popping up. Trying to fix remaining bugs and reduce the noise of our logging.

See individual commit messages for details

### DIFF
--- a/src/app/core/common-components/fa-dynamic-icon/fa-dynamic-icon.component.ts
+++ b/src/app/core/common-components/fa-dynamic-icon/fa-dynamic-icon.component.ts
@@ -42,7 +42,8 @@ export class FaDynamicIconComponent {
     if (!definition) {
       definition = FaDynamicIconComponent.fallbackIcon;
       Logging.warn(
-        `Tried to set icon "${icon}" but it does not exist as a font awesome regular item nor is it registered as an alias.`,
+        "Tried to set icon but it does not exist as a font awesome regular item nor is it registered as an alias.",
+        icon,
       );
     }
     return definition;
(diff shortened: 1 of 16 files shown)


sleidig:
@coderabbitai review

src/app/core/common-components/fa-dynamic-icon/fa-dynamic-icon.component.ts
```diff
@@ -42,7 +42,8 @@ export class FaDynamicIconComponent {
     if (!definition) {
       definition = FaDynamicIconComponent.fallbackIcon;
       Logging.warn(
-        `Tried to set icon "${icon}" but it does not exist as a font awesome regular item nor is it registered as an alias.`,
```
Abhinegi2, line 45: Why did we remove "${icon}"? Wouldn’t it be helpful to keep it so we can also know which icon was set?

Abhinegi2 (commented):
Looks good overall, just one question

src/app/core/common-components/fa-dynamic-icon/fa-dynamic-icon.component.ts
sleidig: we are still logging the exact icon name as a "context" in the second arg to the logging method. That way the primary message is the same for any mismatched icon (and automatically grouped into one sentry issue + easy to filter) while that additional context is still shown in the console and in sentry's "extra" details.

Abhinegi2 (approved):

PR #4018, opened 2026-05-28
fix(Public Forms): fix multiple bugs in public form admin Configure Pre-filled Values tab

### LINKED ISSUE #4003: Public Form admin: required entity not committing, prefilled value blank, tabs lock without user edits
**Describe the bug**

While testing PR #3999, three issues were observed in the Public Form admin UI (originally [posted as a comment](https://github.com/Aam-Digital/ndb-core/issues/3999#issuecomment-3261812149) but unrelated to that PR; reproducible on current `master`).

1. **New Public Form — must "capture" the entity twice for Save to work and for other tabs to enable.** After selecting the `Record` (entity type) and clicking Save, the form silently does not save. The user has to re-select the entity for save to actually persist and for the other tabs to become enabled.
2. **Existing Public Form — default value is blank on the "Configure Pre-filled Values" tab.** When opening an existing form, an already-configured prefilled value shows the mode toggle buttons but the value area is empty / wrong.
3. **Existing Public Form — tabs lock after switching to "Configure Pre-filled Values" without user edits.** After making (or no) changes on "General Setting" and switching to "Configure Pre-filled Values", the user can no longer move across tabs, and clicking the back arrow triggers the "Discard Changes?" popup even though they didn't edit anything on this tab.

**To Reproduce**

_Issue 1 (new form):_

1. Open `/c/admin/public-form/new`
2. Fill in Form Link ID, Title, Description, and pick a value for `Record` (e.g. `Child`)
3. Click `Save` — nothing happens; the `Record` field shows "This field is required" even though a value is selected, and other tabs remain disabled
4. Re-select `Record` and click `Save` again — now it saves and the other tabs enable

_Issue 2 (prefilled blank):_

1. Open an existing Public Form that already has at least one prefilled value with mode `dynamic` (or `inherited-field`)
2. Go to the `Configure Pre-filled Values` tab
3. The default-value editor for that field renders blank / shows the wrong sub-editor (static instead of the saved mode)

_Issue 3 (tabs lock without edits):_

1. Open an existing Public Form that has any prefilled value configured
2. Click `Edit`, switch to the `Configure Pre-filled Values` tab without changing anything
3. Try to switch back to `General Setting` — tabs are disabled
4. Click the back arrow — `Discard Changes?` popup appears even though nothing was edited

**Expected behavior**

1. The first `Save` after selecting the entity should persist the form. The `Record` field should not show a "required" error after a value has been selected from the dropdown.
2. The previously saved default value (in its correct mode) should be displayed when opening the form.
3. Simply viewing/switching to the "Configure Pre-filled Values" tab without editing should not mark the form as having unsaved changes.

**Screenshots**

_Issue 1:_

Image

_Issue 2:_

Image

_Issue 3:_

Image

Image

**Desktop**

- OS: macOS
- Browser: Chrome
- Build: current `master` (HEAD at `8f915f747`)

**Additional context — suspected root causes**

After tracing through the code:

- **Issue 1**: The `entity` field on `PublicFormConfig` declares `validators: { required: true, readonlyAfterSet: true }`. `readonlyAfterSet` is implemented as an **async** validator (`buildReadonlyValidator` in [`readonly-after-set.validator.ts`](https://github.com/Aam-Digital/ndb-core/blob/master/src/app/core/common-components/entity-form/dynamic-form-validators/readonly-after-set.validator.ts)). In `dynamic-validators.service.ts` (around L181-183), when any async validator is present the FormControl is built with `updateOn: \"blur\"`. So the value selected in the entity-type autocomplete is not committed to the FormControl until blur fires. On the first Save click the form still sees the entity as empty → `checkFormValidity` throws `InvalidFormFieldError` (silently swallowed in `form.component.ts saveClicked()`) → nothing happens. A second selection (after the blur from the first click commits the value) lets the save succeed, which then flips `entity.isNew` to false and enables the other mat-tabs (which are gated by `entity()?.isNew || unsavedChanges.pending()`).

- **Issue 2**: Likely a regression from PR #3978 (commit `228c72d75`, signal refactor). `AdminDefaultValueComponent` now exposes `selectedMode = toSignal(this.modeControl.valueChanges, { initialValue: this.modeControl.value })`. But `syncFormWithValue` — called from `writeValue` when loading an existing entity — uses `this.form.setValue(newFormValue, { emitEvent: false })`. `emitEvent: false` suppresses child `valueChanges`, so `modeControl.valueChanges` never fires for the initial load → `selectedMode()` stays at its `initialValue` of `null`. The template `@switch (selectedMode())` then falls into `@default` (the static editor) regardless of the saved mode, and any non-static default value (e.g. dynamic `$now`, inherited-field) renders with the wrong sub-component and shows up blank. The existing spec at `admin-default-value.component.spec.ts` only asserts `form.get(\"mode\").value`, not `selectedMode()`, which is why this slipped through.

- **Issue 3**: Likely a downstream consequence of Issue 2. When the prefilled tab renders an existing non-static value through the wrong (static) sub-component, the value goes through `valueToEntityFormat` / `valueToDatabaseFormat` and the `setTimeout(() => formControl.updateValueAndValidity(), 0)` in `AdminDefaultValueStaticComponent`'s constructor emits a `valueChanges`. If the round-trip transforms the value (e.g. a date schema can't parse `$now`), the new (different) value propagates up through `value` setter → `writeValue` → `onChange` chain all the way to `EditPrefilledValuesComponent.updateFieldGroups`, which calls `this.formControl.setValue(updatedFields); this.formControl.markAsDirty();`. That `markAsDirty()` flips `unsavedChanges.pending()` to true via the subscription in `entity-form.service.ts` — so the tab disable predicate `entity()?.isNew || unsavedChanges.pending()` evaluates to true and all tabs lock, even though the user did not edit anything. Back navigation then triggers the discard confirmation.

### LINKED ISSUE #4014: public form "route" field should not allow spaces or special characters

Abhinegi2:
- closes #4003
- closes #4014

### PR Checklist

- [ ] manually tested (all) required functionality described in the issue manually yourself on the final version of the code (developer)
  - please also avoid changes that are not directly related to the issue of the PR, even small code reformatting changes make the review process much more complex
  - once you are done with your initial work, change to "Ready for Review". This will trigger some additional automated checks.
- [ ] add label **"Final Review"** to trigger Argos visual review and CodeRabbit AI review (can also be done while still in Draft)

---

### Remaining TODOs:

- [ ] unit tests

## Summary by CodeRabbit

* **Bug Fixes & Improvements**
  * Stronger URL/id validation for public forms (allow only letters, digits, hyphens, underscores).
  * Form label display and disabled-state handling improved for more consistent UX across editors.
  * Prefilled/default value validation tightened to prevent incomplete values and avoid unintended updates.
  * Read-only behavior refined so fields stay disabled once set.

* **Tests**
  * Added and expanded unit tests covering validators, default-value behavior, and prefilled-values component state.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Aam-Digital/ndb-core/pull/4018?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


src/app/features/public-form/edit-prefilled-values/edit-prefilled-values.component.ts
```diff
@@ -32,6 +36,19 @@ import { EntitySchemaField } from "app/core/entity/schema/entity-schema-field";
...
+): ValidationErrors | null => {
+  const val: DefaultValueConfig = control.value;
+  if (!val) {
+    return { incompleteDefaultValue: true };
+  }
+  if (val.mode && !val.config) {
+    return { incompleteDefaultValue: true };
+  }
+  return null;
+};
```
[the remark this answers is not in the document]
Abhinegi2: Fixed. Remove button also updated to use `isDisabled()`. Validator extended to catch `config: { value: null }` case.

src/app/features/public-form/edit-publicform-route/edit-publicform-route.component.ts
```diff
@@ -27,10 +27,10 @@ import { PublicFormConfig } from "../public-form-config";
 
 const noSpecialUrlChars: ValidatorFn = (control: AbstractControl) => {
   const value: string = control.value;
-  if (value && /[/?#%&+\\]/.test(value)) {
+  if (value && /[ /?#%&+\\]/.test(value)) {
     return {
       pattern: {
-        errorMessage: $localize`The link ID must not contain special characters like / ? # % & + \\`,
+        errorMessage: $localize`The link ID must not contain spaces or special characters like / ? # % & + \\`,
```
sleidig: I think we better go with an allow-list here, maybe a regex like `[a-z\d\-_]`?

sleidig (changes requested):
Clean, good solutions despite some challenging deeply technical issues. Nice work!

(I was able to verify all bugs fixed)

src/app/features/public-form/edit-publicform-route/edit-publicform-route.component.ts
Abhinegi2: Done. Switched to allow-list `^[a-z\d\-_]+$` with anchoring so the full string is validated. Added uppercase and other chars to the test cases.

sleidig (approved):
Thanks!

PR #4029, opened 2026-06-02
fix: try to load site name for login page with anonymous access

sleidig:
on the first visit of a system, the login page shows the default site styles and title ("Aam Digital Demo") which is confusing.

This is trying to load SiteSettings even before the usual DB init that happens after login.


sleidig:
Could you review and test this in a synced system for me, @Abhinegi2 ? :blush:

Abhinegi2:
Functionality is working as expected

Abhinegi2 (approved):
The code also looks good

PR #4046, opened 2026-06-07
DB pagination and sorting 

### LINKED ISSUE #3943: Entity List Component Filter + Paginate directly on the database

TheSlimvReal:
- closes #3943 

## Approach
The idea to implement pagination and filtering as an opt-in solution is based on datasources.
The goal would be to define in the config whether a normal or a paginated datasource should be used inside a component.
For this all the data management logic (fetch, sort, filter, paginate) needs to be moved from the component (entity-table, entity-list) to the datasource. 
The normal datasource should resemble the current status as close as possible, while the paginated one should provide the same functionality but transfer less data between the DB and the application.

## Implemented
- [puchdb-find](https://pouchdb.com/guides/mango-queries.html) has been installed
- A new paginated datasource has been implemented
- The paginated datasource supports pagination, sorting and filtering (no free-text)
- Automatically creates indices (required for sorting on columns)
- Performance improves for loading children list from ~15s to almost instantly even without an index on [load-test](https://load-test.aam-digital.net) 
- Performance for complex filters (`$elemMatch`, `$or`) is really bad (1:40 minutes for 100 elements out of 1 million), with index for some reason even worse (5-6 minutes for same query). This seems to be a [known issues](https://github.com/apache/couchdb/issues/2236#issuecomment-538928959) and won't like change any time

## Open problems

- [ ] PouchDB/CouchDB does not provide information about how many total docs match the filter. So far, one element more than necessary is requested to detect whether there are further elements to query. With this approach it is not possible to jump to the last page or show how many total elements are available. A db request with `include_docs: false` could be used to get the total number of docs
- [ ] The current concept of filter for active documents does not work on the database, because the `isActive` property is dynamic
- [ ] The `ConfigurableEnum` filter currently works in the fronted/object format, not just with the ID. In the database only the ID is needed. The frontend implementation needs to be adjusted]
- [ ] Some columns cannot be sorted properly. E.g. `ConfigurableEnums` will be sorted according to their database ID value instead of their display value in the app (which might differ, especially as the display value can be changed later). Sorting should probably be turned off for these kind of properties/columns.


sleidig:
filtering doesn't seem to work for me yet - or is it just extremely slow and keeps showing existing records until then?
(I've tried in online-only mode so far)

PR #4063, opened 2026-06-10
fix: export age and entity readable-name columns in list download (#4057)

### LINKED ISSUE #4057: Export column "age" is empty, and no human-readable columns for entity fields
**Describe the bug**
The "age" column (and possibly other special columns not part of the schema directly?) remains empty in the downloaded file.

Also, fields like "entity" datatype do not export their usual additional columns, like the human-readable name.

**To Reproduce**

1. With the new column selection dialog for export/download in the list view
2. Go to the students list in "education" demo use case
3. The "age" values are displayed in the app
4. Click "Download" and keep the columns selected including "age"
5. In the downloaded file, the age column is included but empty

**Context**
Probably there is a problem with fields that are added specifically only in the view config?

--> I think we need to change the column selection dialog to show not only fields of the entity type, but somehow load all export columns (some fields offer more than one exported column) and let the user choose among all of these.

May be also related to
- https://github.com/Aam-Digital/ndb-core/issues/3653

Abhinegi2:
- closes #4057

## PR Checklist

- [x] added label **"Final Review"** to trigger UI regression testing and CodeRabbit AI review
- [x] marked each code review comment as resolved (or commented on it with a question, if unsure)

---

### What & why

The new list-view export column dialog had two bugs:
1. The **"age"** column downloaded **empty** — age is a virtual property (`dateOfBirth.age`) with no export column, and the download went through the query path that could not resolve it.
2. **Entity fields** (e.g. school) lost their **human-readable name** column — the query path never resolved entity ids to names.

### Approach

Build the export column dialog and the download from the schema datatype **export resolvers** (`getExportColumns`) instead of the query path:
- `date-with-age` datatype now contributes a derived **age** export column.
- Entity fields keep offering their **readable-name** column (preselected by default instead of the raw id; the raw id stays selectable).
- Virtual display columns (e.g. `DisplayAge`) map to their underlying field.
- Preselected columns and export headers use the **list-view label**, so the file matches what the user sees ("Age", "School"). Label precedence: admin `exportConfig` label > list-view label > datatype label.

### Manual testing steps

1. Education demo use case → Students list.
2. Click **Download**, keep default selection → CSV/XLSX.
   - **Age** column header is "Age" and values are filled (not empty).
   - **School** column header is "School" with readable school names (no raw id by default).
3. Open the dialog → autocomplete lists all export columns incl. raw-id / "(readable)" / "(age)" variants for manual selection.
4. Deselect/Clear/Include-all and column-group presets still work; CSV and XLSX both correct.
5. Filtered vs all scope exports the right record set.
6. Regression: a list with configured `exportConfig` (e.g. RecurringActivity) still shows admin custom labels.

## Summary by CodeRabbit

* **New Features**
  * Exports now include computed age columns alongside date-with-age fields.
  * Export dialog supports passing explicit selected columns for downloads.

* **Improvements**
  * Entity reference fields present both raw and human-readable export options; readable variants are preselected when visible.
  * Column labels respect admin overrides and visible-list labels; column ordering and explicit selections are preserved during export.

* **Tests**
  * Coverage added for age export, entity references, and export selection behaviors.


src/app/core/export/download-service/download.service.ts
```diff
@@ -233,8 +250,12 @@ export class DownloadService {
       data.map((item) => this.mapEntityToExportRow(item, columnResolvers)),
     );
 
-    const columnKeys = Array.from(columnLabels.keys());
-    const headers = Array.from(columnLabels.values());
+    const columnKeys = selectedKeys
```
sadaf895: Here we filter selected columns to [columnLabels.has(key)], so non-schema query columns are dropped at export time.

Was this intentional? If not, we may need a fallback path for query-expression columns so configured exports remain backward compatible.

sadaf895 (commented):
Functionality works as expected. Just one question I have posted regarding the code.

sleidig (approved):

sadaf895 (approved):

PR #4093, opened 2026-06-22
fix: normalize enum option IDs

sleidig:
**Issue & Reproduction / Testing steps:**
Creating new enum (dropdown) options automatically generates IDs. Until now, they contained the same special characters that the label text entered by the user has. This caused some weird behavior with accents during filtering / reports. We now remove special characters from the ID of options (which are only used in the background)

--> to test:
create new dropdown options
- with some special characters (*, é, ...)
- with only special characters
...
- [ ] directly from the field by typing new text
- [ ] openening the dialog to edit options and adding new option there
- [ ] copying multiple options into the dialog add new (which creates multiple options at once)

---

## Summary

* **Bug Fixes**
  * Enum option IDs are now generated/normalized from labels (lowercase), instead of simply uppercasing input.
  * Inputs that contain no usable characters for an option ID (e.g., symbol-only lines) are now rejected consistently and won’t be added or saved.

* **New Features**
  * During batch creation, users now get a confirmation message that reports how many invalid entries were skipped, including a list of the problematic lines.

* **Tests**
  * Updated and added unit tests for normalization and invalid-entry handling.


george-neha:
Suggestion: Display the skipped entry. It will help the user understand which entry is skipped, rather than deciphering it themselves. 
Screenshot 2026-06-23 at 11 05 15 AM

src/app/core/basic-datatypes/configurable-enum/configure-enum-popup/configure-enum-popup.component.ts
```diff
@@ -283,4 +291,12 @@ export class ConfigureEnumPopupComponent {
...
     );
   }
+
+  private async showInvalidSkippedMessage(invalid: number) {
+    await this.confirmationService.getConfirmation(
+      $localize`:@@invalidOptionsSkippedTitle:Some entries skipped`,
+      $localize`:@@invalidOptionsSkipped:Skipped ${invalid} entr${invalid === 1 ? "y" : "ies"} without usable letters or numbers for ID.`,
+      OkButton,
+    );
+  }
```
Abhinegi2: I like Neha's suggestion as well. Maybe we could include that too?

https://github.com/Aam-Digital/ndb-core/pull/4093#issuecomment-4777553607

Abhinegi2 (commented):
Looks good overall. Just one suggestion, could you check this as well?

sleidig:
Thanks, @george-neha and @Abhinegi2 , good suggestion. Implemented now

PR #4102, opened 2026-06-29
fix: prevent entity context menu from breaking on PublicForm with unknown entity type (#4096)

### LINKED ISSUE #4096: Context menu of detail view doesn't open
Context menu of detail view doesn't open

Below is the error

Image

The system had a PublicForm config for an entity type that did not exist (but was configured to have a ID-param link for this entity)

sleidig:
## Summary

Fixes #4096 — the entity detail **context menu (three-dot menu) failed to open** when the system had a `PublicFormConfig` referencing an entity type that does not exist in the registry (e.g. a leftover `participantSurvey` config with an ID-param link).

## Root cause analysis

1. **Direct cause — throwing registry lookup with a dead guard.**
   `PublicFormsService.findMatchingLinkedFieldIds` called `EntityRegistry.get(formConfig.entity)` directly. `Registry.get()` **throws** when a key is not registered — it never returns `undefined` — so the existing `if (!entityConstructor) return []` guard was dead code. The thrown error escaped the action's `visible()` callback:

   ```
   Requested item is not registered in EntityRegistry. Key: participantSurvey
     at Registry.get (dynamic-registry.ts)
     at PublicFormsService.findMatchingLinkedFieldIds (public-forms.service.ts:199)
     at PublicFormsService.getRelevantFormForEntity
     at PublicFormsService.isEntityTypeLinkedToConfig
     at action.visible (entity-actions-menu.service.ts:45)
   ```

2. **Aggravating cause — no isolation of per-action failures.**
   `EntityActionsMenuService.getVisibleActions` awaited each action's `visible()` in a loop with no try/catch. A single throwing `visible()` rejected the whole `getActionsForSingle()` promise; the consuming `EntityActionsMenuComponent` uses an Angular `resource()`, so on rejection `actions` fell back to `[]` → the menu rendered empty / appeared broken.

## Fix

- **`public-forms.service.ts`** — guard the registry lookups with `has()` before `get()` (matching the existing pattern in `entity-config.service.ts`), plus a `Logging.warn` so the misconfiguration is discoverable by whoever can fix the config. Applied to both `findMatchingLinkedFieldIds` and `saveCustomFieldsToEntityConfig`.
- **`entity-actions-menu.service.ts`** — wrap each per-action `visible()` evaluation in try/catch (log `warn`, hide only that action). Defense-in-depth so any future misbehaving action can never take down the entire context menu.

## Tests (TDD)

Added tests written **red-first**, then made green:

- `public-forms.service.spec.ts` — a config referencing an unregistered entity type resolves to `false` without throwing. (The existing mock returned `undefined` from `get` instead of throwing like the real registry — which is why this bug escaped unit tests; the new test replicates the real throwing behaviour.)
- `entity-actions-menu.service.spec.ts` — an action whose `visible()` throws is hidden, and the menu still resolves with the other healthy actions.

All 20 tests in the two affected specs pass.

## Manual verification

Create a `PublicFormConfig` whose `entity` / `forms[].entity` points to a non-existent entity type (with `linkedEntities` set), open an entity detail view, and confirm the three-dot context menu opens normally (without the copy-form action) with a single warning logged instead of the previous uncaught error.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience when action visibility checks fail, so one bad action no longer breaks the full entity menu.
  * Public form fields now handle unknown or unregistered entity types more safely, preventing invalid links from being applied.
  * Added warning messages for these edge cases to help identify configuration issues.


Abhinegi2 (approved):
Working as expected, and the code also looks good to me.

While testing, I was thinking that this type of broken config can't be created through the Public Form UI and only can be introduced by updating the document directly. we might run into similar issues in the future. We should also consider whether we still want to expose  **Adding application config** option in the admin section.

sleidig:
> While testing, I was thinking that this type of broken config can't be created through the Public Form UI and only can be introduced by updating the document directly. we might run into similar issues in the future. We should also consider whether we still want to expose **Adding application config** option in the admin section.

Yes, that's true. Good discussion to raise.
Ideally we could have some consistency checks for config scans, I think. But even though it can break config, I think we do need an option to edit raw config for special, advanced requirements and AI-assisted configuration, though.

PR #4103, opened 2026-06-29
feat(reporting): warn on outdated report calculation and auto-refresh

### LINKED ISSUE #4036: Report needs to show more clearly if displaying an old (possibly outdated) report calculation
We sometimes show an existing report calculation to give the user faster results:

Image

However, the hint about that is easy to miss. And it is displayed even when the results have actually been calculated freshly right now.

Make the UI clearer:
- if the report was not calculated fresh upon selecting/clicking right now, show a warning icon and red text with something like "You are seeing a previous calculation. Data may be outdated." above the hint about recalculating

Abhinegi2:
- closes #4036

## What

When a SQL report is served from a cached server-side calculation, the previous-vs-fresh state was indistinguishable — the "calculated at … / click again to re-calculate" hint showed even for freshly computed results, and there was no signal that the data might be old.

This adds:
- A red warning **"You are seeing a previous calculation. Data may be outdated."** above the existing hint, shown only when the displayed calculation is **older than 1 minute**.
- **Automatic background re-fetch** of the calculation when it is stale: the cached result shows immediately (with an "(updating…)" indicator), a fresh calculation is recomputed in the background, and the result swaps in when ready (the warning then clears). A request token guards against a slow refresh overwriting a report the user has switched away from.

**Scope:** this only applies to **date-range SQL reports**, where the backend can serve a cached calculation that matches the requested report + date range. Reports run **without** a date filter always recompute fresh (no cache match), so they are always realtime and never show the warning. Non-SQL report modes compute client-side and are unaffected.

Client-side only (`ReportingComponent`); no service/API/data-model changes.

## Manual testing

1. Open a **date-range** SQL report (e.g. **Overview SQL v2**) for a date range that already has a calculation **older than 1 minute**.
2. Click **Calculate** → the cached result shows immediately with the red warning and "(updating…)".
3. The background recompute finishes → the warning disappears and "calculated at" updates to now.
4. Click **Calculate** again → fresh result, **no** warning.
5. Run a SQL report **without** a date filter, or a non-SQL report → no warning, recomputed fresh.

## PR Checklist

---


sadaf895 (approved):

src/app/features/reporting/reporting/reporting.component.ts
```diff
@@ -161,6 +186,7 @@ export class ReportingComponent {
     fromDate: Date,
     toDate: Date,
   ): Promise<void> {
+    const token = ++this.refreshToken;
```
sleidig, line 189: I see the danger of the endless loop now, that you mentioned.

Is there any good reason to re-calculate more than once? I think after doing one background refresh we should probably just stop and let the user click recalculate if needed.

What do you say, @Abhinegi2 ?

Abhinegi2: actully its only recalculates once. backgroundRefresh is fired a single time from calculateResults  and never re-triggers itself, so there's no loop , after the one refresh we stop and the user can click Calculate to recompute again.

++this.refreshToken isn't a retry counter; it's a race guard. It's bumped whenever a new calculation starts or the report/date-range changes, so if a slow background refresh returns after the user has switched reports, its result is discarded instead of overwriting the newly selected one. i can add a comment to make that clearer.

sleidig: ah, true. I see now. Working with observables we might have implemented a structure that cancels the subscription without the need to cache a token but this approach here should also work 👍

sleidig (approved):

Issue #3091, opened 2025-06-23
Conditional Fields: hide fields based on the value of other fields in the record


sleidig:
_As a user I want to see only those fields that are currently relevant in order to have a simpler interface and avoid errors._

Sometimes fields are only relevant in a certain context. For example, a "if selected 'other', please specify" or other detailed fields that are only relevant if a record has a certain status.
Common survey form builders allow to define conditions when a field is visible. We should add a similar functionality to our field configurations.

-----
## Implementation steps (suggested)
- [ ] extend EntitySchemaField configuration with a new property (or adapt one of the existing flags like "hideFromForm") to hold a filter condition
   - the filter condition should use the same filter format that we use for other places, like permissions or list filter values (i.e. MongoQuery syntaxt)
 - [ ] design a custom Admin Component to configure such filters in relation to other fields in the entity type
 - [ ] write unit tests to test hide/show conditions
 - [ ] implement the additional logic and integrate it in `EntityFormComponent`
 - [ ] ensure this is working for Public Forms and also when editing / viewing entity details

PR #4105, opened 2026-06-29
feat(reporting): Admin UI to manage report configs — Part A (#4085)

### LINKED ISSUE #4085: Admin UI: Report management (create and edit report configs)

While a full "report builder" (#2404) that visually assists the query building is more complex, we should provide some easier control to admins to manage reports.

### Proposed Solutions & Alternatives

- Add a list of ReportConfig entities to the "Templates and Forms" section in Admin Overview
- Add a list view showing all ReportConfig records and allowing to create new ones (similar to Public Forms, etc.)
- link to the Admin view from the Report Component context menu
- Add a details view allowing to edit some report parts more easily:
  - title, etc. can be simple EditText fields
  - the query part can be a json field
  - introduce a new "description" long-text field to ReportConfig that allows to document more details about this report. Display this on the report component (with calculated reports) also, maybe as a collapsible section above the results?

If feasible, each query string could be its own input field. Ideally we find a package that can pretty-print or even syntax highlight SQL queries, so that it automatically adds useful linebreaks (and removes them again during save). This would probably be a new custom EditComponent.

---

## Implementation Approach

TODO

sleidig:
Part A of #4085 — **Admin UI: Report management**.

Adds a config-driven admin **list + details** view to create and edit `ReportConfig` entities, modeled on the EmailTemplate admin pattern. Part B (stacked on this branch) adds a SQL syntax-highlighting editor.

### What's included
- **Discoverability:** a **Reports** entry under Admin Overview → *Templates and Forms*, plus a **Manage Reports** link in the Reporting view's context menu.
- **List + details:** an `EntityList` (Title / Mode / Description) and an `EntityDetails` form, registered in `ReportingModule` (mirrors `EmailClientServiceModule`).
- **New `description` field** (long-text) on `ReportConfig`, shown to users as a collapsible *"About this report"* panel above the results on the Reporting view.
- **Friendly field editors:**
  - `EditReportMode` — dropdown over `reporting` / `exporting` / `sql`.
  - `EditReportPeriodToggle` — a single toggle mapping the SQL `transformations` to the report period (start & end date).
  - `EditJson` — reuses the existing `JsonEditorComponent` as a form-field editor for the raw `reportDefinition` / `aggregationDefinitions`. (Part B replaces `reportDefinition`'s editor with a structured SQL editor.)

### Testing
- Unit tests for the three new edit components + `ReportConfig` statics/`description` (all green); existing reporting specs still pass.
- `npm run build` and `npm run lint` pass locally.
- E2E (`e2e/tests/reports.spec.ts`): admin-overview discoverability + list, editing a report's description via the admin view, and the description panel on the Reporting view. (Run in CI.)

Implementation plan: https://github.com/Aam-Digital/ndb-core/issues/4085#issuecomment-4831355007

## Summary by CodeRabbit

* **New Features**
  * Added a Reports admin area with direct navigation from the admin overview.
  * Report records can now show and edit descriptions in the admin UI.
  * Report pages can display an “About this report” expandable section when a description is available.
  * Added new form controls for selecting report mode and toggling report-period settings.
  * Added support for editing raw JSON fields in admin forms.

* **Bug Fixes**
  * Improved layout and wrapping for report description text.


src/app/features/reporting/reporting/reporting.component.html
sleidig:
```suggestion
```

Let's remove this old option. (need to check the ts file if other variables are then unused after deleting it here in template)

Abhinegi2:
@sleidig  While doing testing, I noticed a few things on the Report admin UI - 

- [x] We are currently showing both JSON editors in all modes. Instead, we should display the Report Definition editor only for SQL mode, and the Aggregation Definition editor only for the Exporting and Reporting modes.
- [x] For existing reports, if someone changes the mode (e.g., from Exporting to Report), the report is saved successfully, but downloading it from the Reports screen results in a broken report. We should show a warning pop-up whenever the mode is changed and reset the Aggregation Definition or Report Definition accordingly to prevent invalid configurations.
- [x] The Use Report Period option is currently not working when creating a new report. When editing an existing report, changes to this setting are not saved—it always reverts to its previous value after saving.
- [x] The JSON editors are always editable, even without clicking the Edit button. They should be read-only until the user explicitly enters edit mode
- [x] righjt now, it's possible to create a report config without filling in any fields. At a minimum, Name, Mode should be mandatory. Date and Description should remain optional. (and in 2nd pr we can add the SQL/Aggregation Definition mandatory)

We should also add Chatwoot doc links for different use cases and include them somewhere in the Reports section, as this could otherwise be confusing for users.

sleidig:
very good points, @Abhinegi2 , thanks!



> * [ ]  We are currently showing both JSON editors in all modes. Instead, we should display the Report Definition editor only for SQL mode, and the Aggregation Definition editor only for the Exporting and Reporting modes.

Agreed. In the longer run this would be a great use for "conditional fields" (https://github.com/Aam-Digital/ndb-core/issues/3091). Until we have that, I am not sure what's the best workaround. Maybe within the Edit component detect the mode and show a static "not applicable in this mode" in its template, or something?

> * [ ]  For existing reports, if someone changes the mode (e.g., from Exporting to Report), the report is saved successfully, but downloading it from the Reports screen results in a broken report. We should show a warning pop-up whenever the mode is changed and reset the Aggregation Definition or Report Definition accordingly to prevent invalid configurations.

Would it help if we refactor the `AggregationReport` and `ExportingReport` interfaces, so that they also use the `reportDefinition` property instead of `aggregationDefinitions`? Then one single Edit component could maybe handle the switching and validation internally?

> * [ ]  The Use Report Period option is currently not working when creating a new report. When editing an existing report, changes to this setting are not saved—it always reverts to its previous value after saving.
> * [ ]  The JSON editors are always editable, even without clicking the Edit button. They should be read-only until the user explicitly enters edit mode
> * [ ]  righjt now, it's possible to create a report config without filling in any fields. At a minimum, Name, Mode should be mandatory. Date and Description should remain optional. (and in 2nd pr we can add the SQL/Aggregation Definition mandatory)

Those all sound like very valid concerns and suggestions. Please go ahead with the implementation :blush: 

> We should also add Chatwoot doc links for different use cases and include them somewhere in the Reports section, as this could otherwise be confusing for users.

Yes, I'm preparing a page and sharing the link: https://chatwoot.help/hc/aam-digital/articles/1782889956-analyzing-data-through-advanced-reports

Abhinegi2:
@sleidig  For migration, I'm thinking of the same two-part pattern we already use for PublicFormConfig and the config doc -

- On-read migration (like migratePublicFormConfig) -  when a report is loaded, convert the old field to the new one on the fly. This keeps existing reports working immediately.
- One-time CLI migration (like latest-config-formats): a script runs once per environment to actually rewrite the stored docs (move aggregationDefinitions → reportDefinition). It's idempotent, so is it safe to re-run?

sleidig:
> @sleidig For migration, I'm thinking of the same two-part pattern we already use for PublicFormConfig and the config doc -
> 
> * On-read migration (like migratePublicFormConfig) -  when a report is loaded, convert the old field to the new one on the fly. This keeps existing reports working immediately.
> * One-time CLI migration (like latest-config-formats): a script runs once per environment to actually rewrite the stored docs (move aggregationDefinitions → reportDefinition). It's idempotent, so is it safe to re-run?

Sounds good :+1: 

Do we have a good place to hook in the on-read migration in the app?
Otherwise we could even consider just doing cli migration (maybe adding the new property without deleting the old one? :bulb: ) and flipping that together with deployment.

src/app/features/reporting/edit-report-mode/edit-report-mode.component.ts
```diff
@@ -0,0 +1,38 @@
...
+export class EditReportModeComponent
+  extends CustomFormControlDirective<string>
+  implements EditComponent
+{
+  formFieldConfig = input<FormFieldConfig>();
+
+  readonly modeOptions: string[] = ["reporting", "exporting", "sql"];
+  private readonly modeLabels: Record<string, string> = {
+    reporting: $localize`:ReportConfig mode option:Reporting`,
+    exporting: $localize`:ReportConfig mode option:Exporting`,
```
sleidig: are we able to display tooltips for these, explaining what their difference is? (I actually don't know myself exactly 🙈 )

(if not, maybe building a mat-select directly, without our autocomplete component is an alternative that allows us to put these directly here in the template)

src/app/features/reporting/edit-report-mode/edit-report-mode.component.ts
```diff
@@ -0,0 +1,38 @@
...
+    sql: $localize`:ReportConfig mode option:SQL`,
```
sleidig: We should show a warning using the `FeatureDisabledInfoComponent` if SQL is selected but feature not enabled. (Maybe showing that instead of just disabling the dropdown option gives more options to the user, like getting details on how to enable the feature)

src/app/features/reporting/edit-report-period-toggle/edit-report-period-toggle.component.ts
```diff
@@ -0,0 +1,71 @@
...
+  formFieldConfig = input<FormFieldConfig>();
+
+  /** the canonical transformation object enabling the report period (start & end date) */
+  static readonly REPORT_PERIOD_TRANSFORMATION: Transformations = {
+    startDate: ["SQL_FROM_DATE"],
+    endDate: ["SQL_TO_DATE"],
+  };
+
+  /** the toggle is "on" when the report-period (start & end date) transformation is present */
+  readonly checked = computed<boolean>(() => {
```
sleidig: for non-sql reports the date range is implied from the query, I think? At least there are reports like the education demo "Event Report" that allow to select dates but show a turned off toggle. Maybe we can hide the whole thing for modes other than SQL or somehow make the UI more consistent here?

src/app/features/reporting/edit-report-period-toggle/edit-report-period-toggle.component.ts
```diff
@@ -0,0 +1,115 @@
...
+  changeDetection: ChangeDetectionStrategy.OnPush,
+  imports: [ReactiveFormsModule, MatSlideToggleModule],
+  providers: [
+    {
+      provide: MatFormFieldControl,
+      useExisting: EditReportPeriodToggleComponent,
+    },
+  ],
+})
+export class EditReportPeriodToggleComponent
```
sleidig, line 45: seeing this UX further, I think we could actually automate this in the UI completely: If placeholders are used in the queries, then we need to add the transformations - otherwise not.

Maybe just display it read-only as an information to the user?

And if we do, ideally also cover the old (non-sql) reports so that the toggle also reflects if such a report will show a date select or not.

Otherwise UX looks good to me overall :+1:

sleidig: @Abhinegi2 , this does _not_ work correctly yet. "Basic Report" has no `?` placeholders but shows the toggle as active.
(although it also shows the date picker in the report, so the detection if report period is relevant may already have been broken or missing before)

e2e/tests/reports.spec.ts
```diff
@@ -66,3 +66,115 @@ test("Generate a configured aggregation report and download CSV", async ({
...
+
+  // The new collapsible "About this report" panel appears above the results.
+  await expect(
+    page.getByRole("button", { name: /About this report/ }),
+  ).toBeVisible({ timeout: 10_000 });
+  await page.getByRole("button", { name: /About this report/ }).click();
+  await expect(page.getByText(description)).toBeVisible();
+
+  await argosScreenshot(page, "report-description");
+});
```
sleidig: let's include that in the standard report scenario (set up a sample report already including description and just have the normal reports screenshot include the description part also)

src/app/core/admin/json-editor/edit-json/edit-json.component.ts
```diff
@@ -0,0 +1,30 @@
...
+import { DynamicComponent } from "../../../config/dynamic-components/dynamic-component.decorator";
+import { EditComponent } from "../../../entity/entity-field-edit/dynamic-edit/edit-component.interface";
+import { JsonEditorComponent } from "../json-editor.component";
+
+/**
+ * Edit a field's value as raw JSON, using the shared {@link JsonEditorComponent}
+ * as a form-field edit component.
+ *
+ * Reuses the existing vanilla-jsoneditor wrapper (including its tree-mode fallback for
+ * very long strings, see #3821) by binding it through the standard `[formControl]`.
```
sleidig:
```suggestion
```

Clean Code best practice: comments should not state how things are implemented (read the code for that), only why or at an abstract external level

src/app/features/reporting/edit-report-mode/edit-report-mode.component.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  /** short explanation of each mode, shown as a tooltip on the dropdown option */
+  readonly modeDescriptions: Record<string, string> = {
+    reporting: $localize`:ReportConfig mode description:Aggregated statistics (counts, groupings) calculated in the browser. No server-side reporting backend required.`,
+    exporting: $localize`:ReportConfig mode description:A flat export of the matching records. No server-side reporting backend required.`,
+    sql: $localize`:ReportConfig mode description:Custom server-side SQL queries for advanced reports. Requires the reporting backend to be enabled.`,
+  };
+
+  readonly modeToString = (mode: string): string =>
+    this.modeLabels[mode] ?? mode;
```
sleidig: why not just hard-code the three options in the html template to avoid all those `$localize` tags and variables?

src/app/features/reporting/sql-report/sql-report.service.ts
```diff
@@ -53,6 +54,30 @@ export class SqlReportService {
...
   static readonly API_URL = environment.API_PROXY_PREFIX + "/v1/reporting";
 
+  private _isReportingBackendEnabledPromise: Promise<boolean> | undefined;
+
+  /**
+   * Whether the server-side reporting backend (SQS) required for "sql" reports is available,
+   * as reported by the `/actuator/features` endpoint (`reporting` feature flag).
+   * "reporting" and "exporting" reports are calculated in-browser and do not need this.
+   */
+  isReportingBackendEnabled(): Promise<boolean> {
```
sleidig, line 64: we should maybe also show a feature-disabled component on the reporting.component view? 🤔

sleidig (commented):
code looks good overall 👍 

Just a few last details

src/app/features/reporting/edit-report-period-toggle/edit-report-period-toggle.component.ts
sleidig: Thanks for fixing this :blush:

e2e/tests/reports.spec.ts
sleidig, line 25: add `reportConfig.description = "This report is created for e2e testing ...";`
so that the screenshot includes that UI element also

e2e/tests/reports.spec.ts
```diff
@@ -36,12 +36,7 @@ test("Generate a configured aggregation report and download CSV", async ({
   await page.getByRole("combobox", { name: /Select Report/i }).click();
   await page.getByRole("option", { name: "E2E Basic Report" }).click();
 
-  // The aggregation report shows a date range selector and Calculate is
-  // disabled until a range is set. Fill the date range.
-  await page.getByRole("textbox", { name: "Start date" }).fill("01.01.2024");
-  await page.getByRole("textbox", { name: "End date" }).fill("31.12.2025");
-  await page.getByRole("textbox", { name: "End date" }).blur();
```
sleidig, line 43: do we still have an e2e screenshot that includes the time range picker now?

PR #4109, opened 2026-06-30
fix: filter by enum option IDs that contain a comma (#4104)

### LINKED ISSUE #4104: enum filter with "," in ID cannot correctly filter
**Describe the bug**
A enum that includes `,` in its ID cannot correctly filter.

e.g. see https://dev.aam-digital.net/c/child?center=option,%20with,%20comma
(select the "option, with, comma" value for the "Center" filter in this example)

This is a legacy issue because we do not allow special characters in IDs anymore.
However, we need to fix this for backwards compatibility.

Abhinegi2:
- closes #4104

## Summary

A configurable-enum option whose **ID contains a comma** (legacy data, e.g. `OPTION, WITH, COMMA`) could not be filtered in list views. List-view filter values are stored in the URL joined by `,` and parsed back with `split(",")`, so the comma inside the ID collided with the multi-value separator: the ID was split into fragments that matched no option, and the filter silently matched nothing (showing all records).

Legacy IDs originate from before PR #4093 (which only normalizes _new_ IDs and ships no migration), so they must keep working for backwards compatibility.

## Fix

`FilterComponent.loadUrlParams` now resolves the raw URL value against the filter's known option keys instead of blindly splitting on `,`:

1. Filters without a known option set (e.g. date filters) keep the legacy `split(",")`.
2. If the whole value matches a valid option key, it is kept as a single value (fixes single comma-IDs, incl. dashboard / menu navigation links).
3. Otherwise tokens are greedily resolved to the longest matching option key; tokens that match no key are preserved as-is, so stale/unknown values from old URLs behave exactly as before.

The URL format is unchanged (`toString()` → explicit `join(",")`), so existing bookmarks and links keep working. No data migration.

## Manual testing

> A legacy comma-containing center option is pre-seeded in the demo data on this PR's preview deployment (temp commit, see note below), so no setup is needed.

1. Open the preview deployment and go to the **Children** list.
2. Open the **Center** filter dropdown — note the option `option, with, comma (legacy test fixture)` (its ID literally contains commas).
3. Select it.
   - **Before this fix:** the filter chip shows `, ,` and the list is **not** filtered (all records shown).
   - **With this fix:** the list is filtered to only children with that center, the chip shows the option as a single value, and the URL becomes `?center=option, with, comma`.
4. Reload the page (and/or open `…/c/child?center=option, with, comma` directly) → the option is still selected and the list stays correctly filtered (it round-trips as one value, not split on the commas).
5. Regression check — these still behave as before: a normal center value (e.g. Alipore), selecting multiple values, clearing the filter, and the date-range filter.

Also covered by new + existing unit tests in `filter.component.spec.ts`.

> ⚠️ **Note:** commit `afda61f7e` ("TEMP legacy comma center option for PR test deploy") only adds the demo fixture above so this is testable on the preview deployment. **It will be reverted before merge** — it is not part of the fix.

## PR Checklist

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed URL-based filter selection so option IDs containing commas are preserved exactly when sharing and reloading links.
  * Added coverage for complex query-param cases, including overlapping comma-containing option IDs.
  * Updated filter-related sample data and enum fixtures to include comma-containing option values for better real-world testing.

### DIFF
--- a/src/app/core/filter/filter/filter.component.spec.ts
+++ b/src/app/core/filter/filter/filter.component.spec.ts
@@ -137,19 +137,23 @@ describe("FilterComponent", () => {
     expect(component.filterSelections()[0].selectedOptionValues[1]).toBe("bar");
   });
 
-  it("should load url param with a comma inside an option key as a single value", async () => {
-    // legacy enum/string option ids can contain a comma (e.g. "option, with, comma").
-    // such an id must round-trip as ONE selected value, not be split on the comma.
-    const e1 = TestEntity.create({ other: "option, with, comma" });
-    const e2 = TestEntity.create({ other: "Delhi" });
+  it("should parse url params for option keys that contain commas (#4104)", async () => {
+    // Legacy enum/string option ids can contain a comma. Selected ids are
+    // joined by "," in the url, so a comma inside an id must not be mistaken
+    // for the multi-value separator. Here keys "A" and "B,C" are both valid;
+    // a naive greedy longest-match would parse "A,B,C" as ["A,B","C"] and drop
+    // the valid "B,C" - the values must instead resolve to ["A", "B,C"].
+    const e1 = TestEntity.create({ other: "A" });
+    const e2 = TestEntity.create({ other: "A,B" });
+    const e3 = TestEntity.create({ other: "B,C" });
 
     activatedRouteMock.snapshot = {
-      queryParams: { other: "option, with, comma" },
+      queryParams: { other: "A,B,C" },
     };
 
     await setComponentInputs({
       entityType: TestEntity,
-      entities: [e1, e2],
+      entities: [e1, e2, e3],
       useUrlQueryParams: true,
       filterConfig: [{ id: "other" }],
     });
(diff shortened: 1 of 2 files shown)


src/assets/base-configs/all-features/configurable-enums.json
```diff
@@ -16,6 +16,11 @@
         "id": "C3",
         "label": "Barabazar",
         "color": "#94DDFF"
+      },
+      {
+        "id": "option, with, comma",
+        "label": "option, with, comma (legacy test fixture)",
+        "color": "#FFD8B1"
```
Abhinegi2: need to revert these after testing

src/app/child-dev-project/children/demo-data-generators/fixtures/centers.ts
```diff
@@ -6,4 +6,6 @@ export const centersUnique: ConfigurableEnumValue[] = enums.find(
 ).values;
 
 // multiple entries for the same value increase its probability
-export const centersWithProbability = [0, 0, 1, 2].map((i) => centersUnique[i]);
+export const centersWithProbability = [0, 0, 1, 2, 3, 3].map(
+  (i) => centersUnique[i],
+);
```
Abhinegi2: this one also

sadaf895 (approved):
Looks good to me!

src/app/core/filter/filter/filter.component.spec.ts
```diff
@@ -137,6 +137,78 @@ describe("FilterComponent", () => {
...
+      entities: [e1, e2, e3],
+      useUrlQueryParams: true,
+      filterConfig: [{ id: "other" }],
+    });
+
+    const otherFilter = component
+      .filterSelections()
+      .find((f) => f.name === "other");
+    expect(otherFilter.selectedOptionValues).toEqual(["A", "B,C"]);
+  });
```
sleidig: I personally prefer to limit verbose unit tests 😊 

Maybe we could just combine all these cases into a single unit test? (I think the last one already inherently covers the other two, probably?)

sleidig (approved):
can you clean up the temporary configs for final merge for us, @Abhinegi2 ?

src/app/core/filter/filter/filter.component.spec.ts
Abhinegi2: done, combined them into a single test

PR #4111, opened 2026-06-30
feat: allow manual editing of individual address parts

### LINKED ISSUE #4065: Allow manually editing individual address parts (road, city, postcode, …)
## User requirement

> As a user editing a location, I want to see and correct the individual address parts (road, house number, postcode, city, country) — not just the full address text — so that I can fix mistakes the automatic address lookup got wrong (e.g. a missing house number or a wrong city) and have correct structured data for documents and exports.

When I change a part and it no longer matches the displayed address text, the app should ask me whether to also update that text — the same way it already asks when a map click changes the mapped location.

## Background

PR #4006 added structured address fields — `road`, `house_number`, `postcode`, `city`, `country` — to the `GeoLocation` interface (`src/app/features/location/geo-location.ts`) for PDF/document generation. These parts are currently **only auto-derived** from the API lookup via `enrichGeoLocation()` and there is **no UI to view or correct them**. When the address lookup mis-parses an address, the user cannot fix the individual parts.

## Proposed implementation

Parts stay as independent structured fields on `GeoLocation` (consistent with `enrichGeoLocation` preserving manual values via `??`, and `mapClicked` dropping stale parts on a new lookup). The free-text `locationString` is not auto-rewritten silently; instead we confirm with the user on save.

1. **Part inputs** in `AddressEditComponent` (hosted in the map dialog): a compact, collapsible "Address details" section with inputs for `road`, `house_number`, `postcode`, `city`, `country`, bound to `selectedLocation()`. A single `updateAddressPart()` handler spreads the current value and sets the edited field (so siblings + `locationString` are preserved).
2. **Compose helper** `composeAddressFromParts()` in `GeoService`, reusing the existing `formatStreet` / `formatPostcodeCity` formatters.
3. **Confirm on save** in `MapPopupComponent`: when edited parts diverge from `locationString`, reuse the existing "Address Mismatch" dialog pattern — **Keep current address text** vs **Update address text to match details** — mirroring the current map-click flow.

### Behavior notes

- Picking a new search result / map click / GPS resets manual part edits (by design — `mapClicked` deliberately drops stale parts).
- No data migration needed: existing stored locations get parts derived on load via `transformToObjectFormat` → `enrichGeoLocation`.

### Files

- `src/app/features/location/address-edit/address-edit.component.{ts,html,scss}`
- `src/app/features/location/geo.service.ts`
- `src/app/features/location/map-popup/map-popup.component.ts`

sadaf895:
- closes https://github.com/Aam-Digital/ndb-core/issues/4065

## PR Checklist

---

## Summary by CodeRabbit

* **New Features**
  * Improved address editing with separate fields for road, house number, postcode, city, and country.
  * Added clearer address display, including a fallback when no location is selected.
  * Added a confirmation prompt when saving edited address parts if the free-text address and structured address no longer match.

* **Bug Fixes**
  * Better preserves address details while editing individual parts.
  * Adds clearer mismatch hints to help spot inconsistencies before saving.

### DIFF
--- a/src/app/features/location/address-edit/address-edit.component.scss
+++ b/src/app/features/location/address-edit/address-edit.component.scss
@@ -79,7 +79,6 @@
 
 .parts-mismatch-hint {
   color: colors.$warn;
-  font-size: 0.85rem;
 }
 
 .panel-footer {
--- a/src/app/features/location/address-edit/address-edit.component.scss
+++ b/src/app/features/location/address-edit/address-edit.component.scss
@@ -31,8 +31,6 @@
 
   .location-label {
     display: block;
-    font-size: 0.75rem;
-    color: colors.$muted;
   }
 
   .location-value {
@@ -41,11 +39,6 @@
     .placeholder {
       color: colors.$muted;
     }
-  }
-
-  .location-hint {
-    font-size: 0.75rem;
-    color: colors.$muted;
   }
 }
 
--- a/src/app/features/location/address-edit/address-edit.component.scss
+++ b/src/app/features/location/address-edit/address-edit.component.scss
@@ -1,4 +1,5 @@
 @use "variables/breakpoints";
+@use "variables/colors";
 
 .address-edit-container {
   display: flex;
@@ -23,4 +24,72 @@
 .address-input {
   flex: 1;
   min-width: 0;
+}
+
+.selected-location-display {
+  flex: 1 1 100%;
+
+  .location-label {
+    display: block;
+    font-size: 0.75rem;
+    color: colors.$muted;
+  }
+
+  .location-value {
+    white-space: pre-line;
+
+    .placeholder {
+      color: colors.$muted;
+    }
+  }
+
+  .location-hint {
+    font-size: 0.75rem;
+    color: colors.$muted;
+  }
+}
+
+.address-details-panel {
+  flex: 1 1 100%;
+}
+
+.address-parts-grid {
+  display: grid;
+  // 3-column base: Road gets 2/3, House no. 1/3; Postcode 1/3, City 2/3
+  grid-template-columns: 2fr 1fr 2fr;
+  gap: 0 16px;
+
+  @media (max-width: breakpoints.$md) {
+    grid-template-columns: 1fr;
+  }
+
+  .road-field {
+    grid-column: 1 / 3;
+  }
+
+  .house-number-field {
+    grid-column: 3 / 4;
+  }
+
+  .postcode-field {
+    grid-column: 1 / 2;
+  }
+
+  .city-field {
+    grid-column: 2 / 4;
+  }
+
+  .full-row {
+    grid-column: 1 / -1;
+  }
+}
+
+.parts-mismatch-hint {
+  color: colors.$warn;
+  font-size: 0.85rem;
+}
+
+.panel-footer {
+  display: flex;
+  justify-content: flex-end;
 }
--- a/src/app/features/location/address-edit/address-edit.component.scss
+++ b/src/app/features/location/address-edit/address-edit.component.scss
@@ -80,8 +80,3 @@
 .parts-mismatch-hint {
   color: colors.$warn;
 }
-
-.panel-footer {
-  display: flex;
-  justify-content: flex-end;
-}
--- a/src/app/features/location/address-edit/address-edit.component.ts
+++ b/src/app/features/location/address-edit/address-edit.component.ts
@@ -3,6 +3,7 @@ import {
   ElementRef,
   input,
   model,
+  output,
   viewChild,
   ChangeDetectionStrategy,
   inject,
@@ -43,6 +44,9 @@ export class AddressEditComponent {
    */
   selectedLocation = model<GeoLocation>();
 
+  /** Emitted when the user explicitly edits a structured address part field. */
+  readonly partEdited = output<void>();
+
   /**
    * Whether the search box is enabled and visible.
    */
@@ -82,6 +86,7 @@ export class AddressEditComponent {
     value: string,
   ) {
     this.partsEdited = true;
+    this.partEdited.emit();
     this.updateLocation({
       locationString: this.selectedLocation()?.locationString,
       geoLookup: this.selectedLocation()?.geoLookup,
--- a/src/app/features/location/address-edit/address-edit.component.ts
+++ b/src/app/features/location/address-edit/address-edit.component.ts
@@ -71,7 +71,9 @@ export class AddressEditComponent {
    * location. Inferred (no flag needed): a meaningful manual edit always makes
    * the text diverge from both derivations, so this is self-detecting.
    */
-  private isTextManuallyOverwritten(location: GeoLocation | undefined): boolean {
+  private isTextManuallyOverwritten(
+    location: GeoLocation | undefined,
+  ): boolean {
     const text = location?.locationString?.trim();
     if (!text) {
       return false;
--- a/src/app/features/location/address-edit/address-edit.component.ts
+++ b/src/app/features/location/address-edit/address-edit.component.ts
@@ -3,7 +3,6 @@ import {
   ElementRef,
   input,
   model,
-  signal,
   viewChild,
   ChangeDetectionStrategy,
   inject,
@@ -14,6 +13,7 @@ import { GeoLocation } from "../geo-location";
 import { MatFormField, MatHint, MatLabel } from "@angular/material/form-field";
 import { MatInput } from "@angular/material/input";
 import { MatTooltip } from "@angular/material/tooltip";
+import { MatExpansionModule } from "@angular/material/expansion";
 import { AddressGpsLocationComponent } from "../address-gps-location/address-gps-location.component";
 
 /**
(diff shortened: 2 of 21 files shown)


Abhinegi2:
In the address details panel, when someone edits the Full address text, I was assuming it would trigger the nominatim search and also update the mapped location. Right now it only overwrites the text and the map location stays the same. Is that intended? If yes, maybe we should either run a search from here too, or make this field read-only/disabled so it's not confusing that typing here doesn't move the pin?

src/app/features/location/map-popup/map-popup.component.ts
Abhinegi2: `partsJustEdited` is set to true when a part is edited, but is only reset in `saveAndClose()`. So if a user edits a field (e.g. city) and then clicks a new spot on the map without saving, the flag is set to true. On save, the normal address-mismatch dialog gets suppressed (!this.partsJustEdited) and the parts dialog shows instead — telling the user "You changed the address details" even though they just clicked the map. Can we reset partsJustEdited here in mapClicked?

src/app/features/location/address-edit/address-edit.component.ts
```diff
@@ -43,33 +44,61 @@ export class AddressEditComponent {
...
   disabled = input<boolean>(false);
 
-  manualAddressEnabled = signal(false);
-
   private readonly geoService = inject(GeoService);
 
-  enableManualAddressEditing() {
-    this.manualAddressEnabled.set(true);
-    // switch focus only after input has been enabled
+  partsEdited = false;
```
Abhinegi2: may be use signals?

src/app/features/location/address-edit/address-edit.component.ts
```diff
@@ -43,33 +44,61 @@ export class AddressEditComponent {
...
+    this.updateLocation({
+      locationString: this.selectedLocation()?.locationString,
+      geoLookup: this.selectedLocation()?.geoLookup,
+      road: this.selectedLocation()?.road,
+      house_number: this.selectedLocation()?.house_number,
+      postcode: this.selectedLocation()?.postcode,
+      city: this.selectedLocation()?.city,
+      country: this.selectedLocation()?.country,
+      [key]: value,
+    });
```
Abhinegi2: Could simplify to -  `this.updateLocation({ ...this.selectedLocation(), [key]: value })`?

src/app/features/location/address-edit/address-edit.component.scss
```diff
@@ -24,3 +25,64 @@
...
+  }
+
+  .full-row {
+    grid-column: 1 / -1;
+  }
+}
+
+.parts-mismatch-hint {
+  color: colors.$warn;
+  font-size: 0.85rem;
```
Abhinegi2: .hint-text (global class) for the font sizing

src/app/features/location/geo.service.ts
```diff
@@ -209,6 +209,23 @@ export class GeoService {
...
+   * (the reverse of what {@link enrichGeoLocation} derives from a lookup).
+   */
+  composeAddressFromParts(location: GeoLocation | undefined): string {
+    if (!location) {
+      return "";
+    }
+    return [
+      this.formatStreet(location),
+      this.formatPostcodeCity(location),
+      location.country,
```
Abhinegi2: `composeAddressFromParts` and `reformatDisplayName` never match because one adds country and the other doesn't.

Abhinegi2 (changes requested):

sadaf895:
> In the address details panel, when someone edits the Full address text, I was assuming it would trigger the nominatim search and also update the mapped location. Right now it only overwrites the text and the map location stays the same. Is that intended? If yes, maybe we should either run a search from here too, or make this field read-only/disabled so it's not confusing that typing here doesn't move the pin?

Yes, intentional — pre-existing behavior, not new in this PR. That field is for manually overwriting the display text without moving the pin (there's a tooltip explaining it). It's just more visible now since it sits inside the expanded panel. Happy to raise a follow-up if we want to change that behavior. What are your thoughts @sleidig?

sleidig:
> > In the address details panel, when someone edits the Full address text, I was assuming it would trigger the nominatim search and also update the mapped location. Right now it only overwrites the text and the map location stays the same. Is that intended? If yes, maybe we should either run a search from here too, or make this field read-only/disabled so it's not confusing that typing here doesn't move the pin?
> 
> Yes, intentional — pre-existing behavior, not new in this PR. That field is for manually overwriting the display text without moving the pin (there's a tooltip explaining it). It's just more visible now since it sits inside the expanded panel. Happy to raise a follow-up if we want to change that behavior.

I don't think this makes sense or was the previously intended flow. My understanding is the following:

Selecting a search result should
- _always_ update the map location coordinates
- ... and I think also always update the text + address parts

Selecting a location clicking on the map should
- _always_ update the map location coordinates
- ... ask the user if they want to update the text + address parts

Changing the overall address text should
- not change coordinates / location
- not change address parts
- ... but maybe show a hint in the field or warn the user that these other parts + coordinates are now "diverging"

Changing an address part should
- not change coordinates / location
- change the address summary automatically, if it wasn't manually overwritten by the user before
- ask the user if she wants to update the summary text, otherwise

What do you thing, @Abhinegi2 , @george-neha , @sadaf895 ?

Abhinegi2:
> * not change coordinates / location
> * change the address summary automatically, if it wasn't manually overwritten by the user before
> * ask the user if she wants to update the summary text, otherwise
> 
> What do you thing, @Abhinegi2 , @george-neha , @sadaf895 ?

@sleidig  this flow makes sense to me and is cleaner than what we have now 👍

User searches an address (types in the search box)
  - move the map location to that result
  - fill in the text + all address parts from it

User clicks on the map 
  - move the map location
  - ask the user: "update the text + address parts to this new spot?"

User edits the Full address text box
  - don't move the map, don't change the parts
  - just show a hint that the text and the location are now diverging

User edits one single field (road, house no., postcode, city, country)
  - don't move the map
  - if they never manually typed the text before → update the text automatically
  - if they had manually typed the text → ask before overwriting it

george-neha:
@sadaf895 only one scenario is not covered yet(highlighted below)
Scenario:
User edits the Full address text box
don't move the map, don't change the parts
**just show a hint that the text and the location are now diverging**

src/app/features/location/address-edit/address-edit.component.ts
```diff
@@ -48,28 +49,101 @@ export class AddressEditComponent {
...
+    }
+    const composedParts = this.geoService
+      .composeAddressFromParts(location)
+      .trim();
+    const displayName = location?.geoLookup?.display_name?.trim();
+    return text !== composedParts && text !== displayName;
+  }
+
+  /** Drives the in-panel hint that text and structured details are diverging. */
+  hasDivergingText(): boolean {
```
Abhinegi2: `hasDivergingText()` is a plain method called in the template (`@if (hasDivergingText())` at `address-edit.component.html:85`). It re-runs on every change-detection cycle. The component is fully signal-based (the leftover `signal` import was even removed here) — make it a `computed`:

```ts
readonly hasDivergingText = computed(() =>
  this.isTextManuallyOverwritten(this.selectedLocation()),
);
```

src/app/features/location/address-edit/address-edit.component.ts
```diff
@@ -48,28 +49,101 @@ export class AddressEditComponent {
...
+    key: "road" | "house_number" | "postcode" | "city" | "country",
+    value: string,
+  ) {
+    const current = this.selectedLocation();
+    // Decide BEFORE applying the change: compare the current text against the
+    // OLD parts. Comparing against the new parts would always differ and would
+    // falsely ask on every edit.
+    const textOverwritten = this.isTextManuallyOverwritten(current);
+
+    const updated: GeoLocation = {
```
Abhinegi2: `updated` re-lists every `GeoLocation` field by hand. Fragile: a new field added to `GeoLocation` gets silently dropped here. Same result with a spread:

```ts
const updated: GeoLocation = { ...current, [key]: value };
```

src/app/features/location/address-edit/address-edit.component.scss
```diff
@@ -24,3 +25,63 @@
...
+  .full-row {
+    grid-column: 1 / -1;
+  }
+}
+
+.parts-mismatch-hint {
+  color: colors.$warn;
+}
+
+.panel-footer {
```
Abhinegi2: `.panel-footer { display: flex; justify-content: flex-end }` is bespoke layout scss. Could a global flex utility class in the template cover this instead? (The `.address-parts-grid` grid below is genuinely custom, leave that.)

Abhinegi2 (approved):
looks good to me

src/app/features/location/map-popup/map-popup.component.ts
```diff
@@ -121,92 +116,53 @@ export class MapPopupComponent {
...
+    this.updateLocation({ geoLookup: geoResult });
   }
 
-  private saveAndClose() {
-    this.lastSavedLocation = { ...this.selectedLocation };
-    this.manualAddressJustEdited = false;
+  onSave() {
+    // All divergence between the address text, the structured parts and the
+    // mapped location is now resolved at the moment of each edit (part edits
+    // and map clicks ask immediately), so saving just persists the result.
```
sleidig: @Abhinegi2 , on final reviews, let's remove any code comments that do refer to previous behavior or explain changes.
Code should only explain the current state and reasons, otherwise things will get bloated and unnecessarily complex

```suggestion
```

sleidig:
Nice UX now, really like it.

Good job, @sadaf895 , @george-neha  :1st_place_medal:

PR #4115, opened 2026-07-01
refactor: make "unsaved changes" tracking cleanly separated per form

### LINKED ISSUE #4113: dialog unsaved-changes confirmations are unreliable sometimes
## Summary

The e2e test `entity.spec.ts › "Dialog Cancel discards edits; Dialog Save persists changes"` is flaky (observed on [CI run 28494795338](https://github.com/Aam-Digital/ndb-core/actions/runs/28494795338), PR #4082). The flake is a symptom of a **real, pre-existing bug** in how unsaved-changes state is tracked: `EntityFormService` leaks form subscriptions, and `UnsavedChangesService` models a *stack* of dirty layers with a single global boolean.

## The flaky failure

At `entity.spec.ts:330`:

```
expect(dialog).not.toBeVisible() failed
strict mode violation: getByRole('dialog') resolved to 4 elements
```

From the Playwright trace, the dialog-container count is a clean `0 ↔ 1` for the whole test, then jumps to **4** immediately after the final `Save` click and churns `4 → 1 → 4` for the full 5s assertion window. The 4 elements are:

- **1** note detail dialog (overlay `width: 99%`, matching `FormDialogService.openView`), plus
- **3** default-width **"Discard Changes?"** confirmation dialogs (matching `ConfirmationDialogService`).

No JS errors and no URL change (so not a `canDeactivate` guard) — the confirmations come from the note dialog's `disableClose=true` + `backdropClick → checkUnsavedChanges`, which only fires when `unsavedChanges.pending` is `true`.

## Root cause

1. **`EntityFormService` is `providedIn: "root"` (a global singleton)** and, for every form it creates, subscribes to `group.valueChanges` / `group.statusChanges`, pushing each subscription onto a shared `subscriptions` array (`entity-form.service.ts:177-188`). Each of these writes the **single global** signal `unsavedChanges.pending`.

2. **That array is only torn down on router `ActivationStart`** (`entity-form.service.ts:42-49`). Opening an entity in a **dialog** (e.g. clicking a Note row) does **not** navigate, so the subscription **leaks** and stays live after the dialog closes — with its form still dirty.

3. The test's unique trigger is **cancel → reopen → save**. By the save step, multiple leaked `valueChanges` subscriptions (from the cancelled form *and* the reopened form) all race to write the shared `pending` signal as `saveChanges` runs `pending.set(false)` / `markAsPristine()` / `disable()` (the last re-emits `valueChanges`). When the race leaves `pending` momentarily `true`, the backdrop guard opens discard-confirmation dialogs. It's flaky because it only manifests when the race lands the wrong way.

PR #4082 (auto-trim) is **not** the cause — its only change to this path is `.trim()`, a no-op for the test's whitespace-free subject.

## The deeper design flaw

`unsavedChanges.pending` is a single global `boolean` read by all route guards (`app.routing.ts`, `attendance.routing.ts`, `router.service.ts`, de-duplication), `window.onbeforeunload`, the dialog `backdropClick` guard, and `update-manager`. A boolean cannot represent a **stack** of dirty layers (e.g. a dirty main view with a dirty dialog form opened on top). That limitation is why `dialog-buttons` carries the fragile `pendingStateBeforeDialogOpen` / `restorePendingChanges` snapshot-and-restore hack (`dialog-buttons.component.ts:50-51, 132-134`) — and that shared last-writer-wins state is the race behind this flake.

## Proposed fix

Replace the single global boolean with a **registry of dirty sources** in `UnsavedChangesService`:

- Each form/dialog registers itself (keyed by its own `EntityForm`/component instance) when dirty and **deregisters on its own `DestroyRef`** (and on save/reset).
- `pending` becomes a computed `anySourceDirty`; guards and `onbeforeunload` ask "is any source dirty?".

This fixes all three problems at once:

- **Leak** — each source cleans up via its own lifecycle, independent of navigation; the router-`ActivationStart` cleanup and the `subscriptions` array can be removed.
- **Race** — no shared last-writer-wins boolean.
- **Multi-layer** — N simultaneous sources are represented honestly; closing an inner dialog removes only its entry, leaving the outer view dirty, so the `dialog-buttons` restore hack can be deleted.

## Affected files (initial survey)

- `src/app/core/entity-details/form/unsaved-changes.service.ts` (core change)
- `src/app/core/common-components/entity-form/entity-form.service.ts` (leak + writes)
- `src/app/core/form-dialog/dialog-buttons/dialog-buttons.component.ts` (remove restore hack)
- Route guards: `src/app/app.routing.ts`, `src/app/features/attendance/attendance.routing.ts`, `src/app/core/config/dynamic-routing/router.service.ts`, `src/app/features/de-duplication/de-duplication-module.ts`
- Other `pending.set(...)` call sites: `public-forms.service.ts`, `automated-field-update-config.service.ts`, `bulk-merge-service.ts`, `notification-settings.component.ts`, `roll-call.component.ts`, `entity-edit.service.ts`, `entity-bulk-edit.component.ts`, `entity-inline-edit-actions.component.ts`
- `src/app/core/ui/latest-changes/update-manager.service.ts` (reader)

## Interim option

A minimal lower-risk fix (scoping the `EntityFormService` subscriptions to the caller's `DestroyRef` via `takeUntilDestroyed`) would stop the CI flake, but keeps the single-boolean/restore-hack fragility, so multi-layer stays fragile. The registry refactor above is the full fix.

sleidig:
closes #4113

## Summary by CodeRabbit

* **New Features**
  * Unsaved changes are now tracked per form/dialog layer for more accurate discard prompts and safer navigation.
  * Forms now automatically clear draft/unsaved state when their lifecycle ends (such as closing dialogs or removing components).

* **Bug Fixes**
  * Improved cancel/backdrop behavior so discarding affects only the current form layer, not underlying dialogs.
  * Fixed stale unsaved-change warnings after save/reset and during form re-initialization.
  * Updated inline edit, bulk edit, notification settings, attendance, and other form flows to reliably clear unsaved state.

### DIFF
--- a/e2e/tests/entity.spec.ts
+++ b/e2e/tests/entity.spec.ts
@@ -336,3 +336,66 @@ test("Dialog Cancel discards edits; Dialog Save persists changes", async ({
     page.getByRole("cell", { name: DIALOG_INITIAL_SUBJECT }),
   ).not.toBeVisible();
 });
+
+const NO_CHANGES_CHILD_NAME = "<NO CHANGES CHILD>";
+
+test("Closing an untouched new-entity dialog does not prompt discard-changes, but an edited one does", async ({
+  page,
+}) => {
+  const users = generateUsers();
+  const child = generateChild({ name: NO_CHANGES_CHILD_NAME });
+
+  await loadApp(page, [...users, child]);
+
+  await page.getByRole("navigation").getByText("Children").click();
+  await page.getByRole("cell", { name: NO_CHANGES_CHILD_NAME }).click();
+  await page.getByRole("tab", { name: "Education", exact: true }).click();
+
+  const schoolHistoryHeaderRow = page
+    .getByRole("row")
+    .filter({ hasText: "School Class" });
+  const openNewSchoolEnrollmentDialog = () =>
+    schoolHistoryHeaderRow
+      .getByRole("button", { name: /add element/i })
+      .click();
+
+  const discardDialogContainer = page
+    .getByRole("dialog")
+    .filter({ hasText: "Discard Changes?" });
+  const discardDialog = discardDialogContainer.getByRole("heading", {
+    name: "Discard Changes?",
+  });
+
+  // Case 1: open the dialog, touch no fields, and close it via the backdrop
+  // (equivalent to clicking anything behind the dialog, e.g. the page's back
+  // arrow). No discard-changes prompt should appear.
+  await openNewSchoolEnrollmentDialog();
+  const dialog = page.getByRole("dialog");
+  await expect(dialog).toBeVisible();
+
+  await page
+    .locator(".cdk-overlay-backdrop")
+    .click({ position: { x: 5, y: 5 } });
+
+  await expect(discardDialog).not.toBeVisible({ timeout: 3000 });
+  await expect(dialog).not.toBeVisible();
+
+  // Case 2: open the dialog again and actually edit a field — closing it now
+  // must still trigger the discard-changes prompt.
+  await openNewSchoolEnrollmentDialog();
+  await expect(dialog).toBeVisible();
+
+  await dialog
+    .locator("#entity-field__start")
+    .getByRole("textbox")
+    .fill("01.09.2024");
+
+  await page
+    .locator(".cdk-overlay-backdrop")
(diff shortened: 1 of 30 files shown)


sadaf895 (approved):
Code looks good.

PR #4122, opened 2026-07-02
refactor(tables): In-memory data source

### LINKED ISSUE #2161: Implement a custom DataSource to simplify EntitiesTableComponent
A custom DataSource implementation could handle all filtering (maybe even replacing FilterService?) and split code further
--> see https://blog.angular-university.io/angular-material-data-table/

TheSlimvReal:
- First step towards #3943
- closes #2161

This is just a refactoring, keeping all functionality the same. The main goal was to move the data-fetching, sorting, filtering and pagination logic from the table components into the data source. This was done so that in a next step a second datasource can be added which uses database side pagination/sorting/filtering.

## Summary by CodeRabbit

* **New Features**
  * Entity tables across the app now use a shared live data source (`recordsDataSource`) for consistent loading, filtering, sorting, and pagination.
  * Added dedicated “related notes” and “related todos” loading modes, backed by new loader support and indexed todo fetching.

* **Bug Fixes**
  * Improved table selection behavior (including shift-click range selection) and reduced unwanted browser text selection.
  * Updated related-entity and import/review flows to correctly reflect data source loading state and rendered content.

* **Tests**
  * Refreshed unit tests to validate behavior through the new data source and interaction updates.


src/app/core/common-components/entities-table/entities-table.component.ts
```diff
@@ -163,46 +159,45 @@ export class EntitiesTableComponent<
...
+      );
+    });
+
+    effect(() => {
+      this.recordsDataSource().dataFilter.set(this.effectiveFilter());
+    });
+
+    effect(() => {
+      // TODO somehow on resizing the screen this throws an error (cannot read length of undefined)
+      this.recordsDataSource().filter = this.filterFreetext() ?? "";
```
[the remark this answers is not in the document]
TheSlimvReal: This will not be relevant anymore once the free text filter has been replaced and aligned with the normal filters.

src/app/core/common-components/entities-table/in-memory-data-source.ts
```diff
@@ -0,0 +1,171 @@
...
+      const records = this.allRecords();
+      const filter = this.dataFilter();
+      const predicate = this.filterService.getFilterPredicate(filter);
+      this.filteredRecords.set(records.filter(predicate));
+    });
+    effect(() => {
+      if (this.loadRecordConfig()) {
+        // If config is provided, this class loads the data and listens to updates
+        this.getRecords().then((records) => this.allRecords.set(records));
+        this.listenToEntityUpdates();
```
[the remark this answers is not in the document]
TheSlimvReal: Currently, loading more than once is not used in the app.

src/app/core/common-components/entities-table/entities-table.component.ts
```diff
@@ -88,7 +84,7 @@ export class EntitiesTableComponent<
   ) as EntitiesTableSelectionStore<T>;
 
   // --- Inputs ---
-  records = input<T[]>();
+  recordsDataSource = input.required<InMemoryDataSource<T>>();
```
sleidig, line 87: Do you already want to anticipate different datasources and use an interface/abstract class here?

Can touch this later also, though

src/app/core/common-components/entities-table/entities-table.component.ts
```diff
@@ -163,46 +159,45 @@ export class EntitiesTableComponent<
...
     this.sortStore.attachSort(sort);
     if (sort) {
-      this.recordsDataSource.sort = sort;
+      this.recordsDataSource().sort = sort;
     }
   }
 
   constructor() {
+    effect(() => {
+      this.filteredRecordsChange.emit(
```
sleidig: do we still need this or could filter changes be managed exclusively via the data-source now, so that we can remove this from the entities-table component?

src/app/core/common-components/entities-table/in-memory-data-source.ts
```diff
@@ -0,0 +1,171 @@
...
+import { take } from "rxjs/operators";
+
+export interface LoadRecordConfig<T extends Entity> {
+  entityCtr: EntityConstructor<T>;
+  forEntity?: Entity;
+  relationProperty?: keyof Entity;
+  loaderMethod?: LoaderMethod;
+}
+
+export class InMemoryDataSource<T extends Entity> extends MatTableDataSource<
```
sleidig, line 55: add a quick jsDoc explaining its role (and maybe referencing any documentation about implementing a custom mat-data-source, if relevant?)

src/app/core/common-components/entities-table/in-memory-data-source.ts
```diff
@@ -0,0 +1,171 @@
...
+import { EntityMapperService } from "#src/app/core/entity/entity-mapper/entity-mapper.service";
+import {
+  applyUpdate,
+  UpdatedEntity,
+} from "#src/app/core/entity/model/entity-update";
+import { skip, Subscription } from "rxjs";
+import { BulkOperationStateService } from "#src/app/core/entity/entity-actions/bulk-operation-state.service";
+import { take } from "rxjs/operators";
+
+export interface LoadRecordConfig<T extends Entity> {
```
sleidig, line 33: short JsDocs for the interface and each property?

src/app/core/entity-list/entity-list/entity-list.component.ts
```diff
@@ -256,84 +250,6 @@ export class EntityListComponent<T extends Entity> implements OnInit {
...
-    const entityConstructor = this.entityConstructor();
-    if (!entityConstructor) {
-      return Promise.resolve([]);
-    }
-    return this.entityMapperService.loadType(entityConstructor);
-  }
-
-  private updateSubscription: Subscription;
-
-  private listenToEntityUpdates() {
```
sleidig, line 314: I love how we can reduce all that complexity and duplication with the datasource 😍

src/app/features/matching-entities/matching-entities/matching-entities.component.ts
```diff
@@ -234,9 +242,19 @@ export class MatchingEntitiesComponent implements OnInit {
     }
 
     if (!newSide.selected) {
-      newSide.availableEntities = await this.entityMapper.loadType(
-        newSide.entityType,
+      runInInjectionContext(
```
sleidig, line 244: why not a regular class level DI injection?

sleidig (commented):
Looks like a really, really nice cleanup. Thanks, @TheSlimvReal !

src/app/core/common-components/entities-table/entities-table.component.ts
sleidig: Yes, largely. In every usage the parent that binds the table also *instantiates and owns* the `InMemoryDataSource` it passes in, so several of the table's remaining filter inputs/outputs are now just pass-through plumbing into that same object. Moving them onto the data source matches this PR's "filtering lives in the data source" goal — and this line is the clearest case.

### This output (`filteredRecordsChange`) — redundant, and currently subtly wrong
It re-emits `recordsDataSource().filteredRecords()` back to the parent that already owns that data source. The two consumers — entity-list's export and `related-entities-with-summary`'s summary — can read the data-source signal directly and drop the output entirely (plus the `@ViewChild` + `ngAfterViewInit` subscription in the summary component).

This also converges with the free-text finding at line 192: `filteredRecords()` is the *structured*-filter result and **excludes** the free-text `filter` (applied separately by `MatTableDataSource`). Since only entity-list binds `filterFreetext`, it's entity-list's export that silently gets the wrong set while a search is active. Rather than patching the output, I'd expose **one canonical signal** on the data source — "records matching all active filters (structured + free-text + `isActive`)" — and have both consumers read it. That fixes the correctness gap and removes the redundant round-trip in a single move.

### The filter *inputs* too — `filter` / `filterFreetext` / `showInactive` (follow-up)
Today these are translated by effects in the table into `dataSource.dataFilter` / `dataSource.filter`. Since the parent owns the data source, it could set them directly:
- **`filterFreetext`** is cleanly separable (one consumer, entity-list) — an easy first step.
- **`filter` + `showInactive`** are welded together by the `isActive` merge in `effectiveFilter`, so they move as a unit: give the data source a `showInactive` signal and do the merge there. The table keeps the "include archived" toggle UI but binds it to the data-source signal.

### Suggestion
Land the **free-text correctness fix** in this PR (it's a behaviour bug), and treat the rest — removing this output and moving the filter inputs onto the data source — as a focused follow-up so this PR stays a clean, behaviour-preserving refactor.

src/app/core/common-components/entities-table/entities-table.component.ts
TheSlimvReal: Will be done in #4128 once its clear what shared logic exists

src/app/core/common-components/entities-table/entities-table.component.ts
TheSlimvReal: done

sleidig:
...

src/app/core/common-components/entities-table/entities-table.component.ts
sleidig, line 56: @Abhinegi2 , this refactoring somehow broke the Ctrl+Shift bulk select. Do you have any thoughts?

sleidig (commented):
Thanks a lot for the further simplifications. Code looks good to me. We only have one regression for bulk select to look into ...

src/app/core/common-components/entities-table/entities-table.component.ts
Abhinegi2: found it. Sorting moved into the data source, so `dataSource.data` is no longer in display order (Material sorts and paginates internally into `_renderData`), but the selection wiring still derived the page from it: `getCurrentPageRows()` sliced `data` by `pageIndex * pageSize`, and the store got `sortedRows: displayedData`. The clicked row is then usually not in that slice; `currentIndex` is -1, the range branch is skipped, and the fallback resets `lastSelectedRow` to null, so the anchor never survives one click and shift-select looks completely dead.

@sleidig  I've pushed a fix to this branch. Could you please verify that everything is working as expected and that I haven't introduced any regressions?

sleidig:
@coderabbitai review

sleidig (approved):
works correctly now, thanks a lot @Abhinegi2  🙏

PR #4123, opened 2026-07-03
feat(filter): added a string filter as alternative to the free text entry

### LINKED ISSUE #3945: Entity List Component freetext filter through database query
Freetext filter for Entity List Component on the database
→ may need some different approach than field-wise filters;
somewhat lower priority

TheSlimvReal:
- closes #3945

This is a draft for an alternative to the existing free text filter. The idea is to use our existing filter system and when a normal "text" property like name or subject is selected, a text input field allows to enter any kind of text and filter based on that. Under the hood it uses the MangoQuery `$regex` to build the filter.

This change is necessary because in the future filtering should also be possible on the server side (see #4046)

The biggest difference to the existing approach is, that the text is only compared with values of one column instead of the whole entity. If this is too limiting then an alternative/more powerful version could be that it is possible to select multiple properties for which the regex would be created. The UI could be similar to how it is right now with the big free text filter input and in the admin UI it is possible to select multiple string-properties of the entity which should be considered for the filter. Under the hood it would just create a big `$or: [...]` block for all these properties, that would also work on the database. This approach would be more similar to the current implementation but will also keep the free text filter somewhat separate of the filters.

---

To test:

1. Go to a list view
2. 3-dot menu -> "Configure Data Structure"
3. "Add Filter" select a property with a string type (e.g. "Name")
4. "Save"
5. Type into the newly created filter

---

- disable free text filter in paginated mode


src/app/core/filter/filter-generator/filter-generator.service.ts
```diff
@@ -248,6 +254,29 @@ export class FilterGeneratorService {
...
 
+  /**
+   * Whether the field is a free-text field that should be filtered
+   * with a text-search input (see {@link StringFilter}):
+   * dataType "string" or "long-text", edited with a plain text edit component.
+   */
+  private isFreeTextField(type: string, schema: EntitySchemaField): boolean {
+    if (
+      type !== StringDatatype.dataType &&
+      type !== LongTextDatatype.dataType
```
TheSlimvReal: Check if this can be done with an `instanceof` check

TheSlimvReal: Unfortunately not, as nearly all datatypes inherit from `StringDatatype`

sleidig:
@TheSlimvReal , do you have any insights about performance of $regex queries? I imagine that index may also be slow / expensive for large databases and we might need to go for a prefix only (or an index that splits by word boundaries, similar to our search index).

But maybe the index optimization is a separate question to look into at a later stage.

src/app/core/filter/filter-generator/filter-generator.service.ts
sleidig: email also has its own datatype, and so does url ...

src/app/core/filter/filters/stringFilter.ts
```diff
@@ -0,0 +1,53 @@
...
+    public override name: string,
+    public override label: string = name,
+  ) {
+    super(name, label);
+    this.selectedOptionValues = [];
+  }
+
+  /**
+   * The current search text of this filter.
+   * (URL parsing splits values at "," - join them back to restore the original text)
```
sleidig: I didn't fully get that: Shouldn't we encode such characters in the URL then, rather than heuristically recombine?

sleidig (approved):
Works well and looks clean. Nice work 👍

src/app/core/filter/filter-generator/filter-generator.service.ts
sleidig: fixed now

src/app/core/filter/filters/stringFilter.ts
TheSlimvReal: I have implemented the url-encoding for all filters now. This leaves the issue that legacy links with filters might not work anymore

sleidig (approved):
Thanks for the further cleanup 😍

PR #4127, opened 2026-07-06
feat: show permission rule conditions as form field validation errors

### LINKED ISSUE #2264: Permission conditions are not checked when saving
related to #2262, #2263

**Describe the bug**
When creating a new entity while having special "conditions" in the permission rule, the user can currently save the entity even if it is not meeting the conditions. The EntityFormService should validate this and show a useful error (ideally as a validator attached to the relevant form field, otherwise for the form overall as part of "checkFormValidity").

**To Reproduce**
see #2262

Abhinegi2:
- closes #2264

Permission rule `conditions` restricting a field are now translated into normal per-field form validators, so users see an inline error message on the affected field instead of only a technical rejection toast when saving. The save-time error (still needed as backstop, e.g. for fields not part of the form) is also humanized, showing field labels and required values instead of raw JSON. Configurable-enum ids in these messages are resolved to their display labels.

A validator is only created for a field if _every_ granting rule constrains it — if any rule grants access regardless of the field's value, the decision is left to the save-time permission check.

## PR Checklist

---

## Manual Testing Steps

1. Open the PR deployment and log in as `demo`.
3. Open a child whose **Center = Alipore**, edit the form and change Center to *Tollygunge*.
   → The Center field immediately shows the error _"Your permissions only allow the following value(s) here: Alipore"_ (enum label, not raw id `C1`) and saving is blocked like any invalid form.
4. Change Center back to *Alipore* → error disappears, saving works. Editing unrestricted fields (e.g. Name) is unaffected.
5. Create a **new child** and select Center *Tollygunge* → same inline error on the field; with *Alipore* it saves.
6. Open a child whose Center is **not** Alipore → the form stays read-only (existing permission enforcement, unchanged).
7. Cleanup: save `{}` in the permissions JSON editor to restore full access.

## Summary by CodeRabbit

* **New Features**
  * Forms now validate fields against permission-based conditions, showing clearer errors when a value doesn’t match allowed rules.
  * Permission-related save errors now display more readable required-condition details.
* **Bug Fixes**
  * Fields covered by conditional permissions now correctly stay invalid until the permitted value is entered.
  * Error messages now better handle multiple allowed values and formatted labels for clearer guidance.


src/app/core/common-components/entity-form/dynamic-form-validators/permission-condition-validators.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+  });
+
+  it("should not create a validator if the field is not restricted for every granting rule", () => {
+    // no rules at all
+    expect(buildPermissionConditionValidator([], "language")).toBeNull();
+
+    // an unconditional rule grants access regardless of the field value
+    expect(
+      buildPermissionConditionValidator(
+        [{ conditions: { language: "Bengali" } }, {}],
```
sleidig, line 67: Is the `{}` the unconditional rule? A bit hard to understand the reduced down test cases without their subject / action blocks, I think. Maybe could add that in a comment or other way to be more clear

src/app/core/common-components/entity-form/dynamic-form-validators/permission-condition-validators.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+        [
+          { conditions: { language: "Bengali" } },
+          { conditions: { name: "x" } },
+        ],
+        "language",
+      ),
+    ).toBeNull();
+
+    // inverted (cannot) rules are not translated into field validators
+    expect(
```
sleidig: In theory we could do that though, or does that make it much more complicated?

src/app/core/common-components/entity-form/entity-form.service.ts
```diff
@@ -223,9 +237,65 @@ export class EntityFormService {
...
+    schemaField: EntitySchemaField | undefined,
+  ): ((value: any) => string) | undefined {
+    if (schemaField?.dataType !== ConfigurableEnumDatatype.dataType) {
+      return undefined;
+    }
+    return (value) =>
+      this.enumService
+        .getEnumValues(schemaField.additional)
+        .find((option) => option.id === value)?.label ?? String(value);
+  }
```
sleidig: would it be feasible to move more of this code to the validators file to keep entity-form as simple as possible, free of the complex validator specifics?
(maybe even turning that into a service, if necessary)

src/app/core/common-components/entity-form/dynamic-form-validators/permission-condition-validators.spec.ts
Abhinegi2: feasible in theory, but it would make things quite complicated:
 inverted rules only deny when their conditions match the whole entity (often multiple fields), so a per-field validator would need cross-field re-validation and would have to replicate CASL's rule-order precedence (save-time check already covers these cases, so I'd rather keep it out of this PR)

sleidig: Wouldn't a rule like the following just check exactly one field?

```
"conditions": {
    "center": { "$eq": "BERLIN"}
},
"inverted": true
```

--> "no permission to update records for Center "Berlin"."

Abhinegi2: Implemented, pushing now

sleidig (approved):
Looks good to me :-)

sleidig:
@coderabbitai review

PR #4128, opened 2026-07-06
feat(perf): Paginated data source

### LINKED ISSUE #3570: long entity lists load slow
If we have a looot of entities (i.e. >10k records) of a type, that list starts to get slow on initial loading. Pin down the causes (DB read? data filtering? UI - shouldn't be the problem as we have pagination?) and consider performance improvements

---
## Analysis:
- archiving records does not seem to help reduce the initial loading time. It does help when later filtering on the list, however
- both app-filter (entities) and app-entities-table (records) receive the full array of entities of the given type. But removing `<app-filter>` does not improve performance. The bottleneck here really is the database access. This is also confirmed with benchmarking using our `@PerformanceAnalysisLogging`
- using a pouchdb query index to skip `inactive` entities can improve performance hugely, if there are many inactive

## possible solutions:
- [ ] only load entities that are not archived (= `!e.inactive`) using an index
   - requires additional loading when the user toggles the "include inactive" toggle
   - users must archive less relevant records for this to help (for Notes this is not typically done yet)
- [ ] paginate loading of entities. First load one page (<100 docs) and display these, then gradually load more
   - the flexible sorting and filtering is only possible in memory after loading all records. This means we may not be able to load the correct records that need to be displayed
   - we could sort by the "last updated at" date and initially show most recent entities, then load more and update the view with new filtered data as more and more data comes in.
- [ ] similarly, only load active records in entity-select field initially. always display "show inactive" option in the dropdown and only load and search those once clicked initially

### LINKED ISSUE #3943: Entity List Component Filter + Paginate directly on the database

TheSlimvReal:
- closes #3943 

based on PR #4122 this implements the proper `PaginatedDataSource` using the results from #4046 

## Implemented Feature
- [x] Set `"dataSource": "paginated"` for `EntityList` or `RelatedEntities` to activate the paginated data source for this component. -> In online only mode always use paginated and only make it overwritable
An open question is whether this should only be available in the online only mode? -> yes

## Open problems
- [x] sort + filter throws errors on local PouchDB (not online-only) -> use only in online only mode
- [x] Export dialog required all entities (should it only use available or warn?) -> should load full data
- [x] Generating some filters requires all entities
  - For empty option -> always add
  - ConfigurableEnum not existing options -> keep current and undetermined behaviour in paginated case?
  - SelectionFilter for creating all options (is it even used?) -> Use text filter instead
- [x] indicator for filter request in progress
- [x] Adjust table description (x of y) when fully amount of entities is not known
- [x] Tables first send request without filter and then with filter. Check if first requests can be skipped
- [x] Transformation of string filter queries for DB
- [ ] Text sorting does not respect parsable numbers (`"1" < "10" < "2"`) and omits documents that are missing the property
- [ ] Cannot use sort together with map-reduce queries (which are used to speed up children of a school, notes of a child, ...) - in these cases using the in-memory data source gives better performance
- [x] Pagination works with permission (similar to changes feed) see [backend PR](https://github.com/Aam-Digital/replication-backend/pull/321)
- [x] Bulk select is reset on filter change
- [x] Applying filter moves pagination back to first page

## Summary by CodeRabbit

* **New Features**
  * Added support for in-memory and paginated entity-table data sources.
  * Added server-side querying, sorting, filtering, and full-data loading for large datasets.
  * Added clearer pagination for unknown totals, including “+” range labels.
  * Added configurable data sources for entity lists and related-entity views.
  * Export operations can now retrieve all or filtered records asynchronously.

* **Bug Fixes**
  * Filtering and data changes now reset pagination and clear selections.
  * Added missing-field matching for empty-value filters.
  * Improved loading feedback with an in-table progress bar.


src/app/core/common-components/entities-table/data-source/paginated-data-source.ts
```diff
@@ -0,0 +1,140 @@
...
+      }
+    } else {
+      // We don't really know how it might affect the pages -> full reload
+      await this.setRecords();
+    }
+  }
+
+  private processFilterForDB(filter: DataFilter<T>): EntityFilter<T> {
+    // isActive is not available in the database
+    delete filter["isActive"];
```
TheSlimvReal: replace with `inactive`

src/app/core/common-components/entities-table/data-source/available-data-sources.ts
```diff
@@ -0,0 +1,9 @@
+import { InMemoryDataSource } from "#src/app/core/common-components/entities-table/data-source/in-memory-data-source";
+import { PaginatedDataSource } from "#src/app/core/common-components/entities-table/data-source/paginated-data-source";
+
+export const availableDataSources = {
+  "in-memory": InMemoryDataSource,
+  paginated: PaginatedDataSource,
+};
+
+export type DataSourceType = keyof typeof availableDataSources;
```
sleidig, line 9: Is this different from an enum with explicit value assignment, like `SessionType`?
https://github.com/Aam-Digital/ndb-core/blob/f31f92e437f8968fd200c60e8df53412d40ae5d1/src/app/core/session/session-type.ts#L9-L29

src/app/core/common-components/entities-table/data-source/data-source-resolver.ts
sleidig, line 1: do we use `data-source` or `datasource`? 😊

src/app/core/common-components/entities-table/data-source/entities-table-data-source.ts
```diff
@@ -0,0 +1,190 @@
...
+   * Property through which the relation can be resolved
+   */
+  relationProperty?: keyof Entity;
+  /**
+   * Select if a special loader method should be used for this entity
+   */
+  loaderMethod?: LoaderMethod;
+}
+
+export abstract class EntitiesTableDataSource<
```
sleidig, line 42: add brief JsDoc? (at first glance I am wondering what's the difference of this one compared to InMemory and Paginated DataSource)

src/app/core/database/database.ts
```diff
@@ -91,6 +91,13 @@ export abstract class Database {
     throw new Error("purge() is not supported by this database implementation");
   }
 
+  abstract find(
```
sleidig, line 104: Add JsDoc?

src/app/core/database/pouchdb/pouch-database.ts
```diff
@@ -385,6 +387,49 @@ export class PouchDatabase extends Database {
...
+    if (Number.isInteger(page?.limit)) {
+      findOptions.limit = page.limit;
+    }
+    if (Number.isInteger(page?.skip)) {
+      findOptions.skip = page.skip;
+    }
+    const pouchDB = await this.getPouchDBOnceReady();
+    if (sort?.prop) {
+      // TODO delete indexes at one point? e.g. when column is removed
+      const indexRes = await pouchDB.createIndex({
```
sleidig: what's this index in contrast to the built-in pouchdb.find?

maybe you can share some of your insights on how this all works with me 😊

src/app/core/entity-details/related-entities/related-entities.component.ts
```diff
@@ -80,6 +79,11 @@ export class RelatedEntitiesComponent<E extends Entity> {
    */
   columns = input<ColumnConfig[]>([]);
 
+  dataSource = input<DataSourceType>();
+  recordsDataSource = computed(() =>
+    resolveDataSource<E>(this.injector, this.dataSource()),
+  );
```
sleidig, line 85: could it be easier to have this as a Angular Service that functions as a Factory, so we don't have to juggle with injector? Or would that have other issues?

src/app/core/entity/entity-mapper/entity-mapper.service.ts
```diff
@@ -87,6 +88,19 @@ export class EntityMapperService {
     return records.map((rec) => this.transformToEntityFormat(rec, ctor));
   }
 
+  public async findType<T extends Entity>(
```
sleidig, line 98: JsDoc? The name is not fully intuitive to me yet

... ah, I do see the parallel to `loadType` now, though!

src/app/core/filter/filter/filter.component.spec.ts
```diff
@@ -284,11 +284,7 @@ describe("FilterComponent", () => {
     component.filterOptionSelected(avilableOptions, [t1.id]);
 
     expect(emittedFilterObj).toEqual({
-      $and: [
-        {
-          $or: [{ "category.id": t1.id }],
-        },
-      ],
+      $and: [{ "category.id": t1.id }],
```
sleidig: does that also work for multi-selecting (filter to show results matching one of two options)?

src/app/features/matching-entities/matching-entities/matching-entities.component.ts
```diff
@@ -58,7 +58,7 @@ import {
   MatchingSideConfig,
   NewMatchAction,
 } from "./matching-entities-config";
-import { InMemoryDataSource } from "#src/app/core/common-components/entities-table/in-memory-data-source";
+import { InMemoryDataSource } from "#src/app/core/common-components/entities-table/data-source/in-memory-data-source";
```
sleidig, line 61: does this still explicitly depend on the in-memory datasource? What happens in online-only mode with this component currently?

sleidig (commented):
Looks quite good to me already :smiley: . Did not do functional testing yet, however.

@Abhinegi2 , could you also have a look and especially try if you can break anything in online-only mode (i.e. using the new paginated data-source)?
Testable at https://load-test.aam-digital.net/ (demo / pass)

sleidig:
@coderabbitai review

Abhinegi2:
@TheSlimvReal. tested on [load-test ](https://load-test.aam-digital.ne)with the demo user. A few things:

- I removed the gender from one child, then filtered the Children list by gender = "not defined". The child did not show up, list came back empty.
- Participants → open a child → Groups section is very slow, even for a brand new child with no groups. The loading bar keeps running. Linking a new group saves fine, but the table takes very long before the new group appears. Only 1 record, so it is not a data volume issue.
- Download → "All records" only exports what is visible in the list. earlier it exported all records.
- Bulk action -> select all rows checkbox at the top of the list now only selects the current page. earlier it selected everything.
 - Went to page 4 of a long list, then applied a filter matching only a few records (e.g., Aadidev Kaur which is present). The table came back empty, with "31 - 30 of 30" at the bottom. 
 - Page count/ paginator now reads "1 - 10 of 10+" instead of "1 - 10 of 120". not broken, but users lose the total count, which people rely on to know how many records they have. And "10+" reads like there are barely more than 10. Could we still show the real total somewhere?
- site crashed while I was testing. I was the only user at that time. Could not capture the cause Will try to reproduce.

Abhinegi2:
- When applying date sorting to the **Age** column, it did not work and resulted in the following error:

```
{
  "error": "timeout",
  "reason": "The request could not be processed in a reasonable amount of time."
}
```
also, sometimes the sorting appears to work, but not all records are displayed afterward. For example, in the ss, only 2 records are shown even though there should be more records available.
image
image

src/app/core/common-components/entities-table/data-source/available-data-sources.ts
TheSlimvReal: It allows to directly assign an Object as the value instead of just a string which then needs to be mapped again to the matching object.

src/app/core/filter/filter/filter.component.spec.ts
TheSlimvReal: Yes, this is just a fix of the test because CouchDB does not handle it well when there is only one element in a `$or` but I expanded the test

src/app/features/matching-entities/matching-entities/matching-entities.component.ts
TheSlimvReal: yes, for this component to display all entities in the map it also needs to have them all loaded and stored in-memory. It always uses the in-memory approach which means in online-only mode it will just request all entities instead of the paginated subset.

src/app/core/common-components/entities-table/data-source/paginated-data-source.ts
```diff
@@ -0,0 +1,223 @@
...
+        filter["inactive"] = true;
+      }
+      delete filter["isActive"];
+    }
+    filter = convertToCouchRegex(filter);
+    // Mango queries need `$options: "i"` while CouchDB only supports `$regex: "(?i)..."`
+    const filterString = JSON.stringify(filter);
+    // replace e.g. "gender.id" with "gender" as configurable enums are only stored with id value
+    const updatedString = filterString.replace(/("\w+)\.id(?=":)/g, "$1");
+    return JSON.parse(updatedString);
```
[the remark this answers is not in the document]
TheSlimvReal: Only affects mock entity mapper. Will not provide fix for now. Might have an issue if a test is created in online-only mode that uses mock-entity-mapper.

src/app/core/entity-details/related-entities/related-entities.component.ts
TheSlimvReal: Unfortunately not, because the creation needs the surrounding components injection context for `effect(() => ...)` and other hooks to work.

src/app/core/filter/string-filter/string-filter.component.ts
```diff
@@ -32,7 +40,11 @@ export class StringFilterComponent<E extends Entity> {
 
   constructor() {
     this.textControl.valueChanges
-      .pipe(takeUntilDestroyed(this.destroyRef))
+      .pipe(
+        debounceTime(STRING_FILTER_DEBOUNCE_MS),
+        distinctUntilChanged(),
+        takeUntilDestroyed(this.destroyRef),
+      )
```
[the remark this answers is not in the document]
TheSlimvReal: @CodeRabbitAI I do not understand the issue here. Please describe it more clearly.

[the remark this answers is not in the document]
TheSlimvReal: @coderabbitai I do not understand the issue here. Please describe it

TheSlimvReal:
@coderabbitai review

e2e/tests/bulk-operations.spec.ts
```diff
@@ -274,7 +274,7 @@ test("Bulk selection follows the rendered order through sorting, filtering and p
   await rows.nth(1).click();
 
   // Bulk-editing them out of the active filter must update the filtered result
-  // and the paginator while we are on the second page.
```
sleidig, line 277: ah, this became general behavior for all data sources now, right? Okay

sleidig (approved):
Code looks good to me!
(no functional testing done but we have e2e tests that should cover most, I think)

@Abhinegi2 , maybe you can also do one more review of this key piece of code

src/app/core/filter/string-filter/string-filter.component.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    component.textControl.setValue("");
+    tick(STRING_FILTER_DEBOUNCE_MS);
+    expect(emitted).toEqual([["abc"], []]);
+  }));
+
+  it("should reflect external changes without echoing them back through the debounce", fakeAsync(() => {
+    // e.g. "clear all filters" emits an external change
+    filter.selectedOptionChange.emit([]);
+    tick(STRING_FILTER_DEBOUNCE_MS);
```
Abhinegi2: we should use `vi.useFakeTimers()` / `vi.advanceTimersByTime() `instead f`akeAsync/tick` ?

e2e/tests/bulk-operations.spec.ts
TheSlimvReal: Yes, the default `MatPaginator` works a little different but I did not see a use case where it is important to stay on a higher page after filtering.

TheSlimvReal:
[Demo with 10k Docs](https://load-test-4.aam-digital.net/) (1.000 Children, 10 Notes per Child)
[Demo with 100k Docs](https://load-test-3.aam-digital.net/) (10.000 Children, 10 Notes per Child)
[Demo with 1M Docs](https://load-test.aam-digital.net/) (10.000 Children, 100 Notes per Child)

- All views load pretty fast after the index is built, feels faster than in synced mode with all data
- Sorting by a new row requires a new index and takes a while until index is built
- Filter on a property that is not the active sort column takes considerably longer

Abhinegi2:
did the functional testing, and it seems to be working as expected, apart from some known limitations

PR #4136, opened 2026-07-09
fix: retry transient network aborts on idempotent DB reads (AAM-DIGITAL-73M)

### LINKED ISSUE #4135: Transient network aborts on online-mode DB reads surface as error-level DatabaseException
## Summary

In online mode (`RemotePouchDatabase`), an idempotent read (`get` / `query` / `allDocs`) can fail with a **transient network abort** — e.g. the response-body read is aborted on a connection that went stale after the browser tab was suspended (laptop sleep / long idle) — and this surfaces as an **error-level `DatabaseException`** in our error monitoring, even though the failure is benign and would succeed on a retry.

**Sentry:** `AAM-DIGITAL-73M` — `DatabaseException: The operation was aborted.`

## Root cause

Reads over HTTP go through `RemotePouchDatabase.defaultFetch`, wrapped by a retry helper that applies a per-request abort timeout and retries transient errors. That helper only wraps **acquiring the response headers** — an abort during **response-body streaming** happens *after* the fetch has already returned, outside the helper's `try/catch`. As a result:

- the retry loop never sees it, and
- `defaultFetch`'s `"Failed to fetch from DB"` normalization (which runs on the header-acquisition failure) is also bypassed,

so the raw abort propagates up to `PouchDatabase.get()` and becomes a terminal `DatabaseException` → global error handler → Sentry, on the **first** occurrence (it is *not* gated by exhausting retries).

The physical trigger is benign: a tab left open across a long idle period whose connection is stale on resume; the next read aborts and then recovers.

## Why retrying — not just filtering — is the right fix

Hardening the fetch-layer retry would **not** prevent this, because the failure bypasses that layer entirely. Re-issuing the whole read (fresh fetch + body) on a fresh connection recovers it transparently. This belongs at the **operation layer**, which is the only layer that sees the read's final outcome (headers *and* body).

Only **idempotent reads** are safe to retry. Writes must run exactly once: a network failure gives no signal about whether the server already committed, so retrying a committed write can duplicate a `POST` create, conflict a `PUT`, or turn a create-only write into a forbidden update.

## Fix

Move retry from the fetch layer to the operation layer:

- add a `withReadRetry` hook on `PouchDatabase` (base implementation: no retry) that `get` / `allDocs` / `query` route through;
- `RemotePouchDatabase` overrides it to retry reads on `isConnectivityError` (aborts / timeouts / network errors), up to the existing retry budget;
- the fetch wrapper (renamed `fetchWithTimeout`) keeps only the per-request abort timeout — writes still run once with no timeout.

## Follow-up (tracked separately)

A residual safety net for *persistent* transient failures that still throw terminally after retries are exhausted:

- preserve the original error `name` on `DatabaseException` so wrapped aborts remain classifiable as connectivity errors, and
- extend the Sentry `beforeSend` filter to drop abort/transient events regardless of `navigator.onLine` (a tab that just reconnected reports `onLine: true` while requests still time out).

sleidig:
Closes #4135 · Sentry: `AAM-DIGITAL-73M`

## Problem

Online-mode reads (`get` / `query` / `allDocs` over HTTP) could surface a **transient network abort** as an error-level `DatabaseException` reaching Sentry, even though the failure is benign and recovers on retry — e.g. when a response-body read is aborted on a connection gone stale after the browser tab was suspended (laptop sleep / long idle).

The existing retry lived in the fetch wrapper, which only covers **acquiring the response headers**. An abort during **response-body streaming** happens *after* the fetch returns — outside the wrapper's reach — and also bypasses the `"Failed to fetch from DB"` normalization, so the raw abort became a terminal error on its first occurrence. Full analysis in #4135.

### 1. Recover — move retry to the operation layer

Retry now lives at the **operation layer** — the only layer that sees a read's final outcome (headers *and* body):

- **`PouchDatabase`**: new `withReadRetry` hook (base implementation runs the read once, no retry); `get` / `allDocs` / `query` route their DB call through it. Behaviour for local/synced DBs is unchanged.
- **`RemotePouchDatabase`**: overrides `withReadRetry` to re-issue idempotent reads on `isConnectivityError` (aborts / timeouts / network errors), up to the existing retry budget. The predicate runs on the raw error, so it catches both the bare `AbortError` and the normalized `"Failed to fetch from DB"`.
- **`fetchWithTransientRetry` → `fetchWithTimeout`**: keeps only the per-request abort timeout, no retry. Writes still run exactly once with no timeout (a network failure gives no signal whether the server already committed, so retrying could duplicate a create or turn a create-only write into a forbidden update).

### 2. Classify — preserve the wrapped error name on `DatabaseException`

`DatabaseException` keeps its `name` as `"DatabaseException"` for stable Sentry grouping, which erased the wrapped error's own name (e.g. `"AbortError"`). It now also preserves that name in a new `originalName` field, and `isConnectivityError` checks both `name` and `originalName`. This is the residual safety net: a persistent transient failure that still throws terminally after retries are exhausted stays classifiable as a connectivity error downstream.

Net result: retry lives in exactly **one** place; the abort-during-body-read failure mode is recovered transparently on a fresh connection; and any residual terminal transient error remains recognizable instead of masquerading as a generic DB error.

## Note on sync / replication

The operation-layer retry only wraps `get` / `allDocs` / `query`, so it does **not** apply to sync — `SyncedPouchDatabase` runs PouchDB replication (`localDb.sync(remote, …)`), whose HTTP calls go straight through the remote's `fetch`, bypassing the operation layer.

A side effect of dropping the fetch-layer retry: GET-based replication requests (`_changes`, checkpoint reads, …) previously got a per-request retry via `fetchWithTransientRetry` and no longer do. Writes are unaffected (only GET/HEAD were ever retried). Sync now relies on the retry it already had — `liveSync()` re-runs the whole sync every 30s and on any local change, and treats connectivity errors as expected (debug-level, not Sentry noise). This is a minor, deliberate trade-off to keep retry in a single place and avoid double-retrying online-mode reads (which would otherwise retry at both the fetch and operation layers). If per-request sync resilience is wanted back, it can be restored on the sync-only remote instance without re-introducing double-retry.

## Testing

- `pouch-database.spec.ts`: `DatabaseException` unit tests (name stays `"DatabaseException"`, `originalName` preserved, not set when wrapping another `DatabaseException`).
- `connectivity-error.spec.ts` (new): `isConnectivityError` incl. a `DatabaseException` wrapping an `AbortError` classified via `originalName`.
- `remote-pouch-database.spec.ts`: operation-level `withReadRetry` tests, an end-to-end `get()` retry/recover test reproducing the Sentry issue, and timeout-only `fetchWithTimeout` tests.
- `synced-pouch-database.spec.ts`, `logging.service.spec.ts`, `keycloak-auth.service.spec.ts` (consumers): all green.
- Full run: 107 passing across the affected specs; `tsc --noEmit`: clean.

## Follow-up (separate)

The last piece of the safety net: extend the Sentry `beforeSend` filter to drop abort/transient events regardless of `navigator.onLine` (a tab that just reconnected reports `onLine: true` while requests still time out). Noted in #4135.

### DIFF
--- a/src/app/core/database/pouchdb/pouch-database.ts
+++ b/src/app/core/database/pouchdb/pouch-database.ts
@@ -526,6 +526,16 @@ export class PouchDatabase extends Database {
 export class DatabaseException extends Error {
   entityId?: string;
 
+  /**
+   * The `name` of the underlying error that was wrapped (e.g. "AbortError",
+   * "TimeoutError", "not_found"). Preserved separately because {@link name} is
+   * deliberately kept as "DatabaseException" for stable Sentry grouping — which
+   * would otherwise erase the original name and hide that a failure is actually
+   * a transient network abort. Read by {@link isConnectivityError} so wrapped
+   * transient errors stay classifiable.
+   */
+  originalName?: string;
+
   constructor(
     error: PouchDB.Core.Error | { message: string; [key: string]: any },
     entityId?: string,
@@ -533,8 +543,13 @@ export class DatabaseException extends Error {
     super(error?.message || "Database error");
 
     this.entityId = entityId;
+    // Capture the wrapped error's name before Object.assign / `this.name` erase it.
+    const originalName = (error as { name?: string })?.name;
     Object.assign(this, error);
     // Restore class name after Object.assign overwrites it with PouchDB's name (e.g. "not_found")
     this.name = "DatabaseException";
+    if (originalName && originalName !== "DatabaseException") {
+      this.originalName = originalName;
+    }
   }
 }
--- a/src/app/core/database/pouchdb/pouch-database.ts
+++ b/src/app/core/database/pouchdb/pouch-database.ts
@@ -103,6 +103,20 @@ export class PouchDatabase extends Database {
   }
 
   /**
+   * Hook wrapping an idempotent read (`get`/`allDocs`/`query`) so subclasses
+   * can transparently retry transient failures.
+   *
+   * The base implementation runs the operation exactly once (no retry).
+   * {@link RemotePouchDatabase} overrides this to re-issue reads that fail with
+   * a transient network abort/timeout, which recovers e.g. a connection gone
+   * stale after the tab was suspended. Only reads use this hook — writes must
+   * run exactly once and are never routed through it.
+   */
+  protected async withReadRetry<T>(operation: () => Promise<T>): Promise<T> {
+    return operation();
+  }
+
+  /**
    * Load a single document by id from the database.
    * (see {@link Database})
    * @param id The primary key of the document to be loaded
@@ -115,7 +129,9 @@ export class PouchDatabase extends Database {
     returnUndefined?: boolean,
   ): Promise<any> {
     try {
-      return await (await this.getPouchDBOnceReady()).get(id, options);
+      return await this.withReadRetry(async () =>
+        (await this.getPouchDBOnceReady()).get(id, options),
+      );
     } catch (err) {
       if (err.status === 404) {
         Logging.debug("Doc not found in database: " + id);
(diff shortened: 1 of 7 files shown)


src/app/core/database/pouchdb/pouch-database.ts
```diff
@@ -508,15 +526,30 @@ export class PouchDatabase extends Database {
 export class DatabaseException extends Error {
```
sleidig: @copilot , move `DatabaseException` to its own file

Abhinegi2 (approved):

PR #4141, opened 2026-07-10
test: prevent shared TestEntity.schema pollution causing flaky tests

Abhinegi2:
## PR Checklist

- [x] added label **"Final Review"** to trigger UI regression testing and CodeRabbit AI review
- [x] marked each code review comment as resolved (or commented on it with a question, if unsure)

---

## Manual Testing Steps

This is a test-only change; verify via the test suite:

1. `npm run test-ci` passes.
2. To confirm the flake is gone, temporarily set `sequence.shuffle: true` in `vitest-base.config.ts` and run the suite: `AdminEntityFormComponent` and `EntityFormService` schema tests pass regardless of order (other, unrelated specs may fail under shuffle; that is the separate pre-existing issue noted above).

## Summary by CodeRabbit

* **Tests**
  * Improved test isolation by preventing shared entity schema changes from affecting other tests.
  * Added reliable cleanup for temporary schema modifications, including when tests fail.
  * Updated filter tests to use dedicated entities for custom fields.

* **Documentation**
  * Added guidance for creating dedicated entities when testing custom fields to reduce flaky, order-dependent test failures.

### DIFF
--- a/src/app/core/filter/filter-generator/filter-generator.service.spec.ts
+++ b/src/app/core/filter/filter-generator/filter-generator.service.spec.ts
@@ -17,16 +17,14 @@ import {
   SelectableFilter,
 } from "../filters/filters";
 import { Entity } from "../../entity/model/entity";
+import { DatabaseEntity } from "../../entity/database-entity.decorator";
 import { DateFilter } from "../filters/dateFilter";
 import { BooleanFilter } from "../filters/booleanFilter";
 import { ConfigurableEnumFilter } from "../filters/configurableEnumFilter";
 import { EntityFilter } from "../filters/entityFilter";
 import { FormFieldConfig } from "../../common-components/entity-form/FormConfig";
 import { TestEntity } from "../../../utils/test-utils/TestEntity";
-import {
-  EntitySchemaField,
-  PLACEHOLDERS,
-} from "../../entity/schema/entity-schema-field";
+import { PLACEHOLDERS } from "../../entity/schema/entity-schema-field";
 import { CurrentUserSubject } from "app/core/session/current-user-subject";
 import { expectArrayWithExactContents } from "../../../utils/test-utils/array-test-utils";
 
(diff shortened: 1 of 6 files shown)


Abhinegi2:
`TestEntity.schema` is shared across all spec files. A couple of specs changed it without restoring it — `entity-form.service.spec` overrode the name field (from the trim change), and `filter-generator.spec ` added fields and never cleaned up. When those ran before admin-entity-form, they read a broken schema and failed. Test order varies per run, so it passed/failed randomly( that's why it slipped into master)

Fix:
- removed the unnecessary schema.set("name") (name is already a string field)
- clean up added fields in afterEach in the other specs
- safety net in test-setup.ts: restore TestEntity.schema after every test, so no spec can leak into others even if it forgets to clean up

src/app/core/filter/filter-generator/filter-generator.service.spec.ts
```diff
@@ -42,6 +42,15 @@ describe("FilterGeneratorService", () => {
     filterService = TestBed.inject(FilterService);
   }));
 
+  afterEach(() => {
+    // remove fields individual tests add to the shared static schemas so they
+    // don't leak into later-running specs (runs even if a test throws first)
+    TestEntity.schema.delete("test");
+    TestEntity.schema.delete("enumField");
+    TestEntity.schema.delete("tags");
```
sleidig: can you instead create a custom Entity class for these specific tests, if they need custom fields, instead of overwriting TestEntity?

src/app/utils/test-utils/TestEntity.ts
```diff
@@ -71,3 +71,13 @@ export class TestEntity extends Entity {
...
+
+/**
+ * Pristine snapshot of TestEntity's schema, captured once at module load right
+ * after the decorators populate it (before any spec can mutate the shared static
+ * schema). test-setup restores from this after every test so a spec that adds,
+ * replaces or removes fields without cleaning up (or that throws before its
+ * inline cleanup runs) cannot leak into later specs and cause order-dependent
+ * flakiness.
+ */
+export const PRISTINE_TEST_ENTITY_SCHEMA = new Map(TestEntity.schema);
```
sleidig: Let's instead _always_ create custom entity classes if we need special fields in a test. That pattern is simpler and less error-prone - and used in many tests already.

(one exception may need to be tests specifically for a special entity like `Note`, if that relies on it being Note, not some other entity type)

sleidig (changes requested):

src/app/utils/test-utils/TestEntity.ts
sleidig: and maybe we add this to the README / AGENTS.md / instructions file? :blush:

Abhinegi2:
@coderabbitai review

sleidig:
@Abhinegi2 , not related but could you investigate the flaky UI regression tests, like this one: https://app.argos-ci.com/sleidig/ndb-core/builds/3456/359309222
regularly, we have a tiny layout shift that breaks screenshot comparison

sleidig (approved):

PR #4144, opened 2026-07-13
feat(Import): guide users who lack permission before an import fails

### LINKED ISSUE #4133: Import: permission errors show raw JSON, and some failures leave a partial import
## What happens
A user goes to **Import**, selects a file, picks the record type, maps the columns, and clicks **Start Import**. Instead of a normal result, they get an "Import failed" popup with a raw error dump like this:

```
Sorry, some error occurred during import. Please try again. If the problem persists, contact support. [{"action":"create","entityId":"ImportMetadata:…","entityType":"ImportMetadata","name":"EntityPermissionError"}]
```

Nothing warns them beforehand, and in some cases the records are actually created even though it says failed. Digging into it, there are a few related problems.

## 1. Raw error object shown to the user
`ImportConfirmSummaryComponent.showImportErrorMessage` builds the message with `JSON.stringify(error)`. For an `Error` this drops the (non-enumerable) `message`, so the user sees the serialized fields instead of a readable sentence (the popup above).

## 2. Partial import reported as total failure
The import saves the entities first, then writes an `ImportMetadata` history record (used for undo). Non-admin roles don't have `create` permission for `ImportMetadata`, so that last write throws `EntityPermissionError` — **after the actual data is already saved**.

Reproduced: a user with `create` on `Child` but not on `ImportMetadata` imported 3 children. All 3 records were created in the DB, but the dialog said "Import failed". Risk: the user retries and ends up with duplicates.

## 3. Read-only import isn't blocked early (but is data-safe)
If a user has only `read` on a type and imports that type, nothing stops them through the wizard — it fails only at "Start Import". `EntityMapperService.saveAll` checks permissions on all rows before writing, so it throws before any `putAll` and **nothing is saved**. This case is safe on the data side, but still shows the same raw error dialog and gives no early feedback.

Reproduced: read-only user importing children → `EntityPermissionError` (`entityType: "Child"`), 0 records saved.

## Suggested improvements
- **Readable error messages.** Map `EntityPermissionError` (and errors in general) to a clear message; keep full detail in the logs instead of dumping the serialized error. Likely worth a shared helper since other flows may `JSON.stringify` errors too.
- **Fail fast, before writing anything.** Check create permission on the target type *and* `ImportMetadata` up front (e.g. on the Review step). If something's missing, block "Start Import" with a clear message so an import can't half-complete.

## Open question
Should being able to import an entity imply `create` permission for its `ImportMetadata`? Right now only admins have it, so any non-admin import fails at the history step. Related: Aam-Digital/replication-backend#236.

## Affected code
- `import-confirm-summary.component.ts` — `showImportErrorMessage`
- `import.service.ts` — `saveImportHistory`
- `entity-mapper.service.ts` — `assertPermission` / `saveAll` / `save`

Abhinegi2:
- closes #4133

## PR Checklist

---

## Manual Testing Steps

Requires a real (non-demo) system so permission rules and roles apply. Log in as a user whose role can read but **not** create the relevant records (e.g. a `user_app`-only account that lacks `create` on `Child` and on `ImportMetadata`).

1. **Entity list** – open the Children list, open the "..." action menu. Expect **no** "Import from file" entry. Log in as an admin and confirm the entry **is** present.
2. **Step 2 gating** – go to Import, upload any `.csv`, continue to "Select Import Type". Open the "Import as" dropdown. Expect types the user cannot create to be greyed out and labelled "(no permission)" and to be unselectable; types they can create remain selectable.
3. **Review step** – as a user who can create the target type but not `ImportMetadata`, complete steps 1-3 and reach "Review & Edit Data". Expect a "You don't have the required permissions to import." warning and a disabled "Start Import" button, so no partial import occurs.

## Summary by CodeRabbit

* **New Features**
  * Import options now reflect the current user’s permissions across the workflow.
  * Users without import access see a permission-not-allowed message instead of the stepper.
  * Entity type selection is permission-filtered, with hints shown for available record types.
  * “Continue,” “match existing,” and additional import actions are enabled/hidden based on create/update permissions.
  * Import existing selections are reset when the selected type changes.
* **Tests**
  * Added/updated permission-focused test coverage for import gating and action filtering.


Abhinegi2:
@george-neha Please test this on the dev site using the account `georgeneha92@gmail.com`. This account has permission to create Child records, but it intentionally does not have permission to create the internal ImportMetadata (Import History) record.

1. Log in as `georgeneha92@gmail.com`.
2. Go to the Children list and click the three-dot (⋮) menu.
3. Expected: The "Import from file" option is not visible because this user cannot create the required ImportMetadata record.
4. Open the import page directly: [https://dev.aam-digital.net/import](https://dev.aam-digital.net/import)
5. In the Entity dropdown, you'll notice that several record types are disabled and labeled "(no permission)". These are entities the user is not allowed to create.
6. Select an enabled entity (for example, Child) and continue through the import flow until Step 4 (Review).
7. Expected: A warning banner appears at the top saying, "You don't have the required permissions to import."
8. Expected: The "Start Import" button is disabled, so the import cannot be started.

Note: Every import creates an internal ImportMetadata record (the import history entry used for the "Undo Import" feature). Since this user does not have permission to create this record, the import is blocked before any data is saved.

src/app/core/import/import-review-data/import-review-data.component.html
```diff
@@ -4,10 +4,22 @@
     <mat-progress-bar mode="indeterminate"></mat-progress-bar>
   </div>
 } @else {
+  @if (cannotCreateImportMetadata()) {
+    <app-hint-box type="warning" i18n>
+      You don't have the required permissions to import.
+    </app-hint-box>
+  }
```
sleidig: I think that should show on the first, not the last step?

sleidig: if we lock down the parent component (import.component.ts) then we probably don't need checks here anymore

sleidig (commented):
code looks good. Let's briefly check if we can reduce the levels at which we check permissions

src/app/core/import/import-entity-type/import-entity-type.component.ts
```diff
@@ -37,6 +42,16 @@ export class ImportEntityTypeComponent {
...
 
+  disableTypeOption = (type: EntityConstructor) =>
+    this.ability.initialized && this.ability.cannot("create", type);
+
+  typeOptionToString = (type: EntityConstructor) => {
+    const label = type.label ?? type.ENTITY_TYPE;
+    return this.disableTypeOption(type)
+      ? $localize`${label} (no permission)`
+      : label;
+  };
```
sleidig: If a user doesn't even see (`read` permission) data of a type, it probably shouldn't appear in the list at all.

Alternatively we could (probably easier to implement) simply hide any, not showing them as disabled. And add a mat-hint to the selection stating that you need create permissions to use the import.

src/app/core/import/import-entity-type/import-entity-type.component.html
```diff
@@ -8,6 +8,8 @@
       (ngModelChange)="onSelectedTypeChange($event)"
       [multi]="false"
       [showInternalTypes]="expertMode()"
+      [disableOption]="disableTypeOption"
+      [optionToString]="typeOptionToString"
     ></app-entity-type-select>
   </mat-form-field>
```
sleidig, line 17: we also need to guard the "Check/Update existing instead of creating new records" section, which needs `update` rather than `create` permissions

src/app/core/import/import-entity-type/import-entity-type.component.html
sleidig, line 23: what about "Additional Actions" for imports? They need `update` permissions on the related type (e.g. I need to have permissions to edit Notes if as an additional action I am linking imported Children to a selected note)

sleidig (changes requested):

src/app/core/import/import-entity-type/import-entity-type.component.html
Abhinegi2: "Check/Update existing instead of creating new records" section is now only shown when the user has update permission on the selected target type

sleidig:
@coderabbitai review

sleidig (approved):
Looks good to me. Thanks for the good work 👍

src/app/core/import/import/import.component.ts
```diff
@@ -61,11 +62,34 @@ export class ImportComponent {
...
 
+  /**
+   * Whether the current user lacks permission to create ImportMetadata records.
+   * Every import writes an ImportMetadata history entry at the end, so without this
+   * permission no import can succeed - block the whole flow up front rather than
+   * letting the user prepare an import that would fail.
+   */
+  cannotImport =
+    this.ability.initialized && this.ability.cannot("create", ImportMetadata);
+
```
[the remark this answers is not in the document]
sleidig: @Abhinegi2

Abhinegi2:
@coderabbitai review

sleidig (approved):

PR #4160, opened 2026-07-21
feat(admin): structured UI to manage user role permissions

### LINKED ISSUE #4066: Admin UI: manage user role permissions
_As an admin I want to see and edit user roles in an intuitive UI in order to adapt the Config:Permissions config._

- UI that manages roles
  - create / edit / delete roles
- view + edit role permissions
  - visually presents all permissions
  - maybe use our conditions editor component to configure conditional rules?
  - add / remove subjects (using their entity labels)
- create/delete must edit the Config:Permissions doc but also use the Keycloak API to create a matching role

### Additional context

- restrict to only users with account_manager role
- consider the wildcard conditions: subject `all`, action `manage`

---

## Implementation Approach

- TODO

Abhinegi2:
- closes #4066

> [!IMPORTANT]
> **Depends on the reserved-role rename + config migration (#4168).** This UI uses the reserved keys `_default` / `_public`, but the runtime permission enforcer still reads the legacy `default` / `public` keys until that rename ships. Edits to the reserved roles here only take effect after the migration, so #4168 should merge first. Until then this PR reads the legacy keys as a fallback and never lists them as ordinary, deletable roles.

## PR Checklist

---

## Manual Testing Steps

Creating, deleting and describing roles happen in the user account server, so test on the dev system deployment of this branch, not in the demo-mode preview.

1. Log in as a user with the `admin_app` (or `account_manager`) role.
2. Open Admin > Roles & Permissions: the list shows the reserved `_default` and `_public` roles first, then all other roles with descriptions and rule counts. Reserved and technical roles (`_default`, `_public`, `account_manager`, `no-email-2fa`) show a lock icon and cannot be deleted. Roles without configured permissions show a fallback hint (falls back to the base role).
3. Click a role (e.g. `user_app`): the permission matrix shows Read / Create / Update / Delete / Manage (all) per record type. Internal system types (e.g. ConfigurableEnum) are greyed out with a lock icon and a readable label. Cells restricted by conditions show a filter chip.
4. Click "Edit": tick / untick action cells, tick "Manage (all)" on a row (the individual actions then show as covered), add a record type through the dropdown (internal types included), add the "All record types (wildcard)" row, remove a row, then Save. The "Undo" snackbar restores the previous state.
5. Conditions: on an entity row cell click "only where…", add one or more field conditions, switch between "Any" / "All", Apply and Save. Changing a field and then pressing Cancel must leave the matrix unchanged. Log in as a user with that role and verify only matching records are accessible.
6. Create a role: "Add New Role", check name validation (invalid characters, a leading `_`, duplicate name). The realm role is created first; if that fails a clear error is shown and nothing is written. Verify a valid new role appears in the list and can be assigned in Admin > User Accounts.
7. Delete a non-protected test role from its details page: a confirmation warns about the irreversible action; afterwards the role is gone from the list and from account role assignment. Protected roles have no Delete button and a read-only description.
8. "Edit JSON" (now in the three-dot menu on the list page) still allows raw editing of the complete permissions configuration, with backup + undo.
9. If the logged-in account lacks role-management permission on the user account server, the "Add New Role" and "Delete" actions are disabled with an explaining tooltip.

## Summary by CodeRabbit

* **New Features**
  * Added role management for creating, editing, deleting, and describing roles.
  * Added a permission matrix for record types, actions, wildcard access, and conditions.
  * Added condition editing with “all” and “any” matching options.
  * Added role pagination, permission counts, protected-role indicators, and raw JSON editing.
  * Added unsaved-change protection, validation, and reserved-role visibility.
* **Bug Fixes**
  * Improved authorization, accessibility messaging, and handling of missing or duplicate roles.
  * Preserved unsupported permissions and condition details during edits.
  * Improved compatibility with legacy permission configurations and prevented data loss after loading errors.


sleidig:
UI looks really good to me :heart: 

- [x] Maybe move the "Edit JSON" into a three-dot menu to hide the technical option from normal users?
- [x] show an icon and some slightly different coloring (?) for internal entity types in the list (e.g. slightly greyed out to stress the difference of those from custom entity types). And we probably need human readable labels for ones like ConfigurableEnum
- [x] padding missing:  image
- [x] `no-email-2fa` and `account_manager` have to be protected as internal roles (cannot delete).
  - Should we prevent these special roles to be used for other permission rules completely? (so that they only grant their other technical permission without overlap to data access) Or let them also serve as normal user roles in addition to their special function?
- [x] descriptions of protected roles (e.g. default, public, etc.) should maybe also remain read-only?

I haven't tested the full flows of creating, editing and deleting roles yet, however.

src/app/core/admin/admin-role-permissions/role-permissions.service.ts
```diff
@@ -0,0 +1,243 @@
...
+  DatabaseRules,
+} from "../../permissions/permission-types";
+import { Role } from "../../user/user-admin-service/user-account";
+import { UserAdminService } from "../../user/user-admin-service/user-admin.service";
+
+/**
+ * Reserved permission rules that apply to every logged-in user, combined with
+ * their assigned roles. Stored under this key in the Config:Permissions document.
+ */
+export const DEFAULT_ROLE = "_default";
```
Abhinegi2: **This UI reads/writes `_default` / `_public`, but the runtime permission system uses the unprefixed keys.**

- `ability.service.ts` resolves rules via `rules.public ?? []` and `if (rules.default) { ... }`
- `DatabaseRules` in `permission-types.ts` declares `public?` and `default?`
- `public-form-permission.service.ts` writes `permissionsConfig.data.public` / `.default`

Two consequences on any existing config:

1. The two virtual rows always render empty, and rules saved into them are read by nobody.
2. The real `default` / `public` keys then arrive through the `Object.keys(rules)` loop below as ordinary roles with `isProtected: false`, so they are **deletable from this UI**. `deleteRole` then issues `DELETE /roles/default` to Keycloak, gets a 404, treats it as success (intentionally, for orphan cleanup), and removes the `default` key from the config. Every logged-in user loses their base permissions.

Either land the reserved-role rename plus its config migration first and depend on it explicitly, or accept both spellings here until the migration has run.

src/app/core/admin/admin-role-permissions/condition-dialog/permission-condition-dialog.component.ts
```diff
@@ -0,0 +1,162 @@
...
+
+/**
+ * Convert any stored conditions shape into the { $or: [...] } row format
+ * that the conditions editor works with.
+ */
+function toEditorFormat(conditions: any): any {
+  if (!conditions || typeof conditions !== "object") {
+    return {};
+  }
+  if (Array.isArray(conditions.$or)) {
```
Abhinegi2: **`toEditorFormat` shares references with the matrix model, so the dialog edits it in place and Cancel does not cancel.**

`{ $or: conditions.$or }` reuses the same array *and* the same row objects that live in `model().rows[i].cells[action].conditions`. `ConditionsEditorComponent` mutates those directly: `conditions.splice(index, 1)` in `deleteCondition`, `delete condition[key]` / `condition[fieldKey] = null` in `onConditionFieldChange`, and `condition[fieldKey] = dbValue` in `onFormValueChange`.

So changing a field or deleting a row and then pressing Cancel leaves the model already modified. It isn't visible because the signal reference never changes and `viewRows` doesn't recompute, but the next Save persists it.

Deep-copying at the boundary fixes all three branches:

```ts
function toEditorFormat(conditions: any): any {
  if (!conditions || typeof conditions !== "object") return {};
  const copy = structuredClone(conditions);
  ...
}
```

src/app/core/admin/admin-role-permissions/roles-list/admin-roles-list.component.html
```diff
@@ -0,0 +1,109 @@
...
+        @if (role.rules) {
+          <span i18n>
+            {role.rules.length, plural,
+              one {1 rule}
+              other {{{ role.rules.length }} rules}
+            }
+          </span>
+        } @else if (role.name === publicRole) {
+          <span class="text-secondary" i18n>No permissions defined</span>
+        } @else {
```
Abhinegi2, line 81: The `_default` row itself lands in this branch and tells the user its permissions fall back to `_default`. Only `publicRole` is special-cased above, but `_default` has no rules of its own either.

Needs a `DEFAULT_ROLE` branch alongside `publicRole` (the component already exposes `publicRole` for exactly this).

src/app/core/admin/admin-role-permissions/permission-matrix/permission-matrix.component.html
```diff
@@ -0,0 +1,197 @@
...
+        [class.wildcard-row]="row.row.subject === 'all'"
+      ></tr>
+    </table>
+  </div>
+
+  @if (viewRows().length === 0) {
+    <div class="empty-state text-secondary padding-regular">
+      @if (editable()) {
+        <span i18n>No permissions yet. Add one below.</span>
+      } @else {
```
Abhinegi2, line 228: Same self-reference as in the roles list: when this matrix is showing the `_default` role, the empty state tells the user they "only have the permissions of the `_default` role".

Worth passing the role name in (`roleName()` is already an input) and skipping the fallback sentence when it *is* the default role.

src/app/core/admin/admin-role-permissions/role-details/admin-role-details.component.ts
```diff
@@ -0,0 +1,246 @@
...
+    this.nameControl.markAsTouched();
+    if (this.nameControl.invalid) return;
+
+    const name = this.nameControl.value;
+    try {
+      await this.rolePermissionsService.createRole(
+        name,
+        this.descriptionControl.value ?? "",
+        matrixToRules(this.model()),
+      );
```
Abhinegi2, line 242: This discards the specific error the service just built. `KeycloakAdminService.createRole` maps a 409 to `UserAdminApiError(409, $localize`A role with this name already exists.`)`, and that message is replaced here by "your account may not have permission".

It is reachable: `existingRoleNames` is populated asynchronously in `initNewRole`, so a user who types a name and saves before `loadRoles()` resolves skips the duplicate validator and gets the misleading message.

Prefer surfacing the API message when there is one, e.g. `err instanceof UserAdminApiError ? err.message : <generic>`.

src/app/core/admin/admin-role-permissions/condition-dialog/permission-condition-dialog.component.html
```diff
@@ -0,0 +1,44 @@
+<h2 mat-dialog-title>
+  <span i18n>Add Condition</span>
+  <app-dialog-close mat-dialog-close></app-dialog-close>
+</h2>
+
+<mat-dialog-content>
+  <p i18n>
```
Abhinegi2: This sentence is assembled from three placeholders plus a separately-translated verb, so it extracts as `{$INTERPOLATION} {$INTERPOLATION_1} {$INTERPOLATION_2} only where…` and `actionLabel` is its own `$localize` string ("can read"). Translators can't reorder it, which breaks most languages that don't follow subject-verb-object.

One parameterized message per action instead, e.g. `$localize`:permission condition:${roleName} can read ${entityLabel} only where…``, and drop `getActionLabel`.

src/app/core/admin/admin-role-permissions/roles-list/admin-roles-list.component.scss
```diff
@@ -0,0 +1,33 @@
+@use "variables/colors";
+@use "@angular/material/core/style/elevation" as mat-elevation;
+
+// same table styling as the user accounts list (user-list component)
+.table-container {
```
Abhinegi2: `.table-container`, `.table-footer`, `.table` and `.clickable-row` here are a verbatim copy of `user-list.component.scss`, and `permission-matrix.component.scss` copies part of the same block again. That's three copies of the list-table look.

Worth extracting into a shared partial (or a global class) so the tables can't drift apart.

src/app/core/admin/admin-role-permissions/permission-matrix/permission-matrix.component.ts
```diff
@@ -0,0 +1,313 @@
...
+              ? this.describeConditions(cell.conditions, row.subject)
+              : "",
+          };
+          return [action, state];
+        }),
+      ) as Record<EntityActionPermission, CellState>,
+    })),
+  );
+
+  actionLabel(action: EntityActionPermission): string {
```
Abhinegi2: This switch and `getActionLabel` in `permission-condition-dialog.component.ts` both enumerate the same five actions, one as "Read" and one as "can read". Adding a sixth action means remembering both.

Could both read from one label map kept next to `EntityActionPermission`?

src/app/core/admin/admin-role-permissions/role-details/admin-role-details.component.ts
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  async loadRole() {
+    const roles = await this.rolePermissionsService.loadRoles();
+    const role = roles.find((r) => r.name === this.roleName());
+    this.role.set(role);
+    this.model.set(rulesToMatrix(role?.rules ?? []));
+    this.descriptionControl.setValue(role?.description ?? "");
+    this.descriptionControl.markAsPristine();
+  }
```
Abhinegi2, line 138: `canManageRoles()` reads `sessionInfo.value` once at construction, so this never updates if the session resolves later, and the same snapshot plus a near-identical tooltip string is repeated in `AdminRolesListComponent`.

A `computed` on `RolePermissionsService` would stay reactive and let both components drop their local copy.

Abhinegi2 (commented):
Pre-review pass before the human review. Grouped as **should fix** vs **worth a look** below; inline comments carry the detail.

### Should fix

1. **`_default` / `_public` vs the runtime `default` / `public` keys** (`role-permissions.service.ts`) — the biggest one. The enforcer, `DatabaseRules` and `public-form-permission.service.ts` all use the unprefixed keys, so the two virtual rows are always empty, and the real `default` / `public` entries show up as ordinary deletable roles. Deleting `default` wipes base permissions for every logged-in user.
2. **The condition dialog mutates the matrix model in place**, so Cancel doesn't cancel (`permission-condition-dialog.component.ts`).
3. **The `_default` row tells the user it falls back to `_default`** (roles list + matrix empty state).
4. **The localized 409 "role already exists" message is built and then discarded** (`admin-role-details.component.ts`).
5. **The condition sentence is assembled from three placeholders plus a separately-translated verb**, so it can't be reordered by translators.

### Worth a look

Duplicated table SCSS (three copies now, including `user-list`), duplicated action-label switch, `canManageRoles` snapshot duplicated and non-reactive, `ngOnInit` + `paramMap.subscribe` where `resource()` would fit, method calls in templates, the `setTimeout` select-reset, a role-name pattern that allows the reserved `_` prefix, `unsupportedRules` reordering on round-trip, `EMPTY_MODEL` as a shared mutable const, the widened `permittedUserRoles`, "Keycloak" vs "user account server" wording, and an over-wide `Record<EntityActionPermission, ...>`.

### Not covered here

Visual and a11y items I couldn't check from the diff: the condition chip is a clickable `<span>` and rows are clickable `<tr>`s (neither keyboard-reachable), the dialog title stays "Add Condition" when editing an existing one, and `.cell-condition-chip` caps at 140px so long summaries truncate.

Also: the Manual Testing Steps say the list shows `default` and `public` first, but the code uses `_default` / `_public` — worth updating along with whatever you decide on item 1.

src/app/core/admin/admin-role-permissions/roles-list/admin-roles-list.component.ts
```diff
@@ -0,0 +1,106 @@
...
+    const start = this.pageIndex() * this.pageSize();
+    return this.roles().slice(start, start + this.pageSize());
+  });
+
+  readonly displayedColumns = ["name", "description", "permissions"];
+
+  readonly publicRole = PUBLIC_ROLE;
+
+  /** whether the user may create/delete roles in the authentication server */
+  readonly canManageRoles = this.rolePermissionsService.canManageRoles();
```
Abhinegi2: Same snapshot as in `AdminRoleDetailsComponent` (see the comment there), along with a near-duplicate `addDisabledTooltip` / `deleteDisabledTooltip` pair. One reactive accessor on the service would cover both.

src/app/core/admin/admin-role-permissions/role-details/admin-role-details.component.ts
```diff
@@ -0,0 +1,246 @@
...
+  readonly descriptionControl = new FormControl("");
+
+  private originalModel: MatrixModel = EMPTY_MODEL;
+  private existingRoleNames = new Set<string>();
+
+  constructor() {
+    // keep the description form state in sync with the edit mode
+    effect(() => {
+      if (this.descriptionEditable()) {
+        this.descriptionControl.enable();
```
Abhinegi2, line 95: `route.paramMap.subscribe(...)` has no `takeUntilDestroyed`, and it kicks off an async `loadRole()` whose result lands in signals.

`resource()` keyed off a route signal (`toSignal(route.paramMap)`) would drop the manual subscription, the ordering between `roleName.set` and `loadRole()`, and the leak question in one go. `AdminRolesListComponent.ngOnInit` has the same shape.

src/app/core/admin/admin-role-permissions/permission-matrix/permission-matrix.component.html
```diff
@@ -0,0 +1,197 @@
...
+              }
+            </div>
+          }
+        </td>
+      </ng-container>
+
+      @for (action of crudActions; track action) {
+        <ng-container [matColumnDef]="action">
+          <th mat-header-cell *matHeaderCellDef class="action-col">
+            {{ actionLabel(action) }}
```
Abhinegi2: `actionLabel(action)` re-runs on every change-detection cycle for every header cell, and `hasSubject("all")` further down walks all rows each cycle.

Both are pure derivations of static data / the model, so a `computed` (or a label baked into `crudActions`) would fit better. Same for the three `nameControl.hasError(...)` calls in the role-details template, which `toSignal(nameControl.statusChanges)` could cover.

src/app/core/admin/admin-role-permissions/permission-matrix/permission-matrix.component.ts
```diff
@@ -0,0 +1,313 @@
...
+    // the "all" wildcard row grants full access ({subject: "all", action: "manage"}),
+    // new entity rows start with read only
+    const cells: MatrixRow["cells"] =
+      subject === "all"
+        ? { manage: { allowed: true } }
+        : { read: { allowed: true } };
+    this.emitUpdated((m) => m.rows.push({ subject, cells }));
+  }
+
+  private resetAddSelect() {
```
Abhinegi2, line 519: Destroying and re-creating the select through a `setTimeout` to clear its value is doing a lot of work for a reset.

Can `app-entity-type-select` be cleared by rebinding `[value]`? If the remount really is necessary, a short comment on *why* would help the next reader (the JSDoc on `addSelectVisible` says what it does, not why a remount is required).

src/app/core/admin/admin-role-permissions/role-details/admin-role-details.component.ts
```diff
@@ -0,0 +1,246 @@
...
+  private readonly unsavedChanges = inject(UnsavedChangesService);
+
+  readonly roleName = signal("");
+  readonly role = signal<RoleWithPermissions | undefined>(undefined);
+  readonly model = signal<MatrixModel>(EMPTY_MODEL);
+  readonly editing = signal(false);
+  readonly isNew = signal(false);
+
+  readonly nameControl = new FormControl("", [
+    Validators.required,
```
Abhinegi2, line 80: This allows a leading underscore, so a user can create `_myrole` in Keycloak, even though this PR reserves the `_` prefix for the virtual roles. Blocking it keeps the convention enforceable:

```suggestion
    Validators.pattern(/^(?!_)[a-zA-Z0-9_-]+$/),
```

src/app/core/admin/admin-role-permissions/permission-matrix.ts
```diff
@@ -0,0 +1,188 @@
...
+        subject: f.subjects.length === 1 ? f.subjects[0] : f.subjects,
+        action: f.actions.length === 1 ? f.actions[0] : f.actions,
+        ...(f.conditions !== undefined ? { conditions: f.conditions } : {}),
+        ...(f.extra ?? {}),
+      }) as DatabaseRule,
+  );
+
+  // unsupported rules are appended last on purpose: later rules take precedence
+  // in CASL, so restrictions (inverted rules) always win over the matrix's allow rules
+  return [...rules, ...(model.unsupportedRules ?? [])];
```
Abhinegi2: Appending `unsupportedRules` last changes CASL precedence for configs where an `inverted` rule deliberately came *before* an allow rule: after a round-trip the restriction now wins where it previously didn't (or vice versa).

The comment explains why last is usually right, which is fair, but a config editor silently reordering rules on save is worth a second look. The existing spec asserts with `toContainEqual`, so position isn't covered either way.

src/app/core/admin/admin-role-permissions/role-details/admin-role-details.component.ts
```diff
@@ -0,0 +1,246 @@
...
+  MatrixModel,
+  matrixToRules,
+  rulesToMatrix,
+} from "../permission-matrix";
+import { PermissionMatrixComponent } from "../permission-matrix/permission-matrix.component";
+import {
+  RolePermissionsService,
+  RoleWithPermissions,
+} from "../role-permissions.service";
+
```
Abhinegi2, line 40: `EMPTY_MODEL` is a shared mutable module-level object used both as the signal seed and as `originalModel`. It's safe today only because `PermissionMatrixComponent.emitUpdated` clones before mutating.

A factory (`const emptyModel = () => ({ rows: [], unsupportedRules: [] })`) or `Object.freeze` would make that guarantee local instead of depending on the child's behaviour.

src/app/core/admin/admin.routing.ts
```diff
@@ -15,6 +18,8 @@ import { AdminPrimaryActionComponent } from "./admin-primary-action/admin-primar
 import { AdminAiAgentComponent } from "./admin-ai-agent/admin-ai-agent.component";
 import { AdminConfigCleanupComponent } from "./config-cleanup/admin-config-cleanup.component";
 
+const ROLE_MANAGEMENT_ROLES = ["account_manager", "admin_app"];
```
Abhinegi2, line 21: This widens the `user-roles` route from `["admin_app"]` to also allow `account_manager`. Intentional that account managers can now edit permission rules, not just user accounts?

src/app/core/admin/admin-role-permissions/role-details/admin-role-details.component.html
```diff
@@ -0,0 +1,92 @@
...
+    <input
+      matInput
+      [formControl]="nameControl"
+      i18n-placeholder
+      placeholder="e.g. field_supervisor"
+    />
+    @if (!isNew()) {
+      <fa-icon icon="lock" matSuffix></fa-icon>
+    }
+    <mat-hint i18n>
```
Abhinegi2, line 58: This hint names Keycloak, while every other new user-facing message in the PR says just "user account server". Worth picking one so the wording is consistent.

src/app/core/admin/admin-role-permissions/permission-matrix/permission-matrix.component.ts
```diff
@@ -0,0 +1,313 @@
...
+          const state: CellState = {
+            allowed: !!cell?.allowed,
+            hasCondition: !!cell?.conditions,
+            summary: cell?.conditions
+              ? this.describeConditions(cell.conditions, row.subject)
+              : "",
+          };
+          return [action, state];
+        }),
+      ) as Record<EntityActionPermission, CellState>,
```
Abhinegi2: The assertion claims a `manage` key, but only `crudActions` (read/create/update/delete) are mapped, so `actionStates.manage` is typed as present and is always `undefined`.

A narrow alias would keep it honest, e.g. `type CrudAction = "read" | "create" | "update" | "delete";` with `crudActions: CrudAction[]` and `Record<CrudAction, CellState>` here.

src/app/core/common-components/conditions-editor/conditions-editor.component.ts
```diff
@@ -49,6 +49,12 @@ export class ConditionsEditorComponent implements OnInit {
   @Input() disabled = false;
   @Input() label = $localize`Edit JSON`;
 
+  /** optional text replacing the default "any one criteria matches" explanation */
```
Abhinegi2, line 53: Minor / matching-local-style call: these are new inputs on a component that already mixes `@Input()` with `signal`/`computed`. `input()` would be the current house style, though I appreciate converting the whole component is a bigger change than this PR needs.

Abhinegi2:
@coderabbit review

Abhinegi2:
@coderabbit review

george-neha:
1. In the Add Permission for record type dropdown list, only show the record type in the dropdown that is not listed above.  Only show those record types that dont have any permissions recorded yet.
image

sadaf895 (approved):
This already looks good to me.

src/app/core/admin/admin-role-permissions/condition-dialog/permission-condition-dialog.component.html
```diff
@@ -0,0 +1,46 @@
...
+    <span class="text-secondary" i18n>
+      Records match if this applies to&hellip;
+    </span>
+    <mat-button-toggle-group
+      [value]="combinator()"
+      (change)="combinator.set($event.value)"
+      hideSingleSelectionIndicator
+    >
+      <mat-button-toggle value="any" i18n>Any</mat-button-toggle>
+      <mat-button-toggle value="all" i18n>All</mat-button-toggle>
```
sleidig: maybe these deserve a tooltip, explaining what it exactly means.

Could also move this to the top, above the conditions, so that it reads something like
> Users with role "user_app" can read <Task> only where…
> [Any][All] of the following conditions match
> ...

?

src/app/core/admin/admin-role-permissions/permission-matrix/permission-matrix.component.spec.ts
```diff
@@ -0,0 +1,374 @@
...
+    TestBed.inject(FaIconLibrary).addIconPacks(fas);
+    fixture = TestBed.createComponent(PermissionMatrixComponent);
+    component = fixture.componentInstance;
+    fixture.componentRef.setInput("model", model);
+    fixture.detectChanges();
+  });
+
+  it("renders one row per subject and summarizes conditions as a readable chip", () => {
+    const rows = fixture.nativeElement.querySelectorAll("tr[mat-row]");
+    expect(rows.length).toBe(2);
```
sleidig: very minor, but sonarcloud flags this, recommending to use `toHaveLength`: https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=4160&id=Aam-Digital_ndb-core&open=AZ-DyKWS0D7dobw1ccFK

src/app/core/admin/admin-role-permissions/permission-matrix/permission-matrix.component.html
```diff
@@ -0,0 +1,266 @@
...
+        <td mat-cell *matCellDef="let v" class="subject-col">
+          @if (v.isDefaultRow) {
+            <div class="flex-row align-center gap-regular">
+              <div class="wildcard-badge">
+                <fa-icon icon="lock"></fa-icon>
+              </div>
+              <div>
+                <div>{{ v.label }}</div>
+                <div class="text-secondary wildcard-subtitle" i18n>
+                  applies to any logged-in user, in addition to their roles
```
sleidig: maybe centralize these (name, description) somewhere in an object describing the default roles? (we may want to reuse it in a few places, e.g. in the individual feature permissions dialog)

src/app/core/admin/admin-role-permissions/permission-matrix/permission-matrix.component.html
```diff
@@ -0,0 +1,266 @@
...
+  @if (viewRows().length === 0) {
+    <div class="empty-state text-secondary padding-regular">
+      @if (editable()) {
+        <span i18n>No permissions yet. Add one below.</span>
+      } @else if (isDefaultRole()) {
+        <span i18n>No permissions defined for this role yet.</span>
+      } @else {
+        <span i18n>
+          No permissions defined for this role yet. Users with this role only
+          have the permissions of the base role.
```
sleidig: Let's also add a small note to the end of the table that is always visible, saying "Logged-in users also always have the permissions defined in the [_default role][LINK_TO_ROLE]"

src/app/core/admin/admin-role-permissions/permission-matrix.ts
```diff
@@ -0,0 +1,215 @@
...
+  (rules ?? []).forEach((rule, index) => {
+    if (!isSupported(rule)) {
+      unsupportedRules.push(rule);
+      unsupportedRuleIndices.push(index);
+      return;
+    }
+
+    const subjects = (
+      Array.isArray(rule.subject) ? rule.subject : [rule.subject]
+    ) as string[];
```
sleidig: could use our utils `asArray()` function for this. Makes it slightly faster to read

src/app/core/user/admin-user-roles/admin-user-roles.component.scss
sleidig, line 1: I think I would tend to put the whole user roles admin functionality added in this PR into the `/user` module rather than `/admin` (that already contains a ton of different admin things that are not necessarily connected, while /user has the logic for creating accounts already)

sleidig (changes requested):
Did a final screening of UX and code with some last remarks. Looks pretty good though

src/app/core/admin/admin-role-permissions/condition-dialog/permission-condition-dialog.component.html
sleidig: Done in 3963fb19b — both suggestions.

The toggle now sits above the conditions it applies to, so the dialog reads as one sentence:

> user_app can read Task only where…
> [Any|All] of the following conditions match
> …conditions…

And each option has a tooltip spelling out what it means ("A record matches as soon as one of the conditions below applies (\"or\")" / "…only if every one…(\"and\")").

Two notes:

- The toggle line is a single `i18n` message with the toggle group as an element placeholder, rather than a separate `<span i18n>of the following conditions match</span>` — that fragment would be unreorderable and word order varies by language. Verified via `ng extract-i18n` that it comes out as one reorderable unit.
- I kept passing `[hint]="combinatorHint()"` to the conditions editor. Its own default hint hardcodes "any one … (\"or\" conditions)", which would be actively wrong once "All" is selected.

src/app/core/admin/admin-role-permissions/permission-matrix/permission-matrix.component.spec.ts
sleidig: Done in 89f8e5dd9 — switched all 9 occurrences across the specs added in this PR (`permission-matrix.spec.ts`, `role-permissions.service.spec.ts`, `permission-matrix.component.spec.ts` ×5, `entity-type-select.component.spec.ts`), not just this one.

src/app/core/admin/admin-role-permissions/permission-matrix/permission-matrix.component.html
sleidig: Done in 501651385.

Added a permanent footnote below the matrix:

> Logged-in users also always have the permissions defined in the [_default](/admin/user-roles/_default) role.

Since it is now always visible, the empty state no longer needs to repeat the fallback and collapses to a single "No permissions defined for this role yet." for all roles. The note is hidden on `_default` and `_public`, which do not inherit `_default` themselves.

While wiring this up I found a related bug and fixed it in 04dacc7ad: `AbilityService.getRulesForUser` returns the `_public` rules **on their own** for visitors who are not logged in — `_default` is never prepended there. But `setInheritedRules` excluded only `_default`, so opening `_public` rendered a read-only "Default" row and locked its checkboxes with a tooltip claiming those permissions cannot be revoked. Both places now use the shared `RESERVED_RULE_CONFIG_KEYS` (which also covers the legacy non-prefixed spellings).

src/app/core/admin/admin-role-permissions/permission-matrix.ts
sleidig: Done in 89f8e5dd9 — used `asArray()` here and in the two equivalent spots in `permission-matrix.component.ts` (`inheritedGrants`).

Left alone deliberately: `Array.isArray(selected) ? selected[0] : selected` in `addSubject()` (takes the first element, not a wrap) and `f.subjects.length === 1 ? f.subjects[0] : f.subjects` in `matrixToRules()` (the array→scalar inverse).

src/app/core/admin/admin-role-permissions/permission-matrix/permission-matrix.component.ts
```diff
@@ -317,9 +318,15 @@ export class PermissionMatrixComponent {
...
-    () => this.roleName() === DEFAULT_SECTION_KEY,
+  /**
+   * Whether users of this role also receive the shared "_default" rules.
+   * False for the reserved roles: "_default" cannot inherit from itself and
+   * "_public" applies to visitors who are not logged in, who never receive the
+   * "_default" rules (see AbilityService.getRulesForUser). Their empty state
+   * therefore differs as well, having no base role to fall back to.
+   */
+  readonly inheritsFromDefaultRole = computed(
+    () => !RESERVED_RULE_CONFIG_KEYS.includes(this.roleName()),
```
sleidig: better hard-code _default and _public here? We might add reserved roles later that also inherit _default (because they apply to authenticated users)

sleidig:
@Abhinegi2 , I have made some tweaks. The most fundamental one: move legacy roles migration into a separate function and try to remove the references from other code.

sleidig:
@coderabbitai review

src/app/core/user/admin-user-roles/admin-user-roles.component.scss
sleidig: I'll leave this for you, @Abhinegi2 , so that you can first look over the other tweaks before files move

sleidig:
@coderabbitai review

src/app/core/admin/admin-role-permissions/permission-matrix.ts
```diff
@@ -0,0 +1,212 @@
...
+    return [...rules, ...unsupported];
+  }
+
+  const result = [...rules];
+  unsupported
+    .map((rule, i) => ({ rule, index: indices[i] }))
+    .sort((a, b) => a.index - b.index)
+    .forEach(({ rule, index }) =>
+      result.splice(Math.min(index, result.length), 0, rule),
+    );
```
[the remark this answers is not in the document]
sleidig: This is acceptable here as we strongly advice against using inverted rules

PR #4167, opened 2026-07-22
feat(Admin UI): manage feature permissions from entity admin lists

### LINKED ISSUE #4142: Admin UI to manage permissions for features / internal entity types on the relevant admin view
internal entity types that should prompt admins to review feature permission:
- `ConfigurableEnum`
- `ImportMetadata`
- `EmailTemplate`
- `TemplateExport`
- `PublicFormConfig`
- `ReportConfig`
- `AuditRecord` (?)

--> on their admin screens, include a UI that informs the admin about who has access to this feature and allows to edit that.

e.g. on Lists like "Export Templates" (https://test.aam-digital.net/c/admin/template-export), show something like the following:

Image

sadaf895:
- closes https://github.com/Aam-Digital/ndb-core/issues/4142

## PR Checklist

---

## Where it appears

On internal "feature" entity types (Export Templates, Email Templates, Public Forms, Report Config, ...) a **"Configure Permissions" button** sits next to "Add New", explaining itself through a tooltip ("Access to this feature is controlled by user roles"). On mobile, where the header carries no actions, the same entry is in the list's overflow (⋮) menu.

Which lists offer this follows the existing `isInternalEntity` flag on the entity class, so no extra configuration is needed. Regular entity types deliberately do not offer it for now (their list menu is already crowded).

The action is only shown to users who may edit the permissions config, derived from CASL (`update` on `Config`) rather than a hardcoded role name.

## The dialog

One row per user role, one checkbox per action: **Add / Read / Update / Delete**.

- The shared `_default` section is listed as a read-only first row ("Default", _applies to any logged-in user_), so it is visible what every logged-in user may already do.
- An action granted by `_default` is checked and disabled on every role row, with a tooltip saying why: revoking it for a single role would need an inverted rule, which this dialog cannot express. The role's other actions stay editable.
- A role whose access comes from a rule the dialog does not own (wildcard `all`, grouped subjects, conditions, inverted rules, backend-managed rules) is read-only for the whole row, with a link to the advanced JSON editor. In the default config `user_app` and `admin_app` hold `{ subject: "all", action: "manage" }`, so both appear locked.
- Saving writes one rule per role for this entity type: a plain action string, an array of actions, or `"manage"` when all four are selected, so a config keeps the shape an admin would write by hand. Every other rule is preserved untouched, and actions inherited from `_default` are not duplicated into the role.
- Each save stores a timestamped backup (`Config:Permissions:<timestamp>`) and offers an "Undo".

## Also in here

- Public form permission changes now go through the same shared service, so they are backed up like every other permission change, and a failed config load is no longer reported as "public users have no access".
- Subject/action matching of permission rules lives in one place, which also fixes rules written as `action: ["create", "read"]` not being recognised by the public form check.

## Manual Testing Steps

1. Log in as a user who may edit permissions (e.g. `admin_app`).
2. Open **Admin > Export Templates**. Next to "Add New" there is a "Configure Permissions" button; hovering it explains that access is controlled by user roles.
3. Click it. The dialog lists "Default" first, then every user role with its description, with checkboxes for Add, Read, Update and Delete.
4. Tick "Read" for a role that had no access and confirm. A message appears with an "Undo" action; reopening the dialog shows the checkbox still ticked.
5. Tick all four actions for a role and confirm, then check in **Admin > User Roles** (advanced editor) that the role has a single `manage` rule for this entity type. Untick one action again and it becomes a list of the remaining actions.
6. In the advanced editor, give the `_default` section read access to this entity type. Reopen the dialog: "Read" is now ticked and greyed out for every role, with a tooltip explaining why, while Add/Update/Delete stay editable.
7. Open a regular list such as **Children**: no "Configure Permissions" button in the header and no such entry in the ⋮ menu.
8. Log in as a user without permission to edit permissions: the button is not shown at all.

## Not in this PR

- **Configurable Enums** - no admin list screen exists yet, so there is no place to offer this.
- **Import Metadata** - covered by #4144.
- **AuditRecord** - has no entity type of its own.

## Summary by CodeRabbit

- **New Features**
  - Added per-feature permission management for eligible internal entities.
  - Administrators can grant or edit role-based **Use** and **Manage** access through a dedicated dialog.
  - Added support for permission previews, read-only rules, advanced configuration links, and role descriptions.
  - Added automatic backups and undo support when saving permission changes.
  - Improved permission handling for public forms and administrative role settings.

- **Bug Fixes**
  - Prevented unnecessary role-server requests in demo or preview environments.

- **Style**
  - Standardized header action button appearance and alignment.


Abhinegi2:
@coderabbit review

src/app/core/permissions/feature-permission/feature-permission.service.ts
```diff
@@ -0,0 +1,310 @@
...
+ */
+@Injectable({ providedIn: "root" })
+export class FeaturePermissionService {
+  private readonly entityMapper = inject(EntityMapperService);
+  private readonly sessionInfo = inject(SessionSubject);
+
+  /**
+   * Whether the current user is allowed to edit the permissions config at all.
+   */
+  hasAdminPermission(): boolean {
```
Abhinegi2: **Should fix** – `hasAdminPermission()` is a verbatim copy of [`public-form-permission.service.ts:166`](https://github.com/Aam-Digital/ndb-core/blob/140dc0f687b1a34bf67604bc295c8c160c4fea5d/src/app/features/public-form/public-form-permission.service.ts#L166). Same story for `subjectMatches()` (line 267 here vs [`:35` there](https://github.com/Aam-Digital/ndb-core/blob/140dc0f687b1a34bf67604bc295c8c160c4fea5d/src/app/features/public-form/public-form-permission.service.ts#L35)) and `loadPermissionsConfig()` (line 284 here vs [`:261` there](https://github.com/Aam-Digital/ndb-core/blob/140dc0f687b1a34bf67604bc295c8c160c4fea5d/src/app/features/public-form/public-form-permission.service.ts#L261)).

Could we pull those three into one shared permissions-config service and have both consumers use it? Otherwise a bugfix in one of them lands in only half the app.

src/app/core/permissions/feature-permission/feature-permission.service.ts
```diff
@@ -0,0 +1,310 @@
...
+   * A timestamped backup of the previous config is stored before saving.
+   *
+   * @returns the backup Config that was created, so callers can offer an "undo".
+   */
+  async setPermissions(
+    entityType: string,
+    updates: Pick<RoleFeaturePermission, "role" | "use" | "manage">[],
+  ): Promise<Config<DatabaseRules>> {
+    const config = await this.loadOrInitPermissionsConfig();
+    const backup = new Config<DatabaseRules>(
```
Abhinegi2: **Should fix** – this backup-then-undo flow (here plus `offerUndo()` in `feature-permission-dialog.component.ts:200`) is line-for-line the same as [`admin-user-roles.component.ts:39-57`](https://github.com/Aam-Digital/ndb-core/blob/140dc0f687b1a34bf67604bc295c8c160c4fea5d/src/app/core/user/admin-user-roles/admin-user-roles.component.ts#L39-L57): same `Config.PERMISSION_KEY + ":" + moment().format("YYYY-MM-DD_HH-mm-ss")` id, same 8000ms Undo snackbar, same load / restore / remove sequence.

Worth extracting `saveWithBackup()` + `offerUndo()` once and switching `AdminUserRolesComponent` over to it, so the two undo paths can't drift.

src/app/core/permissions/feature-permission/feature-permission-banner/feature-permission-banner.component.scss
```diff
@@ -0,0 +1,12 @@
+.banner {
```
Abhinegi2: **Should fix** – the whole file is already covered by existing global utility classes, so we can delete it and drop the `styleUrl`:

```html
<app-hint-box type="info" class="banner margin-bottom-regular">
  <div class="flex-row flex-wrap align-center gap-regular">
```

(`.banner` keeps `display: block` from `hint-box`; if it doesn't, `display-block` is available too.)

Abhinegi2 (commented):
Review notes, grouped as **Should fix** (duplication and house-style) and **Worth a look** (design questions worth a decision, not necessarily changes).

Good things first: no leftover console logs, no `markForCheck`, no `fdescribe`/`fit`, no `*ngIf`/`*ngFor`, all `inject()` fields are `readonly`, and every user-visible string carries `i18n`. The `_default`/legacy-`default` fallback and reserved-key handling match `AbilityService.getRulesForUser()` exactly, backup docs use distinct ids so they don't trigger a spurious `AbilityService` reload, and the service has solid branch coverage. The README section is a genuinely good addition.

Two things to check by eye that don't fit on a line:

- On a default install both `user_app` and `admin_app` hold `{subject: "all", action: "manage"}` (`src/assets/base-configs/basic/Config_Permissions.json`), so every shipped role renders locked and the dialog is read-only until an admin has already hand-edited the JSON. The README acknowledges this, but worth confirming with product that the feature still delivers what #4142 asked for out of the box.
- The lock tooltip points at "the advanced permissions editor" but gives no way to get there. Consider linking to `/admin/user-roles`.

src/app/core/permissions/feature-permission/feature-permission-dialog/feature-permission-dialog.component.scss
```diff
@@ -0,0 +1,48 @@
...
+
+  th,
+  td {
+    padding: 8px 12px;
+    text-align: left;
+  }
+
+  thead th {
+    border-bottom: 1px solid rgba(0, 0, 0, 0.12);
+    font-weight: 500;
```
Abhinegi2: **Should fix** – `font-weight` in component scss (also line 40). Those should come from the Material typography classes rather than being set per component.

Also lines 12 and 17 hardcode `rgba(0, 0, 0, 0.12)` / `rgba(0, 0, 0, 0.06)` borders. Please use the `$border-color` variable from `src/styles/variables/_colors.scss` so this follows the theme.

src/app/core/permissions/feature-permission/feature-permission-dialog/feature-permission-dialog.component.html
```diff
@@ -0,0 +1,107 @@
...
+                class="checkbox-tooltip-wrapper"
+                matTooltip="'Use' is included automatically when 'Manage' is selected, so it can't be changed on its own."
+                i18n-matTooltip="Use checkbox disabled tooltip"
+                [matTooltipDisabled]="!row.editable || !row.manage"
+              >
+                <mat-checkbox
+                  [checked]="row.use || row.manage"
+                  [disabled]="!row.editable || row.manage"
+                  (change)="setUse(row.role, $event.checked)"
+                  [attr.aria-label]="'Use ' + entityLabel + ' as ' + row.role"
```
Abhinegi2: **Should fix** – these aria-labels build an English sentence by string concatenation inside the binding (same on line 80), so they can never be extracted for translation. Screen-reader users on a non-English instance get English here.

Build the label with `$localize` when mapping the rows in `ngOnInit` and bind the precomputed value:

```ts
useLabel: $localize`:Use checkbox aria label:Use ${entityLabel} as ${role}`,
```

src/app/core/permissions/feature-permission/feature-permission.service.spec.ts
```diff
@@ -0,0 +1,429 @@
...
+      ],
+    });
+    service = TestBed.inject(FeaturePermissionService);
+  });
+
+  it("should be created", () => {
+    expect(service).toBeTruthy();
+  });
+
+  describe("hasAdminPermission", () => {
```
Abhinegi2: **Should fix** – five nested `describe` blocks in this file (lines 57, 70, 267, 351, 369). House style here is flat `it()` with the setup inlined per test, so the nesting should go.

src/app/core/entity-list/entity-list/entity-list.component.html
```diff
@@ -27,6 +27,12 @@
       }
     </app-view-actions>
 
+    @if (managePermissions() && entityConstructor()) {
```
Abhinegi2: **Should fix** – this identical 5-line block is pasted again at line 95 for the mobile branch. This file already solves that pattern with `<ng-template #filterDialog>` / `#bulkActions` / `#subrecord`, so a `#permissionBanner` template plus `*ngTemplateOutlet` in both places would fit right in.

It also lets you drop the `entityConstructor()!` non-null assertions, since the `@if` guard can live inside the template.

src/app/core/permissions/feature-permission/feature-permission.service.ts
```diff
@@ -0,0 +1,310 @@
...
+      !rule.conditions &&
+      !rule.reason?.includes("[system-default]") &&
+      rule.subject === entityType &&
+      (rule.action === FEATURE_USE_ACTION ||
+        rule.action === FEATURE_MANAGE_ACTION)
+    );
+  }
+
+  /** Whether the rule grants "manage" on this entity type (ignoring conditions). */
+  private grantsManage(rule: DatabaseRule, entityType: string): boolean {
```
Abhinegi2: **Worth a look** – `grantsManage()` and `grantsRead()` (line 258) skip any rule with `inverted: true`. So a role with `{subject: "all", action: "manage"}` plus a later `{subject: ENTITY_TYPE, action: "manage", inverted: true}` deny renders with both boxes checked, i.e. the dialog reports access the user does not actually have.

CASL resolves last-match-wins, so ignoring inverted rules inverts the answer rather than just being incomplete. Two reasonable ways out: honour them, or treat any role that has an inverted rule for this subject as read-only with unknown state. There's no test covering an inverted rule today either.

src/app/core/permissions/feature-permission/feature-permission.service.ts
```diff
@@ -0,0 +1,310 @@
...
+  /**
+   * true if at least one role's access is granted by rules the grid cannot edit
+   * (wildcards, grouped subjects, conditions or shared `_default` rules) - i.e. some
+   * rows are read-only.
+   *
+   * When true the UI should point admins to the advanced (raw JSON) permissions
+   * editor for the full picture. Editing via the grid stays safe either way:
+   * {@link FeaturePermissionService.setPermissions} never mutates those rules.
+   */
+  hasComplexRules: boolean;
```
Abhinegi2, line 88: **Worth a look** – `hasComplexRules` is computed and asserted in 8 tests, but no template or component ever reads it. Its own JSDoc says "the UI should point admins to the advanced editor" and the dialog doesn't.

Either wire it up (a hint linking to `/admin/user-roles` when it's true) or drop the field until it has a consumer.

src/app/core/permissions/feature-permission/feature-permission-banner/feature-permission-banner.component.ts
```diff
@@ -0,0 +1,54 @@
...
+})
+export class FeaturePermissionBannerComponent {
+  private readonly permissionService = inject(FeaturePermissionService);
+  private readonly dialog = inject(MatDialog);
+
+  /** the feature entity type whose permissions are managed here */
+  readonly entityType = input.required<EntityConstructor>();
+
+  /** whether the current user may edit the permissions config */
+  readonly canManage = computed(() =>
```
Abhinegi2: **Worth a look** – this `computed()` wraps `hasAdminPermission()`, which reads `SessionSubject.value`, not a signal. The computed therefore has zero dependencies and will never recompute, so a role change mid-session won't show/hide the banner.

Either make it a plain `readonly` field (honest about being a one-shot read) or derive it reactively the way [`ui.component.ts:181`](https://github.com/Aam-Digital/ndb-core/blob/140dc0f687b1a34bf67604bc295c8c160c4fea5d/src/app/core/ui/ui/ui.component.ts#L181) does with `toSignal(sessionSubject.pipe(map(...)))`.

src/app/core/permissions/feature-permission/feature-permission-dialog/feature-permission-dialog.component.ts
```diff
@@ -0,0 +1,225 @@
...
+
+  readonly entityType = this.data.entityType;
+  readonly entityLabel = this.data.entityLabel ?? this.data.entityType;
+
+  /** `undefined` while loading */
+  readonly roles = signal<RolePermissionRow[] | undefined>(undefined);
+  readonly loadError = signal(false);
+  readonly saving = signal(false);
+
+  async ngOnInit(): Promise<void> {
```
Abhinegi2: **Worth a look** – async load in `ngOnInit` with manual `loadError` / `undefined` state is exactly what `resource()` handles for us. The closest sibling, [`public-form-permission-warning.component.ts:53`](https://github.com/Aam-Digital/ndb-core/blob/140dc0f687b1a34bf67604bc295c8c160c4fea5d/src/app/features/public-form/public-form-permission-warning/public-form-permission-warning.component.ts#L53), already uses `resource()` for the same shape of work.

src/app/core/permissions/feature-permission/feature-permission.service.ts
```diff
@@ -0,0 +1,310 @@
...
+    return Object.keys(rules).filter((role) => !this.isReservedSection(role));
+  }
+
+  /**
+   * Whether the config key carries special semantics instead of naming a user
+   * role, so the per-role grid must neither list nor rewrite it. Covers the
+   * reserved underscore prefix as well as the legacy (non-prefixed) spellings
+   * of not yet migrated configs.
+   */
+  private isReservedSection(key: string): boolean {
```
Abhinegi2: **Worth a look** – `isReservedSection()` repeats the exact guard from [`ability.service.ts:106`](https://github.com/Aam-Digital/ndb-core/blob/140dc0f687b1a34bf67604bc295c8c160c4fea5d/src/app/core/permissions/ability/ability.service.ts#L106). Since the two must always agree about what counts as reserved, exporting it once from `permission-types.ts` next to `RESERVED_RULE_CONFIG_KEYS` would keep them in sync.

src/app/core/permissions/feature-permission/feature-permission-dialog/feature-permission-dialog.component.ts
```diff
@@ -0,0 +1,225 @@
...
+        $localize`Could not save permissions. Please try again.`,
+        undefined,
+        { duration: 5000 },
+      );
+    } finally {
+      this.saving.set(false);
+    }
+  }
+
+  private offerUndo(backup: Config): void {
```
Abhinegi2: **Worth a look** – the undo path is untested. `offerUndo()` writes the permissions doc directly with `save(config, true)` (force-overwriting rev conflicts) and then removes the backup, which makes it the riskiest code in the PR, and neither it, `setUse()`, nor the error branch of `confirm()` has a test. Worth covering at least the restore-then-remove sequence.

src/app/core/permissions/feature-permission/feature-permission.service.ts
```diff
@@ -0,0 +1,310 @@
...
+      }
+
+      if (updated.length > 0) {
+        config.data[role] = updated;
+      } else {
+        delete config.data[role];
+      }
+    }
+
+    await this.entityMapper.save(backup);
```
Abhinegi2: **Worth a look** – every Confirm writes a `Config:Permissions:<timestamp>` doc that is only removed if the user clicks Undo within 8 seconds. Each one is a full copy of the permissions doc and replicates to every client, so on an instance where admins tune permissions regularly these accumulate forever.

Is there a cleanup plan (this predates the PR in `AdminUserRolesComponent` too), or should the backup be dropped once the snackbar dismisses without action?

src/app/core/permissions/feature-permission/feature-permission.service.ts
```diff
@@ -0,0 +1,310 @@
...
+      return await this.entityMapper.load<Config<DatabaseRules>>(
+        Config,
+        Config.PERMISSION_KEY,
+      );
+    } catch {
+      return null;
+    }
+  }
+
+  private async loadOrInitPermissionsConfig(): Promise<Config<DatabaseRules>> {
```
Abhinegi2: **Worth a look** – `loadOrInitPermissionsConfig()` seeds `_default: [{subject: "all", action: "manage"}]` when no config exists, but [`README.md:169`](https://github.com/Aam-Digital/ndb-core/blob/140dc0f687b1a34bf67604bc295c8c160c4fea5d/src/app/core/permissions/README.md#L169) states the dialog never writes `_default`/`_public`. Worth a sentence in the README noting the first-write seeding, so the two don't read as contradicting each other.

src/app/core/permissions/feature-permission/feature-permission.service.ts
```diff
@@ -0,0 +1,310 @@
...
+ * underlying CASL actions stored in the permissions config.
+ *
+ * This is the single source of truth for that mapping: if "Use" should ever grant
+ * more than read access (e.g. also create/update), change {@link FEATURE_USE_ACTION}.
+ *
+ * Semantics (confirmed with product): "read" lets a role *use* a feature
+ * (e.g. send emails from an EmailTemplate) without being able to add/edit its
+ * records, while "manage" grants full control.
+ */
+export const FEATURE_USE_ACTION: EntityActionPermission = "read";
```
Abhinegi2: **Worth a look** – `FEATURE_USE_ACTION` and `FEATURE_MANAGE_ACTION` are exported but referenced only inside this file. Make them module-private unless something else is meant to consume them.

src/app/core/permissions/feature-permission/feature-permission.service.spec.ts
```diff
@@ -0,0 +1,429 @@
...
+      mockConfig({});
+
+      await service.setPermissions(ENTITY_TYPE, [
+        { role: "user_app", use: true, manage: false },
+        { role: "admin_app", use: true, manage: true },
+      ]);
+
+      expect(savedPermissions()).toEqual({
+        user_app: [{ subject: ENTITY_TYPE, action: "read" }],
+        admin_app: [
```
Abhinegi2: **Worth a look** – `setPermissions` writes both a `read` and a `manage` rule when Manage is checked, as asserted here. In CASL `manage` already implies `read`, so the extra rule is redundant in the stored doc. Intentional (e.g. so the grid can read the two flags back independently), or noise we could drop?

src/app/core/entity-list/EntityListConfig.ts
```diff
@@ -40,6 +40,14 @@ export interface EntityListConfig {
...
    */
   defaultSort?: Sort;
+
+  /**
+   * If true, show a banner on this (admin) list that lets admins review and edit
+   * which user roles can "use" or "manage" this feature entity type.
+   * Intended for internal "feature" entity types (e.g. EmailTemplate,
+   * TemplateExport, ReportConfig) whose access is controlled per role.
+   */
+  managePermissions?: boolean;
```
Abhinegi2: **Worth a look** – `managePermissions` is a per-view config flag, which means every feature list has to opt in by config, and per the PR description `ReportConfig` can't use it at all because it registers no `EntityList` view.

Could this be derived from the entity class instead (e.g. `isInternalEntity` plus an opt-in static on the entity), so it also covers screens that aren't `EntityList`-based and doesn't need a config migration per feature?

src/app/core/permissions/feature-permission/feature-permission.service.ts
Abhinegi2: Pre-existing.

Abhinegi2:
@coderabbit review

Abhinegi2:
@coderabbit review

george-neha:
Looks good to me.

src/app/core/entity-list/entity-list/entity-list.component.ts
```diff
@@ -246,6 +268,22 @@ export class EntityListComponent<T extends Entity> implements OnInit {
...
   }
 
+  /** Review and edit which user roles have access to this entity type. */
+  openPermissionsDialog() {
+    const entityConstructor = this.entityConstructor();
+    if (!entityConstructor) {
+      return;
+    }
+
+    this.dialog.open(FeaturePermissionDialogComponent, {
```
sleidig, line 288: Using the "Configure Permissions" button somehow triggers the app to ask for browser permissions for me:
Image

src/app/core/permissions/feature-permission/feature-permission-dialog/feature-permission-dialog.component.html
```diff
@@ -0,0 +1,117 @@
...
+          default or an advanced permission rule.
+        </p>
+      }
+      <a
+        mat-button
+        color="accent"
+        routerLink="/admin/user-roles"
+        mat-dialog-close
+        i18n="Open advanced permissions editor link"
+        >Open advanced editor</a
```
sleidig:
```suggestion
        >Manage user roles</a
```

src/app/core/permissions/feature-permission/feature-permission-dialog/feature-permission-dialog.component.html
```diff
@@ -0,0 +1,117 @@
...
+    <mat-progress-bar mode="indeterminate"></mat-progress-bar>
+  } @else if (permissionRows.error()) {
+    <app-hint-box type="warning">
+      <span i18n>
+        Could not load the list of user roles. Please try again later or edit
+        permissions in the advanced settings.
+      </span>
+    </app-hint-box>
+  } @else {
+    <app-hint-box type="warning" class="permission-hint">
```
sleidig: I think this is not a critical warning and the grey hint box probably is enough. (The orange looks nice but really grabs a lot of attention)

```suggestion
    <app-hint-box class="permission-hint">
```

src/app/core/permissions/feature-permission/feature-permission-dialog/feature-permission-dialog.component.html
```diff
@@ -0,0 +1,117 @@
...
+      <p i18n="Feature permission dialog read hint">
+        Give at least <strong>Read</strong> permissions, if users should be able
+        to use the feature (i.e. create files).
+      </p>
+      @if (hasComplexRules()) {
+        <p i18n="Feature permission dialog complex rules hint">
+          Some permissions are shown read-only because they come from the
+          default or an advanced permission rule.
+        </p>
+      }
```
sleidig: We explain that in the tooltips of disabled checkboxes, right? I would tend to remove that note here, in order to make the dialog less text-heavy

src/app/core/permissions/feature-permission/feature-permission-dialog/feature-permission-dialog.component.ts
```diff
@@ -0,0 +1,338 @@
...
+
+/** one checkbox of a row, with everything the template needs precomputed */
+interface PermissionCell {
+  action: FeatureAction;
+  granted: boolean;
+  editable: boolean;
+  /** why this checkbox cannot be changed; empty when it is editable */
+  lockTooltip: string;
+  ariaLabel: string;
+}
```
sleidig: can we reuse the interfaces from #4160 that are about to be merged?

src/app/core/permissions/feature-permission/feature-permission-dialog/feature-permission-dialog.component.ts
```diff
@@ -0,0 +1,338 @@
...
+      return $localize`:Default permissions row tooltip:These permissions apply to every logged-in user, in addition to their roles. They can only be changed in the advanced permissions editor.`;
+    }
+    return this.cellLockTooltip(permission.actions.read);
+  }
+
+  /** the tooltip explaining why a checkbox cannot be changed */
+  private cellLockTooltip(permission: FeatureActionPermission): string {
+    switch (permission.lockedBy) {
+      case "default":
+        return $localize`:Permission locked by default rules tooltip:Granted to all logged-in users by the default permissions, so it cannot be changed for a single role here.`;
```
sleidig: let's make sure we only reuse texts from the user role UI (#4160) for consistency and to limit translation effort.

src/app/core/permissions/feature-permission/feature-permission-dialog/feature-permission-dialog.component.ts
```diff
@@ -0,0 +1,338 @@
...
+      case "advanced-rule":
+        return $localize`:Permission locked by advanced rule tooltip:This role's access comes from an advanced permission rule and can only be changed in the advanced permissions editor.`;
```
sleidig: "advanced" mostly means subject=all? or are there other common patterns?

src/app/core/permissions/README.md
```diff
@@ -88,6 +89,11 @@ Admins edit this document to control what each role can do.
 
 - Use CouchDB Fauxton GUI or another database tool to edit the document directly
```
sleidig, line 91: we should actually remove this (both from README and from the Admin Overview). Instead, there is the Admin > Users > User Roles screen. And if necessary, that allows to open the JSON from its context menu also.

@Abhinegi2 , can you still do that in the other PR?

src/app/core/permissions/README.md
```diff
@@ -138,6 +144,75 @@ Permissions use JSON format with a role → rules mapping:
...
+
+Who may open the dialog is derived from CASL (`update` on `Config`), not from a hardcoded
+role name, so an instance that grants permission editing to a role other than `admin_app`
+gets the same UI.
+
+_Key files:_ `feature-permission/feature-permission.service.ts` (rule reading/writing),
+`feature-permission/feature-permission-dialog/`, `permissions-config.service.ts`
+(shared loading, backup/undo and admin check), `entity-list/entity-list.component.html`
+(where the dialog is offered).
+
```
sleidig: The UI is designed to be intuitive. No point describing its minute details in the README.

```suggestion
```

sleidig (changes requested):
Preliminary review. I think we should consolidate a few details here with the admin user roles code, which is about to be merged.

src/app/core/entity-list/entity-list/entity-list.component.ts
Abhinegi2: Could you test this again now?

sleidig: seems fixed now :+1: 

(what was the cause? I couldn't see anything in code that seemed related)

Abhinegi2: dialog asks Keycloak for the list of user roles and  Keycloak address comes from `assets/keycloak.json`, in the demo mode, it is   https://keycloak.localhost.

now ([ee0543aa9](https://github.com/Aam-Digital/ndb-core/pull/4167/changes/ee0543aa99efde93e211f1b575a524f8b16e653c)) app does not contact Keycloak at all when there is none, and simply uses the roles it finds in Config:Permissions.

src/app/core/permissions/feature-permission/feature-permission-dialog/feature-permission-dialog.component.ts
```diff
@@ -0,0 +1,324 @@
...
+/** per-role row shown in the dialog grid */
+interface RolePermissionRow {
+  /** the role's technical name, or the reserved key of the shared default section */
+  role: string;
+  /** what to show in the first column */
+  label: string;
+  /** human-readable description from the auth server, if available */
+  description?: string;
+  /** true for the shared `_default` row shown above the roles */
+  isDefaultRow: boolean;
```
sleidig, line 91: Functionally there still seems to be a glitch:
I would expect that I can edit the _default role permissions in this scenario?

Image

The scenario that _default grants a permission and the other roles are then locked for that checkbox works as expected, only editing _default itself seems broken.

Abhinegi2:
@coderabbit review

src/app/core/permissions/feature-permission/feature-permission.service.ts
```diff
@@ -0,0 +1,433 @@
...
+  defaultRules: RoleFeaturePermission;
+
+  roles: RoleFeaturePermission[];
+}
+
+/** the actions of one row as sent back to {@link FeaturePermissionService.setPermissions} */
+export interface RoleFeaturePermissionUpdate {
+  role: string;
+  actions: Record<FeatureAction, boolean>;
+}
```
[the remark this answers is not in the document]
Abhinegi2: @sleidig I guess we’ve already run the `_default` migration in all production environments. Or are there still any systems where the configuration contains only the legacy `default` section?

src/app/core/permissions/feature-permission/feature-permission-dialog/feature-permission-dialog.component.ts
Abhinegi2: could you test this again now?

src/app/core/permissions/feature-permission/feature-permission.service.ts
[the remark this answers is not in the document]
sleidig: Claude these days would say "let me check rather than assume" :laughing:  ...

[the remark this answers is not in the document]
sleidig: yes, all systems have the new _default and _public roles already in their CouchDB. There have been a few users on version `3.88.0` last week still. the ndb-core version is the critical part - if someone is still on an old version that checks `public` rather than `_public` when we have already removed the legacy role in the DB, then there will be problems.

@Abhinegi2 , can you double check which version landed the new roles in the frontend? Maybe we can already remove all legacy role code then :blush:  (--> https://github.com/Aam-Digital/replication-backend/issues/295)

src/app/core/permissions/feature-permission/feature-permission-dialog/feature-permission-dialog.component.ts
sleidig: still not fully functional for me:

GIVEN user_app has no EmailTemplate permissions (and no manage-all permission)
AND _default has an EmailTemplate read-only permission
THEN I would expect that I can edit the "Default" role's create/update/... permission checkbox from the Email Template "Configure Permissions" dailog. They remain locked, currently

src/app/core/permissions/feature-permission/feature-permission.service.ts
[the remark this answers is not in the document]
Abhinegi2: 3.87.2 (, #4168) is where the frontend started reading _default / _public, so the 3.88.0 users are fine , anything below 3.87.2 is the cut-off.

will open a pr for legacy code cleanups

src/app/core/permissions/feature-permission/feature-permission-dialog/feature-permission-dialog.component.ts
sleidig: works now :+1: 
Thanks!

src/app/core/permissions/feature-permission/feature-permission-dialog/feature-permission-dialog.component.ts
```diff
@@ -0,0 +1,395 @@
...
+}
+
+/** one checkbox column of the grid */
+interface ActionColumn {
+  action: FeatureAction;
+  label: string;
+}
+
+/** one checkbox of a row, with everything the template needs precomputed */
+interface PermissionCell {
```
sleidig: Is this different from the main role admin's checkboxes? Or would the same interface work to be reused? (or maybe even a mini-component wrapping the checkbox and handling all the state and tooltip stuff could be useful to ensure consistent UI)

src/app/core/permissions/feature-permission/feature-permission-dialog/feature-permission-dialog.component.ts
```diff
@@ -0,0 +1,395 @@
...
+        Logging.error("Failed to load feature permissions", error);
+        return { rows: [], loadFailed: true };
+      }
+    },
+  });
+
+  /** whether the roles and their permissions could not be loaded */
+  readonly loadFailed = computed(
+    () => this.permissionRows.value()?.loadFailed === true,
+  );
```
sleidig, line 165: Angular `resource` may have a built-in error state that we can use without an error status object to return? Not sure if this would work here: https://angular.dev/api/core/Resource#error

src/app/core/permissions/feature-permission/feature-permission-dialog/feature-permission-dialog.component.ts
```diff
@@ -0,0 +1,395 @@
...
+   * `displayRows` additionally `false` while `_default` grants the action.
+   */
+  editable: boolean;
+  /** why this checkbox cannot be changed; empty when it is editable */
+  lockTooltip: string;
+  ariaLabel: string;
+}
+
+/** per-role row shown in the dialog grid */
+interface RolePermissionRow {
```
sleidig, line 83: thinking a bit further:

The Role-Admin view has one row per entity type, with the "all" type as a default row. The feature permission view here has one row per role, with the "_default" role as a default row. They both apply the same rules of which cells get locked by the default row or the "manage" action, I think. Generalizing this into a class that updates its lock states, so that we can reuse it across both could be nice. But if there are logically differences, then it's probably not worth it.

src/app/core/permissions/feature-permission/feature-permission.service.ts
```diff
@@ -0,0 +1,440 @@
...
+ *
+ * A role that is granted all of them is stored as the CASL `manage` action instead,
+ * so that a config written by an admin keeps the shape it had before.
+ */
+export const FEATURE_ACTIONS = [
+  "create",
+  "read",
+  "update",
+  "delete",
+] as const satisfies readonly EntityActionPermission[];
```
sleidig: as we decided against different permission wording for features, I would like the remove these and use `CRUD_ACTIONS` directly instead.

This kind of consolidation may also apply to the other code in this class.

src/app/core/permissions/permission-types.ts
sleidig: isn't the new `CRUD_ACTIONS` a duplicate of this?

src/app/core/permissions/README.md
```diff
@@ -139,6 +145,40 @@ Permissions use JSON format with a role → rules mapping:
...
+The one case in which the dialog writes a `_default` section is the very first save on
+an instance that has no permissions config at all: an absent config means "everyone may
+do everything", so `_default: [{ subject: "all", action: "manage" }]` is seeded alongside
+the new rules to avoid locking every logged-in user out of everything else.
+
+_Key files:_ `feature-permission/feature-permission.service.ts` (rule reading/writing),
+`feature-permission/feature-permission-dialog/`, `permissions-config.service.ts`
+(shared loading, backup/undo and admin check).
+
```
sleidig: I think this is too much detail for the overview README. The details become clear from the code itself.

```suggestion
```

sleidig (commented):
Thanks, extracting the various logic and texts into clearly named helpers makes the code pretty easy to read also.

Still a few thoughts on code consolidation

src/app/core/admin/admin-role-permissions/permission-matrix/permission-matrix.component.ts
```diff
@@ -52,36 +59,12 @@ type CrudAction = "read" | "create" | "update" | "delete";
 type GrantedBy = "manage" | "wildcard" | "default";
 
 /** display state of one action cell */
```
sleidig:
```suggestion
/**
  * display state of one action cell
  * with additional details required for permission conditions.
  */
```

src/app/core/permissions/feature-permission/feature-permission-dialog/feature-permission-dialog.component.ts
```diff
@@ -0,0 +1,396 @@
...
+   * `displayRows` additionally `false` while `_default` grants the action.
+   */
+  editable: boolean;
+  /**
+   * What is written for this action, i.e. the row's own grant. It differs from
+   * {@link PermissionCellState.allowed} where an uneditable rule decides the
+   * checkbox, so that such a rule is neither duplicated into the row nor
+   * silently dropped from it.
+   */
+  grantedByOwnRule: boolean;
```
sleidig: what's the difference of this to `CellState.ownAllowed`? the names and comments still sounds very similar to me but I guess the three interfaces (`PermissionCellState`, `CellState`, `PermissionCell`) are so significantly different that we do need them to be separate?

src/app/core/permissions/permission-checkbox/permission-checkbox.component.ts
```diff
@@ -0,0 +1,89 @@
...
+  /** whether the checkbox may be changed on this row */
+  editable: boolean;
+  /** why the checkbox cannot be changed; empty when it is editable */
+  lockTooltip: string;
+  /**
+   * id of the hidden element repeating {@link lockTooltip} for screen readers,
+   * which do not announce the tooltip of a checkbox they cannot change.
+   * Empty when the cell is editable.
+   */
+  lockDescriptionId: string;
```
sleidig: as this is only used in the component template - does it really need to be part of the state object and "public" interfaces? Or could the component just separately generate this and keep it only internal to permission-checkbox component?

src/app/core/permissions/feature-permission/feature-permission.service.ts
```diff
@@ -0,0 +1,426 @@
...
+  /** the effective access, i.e. what the checkbox shows when the dialog opens */
+  granted: boolean;
+  /**
+   * Whether the row's *own* rules for this entity type grant the action, ignoring
+   * what the shared `_default` section adds on top.
+   *
+   * The dialog keeps this as the row's own intent, so that unticking an action on
+   * the `_default` row reveals a role's own rule again instead of discarding it.
+   */
+  grantedByOwnRule: boolean;
```
sleidig: what's the difference to `PermissionCell.grantedByOwnRule`?

sleidig (changes requested):
Sorry, one last try of code consolidation. Some of the services and components have really big interfaces, which already now make it hard to follow for me. So I think simplifying this as far as possible now is important to help us maintain this in the future ...

src/app/core/permissions/permission-checkbox/permission-checkbox.component.ts
```diff
@@ -7,38 +7,26 @@ import {
...
 
-import { EntityActionPermission } from "../permission-types";
-
 /**
- * What a permission grid has to tell this component about one checkbox.
- * The grids extend it with the state they need for their own editing.
+ * The state of one cell of a permission grid, as far as the grids have it in
+ * common. They extend it with what they need for their own editing.
  */
 export interface PermissionCellState {
```
sleidig, line 16: @claude , why do we need two similar interfaces with `PermissionCellState` and `FeatureActionPermission`? Could they be combined somehow to simplify the code or would that make the code less readable.

sleidig: They sit on opposite sides of the dialog, and one field settles it: `allowed` does not mean the same thing in both.

- `FeatureActionPermission.allowed` is the *effective* access as stored, `_default` included (`effectiveRules = [...defaultRules, ...roleRules]`).
- `PermissionCell.allowed` is what the checkbox shows, which for an editable cell is deliberately the row's *own* grant **without** `_default`. `applyDefaultGrants()` layers `_default` back on live, so unticking an action on the `_default` row reveals the role's own rule again instead of discarding it.

`toCell()` is exactly that translation. A single merged interface would give `allowed` two meanings depending on which side of it you are on.

Beyond that:

- `PermissionCellState` has two consumers, `FeatureActionPermission` only one. The role matrix builds its `CellState` from `MatrixCell` (conditions, preserved extra properties) and never sees the feature service; its lock reasons are the three-way `GrantedBy` (own `manage` / the "all record types" row / `_default`). A merged type would carry `lockedByAdvancedRule` into a grid where that is not the question being asked.
- The lock fields are cause vs. rendering: the service reports *why* access is fixed (`lockedByAdvancedRule`, derived from the CASL rules), the checkbox needs *what to render* (`editable` plus a localized `lockTooltip`). Merging them pulls `$localize` copy into a service — and the service could not compute `editable` correctly anyway, because whether a checkbox is locked by `_default` depends on the `_default` row **as currently edited**, not as stored.

So I kept the two, but pushed the overlap down: `ownAllowed` sat in `PermissionCellState` although `PermissionCheckboxComponent` never reads it (its template only uses `allowed`, `editable`, `lockTooltip`) - both grids reach it through their own extension. It now lives in `CellState` and `PermissionCell`, so the shared base is exactly "what the checkbox renders" and the overlap with `FeatureActionPermission` is down to the single `allowed` field. The properties that stay similar-but-different (`allowed`, `ownAllowed`, `lockedByAdvancedRule` vs. `editable`/`lockTooltip`, row-level vs. cell-level `editable`) now say in their JSDoc how they differ.

Pushed as 2776191b1.

sleidig (approved):

sleidig:
@coderabbitai review

src/app/features/public-form/public-form-permission.service.ts
```diff
@@ -33,46 +33,22 @@ export class PublicFormPermissionService {
...
-      return false; // If we can't load permissions, assume no access
+    // a load failure is not caught here: treating it as "no permission" would
+    // claim the form is broken while we simply do not know
+    const permissionsConfig = await this.permissionsConfigService.load();
+    if (!permissionsConfig?.data) {
+      return false; // No permissions config means "public" users have no access
     }
+    const publicRules = permissionsConfig.data[PUBLIC_SECTION_KEY] ?? [];
+    return publicRules.some((rule) =>
+      ruleCoversAction(rule, entityType, "create"),
```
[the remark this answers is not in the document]
sleidig: use of inverted conditions is discouraged and not available through UI, we are therefore accepting this edge case as a possible issue

src/app/core/permissions/feature-permission/feature-permission.service.ts
```diff
@@ -0,0 +1,416 @@
...
+    }
+
+    // No permissions config yet means "everyone may do everything". Seed the
+    // `_default` all-access rule so that starting to restrict a single feature
+    // does not accidentally lock every logged-in user out of everything else.
+    // This is the one case in which this UI writes a `_default` section.
+    const config = existing ?? new Config<DatabaseRules>(Config.PERMISSION_KEY);
+    config.data = {
+      [DEFAULT_SECTION_KEY]: [{ subject: "all", action: "manage" }],
+    };
```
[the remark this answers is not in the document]
sleidig: true, the logic and its comment actually does not make sense: If we start adding rules but have _default manage-all permissions, no rules can be added in the UI because everything is already allowed. We should not assume any permissions in such a case ...

sleidig:
@coderabbitai review

PR #4168, opened 2026-07-24
fix: reserved permission keys can no longer collide with Keycloak role names

### LINKED ISSUE #236: Reserved role names (`default`, `public`) collide with Keycloak role names in getRulesForUser
## Problem

In `RulesService.getRulesForUser` (`src/permissions/rules/rules.service.ts`) the lookup of rules for a user is:

```ts
const userRules = user.roles
  .filter((role) => this.permission.hasOwnProperty(role))
  .map((role) => this.permission[role])
  .flat();
if (this.permission.default) {
  userRules.unshift(...this.permission.default);
}
```

The `RulesConfig` document uses two reserved keys with special semantics:

- `default` — rules prepended to **every** authenticated user
- `public` — rules used when the user is **not** logged in

Because the role lookup is `this.permission.hasOwnProperty(role)` against the same object that holds those reserved keys, a Keycloak role literally named `default` or `public` will be picked up as if it were a normal role mapping. In practice that means:

- A user with a Keycloak role `default` would unintentionally re-apply (or duplicate) the `default` rules.
- A user with a Keycloak role `public` would receive the unauthenticated `public` rules in addition to their normal rules — which can be broader than intended (the `public` block is typically used to allow the login page to read minimal config without authentication).

This is a misconfiguration foot-gun, not an exploitable bypass on its own, but it can lead to access being granted that an admin did not intend.

## Suggested fix

- Explicitly skip the reserved keys in the role lookup, e.g. `user.roles.filter(role => role !== 'default' && role !== 'public' && this.permission.hasOwnProperty(role))`.
- Optionally validate the permission document on load and warn if any non-reserved key collides with a known reserved name in the future.
- Add unit tests for both reserved names.

## Related

Came up while reviewing permission-evaluation paths around the v1.6.0–v1.6.5 releases (see also #235 / fail-closed startup hardening).

Abhinegi2:
- closes [#236](https://github.com/Aam-Digital/replication-backend/issues/236)

## What this does

Frontend counterpart of the reserved-role-name fix (backend: Aam-Digital/replication-backend#289). `AbilityService` resolved a user's roles as keys of the permission-rules object, so a Keycloak role literally named `default` or `public` was treated as a real role, duplicating the default rules or granting the anonymous rules to a logged-in user.

Changes:
- Rename the reserved sections to `_default` / `_public`; the underscore marks them internal so they cannot collide with a realm role name.
- The rules lookup skips any role starting with `_` and any reserved section name.
- Reads accept both new and legacy spellings (prefers new) for a safe migration window.
- `PublicFormPermissionService` reads with the fallback and writes the new keys (migrating a legacy `public` section it touches).
- Adds a CLI migration (`cli/migration/permissions-key-rename.migration.ts`) that renames the `Config:Permissions` sections `default`/`public` to `_default`/`_public` (renamed key wins, legacy dropped; idempotent, dry-run-safe), renames the base-config seed asset, and documents the reserved keys + underscore-prefix policy + legacy fallback in the permissions README.

Backward compatible: existing documents keep working via the legacy-key fallback until the migration runs. No Keycloak changes (the reserved names are config keys, not realm roles).

Coordinated rollout: deploy the read-compatible code (this PR + #289) first, then run the migration per instance; a later PR removes the legacy fallback once all instances are migrated.

## PR Checklist

---

## Manual Testing Steps

Verified on a local dev stack (CouchDB + Keycloak + backend from #289) with a `Config:Permissions` document that uses only the renamed `_default`/`_public` keys:

1. Sign in as a user whose only role (e.g. `user_app`) grants nothing beyond one entity type, relying on `_default` for core access. The dashboard, config-driven navigation, site name and entity list views all load correctly, confirming the frontend reads the renamed keys.
2. Before login, the welcome page shows the configured site name, confirming anonymous `_public` reads work.
3. Create a Keycloak realm role literally named `public` and one named `_something`, assign both to the user, and confirm the app still loads normally and neither injects a reserved section into the user's abilities.
4. On an instance whose document still uses the legacy `default`/`public` keys, confirm the app resolves permissions correctly (legacy fallback), then run the `oneoff-20260724-permissions-key-rename` migration and confirm the keys are renamed with role sections untouched.

## Summary by CodeRabbit

- **New Features**
  - Added automatic migration of legacy `default`/`public` permission sections to `_default`/`_public`.
  - Updated permission evaluation to use `_default`/`_public` and ignore reserved role names/sections.
  - Public form permissions now read `_default`/`_public` with legacy fallback.

- **Documentation**
  - Updated permission schema and examples to describe `_default`/`_public` semantics and precedence.

- **Bug Fixes**
  - Improved idempotency and preserved existing `_default`/`_public` when legacy keys are also present.
  - Ensured dry-run/no-change behavior when the permissions document is missing.

- **Tests**
  - Expanded coverage for the new section keys and migration behavior.


Abhinegi2:
@coderabbitai review

sleidig (approved):

PR #4170, opened 2026-07-27
fix(Admin UI): distance column gets missing on Admin UI for matching view 

### LINKED ISSUE #4161: distance column gets missing on Admin UI for matching view
After chaning the visible columns for the Matching View, the special column "distance" gets lost. That column is also not available in the Admin UI dropdown:

Image

Make this available in the Admin UI.

possible could be solved by a more generic approach: 
- #3653

Abhinegi2:
- closes #4161

## PR Checklist

---

## Manual Testing Steps

The special columns of the Matching View ("Distance" and the record preview column) were missing from the "Add columns" dropdown in the Admin UI. Since they could not be selected there, they were also silently dropped from the config as soon as any other column was changed and the config was saved.

Both special columns are covered here: "Distance" was not offered because explicitly configured fields that do not exist in the entity schema were discarded, and "Name (record preview)" was not offered because the internal `_id` field was filtered out even though it was explicitly configured.

While testing this, the calculated "Distance" also turned out to be missing in the top summary comparison view of the Matching View itself, even though it was configured and worked in the tables below. The generated column config was only applied to the tables, not to the summary view. That is fixed here as well.

Use the `[All Features]` demo setup.

### Prerequisite: make the Matching View reachable

The demo config contains a Matching View but no menu entry for it, so add one first:

1. Go to Admin > Admin Overview > "Main Menu".
2. Click **"Add New Item"**.
3. Under **"Link"** select the Matching View entry, enter any **"Label"** (e.g. "Matching") and click **"Apply"**.
4. Save the menu configuration. The Matching View now appears in the left sidebar.

### The actual fix

1. Open the Matching View from the left sidebar.
2. Click the three-dot menu in the top right of the Matching View and select **"Edit Matching Entities"**.
3. Under "Available columns" for either side, open the "Add columns" dropdown.
4. Verify **"Distance"** is listed and can be checked and unchecked.
5. Verify **"Name (record preview)"** is listed as well.
6. Keep both selected, additionally select another column (e.g. "Name") and click "Save".
7. Go back into "Edit Matching Entities" and verify that "Distance" and "Name (record preview)" are still configured. Before this fix they were removed from the config at this point.
8. Back in the Matching View, select a record on one side and verify the Distance column in the table below shows the calculated values. Records without geo coordinates on their location field correctly show "-".
9. With a record selected on both sides, verify that the top summary comparison view now also shows a **"Distance"** row with the calculated value. Before this fix that row stayed empty.

## Summary by CodeRabbit

* **New Features**
  * Custom fields now appear in field-selection menus even without standard labels.
  * Field selection now merges explicitly provided autocomplete options with inferred schema fields, with explicit configuration taking precedence.
  * Matching-entity field selection includes the special selectable `distance` and `_id` columns.
  * Added a **Matching View** link to the **School** navigation submenu.
* **Bug Fixes**
  * Matching comparison tables now render from the correct comparison column set, keeping distance and summary views synchronized.

### DIFF
--- a/src/app/core/entity/entity-field-select/entity-field-select.component.ts
+++ b/src/app/core/entity/entity-field-select/entity-field-select.component.ts
@@ -78,13 +78,22 @@ export class EntityFieldSelectComponent extends BasicAutocompleteComponent<
   });
 
   /**
-   * Explicitly given options take precedence over the entity's schema fields,
+   * Explicitly given options are merged with the fields inferred from the entity's schema,
    * so that callers can offer additional, calculated columns (e.g. "distance")
    * that do not exist as a field in the schema.
+   * An explicit config takes precedence over the inferred field of the same id.
    */
   protected override optionsSource = computed(() => {
-    const explicitOptions = this.options();
-    return explicitOptions?.length ? explicitOptions : this.fieldOptions();
+    const optionsById = new Map<string, FormFieldConfig>();
+    for (const option of this.options() ?? []) {
+      optionsById.set(option.id, option);
+    }
+    for (const inferredOption of this.fieldOptions()) {
+      if (!optionsById.has(inferredOption.id)) {
+        optionsById.set(inferredOption.id, inferredOption);
+      }
+    }
+    return [...optionsById.values()];
   });
 
   private getAllFieldProps(schema: EntitySchema): FormFieldConfig[] {
--- a/src/app/core/entity/entity-field-select/entity-field-select.component.ts
+++ b/src/app/core/entity/entity-field-select/entity-field-select.component.ts
@@ -77,7 +77,15 @@ export class EntityFieldSelectComponent extends BasicAutocompleteComponent<
     return this.getAllFieldProps(entityType.schema);
   });
 
-  protected override optionsSource = computed(() => this.fieldOptions());
+  /**
+   * Explicitly given options take precedence over the entity's schema fields,
+   * so that callers can offer additional, calculated columns (e.g. "distance")
+   * that do not exist as a field in the schema.
+   */
+  protected override optionsSource = computed(() => {
+    const explicitOptions = this.options();
+    return explicitOptions?.length ? explicitOptions : this.fieldOptions();
+  });
 
   private getAllFieldProps(schema: EntitySchema): FormFieldConfig[] {
     return [...schema.entries()]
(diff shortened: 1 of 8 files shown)


Abhinegi2:
@coderabbitai review

Abhinegi2:
While testing, I found that the distance calculation was not showing in the comparison view(top one), so I fixed that as well.

src/app/core/entity/entity-field-select/entity-field-select.component.ts
```diff
@@ -77,7 +77,15 @@ export class EntityFieldSelectComponent extends BasicAutocompleteComponent<
...
 
-  protected override optionsSource = computed(() => this.fieldOptions());
+  /**
+   * Explicitly given options take precedence over the entity's schema fields,
+   * so that callers can offer additional, calculated columns (e.g. "distance")
+   * that do not exist as a field in the schema.
+   */
+  protected override optionsSource = computed(() => {
+    const explicitOptions = this.options();
+    return explicitOptions?.length ? explicitOptions : this.fieldOptions();
```
sleidig: should this really be an either-or? Or rather merging explicit options with the inferred ones from schema?

sleidig (approved):
Works well and code looks good to me

src/app/core/entity/entity-field-select/entity-field-select.component.ts
Abhinegi2: Changed it to a merge, explicitly given options are applied first, then the fields inferred from the schema fill in whatever is not already there, deduplicated by id so an explicit config wins over the inferred field of the same id.

src/assets/base-configs/all-features/Config_CONFIG_ENTITY.json
```diff
@@ -40,7 +40,14 @@
           "entityType": "Child"
         },
         {
-          "entityType": "School"
+          "entityType": "School",
+          "subMenu": [
+            {
+              "label": "Matchig View",
```
Abhinegi2:
```suggestion
              "label": "Matching View",
```

Abhinegi2:
@coderabbit review

src/app/core/entity/entity-field-select/entity-field-select.component.ts
sleidig: Did you push that yet? :blush:

Abhinegi2: It was committed locally but not pushed; now it is

sleidig (approved):
Thanks! looks good

sleidig:
@coderabbitai review

PR #4181, opened 2026-07-28
test(e2e): cover list filtering, sorting and pagination interactions

sleidig:
Extends entity-list e2e coverage from "a filter narrows the list" to the combinations that actually break: several filters at once, a sort on top, and results spanning more than one page.

No new spec files and only one additional `loadApp` — the scenarios chain by clicking, and the bulk cases reuse the existing `beforeEach`.

## 🐛 Finding: descending sort does nothing for records with equal values

While writing these tests I found a genuine bug, **on master, independent of this PR** — reported separately as #4182.

Rows are sorted **twice**:

1. `EntitiesTableSortStore.sortedRows()` calls `tableSort(...)` — [entities-table.component.ts#L238](https://github.com/Aam-Digital/ndb-core/blob/master/src/app/core/common-components/entities-table/entities-table.component.ts#L238)
2. the result is assigned to a `MatTableDataSource` whose `sortData` is overridden to call `tableSort(...)` again — [entities-table.component.ts#L319](https://github.com/Aam-Digital/ndb-core/blob/master/src/app/core/common-components/entities-table/entities-table.component.ts#L319)

`tableSort` sorts and then, for `desc`, `reverse()`s the whole array. Running that twice is harmless for distinct values (the second `sort()` re-normalises, the second `reverse()` flips again), but for **tied** values the sort is stable and preserves the already-reversed input, so the second `reverse()` undoes the first.

**Effect:** sorting descending by a column where records share a value returns them in *ascending* order. `aria-sort` reads `descending`, so the UI claims it worked. This hits any column with duplicates — Center, Gender, Status, Class — and also the `created.at` tie-breaker, which is what orders equal records.

Reproduced by clicking a tied column twice and observing an unchanged order.

I did **not** fix it here to keep this PR test-only, and because #4122 touches exactly this code. Sorting by a tied column is therefore deliberately left unasserted, with a note at the site. Worth deciding whether the fix belongs here, on #4122, or separately — the likely fix is to stop re-sorting in the data source, since the sort store has already sorted.

## What is covered

`entity.spec.ts`
- an enum filter and an entity-reference filter combined, sorted by name, walked to page 2 — including a decoy that sorts *into* page 2, so filtering is proven to apply to the whole dataset rather than the rendered page
- archived records appear only once explicitly included, recomputing the current page
- descending sort, and narrowing the filter, while sitting on page 2
- records without a value for the sorted column appear at either end depending on sort direction
- several options selected within one filter, and the "not defined" option
- exporting with the "Current (filtered)" scope writes only the filtered rows — this path (`EntityListComponent.openExportDialog` → `dataSource.filteredData`) had no e2e coverage; the existing download tests exercise the Reports view instead

`bulk-operations.spec.ts`
- shift-range selection on a filtered, sorted second page
- "select all" spans the whole filtered result, not just the visible page
- bulk-editing records out of the active filter updates both list and paginator

## Verification

- full e2e suite green (40 passed)
- key assertions mutation-checked: dropping the second filter yields 17 instead of 14 records; making the page-2 decoy match both filters makes it appear on page 2
- run against #4122 as well, see the comment below

### DIFF
--- a/e2e/fixtures.ts
+++ b/e2e/fixtures.ts
@@ -1,5 +1,5 @@
 // eslint-disable-next-line no-restricted-imports
-import { Page, test as base } from "@playwright/test";
+import { Download, Page, test as base } from "@playwright/test";
 // eslint-disable-next-line no-restricted-imports
 import {
   argosScreenshot as argosScreenshotBase,
@@ -37,6 +37,7 @@ import { EventAttendanceMapDatatype } from "#src/app/features/attendance/depreca
 
 // eslint-disable-next-line no-restricted-imports
 export { expect } from "@playwright/test";
+export type { Page };
 
 /** The mocked "now" date to which e2e tests are fixed. */
 export const E2E_REF_DATE = "2025-01-23";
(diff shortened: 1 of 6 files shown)


sleidig:
## Result against #4122 (In-memory data source)

I cherry-picked this commit onto `in-memory-data-source` and ran both spec files there. **Two bulk-selection tests fail; everything else passes.**

```
✓ Bulk-edit a field across selected records
✓ Bulk-merge two records combines them into one
✘ Bulk selection with sorting › Bulk selection range with sorting by shift-click rows
✘ Bulk selection with sorting › Bulk selection and editing operate on the filtered, sorted result across pages
✓ Combining filters keeps sorting consistent across paginated pages
✓ List filter narrows results and clears restore full list
   (+ 3 more entity.spec.ts tests)
9 passed, 2 failed
```

Both fail on the same assertion — after shift-clicking a range, the rows rendered in that range are not the ones selected (`expect(locator).toBeChecked()` → received `unchecked`).

### Likely cause

`getCurrentPageRows()` supplies the row list that shift-range selection indexes into.

- on master it slices `sortStore.sortedRows()` — the sorted rows, matching what is rendered
- on #4122 it slices `recordsDataSource().data`, which holds the **filtered but unsorted** rows: `data` is assigned from `filteredRecords()`, and `MatTableDataSource` sorts a *copy* (`_orderData` → `sortData(data.slice(), sort)`), leaving `data` itself unsorted

So the anchor and target indices are looked up in a different order than the user sees, and on page 2 the slice window covers the wrong records entirely. Same likely applies to `EntitiesTableSelectionStore`, which receives `displayedData` under the name `sortedRows`.

### Note on why CI here is currently green

The `Bulk selection range with sorting by shift-click rows` test already exists on master and passes on this PR today — its `beforeEach` generates 10 random-named children, one page. This PR's version bumps that to 14 named children whose load order is the reverse of their name order, plus 4 excluded by a filter, which is what makes sorted order diverge enough to expose the problem. So this is **not** a case of the PR already having a red test; the current data just doesn't reach the defect.

### One difference that looks like an improvement

The "not defined" filter step initially failed on #4122 too, but for a good reason: two records tied under the active sort came back in the opposite order. That is #4122 *fixing* the double-sort described in the PR body — with a single sort pass, tied records genuinely reverse under `desc`. My assertion had implicitly pinned master's buggy order, so I made it order-independent, and that step now passes on both. Worth confirming that removing the double sort was intentional, since it also fixes descending order for tied values.

cc @sleidig — re #4122

sleidig:
@coderabbitai review

Abhinegi2 (approved):

PR #4186, opened 2026-07-29
test: isolate unit test files from each other

sleidig:
## Problem

`qa / test-unit` has been failing intermittently on unrelated PRs and on master, with a **different set of specs each time**. Recent examples:

| run | failures |
|---|---|
| a PR touching only `e2e/` | 17 tests across 2 files |
| master (`30378596351`) | 0 failing tests, but **9 uncaught exceptions** → exit 1 |
| local, repeated | 9, then 4, then 10 — never the same set twice |

## Cause

The Angular unit-test builder defaults Vitest's `isolate` to `false` ([`plugins.js:128`](https://github.com/angular/angular-cli/blob/main/packages/angular/build/src/builders/unit-test/runners/vitest/plugins.js), commented _"Default to `false` to align with the Karma/Jasmine experience"_).

All 446 spec files therefore share one module registry, one `environment` singleton and one TestBed per worker. Async work that outlives a spec file — PouchDB task queues, Angular initializers — throws into whichever file happens to be running next. So failures land on innocent specs and move between runs.

Concretely, in the CI run above: a PouchDB on the `indexeddb` adapter (jsdom has no `indexedDB`) threw from a deferred `setTimeout` while another spec was mid-`beforeEach`, aborting its initializer *after* the TestBed was instantiated. Every later `beforeEach` then hit `Cannot configure the test module when the test module has already been instantiated`.

## Change

`isolate: true`, plus the two specs that only ever passed because another file happened to register an entity for them:

- `primary-action.service.spec.ts` — `getEntityConstructor("Note")` returned `undefined`
- `admin-inherited-field.component.spec.ts` — `Requested item is not registered in EntityRegistry. Key: Note`

Both **failed standalone before this PR** (`--include='**/that-file.spec.ts'`); they are genuine latent bugs that `isolate: false` was masking. They now import the entity models they depend on, so the `@DatabaseEntity` decorator registers them.

Also caps `maxWorkers`, since each worker holds a full Angular TestBed + jsdom. The cap is inert on CI (4 cores → 3, already Vitest's own default there) and only bounds RAM on many-core dev machines.

## Alternatives ruled out

Two cheaper config-only fixes were tried and both made things **worse**, which is what makes `isolate: true` the answer rather than a guess:

| approach | failures |
|---|---|
| baseline (`isolate: false`) | 4 |
| fork ProxyZone from `Zone.root` | 10 — breaks the nesting `fakeAsync`/`waitForAsync` rely on |
| `restoreMocks` + `unstubGlobals` | 16 — restores originals after every test, breaking specs that set spies in outer scope |
| **`isolate: true`** | **0, exit 0** |

## Cost

Isolation is slower — each file gets a fresh environment. Worth watching what `test-unit` reports on this PR. For context, that job ran ~3 min while `test-e2e` ran ~6.5 min in the same workflow, so `test-unit` is not the critical path and the extra time may not extend the pipeline at all.

That this is leakage rather than machine load: the isolated run was **slower overall yet had zero timeouts**. Load would produce more timeouts in a longer run, not none.

## Testing

- `npm run test-ci` on this branch: green, exit 0
- both fixed specs pass standalone *and* together
- `tsc -p .` clean

### DIFF
--- a/.github/workflows/_qa-unit.yaml
+++ b/.github/workflows/_qa-unit.yaml
@@ -37,21 +37,12 @@ jobs:
   test-unit:
     needs: toolchain
     runs-on: ubuntu-latest
+    permissions:
+      id-token: write # for QLTY code coverage (https://docs.qlty.sh/migration/coverage)
     env:
       TZ: Europe/Berlin
       FORCE_COLOR: 1
-      # Each spec file runs in its own environment (see vitest-base.config.ts),
-      # which costs too much wall time for a single runner. Split across shards.
-      # If a shard ever approaches the timeout, add entries to the matrix below.
-      VITEST_SHARD: ${{ matrix.shard }}/${{ strategy.job-total }}
-      # The config caps workers at 4, which resolves to 3 here (4 cores - 1) and
-      # is what Vitest would pick anyway. Runners have 16 GB, so use all 4.
-      VITEST_MAX_WORKERS: 4
     timeout-minutes: 10
-    strategy:
-      fail-fast: false
-      matrix:
-        shard: [1, 2, 3, 4]
     steps:
       - uses: actions/checkout@v4
       - uses: actions/setup-node@v4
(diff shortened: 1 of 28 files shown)


sleidig:
@coderabbitai review

PR #4196, opened 2026-07-30
fix(import): resolve entity references matched through a nested entity reference

### LINKED ISSUE #4195: Import of entity ref by matching an entity ref within that doesn't work (anymore?)
**Describe the bug**
Importing data for an `entity` type field does not match existing records if the criteria is an `entity` type field on the record to be referenced:

for example:

GIVEN `Note` entity has a field "related records" to link to `Child` entities
AND `Child` entity type has a field "responsible user" that links to a `User` entity
AND the database has exactly one Child referencing User "test user" (because import lookups may need to match exactly one record)

WHEN importing a spreadsheet of Note data with a row that has "test user" in its "responsible user of child" column
AND selecting the "related records" field mapping for that column
AND selecting "responsible user" field as the linking field match (the second dropdown that appears to select the field of the `Child` entity to be used for matching
AND selecting "user name" for the third dropdown that appears to select the field on the `User` entity to be used for matching

THEN the preview step of the import should correctly show the `Child` to be found and linked
(this is not the case currently)

quick debugging with breakpoints seems to shows that 
```
const criteria = await this.buildMatchCriteria(
      columns,
      context,
      importProcessingContext,
    );
```
in the Entity Datatype only contains the raw values (user names) and doesn't resolve the imported data to the  user IDs to match.

sleidig:
Fixes #4195

## Problem

`EntityDatatype` overrides `importMatchField` but **not** `importMapFunction`. When the selected matching field is itself an `entity` field, `resolveColumnValue()` runs the raw cell through that field's datatype via `importMapFunction` — which for `EntityDatatype` falls through to the inherited `StringDatatype` implementation and just returns `String(val)`. The match criterion therefore holds the raw import value (e.g. a user name) instead of the resolved `User:…` id, and never matches the id stored on the `Child`.

## Fix

`EntityDatatype.importMapFunction` resolves a single value to the referenced record's id by delegating to its own `importMatchField`. This recurses correctly for deeper nesting and reuses the per-entity-type candidate cache in the import processing context.

A second, related problem had to be fixed along with it: when a value mapping cannot resolve a value, `normalizeValue(undefined)` produced `""`, which matches every candidate whose matching field is *empty*. The fix above makes that path reachable for entity references, so an unresolvable value would have silently linked the wrong record. Unresolvable values are now dropped from the criterion's value list while the criterion itself is kept — an empty list matches nothing rather than relaxing the condition.

## Tests

Two unit tests in `entity.datatype_advanced.spec.ts`:
- matching an entity reference through a nested entity reference (fails before the fix)
- an unresolvable nested entity reference must match nothing (guards the `""` false-match above)

## Sample config (second commit — see review comment)

The All Features config had no `entity` field on `Child` pointing at `User`, so the reported case could not be configured at all. `303b9624` adds a `responsibleUser` field to `entity:Child` and to the Child details form.

This commit is **optional** and can be dropped — see the review comment with a revert suggestion. Note that keeping it adds a visible field to the Child details form and will require updating the Argos e2e screenshots.

## Manual test

Set *Responsible User* on one or two Child records (demo data does not populate the new field), then import as `Note`:

```csv
Subject,Notes,Responsible User of Child
Follow-up call,Discussed progress with the family.,demo
Home visit,Met the family at their home.,demo-admin
Phone call,Should stay unmatched (no such user).,not-an-existing-user
```

Map `Responsible User of Child` → **Children** → *Match by Child property:* **Responsible User** → *Match by User property:* **Name**.
Rows 1 and 2 link the assigned Child records; row 3 links nothing (and must not link a Child that has no responsible user).

> `Note.relatedEntities` ("Related Records") has no `additional` entity type in the All Features config, so no "Match by … property" dropdown renders for it. `Note.children` is the equivalent `entity`/`Child` field.

## Summary by CodeRabbit

* **New Features**
  * Added support for importing entity references by matching configured fields.
  * Added a “Responsible User” field to Child records in the Basic Information form.

* **Bug Fixes**
  * Improved entity matching when referenced values cannot be resolved, preventing invalid matches.
  * Enhanced nested entity reference resolution during imports.

* **Tests**
  * Added coverage for nested entity matching and unresolved reference handling.

### DIFF
--- a/src/app/core/basic-datatypes/entity/entity.datatype.ts
+++ b/src/app/core/basic-datatypes/entity/entity.datatype.ts
@@ -156,6 +156,52 @@ export class EntityDatatype extends StringDatatype {
   }
 
   /**
+   * Resolve a single import value to the id of the referenced entity,
+   * looking that entity up by the property given in `additional.refField`.
+   *
+   * This is relevant when an entity reference is matched through another entity
+   * reference (e.g. import a Note's related Child, identified by the name of the
+   * user linked in the Child's "responsible user" field). The nested value has to
+   * be resolved to an id before it can be compared with the stored reference.
+   *
+   * @returns the id of the single matching entity or undefined if there is no unique match
+   */
+  override async importMapFunction(
+    val: any,
+    schemaField: EntitySchemaField,
+    additional?: any,
+    importProcessingContext?: ImportProcessingContext,
+  ): Promise<string> {
+    const config = normalizeEntityAdditional(additional);
+    if (!config?.refField || !importProcessingContext) {
+      return super.importMapFunction(
+        val,
+        schemaField,
+        additional,
+        importProcessingContext,
+      );
+    }
+
+    const match = await this.importMatchField(
+      // the caller compares against a single stored value, so never return an array here
+      { ...schemaField, isArray: false },
+      [
+        {
+          mapping: {
+            column: "",
+            propertyName: schemaField.id,
+            // the value has already been split by the calling column
+            additional: { ...config, enableSplitting: false },
+          },
+          rawCell: val,
+        },
+      ],
+      importProcessingContext,
+    );
+    return match as string;
+  }
+
+  /**
    * Build one match criterion per mapped column: the referenced field to
    * compare and the acceptable (normalized) values parsed from the column cell.
    */
@@ -182,16 +228,20 @@ export class EntityDatatype extends StringDatatype {
         mapping,
         separator,
       )) {
-        values.push(
-          await this.resolveColumnValue(
-            rawValue,
-            config.refField,
-            config.valueMapping,
-            context,
-            importProcessingContext,
-          ),
+        const value = await this.resolveColumnValue(
+          rawValue,
+          config.refField,
+          config.valueMapping,
+          context,
+          importProcessingContext,
         );
+        if (value !== undefined) {
+          values.push(value);
+        }
       }
+      // the criterion is added even if no value could be resolved:
+      // an empty list matches nothing, whereas dropping the criterion would
+      // silently relax the condition and match unrelated records
       criteria.push({ refField: config.refField, values });
     }
     return criteria;
@@ -216,6 +266,9 @@ export class EntityDatatype extends StringDatatype {
   /**
    * Resolves the effective comparison value for a column,
    * applying any configured value mapping through the referenced field's datatype.
+   *
+   * @returns the normalized value to compare against, or undefined if the value
+   *   mapping could not resolve the import value (which must match nothing)
    */
   private async resolveColumnValue(
     rawValue: any,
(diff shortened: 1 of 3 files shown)


src/assets/base-configs/all-features/Config_CONFIG_ENTITY.json
```diff
@@ -805,7 +805,8 @@
                     {
                       "fields": [
                         "photo",
-                        "schoolId"
+                        "schoolId",
+                        "responsibleUser"
```
sleidig: Revert: remove the field from the Child details form.

```suggestion
                        "schoolId"
```

src/assets/base-configs/all-features/Config_CONFIG_ENTITY.json
```diff
@@ -1342,6 +1343,12 @@
           "label": "Linked School",
           "anonymize": "retain"
         },
+        "responsibleUser": {
+          "dataType": "entity",
+          "additional": "User",
+          "label": "Responsible User",
+          "anonymize": "retain"
+        },
```
sleidig: Revert: remove the `responsibleUser` field from `entity:Child`.

```suggestion
```

sleidig (commented):
The `chore(config)` commit (`303b9624`) is only there so the reported case can be configured and tested manually — the All Features config had no `entity` field on `Child` pointing at `User`.

It is **not** needed for the fix itself. If we'd rather not ship it, the two suggestions below revert it (or just drop the commit). Keeping it means the Child details form gains a visible field, so the Argos e2e screenshots will need updating.

sleidig:
@coderabbitai review

Abhinegi2:
Functionality working as expected.

src/app/core/basic-datatypes/entity/entity.datatype.ts
```diff
@@ -155,6 +155,52 @@ export class EntityDatatype extends StringDatatype {
...
+            // the value has already been split by the calling column
+            additional: { ...config, enableSplitting: false },
+          },
+          rawCell: val,
+        },
+      ],
+      importProcessingContext,
+    );
+    return match as string;
+  }
```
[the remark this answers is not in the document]
Abhinegi2: Fair on the doc/signature mismatch, will align those.

The rest doesn't apply here:

tsconfig.json has "strict": false, so widening to Promise<string | undefined> is a compiler no-op.
The as string isn't about undefined: importMatchField returns string | string[] | undefined, and string[] is not assignable to string either way. It only goes away with a runtime narrowing: return Array.isArray(match) ? match[0] : match;
No need to widen DefaultDatatype.importMapFunction. LocationDatatype already annotates Promise<GeoLocation> and returns undefined, so the base type is imprecise repo-wide already. Separate cleanup.
No behaviour change either way, all callers already handle undefined.

Abhinegi2:
@coderabbit review

Abhinegi2 (approved):

PR #4210, opened 2026-08-03
fix(logging): group monitoring issues by error chain instead of stack trace

sleidig:
## Problem

A single production problem is currently scattered across many monitoring issues, so triage
repeats the same work every week — and archiving one of them does not silence the others.
Aggregating the last 30 days of events by exception chain rather than by issue:

| chain signature | distinct issues | events |
| --- | --- | --- |
| `DatabaseException` (unwrapped) | 14 | 1031 |
| `Error` | 31 | 301 |
| `DatabaseException` → `Error` | 11 | 159 |
| `DatabaseException` → `ConfigLoadError` | 6 | 216 |
| `DatabaseException` → `SiteSettingsLoadError` | 2 | 15 |

**Sentry groups an exception by its stack trace whenever one is available, and ignores the
message.** (Visible on any issue holding events with several different `error.value`s — the
archived `DatabaseException` issue holds three.) Our wrapper errors are thrown from one central
place but reached from many components, routes and async call sites, so the stack differs per
caller — and per build, since releases without source maps report minified frames.
[AAM-DIGITAL-731](https://aam-digital.sentry.io/issues/AAM-DIGITAL-731) and
[72E](https://aam-digital.sentry.io/issues/AAM-DIGITAL-72E) have an identical error chain and
overlapping releases, yet are separate issues; one renders symbolicated, the other minified. At
the same time 731 lumps two different root causes together, and
[735](https://aam-digital.sentry.io/issues/AAM-DIGITAL-735) — an `unauthorized` failure, a
genuine problem rather than offline noise — is indistinguishable by title from its neighbours.

Separately, `Logging.warn`/`Logging.error` messages *are* grouped by their text, because they go
through `captureMessage`. Interpolating an id, URL or status code into one splits the problem per
value and puts record content into issue titles.

## Changes

- `beforeSend` sets an explicit `fingerprint` for our own wrapper error classes
  (`CAUSE_GROUPED_ERROR_TYPES`), built from the thrown error plus its root cause with ids, URLs
  and numbers masked. The thrown error says which operation failed, the root cause says why — so
  offline and unauthorized config failures separate, while the same failure reported from ten
  different screens collapses into one issue.
- The route stays available as the `transaction` tag, so failures can still be filtered by where
  they happened without splitting the issue.
- The session repeat cap reuses the same normalization, so errors interpolating ids no longer get
  a separate flood budget per variant.
- The registry gets dedicated `RegistryLookupError` and `RegistryDuplicateError` classes,
  following `DatabaseException` and `EntityPermissionError`. A missing registration is one problem
  no matter which component or pipe hit it first, but today it opens an issue per call site (73Y on
  `/import/` and 6YW on `/` are the same missing-registration bug). Both set `this.name` to a
  literal — Sentry derives the exception type as `error.name || error.constructor.name`, and a
  subclass otherwise inherits `"Error"` from `Error.prototype`, which would silently drop them out
  of the fingerprint list while the constructor name would be minified. Specs lock that. Their
  messages deliberately keep the requested key — for a fingerprinted error the message *is* part of
  the grouping key, and a missing `Event` component is a different fix from a missing `Child`
  entity. The duplicate-registration message also no longer interpolates the registered element,
  which stringified a whole class constructor or async import function.
- Registries take an explicit `registryName`. This is a **readability** fix, not a grouping one:
  the minified class name is baked into the message when the error is constructed, so unlike a
  stack frame source maps can never repair it and the report permanently reads
  `not registered in u`.
- All `Logging.warn`/`Logging.error` call sites move interpolated values into the context
  argument, including a JSON-stringified HTTP error response that was being dumped into a title.
- The conflicting document's ID is passed as `entityId` — already surfaced as structured extra
  data — rather than appended to the message.
- Convention and rationale documented in `src/app/core/logging/README.md`.

## Deliberate limitations

- **Generic errors keep the default grouping.** `["DatabaseException", "Error"]` (11 issues) stays
  split, because only the *thrown* type is matched and it is a plain `Error` there. For generic
  errors the stack trace is the only thing telling two unrelated bugs apart, so fingerprinting
  them would merge real distinct problems.
- Inconsistent symbolication contributes to the splitting; the sourcemap upload failing on release
  tags is tracked separately and not addressed here.
- `replication-backend` has the same pattern on the message side; handled in
  Aam-Digital/replication-backend#302.

## After deploying

Changing a fingerprint does **not** re-group events already in Sentry. Existing issues stop
receiving events and the merged clusters show up as new, unresolved, high-volume issues — the
already-archived `DatabaseException` issue among them. They need to be archived once more after
the first release containing this.

## Testing

- New specs cover the fingerprint for chained errors, root-cause separation, id/URL masking,
  self-wrapping errors matching their unwrapped form, registry lookups grouping by key rather than
  call site, and generic and message-only events keeping the default grouping.
- Specs lock the registry name, the `RegistryLookupError` name, and the duplicate-key message.
- Full unit suite: 2435 passed, 21 skipped. Lint and typecheck clean.

### DIFF
--- a/src/app/core/config/registry/dynamic-registry.ts
+++ b/src/app/core/config/registry/dynamic-registry.ts
@@ -1,6 +1,52 @@
 import { environment } from "../../../../environments/environment";
 
 /**
+ * Thrown when an item that the config refers to - a dynamic component, an entity
+ * type, a route - is looked up in a {@link Registry} without having been registered.
+ */
+export class RegistryLookupError extends Error {
+  constructor(
+    readonly registryName: string,
+    readonly key: string,
+  ) {
+    super(`Requested item is not registered in ${registryName}. Key: ${key}`);
+
+    // Set explicitly, and to a literal: remote monitoring reads the exception
+    // type as `error.name || error.constructor.name`. A subclass inherits
+    // "Error" from Error.prototype, so without this the error would be reported
+    // as a plain "Error" and drop out of the fingerprint-based grouping
+    // (CAUSE_GROUPED_ERROR_TYPES in logging.service.ts). The constructor name is
+    // no alternative either - it is minified in production builds.
+    this.name = "RegistryLookupError";
+  }
+}
+
+/**
+ * Thrown when the same key is registered twice in a {@link Registry},
+ * which usually means two modules define the same component or entity type.
+ *
+ * Only thrown in production builds (see {@link Registry.allowDuplicates}), where
+ * it aborts bootstrap - so it reaches remote monitoring through the top-level
+ * bootstrap error handler in `main.ts`.
+ */
+export class RegistryDuplicateError extends Error {
+  constructor(
+    readonly registryName: string,
+    readonly key: string,
+  ) {
+    // the already-registered element is deliberately not interpolated here:
+    // stringifying it dumps a whole class constructor or async import function
+    // into the message
+    super(
+      `${registryName}: Duplicate definition, "${key}" is already registered`,
+    );
+
+    // set explicitly for the same reason as in RegistryLookupError above
+    this.name = "RegistryDuplicateError";
+  }
+}
+
+/**
  * A registry is an affordance to register dynamic objects to strings.
  * It is commonly used to dynamically load entities, views or routes from the config
  *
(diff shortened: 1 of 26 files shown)


src/app/core/config/registry/dynamic-registry.ts
```diff
@@ -39,7 +47,7 @@ export abstract class Registry<T> extends Map<string, T> {
   public override get(key: string): T {
     if (!this.has(key)) {
       throw Error(
-        `Requested item is not registered in ${this.constructor.name}. Key: ${key}`,
+        `Requested item is not registered in ${this.registryName}. Key: ${key}`,
```
sleidig: better move the key to a separate error context field also for better grouping?

Issue #4165, opened 2026-07-22
Checkbox field imported as text string instead of boolean (inconsistent stored value)


Abhinegi2:
**Describe the bug**

A checkbox (boolean) field behaves inconsistently around import, in two related ways:

**1. Imported value stored as a text string instead of a boolean**

- **Manual entry** (ticking the checkbox) stores a proper boolean `true`.
- **Import** stores the raw text string `"true"` instead of converting it to a boolean.

Both render identically in the list view (a checkmark), so the difference is hidden in the UI, but it surfaces in SQL reports: boolean `true` shows as `1`, while the imported string `"true"` shows as `true`. This makes reporting and filtering on the field unreliable.

**2. Value-mapping dialog defaults `false` to "checked"**

When importing a column that contains both `true` and `false`, the value-mapping dialog pre-selects **both** values as checked. So the `false` value has to be manually unchecked, otherwise it would be imported as `true`. The mapping checkbox treats any non-empty text as "checked", and since `false` is still a non-empty string, it defaults to checked.

**To Reproduce**

Symptom 1:

1. Import a set of records with a checkbox column whose value is `true`, mapping it to a boolean/checkbox field.
2. Manually create a record of the same type and tick the same checkbox.
3. Open a SQL report that includes this field.
4. See that imported records show `true` (string) while the manual record shows `1` (boolean).

Symptom 2:

1. Import a file with a boolean column containing both `true` and `false` values.
2. Open the value-mapping dialog for that column.
3. See that both `true` and `false` are pre-selected as checked (the `false` value is not recognised as unchecked).

**Expected behavior**

- The value stored is the same regardless of how the record was created, a proper boolean (`true` / `false`) in all cases. Imported string values (`"true"`, `"false"`, etc.) should be converted to booleans on import.
- In the value-mapping dialog, `false` (and equivalents like `0`, `no`) should default to unchecked, and only `true` (and equivalents like `1`, `yes`) to checked.

**Additional context**

- Root cause (both symptoms): the boolean datatype's `transformToObjectFormat` returns the value unchanged, so string values from import are never converted to booleans.
  - Symptom 1: the raw string is stored as-is instead of a boolean.
  - Symptom 2: in the mapping dialog the field value stays the string `"false"`, and the checkbox coerces any non-empty string to checked (`!!"false"` is `true`).
- The list view uses a truthy check (`value ? "✓" : ""`), which is why both the string and the boolean look the same there.
[...]

PR #4217, opened 2026-08-04
feat(import): let user review value mapping config of every mapped column

### LINKED ISSUE #4185: Import: value mapping dialogs can be skipped silently (auto-suggest mappings and require review)
_As a user importing data I want the app to apply and let me confirm value mappings for every column that needs one, in order to avoid importing raw file values that silently end up wrong in the database._

Some column types need an extra configuration dialog during import (checkbox and dropdown value mapping, date format, entity reference lookup, location). Today these dialogs are optional: the column mapping step is considered complete as soon as a column is assigned to a field, and the user can continue to the preview without ever opening "Configure value mapping". A badge with "?" is the only hint.

If the dialog is skipped, the value from the file is taken over as-is. This is not obvious to the user and has already caused real data problems: in #4165 imported checkbox values were stored as plain text instead of true/false, which broke reports for the client. #4166 added a safety net for checkboxes, but the underlying UX gap applies to all of these dialogs.

### Proposed Solutions & Alternatives

Two parts, to be discussed:

1. Auto-suggest mappings without requiring the user to open the dialog. The suggestions already exist inside the dialogs (dropdown options are matched by id or label, checkbox values match true/1/yes), they simply never run when the dialog stays closed.
2. Require the user to review and confirm those suggestions before continuing. For example, block the "next" step until each column with a config dialog has been opened and confirmed once, and keep the warning badge until then.

Preference so far in the discussion: suggestions should be a proposal only, not applied silently, because the user may deliberately want a different mapping and would otherwise notice only after the import.

Alternatives considered:

- Only show a stronger warning before continuing, without blocking. Simpler, but easy to click away.
- Apply auto-mappings silently with no confirmation. Fastest for the user, but hides the cases where the guess is wrong.

Open UX questions:

- How to mark a column as "reviewed" without annoying users who have nothing to configure.
- What to do with values that cannot be mapped at all (currently they are skipped).
- Related known problem in the same dialog: a value mapped to "unchecked" (false) is treated like an unmapped value in the confirmation prompt, so the user gets a misleading "some values don't have a mapping" warning. Listed in #1991 as well.

### Additional context

Came out of a discussion about #4165 / #4166. Part of the umbrella list in #1991 (bullet: "maybe don't allow user to continue if the value mapping dialogs are not opened and confirmed?").

Affected config dialogs: dropdown and checkbox value mapping, date format, entity reference, location.

---

## Implementation Approach

To be filled in after the UX approach is agreed.

Abhinegi2:
- closes #4185

## PR Checklist

---

## What this changes

Columns mapped to a field whose datatype needs extra import configuration (dropdown / checkbox value mapping, date format) could pass the column mapping step without that dialog ever being opened. The values from the file were then taken over as they are, which is what broke the stored checkbox values reported in #4165.

The column mapping step now makes that state visible and, where the user started configuring, mandatory:

| situation | what the user sees |
| --- | --- |
| user selects a field that needs configuration | its dialog opens right away, with the suggested mappings applied, to confirm or refine |
| column was mapped automatically (header matches a field) | no dialog, but a warning badge with tooltip on the button and a message below the field that the values are imported as they are |
| dialog closed with "Save & Close" | badge and message disappear, the mapping can still be reopened and changed |
| dialog closed with "Cancel" | message asks to open and confirm it, and "Continue" stays disabled until that happened |

Automatically mapped columns deliberately do not open their dialogs: a file whose headers match several fields would otherwise bury the user under dialogs before they even see the step.

### Implementation

- New `ImportConfigDialogService` owns opening a column's config dialog: it resolves the dialog component, collects the column's unique values, and returns the resulting `ColumnMapping`. The mapping passed in is no longer modified by the dialog, the result comes back as a new object.
- Datatypes declare their dialog through the new `importConfigDialog` property (`DiscreteImportDialog`, `DateImportDialog`), so nothing outside the datatype needs to know which dialog belongs to which field.
- Both dialogs now close with a result on save, which the service turns into `configReview` on the column mapping (`confirmed` / `cancelled`). Cancelling does not overwrite a configuration that was confirmed before.
- `ImportColumnMappingComponent` opens the dialogs of newly mapped columns one after the other and uses `configReview` to never show the same dialog twice.
- The message below the field uses `mat-error`, like other validation messages, pulled into the space the form field keeps for its own errors so that it belongs to the field above it.
- The date column now has the same badge and tooltip as value mappings, previously it gave no indication at all.
- The date and discrete inline config components delegate to the service, which removes their duplicated dialog setup code.

### Automatically mapped columns are marked, not interrupted

1. Open the Import view and upload a CSV whose headers match existing fields:

```
name,gender,center,dropoutDate
Alice,female,Alpha,2024-01-05
Bob,male,Beta,
Carol,f,Alpha,2023-11-30
```

2. Continue, import as **Child**, Continue again.
3. "Map Columns" opens with **no dialog**. The rows for `gender`, `center` and `dropoutDate` show a red "?" badge on their "Configure value mapping" button and a red message below the field: "Value mapping not configured, the values from the file are imported as they are.". The `name` row has no button and no message.
4. Hover the badge / button of `gender` and of `dropoutDate`: each shows a tooltip explaining what is missing.
5. "Continue" is enabled, an unconfigured column does not block the import.

### A newly selected field opens its dialog

6. Map any column to a dropdown or date field yourself (e.g. reselect "Gender" on the `gender` row). Its dialog opens immediately, exactly once.
7. Press "Save & Close". The message and the "?" badge of that row disappear.

### Cancelling blocks the step

8. Reselect a field on the `center` row so its dialog opens, and press "Cancel" this time.
9. The row now says "Value mapping was not saved. Open it and confirm to continue." and "Continue" is **disabled**.
10. Open the same dialog again through its button and press "Save & Close": the message disappears and "Continue" is enabled again.
11. Reopen that dialog once more and press "Cancel": the column stays confirmed, cancelling an edit does not throw away what was already confirmed.

### Nothing repeats itself

12. Press "Back" and "Continue" again: no dialog opens a second time and the messages are unchanged.
13. Map a column to a plain text field: no button, no message, no dialog.

## Summary by CodeRabbit

* **New Features**
  * Added datatype-specific import configuration for date formats and discrete value mappings.
  * Added badges and tooltips showing configuration status.
  * Added validation errors for missing import configuration.
  * Import completion now requires all mapped fields to be fully configured.
  * Improved mapping updates when column data changes.

* **Bug Fixes**
  * Ensured empty date formats are saved consistently.
  * Prevented duplicate configuration dialogs.

* **Documentation**
  * Added import workflow and configuration guidance.

### DIFF
--- a/src/app/core/basic-datatypes/discrete/discrete-import-config/discrete-import-config.component.ts
+++ b/src/app/core/basic-datatypes/discrete/discrete-import-config/discrete-import-config.component.ts
@@ -1,6 +1,7 @@
 import {
   ChangeDetectionStrategy,
   Component,
+  computed,
   inject,
   input,
 } from "@angular/core";
@@ -11,6 +12,7 @@ import { ImportConfigDialogService } from "../../../import/import-column-mapping
 import { DiscreteColumnMappingAdditional } from "../discrete.datatype";
 import { MatButtonModule } from "@angular/material/button";
 import { MatBadgeModule } from "@angular/material/badge";
+import { MatTooltipModule } from "@angular/material/tooltip";
 import { DynamicComponent } from "../../../config/dynamic-components/dynamic-component.decorator";
 
 /**
@@ -22,7 +24,7 @@ import { DynamicComponent } from "../../../config/dynamic-components/dynamic-com
   changeDetection: ChangeDetectionStrategy.OnPush,
   selector: "app-discrete-import-config",
   templateUrl: "./discrete-import-config.component.html",
-  imports: [MatButtonModule, MatBadgeModule],
+  imports: [MatButtonModule, MatBadgeModule, MatTooltipModule],
 })
 export class DiscreteImportConfigComponent {
   private readonly configDialogs = inject(ImportConfigDialogService);
@@ -34,18 +36,35 @@ export class DiscreteImportConfigComponent {
   additionalSettings = input<ImportAdditionalSettings>();
   onColumnMappingChange = input<(col: ColumnMapping) => void>();
 
-  badge(): string | undefined {
+  /** how many of the file's values have no mapping yet, undefined while nothing is configured */
+  readonly unmappedCount = computed<number | undefined>(() => {
     const additional = this.col()
       ?.additional as DiscreteColumnMappingAdditional;
     const valueMappings = additional?.values;
     if (!valueMappings) {
+      return undefined;
+    }
+    return Object.values(valueMappings).filter((v) => v === undefined).length;
+  });
+
+  readonly badge = computed(() => {
+    const unmapped = this.unmappedCount();
+    if (unmapped === undefined) {
       return "?";
     }
-    const unmappedCount = Object.values(valueMappings).filter(
-      (v) => v === undefined,
-    ).length;
-    return unmappedCount > 0 ? unmappedCount.toString() : undefined;
-  }
+    return unmapped > 0 ? unmapped.toString() : undefined;
+  });
+
+  readonly tooltip = computed(() => {
+    const unmapped = this.unmappedCount();
+    if (unmapped === undefined) {
+      return $localize`:import value mapping tooltip - not configured:The values of this column are not mapped yet. They are imported exactly as they are in the file, which can result in values the system does not recognise.`;
+    }
+    if (unmapped > 0) {
+      return $localize`:import value mapping tooltip - unmapped values:${unmapped}:count: of the values in this column have no mapping and are skipped during import.`;
+    }
+    return $localize`:import value mapping tooltip - configured:All values of this column are mapped. Open to review or change the mapping.`;
+  });
 
   async openConfig() {
     const updated = await this.configDialogs.openConfigDialog(
--- a/src/app/core/basic-datatypes/discrete/discrete-import-config/discrete-import-config.component.ts
+++ b/src/app/core/basic-datatypes/discrete/discrete-import-config/discrete-import-config.component.ts
@@ -7,9 +7,7 @@ import {
 import { ColumnMapping } from "../../../import/column-mapping";
 import { EntityConstructor } from "../../../entity/model/entity";
 import { ImportAdditionalSettings } from "../../../import/import-additional-settings";
-import { MatDialog } from "@angular/material/dialog";
-import { MappingDialogData } from "../../../import/import-column-mapping/mapping-dialog-data";
-import { DiscreteImportDialogComponent } from "./discrete-import-dialog.component";
+import { ImportConfigDialogService } from "../../../import/import-column-mapping/import-config-dialog.service";
 import { DiscreteColumnMappingAdditional } from "../discrete.datatype";
 import { MatButtonModule } from "@angular/material/button";
 import { MatBadgeModule } from "@angular/material/badge";
@@ -27,7 +25,7 @@ import { DynamicComponent } from "../../../config/dynamic-components/dynamic-com
   imports: [MatButtonModule, MatBadgeModule],
 })
 export class DiscreteImportConfigComponent {
-  private readonly dialog = inject(MatDialog);
+  private readonly configDialogs = inject(ImportConfigDialogService);
 
   col = input<ColumnMapping>();
   rawData = input<any[]>([]);
@@ -49,28 +47,13 @@ export class DiscreteImportConfigComponent {
     return unmappedCount > 0 ? unmappedCount.toString() : undefined;
   }
 
-  openConfig() {
-    const col = this.col();
-    const uniqueValues = new Set<any>(
-      this.rawData().map((row) => row[col.column]),
+  async openConfig() {
+    const updated = await this.configDialogs.openConfigDialog(
+      this.col(),
+      this.rawData(),
+      this.entityType(),
+      this.additionalSettings(),
     );
-
-    this.dialog
-      .open<DiscreteImportDialogComponent, MappingDialogData>(
-        DiscreteImportDialogComponent,
-        {
-          data: {
-            col: col,
-            values: [...uniqueValues],
-            totalRowCount: this.rawData().length,
-            entityType: this.entityType(),
-            additionalSettings: this.additionalSettings(),
-          },
-          width: "80vw",
-          disableClose: true,
-        },
-      )
-      .afterClosed()
-      .subscribe(() => this.onColumnMappingChange()?.(col));
+    this.onColumnMappingChange()?.(updated);
   }
 }
--- a/src/app/core/import/column-mapping.ts
+++ b/src/app/core/import/column-mapping.ts
@@ -19,13 +19,4 @@ export interface ColumnMapping {
    * This is used to track if the coloumns are manually updated or not.
    */
   manuallyUpdated?: boolean;
-
-  /**
-   * How the user left the config dialog of this column, if its field requires one.
-   *
-   * `undefined` means the dialog was never opened, so the values are imported as they are
-   * in the file. "cancelled" means the user closed it without saving, which blocks the import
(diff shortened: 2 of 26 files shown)


Abhinegi2:
@coderabbit review

src/app/core/basic-datatypes/discrete/discrete-import-config/discrete-import-config.component.ts
```diff
@@ -36,41 +36,43 @@ export class DiscreteImportConfigComponent {
...
-  badge(): string | undefined {
+  /** how many of the file's values have no mapping yet, undefined while nothing is configured */
+  readonly unmappedCount = computed<number | undefined>(() => {
     const additional = this.col()
       ?.additional as DiscreteColumnMappingAdditional;
     const valueMappings = additional?.values;
     if (!valueMappings) {
+      return undefined;
+    }
+    return Object.values(valueMappings).filter((v) => v === undefined).length;
```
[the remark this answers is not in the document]
Abhinegi2: Valid premise, thanks. `applyPreviousMapping` copies stored mappings verbatim and `ImportMetadata.config` persists the whole `ColumnMapping`, so a previous `values: {male: "M"}` can indeed land on a file that also contains `female`.

Splitting the finding in two:

**The count itself is pre-existing.** `badge()` on master uses the same logic (count the `undefined` entries of `additional.values`), so this PR does not change that behaviour. Deriving from the current `rawData` is the right fix, but the splitting rules (`enableSplitting`, the multi-value separator, `splitArrayValue`) live in `splitAndFlattenValues` inside `DiscreteImportDialogComponent`; reimplementing them in the inline component would duplicate that logic. It is tracked together with the related miscount in #4221, where the intent is to extract that helper and share it between dialog and badge, with the regression test you describe.

**What this PR did introduce, now fixed in 2125d3ac9.** Because `configReview` is persisted too, a loaded mapping came back as `confirmed`, so the new warning message was suppressed as well. Combined with the blind badge, a reused mapping on a new file gave no warning at all. Note that clearing only `configReview` would have been wrong: stored mappings carry `manuallyUpdated: true`, which would make the review queue open a dialog for every loaded column at once. `applyPreviousMapping` now strips both flags, so a loaded mapping behaves like an automatically mapped column: warning shown, no dialog forced. Covered by a new case in `import.component.spec.ts`.

Abhinegi2:
@george-neha, could you do the functional testing for this? PR description includes the details of what needs to be verified.

Abhinegi2:
@coderabbit review

src/app/core/import/column-mapping.ts
```diff
@@ -19,4 +19,13 @@ export interface ColumnMapping {
...
   manuallyUpdated?: boolean;
+
+  /**
+   * How the user left the config dialog of this column, if its field requires one.
+   *
+   * `undefined` means the dialog was never opened, so the values are imported as they are
+   * in the file. "cancelled" means the user closed it without saving, which blocks the import
+   * until the mapping is confirmed.
+   */
+  configReview?: "confirmed" | "cancelled";
```
sleidig: can't we infer this by checking if `additional` has a useful value? (e.g. when that is undefined)

That would also make it simpler in the future if we auto-map certain things

src/app/core/import/import/import.component.html
```diff
@@ -128,11 +128,14 @@ <h2 i18n>You don't have permission to import</h2>
...
-    <mat-step [completed]="mappedColumnsCount() > 0" #step3>
+    <mat-step
+      [completed]="mappedColumnsCount() > 0 && !hasUnconfirmedColumnConfig()"
+      #step3
+    >
       <ng-template matStepLabel>
         <div i18n="Import Step - map columns">Map Columns</div>
 
-        @if (step3.completed) {
+        @if (mappedColumnsCount() > 0) {
```
sleidig: did you intentionally decouple this from reusing the condition from line 131/132 above?

src/app/core/import/import-column-mapping/import-column-mapping.component.ts
```diff
@@ -50,6 +54,8 @@ export class ImportColumnMappingComponent {
     this.entityType() ? this.entities.get(this.entityType()) : undefined,
   );
 
+  private reviewingConfigs = false;
```
sleidig: what exactly does this indicate and how does it change? maybe add a one line JsDoc

src/app/core/import/import-column-mapping/import-column-mapping.component.ts
```diff
@@ -63,16 +69,61 @@ export class ImportColumnMappingComponent {
       if (JSON.stringify(autoMappings) !== JSON.stringify(cm)) {
         untracked(() => this.columnMapping.set(autoMappings));
       }
+      untracked(() => this.reviewConfigsOfMappedColumns());
```
sleidig: why untracked?

src/app/core/import/import-column-mapping/import-column-mapping.component.ts
```diff
@@ -63,16 +69,61 @@ export class ImportColumnMappingComponent {
...
+   * as unconfigured in the UI instead.
+   */
+  private async reviewConfigsOfMappedColumns() {
+    if (this.reviewingConfigs) {
+      return;
+    }
+    this.reviewingConfigs = true;
+
+    try {
+      for (const column of this.columnMapping().map((c) => c.column)) {
```
sleidig: didn't we want to avoid multiple dialogs to show at once?
(and if so, do we not get access to a single column being updated individually?)

i.e. isn't the config for a specific column something that a single EditImportColumnMapping component could handle completely transparently in its internals?

src/app/core/import/import-column-mapping/edit-import-column-mapping/edit-import-column-mapping.component.html
```diff
@@ -13,3 +13,7 @@
     <ng-container [appDynamicComponent]="inlineComponentConfig()" />
   }
 </div>
+
+@if (configHint(); as hint) {
+  <mat-error class="config-hint">{{ hint }}</mat-error>
```
sleidig: maybe we should call it configError or configMissingError, rather than just a hint?

sleidig (commented):
Generally looks good - a few thoughts regarding details of the general architecture only.


---
Also, let's add a basic developer README.md to the import module. Maybe include these basic diagrams as a start:

```mermaid
---
config:
  theme: redux
  layout: dagre
---
flowchart BT
 subgraph s1["ImportComponent"]
        n2["ImportFileComponent"]
        n3["ImportEntityTypeComponent etc."]
        n4["ImportColumnMappingComponent"]
        n5["ImportReviewDataComponent"]
  end
    n2 --> n3
    n3 --> n4
    n4 --> n5
```


```mermaid
---
config:
  theme: redux
  layout: dagre
---
flowchart BT
 subgraph s3["EditImportColumnMapping (one per column)"]
        m1["EntityFieldSelect"]
        m2["Inline Component Config (for additional)"]
  end
 subgraph s2["ImportColumnMappingComponent"]
        s3
  end
```

src/app/core/import/import-column-mapping/import-column-mapping.component.ts
Abhinegi2: It got removed after some refactoring.

src/app/core/import/import-column-mapping/import-column-mapping.component.ts
Abhinegi2: Agreed, moved into EditImportColumnMapping, it opens the config of the field the user just selected itself, so the parent no longer loops over columns at all

Abhinegi2:
@coderabbit review

sleidig (approved):

PR #4219, opened 2026-08-04
fix: add missing translation (distance column)

sleidig:


src/app/features/matching-entities/admin-matching-entities/edit-matching-entity-side/edit-matching-entity-side.component.ts
```diff
@@ -42,7 +42,7 @@ export class EditMatchingEntitySideComponent {
     .map((ctor) => ctor.value.ENTITY_TYPE);
 
   readonly additionalFields = computed<ColumnConfig[]>(() => [
-    { id: "distance", label: "Distance" },
+    { id: "distance", label: $localize`column for matching view:Distance` },
```
sleidig:
```suggestion
    { id: "distance", label: $localize`:column for matching view:Distance` },
```

Abhinegi2 (approved):

PR #4224, opened 2026-08-05
refactor: move bootstrap-time helpers into plain TS modules

### LINKED ISSUE #4222: refactor: move bootstrap-time helpers out of injectable services into plain TS modules
Bootstrap-time helpers currently live as `static` members on `@Injectable` services. Importing them into `main.ts` drags the whole DI graph of those services into the modules that are evaluated **before** Angular bootstraps — 92 files for a single import.

## Current state

`main.ts` calls two statics before `bootstrapModule`:

```ts
PwaInstallService.registerPWAInstallListener();   // must run before the beforeinstallprompt event
await BackupService.runPendingReset();            // must delete IndexedDB before PouchDB opens connections
```

Both ordering constraints are legitimate and documented in the code. The problem is only *where the functions live*.

`BackupService` is an `@Injectable` that imports the database layer, so the import chain is:

```
main.ts
  -> app/core/admin/backup/backup.service.ts
    -> app/core/database/database-resolver.service.ts
      -> app/core/database/database-factory.service.ts
        -> app/core/database/pouchdb/pouchdb-corruption-recovery.service.ts
```

Measured size of the graph evaluated before `initLanguage()`:

| Top-level import of `main.ts` | files it contributes |
| --- | --- |
| `app/core/admin/backup/backup.service.ts` | **92** |
| `bootstrap-environment.ts` | 3 |
| `app/core/logging/logging.service.ts` | 2 |
| `bootstrap-i18n.ts` | 2 |
| `environments/environment.ts` | 1 |
| `app/core/pwa-install/pwa-install.service.ts` | 1 |
| **total eager graph** | **107** |

Everything `runPendingReset()` actually needs is `sessionStorage`, `indexedDB`, and `DatabaseResolverService.clearLastSyncMarkers()` — itself four lines of `localStorage` key removal. None of it requires the injectable graph.

## Why this matters

**1. Silent i18n breakage (the reason this was found).** `main.ts` awaits `initLanguage()` *after* those 107 modules have already been evaluated. Any module-level `$localize` in them resolves before translations are loaded and is therefore permanently English:

```ts
// module scope -> runs at import time, before loadTranslations()
const SOME_HINT = $localize`:...:We are working on improvements to allow this in the future.`;
```

There is no error, no failing test and nothing visible in review — the string simply never gets translated in any locale. A near-miss of exactly this was caught while working on #4156, in `pouchdb-corruption-recovery.service.ts`, which sits at the end of the chain above.

**2. It undermines the deferred app import.** `main.ts` deliberately loads the app via `await import("./app/app.module")`, but the static `BackupService` import already pulls a large part of the app in beforehand.

## Proposed change

Move bootstrap-time logic into plain TS modules with no app dependencies, following the pattern already established by `src/bootstrap-i18n.ts` and `src/bootstrap-environment.ts`.

- [ ] Add `src/bootstrap-reset.ts` owning `RESET_PENDING_KEY`, `runPendingReset()` and the `LAST_SYNC_KEY_PREFIX` cleanup. It must import nothing from `src/app/`.
- [ ] `BackupService` and `DatabaseResolverService` import the key/helper from there instead of owning them.
- [ ] Remove the `BackupService` import from `main.ts`.
- [ ] Same treatment for `PwaInstallService.registerPWAInstallListener()` .
- [ ] Audit the remaining eager graph and keep it small.

Expected result: the pre-i18n graph drops from **107 files to ~15**.

## Guard against regression

Even after the cleanup the trap remains possible, just less likely. Worth adding one of:

- a CI check that fails when a module in the eager graph of `main.ts` contains a module-scope `$localize`, or
- an ESLint rule forbidding `$localize` at module / `static` class scope anywhere (evaluate it in a getter, method or instance field instead).

The second is broader and cheaper to reason about: instance fields run at construction, getters at call time, and both are always after bootstrap.

## Notes

Documenting "never evaluate `$localize` at module or static-class scope" in `doc/compodoc_sources/concepts/i18n.md` should be part of this.

sleidig:
## Summary
- Move `BackupService.runPendingReset()` and `PwaInstallService.registerPWAInstallListener()` out of `@Injectable` services into new dependency-free modules (`src/bootstrap-reset.ts`, `src/bootstrap-pwa-install.ts`), following the pattern of `bootstrap-i18n.ts` / `bootstrap-environment.ts`
- `main.ts` no longer imports `BackupService`, so its DI graph (92 files) is no longer evaluated before `initLanguage()` — eliminates the class of bug where a module-scope `$localize` in that graph silently stays untranslated
- `PwaInstallService.canInstallDirectly` becomes an instance getter instead of a static; `PouchdbCorruptionRecoveryService` now reads `RESET_PENDING_KEY` from `bootstrap-reset.ts` instead of importing `BackupService`, removing another edge into that graph
- Document the bootstrap-order trap in `doc/compodoc_sources/concepts/i18n.md`, with a do/don't example for evaluating `$localize` in a getter/method/instance field rather than module or static scope

No behavior change.

Fixes #4222

## Test plan
- [x] `npx eslint` on all changed files — clean
- [x] `npx tsc --noEmit` — clean
- [x] Unit tests for all touched files pass (`bootstrap-reset`, `pwa-install.service`, `pwa-install.component`, `backup.service`, `pouchdb-corruption-recovery.service`, `support.component` — 6 files, 36 tests)
- [ ] Manual smoke test of PWA install prompt and "Reset Application" flow in a running app

## Summary by CodeRabbit

* **New Features**
  * Improved PWA installation prompt handling for more reliable installation flows.
  * Added safer application reset cleanup while preserving Sentry’s offline queue.
  * Added guidance for initializing translations before application startup.

* **Bug Fixes**
  * Improved cleanup of synchronization markers, cached databases, and service workers during resets.

### DIFF
--- a/doc/compodoc_sources/concepts/i18n.md
+++ b/doc/compodoc_sources/concepts/i18n.md
@@ -219,24 +219,6 @@ and booleans.
 
 ### Do's and don'ts
 
-- In the `bootstrap-*.ts` modules that `main.ts` imports statically (see
-  [bootstrap order](#bootstrap-order-dont-localize-before-translations-are-loaded)), never evaluate
-  `$localize` at module scope or in a `static` class member — translations aren't loaded yet at that
-  point. Use a getter, method or instance field instead, which only run once the app has bootstrapped:
-
-  ```typescript
-  // breaks in a bootstrap module - runs before translations are loaded:
-  const HINT = $localize`:some hint:We are working on improvements.`;
-
-  // works - only evaluated when called, after bootstrap:
-  get hint() {
-    return $localize`:some hint:We are working on improvements.`;
-  }
-  ```
-
-  Regular app code doesn't need this: it's loaded via `await import("./app/app.module")` after
-  translations are ready.
-
 - Try to keep your translatable texts as simple as possible concerning the
   amount of string-interpolation or computation inside the `{{}}`-Blocks. This
   makes it easier for translators
--- a/doc/compodoc_sources/concepts/i18n.md
+++ b/doc/compodoc_sources/concepts/i18n.md
@@ -219,24 +219,23 @@ and booleans.
 
 ### Do's and don'ts
 
-- Prefer evaluating `$localize` **when the text is used, not when the module is loaded**. A getter,
-  a method or an instance field all run after the app has bootstrapped, so translations are
-  guaranteed to be loaded:
+- In the `bootstrap-*.ts` modules that `main.ts` imports statically (see
+  [bootstrap order](#bootstrap-order-dont-localize-before-translations-are-loaded)), never evaluate
+  `$localize` at module scope or in a `static` class member — translations aren't loaded yet at that
+  point. Use a getter, method or instance field instead, which only run once the app has bootstrapped:
 
   ```typescript
-  // avoid in modules that may be loaded during bootstrap:
+  // breaks in a bootstrap module - runs before translations are loaded:
   const HINT = $localize`:some hint:We are working on improvements.`;
 
-  // prefer:
+  // works - only evaluated when called, after bootstrap:
   get hint() {
     return $localize`:some hint:We are working on improvements.`;
   }
   ```
 
-  This only actually breaks for the few modules `main.ts` imports statically (see
-  [bootstrap order](#bootstrap-order-dont-localize-before-translations-are-loaded)), but it is the
-  safer habit everywhere: a module that is lazily loaded today may become part of the bootstrap
-  graph tomorrow, and nothing would report the regression.
+  Regular app code doesn't need this: it's loaded via `await import("./app/app.module")` after
+  translations are ready.
 
 - Try to keep your translatable texts as simple as possible concerning the
   amount of string-interpolation or computation inside the `{{}}`-Blocks. This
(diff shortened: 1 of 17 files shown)


doc/compodoc_sources/concepts/i18n.md
```diff
@@ -202,6 +219,25 @@ and booleans.
...
+  get hint() {
+    return $localize`:some hint:We are working on improvements.`;
+  }
+  ```
+
+  This only actually breaks for the few modules `main.ts` imports statically (see
+  [bootstrap order](#bootstrap-order-dont-localize-before-translations-are-loaded)), but it is the
+  safer habit everywhere: a module that is lazily loaded today may become part of the bootstrap
+  graph tomorrow, and nothing would report the regression.
+
```
sleidig: That makes the code more messy ... If not a complete i18n disaster, I would prefer to remove that recommendation.

sleidig:
@coderabbitai review

doc/compodoc_sources/concepts/i18n.md
```diff
@@ -202,6 +219,24 @@ and booleans.
...
+  // works - only evaluated when called, after bootstrap:
+  get hint() {
+    return $localize`:some hint:We are working on improvements.`;
+  }
+  ```
+
+  Regular app code doesn't need this: it's loaded via `await import("./app/app.module")` after
+  translations are ready.
+
```
sleidig:
```suggestion
```

Abhinegi2 (approved):

PR #4229, opened 2026-08-06
fix(dropdown): keep multi-select dropdown open while selecting options

### LINKED ISSUE #4177: multi-select dropdown field should stay open to select multiple options from the drawer without clicking again
Currently the configurable-enum (and entity?) dropdown field closes the list of options immediately after the user selects one item. For multi-select (`isArray`) fields, this is annoying when trying to select a larger number of options. The field should keep showing options in that case.

---
probably to be implemented in the underlying basic-autocomplete component?

Abhinegi2:
- closes #4177

## PR Checklist

---

Selecting an option is one of the "panel closing actions" of Angular Material's autocomplete, and there is no input to opt out of it. Multi-select fields therefore closed their list of options after every single item.

The new `KeepPanelOpenOnSelectDirective` filters the option selections out of the trigger's closing actions while the field is a multi-select, so the dropdown stays open. Material then no longer runs its own selection handling (its `optionSelected` output does not emit), so the selections are applied through `onSelectionChange` of the individual options instead. The panel is no longer torn down and rebuilt per selection, which keeps the typed filter text, the scroll position within a long list and the keyboard navigation in place.

Further details of the behaviour with the dropdown staying open:

- the focus goes back to the search input after each selection, so the user can keep typing to narrow the options
- for fields showing their selection as text, the search input keeps displaying the updated selection and its text is selected, so typing a filter overwrites it instead of being appended
- the panel is repositioned after each selection, because the field can grow (e.g. an additional row of chips)
- the checkbox state is derived from the value via a signal, so it also updates for "Select All" / "Clear" while the dropdown is open
- creating a new option still closes the dropdown while its form is open, and afterwards reopens it with the filter cleared, so further options can be selected
- single-select fields are unchanged and still close on selection

Note: replacing the trigger's stream of option selections relies on a Material internal. The specs assert the resulting behaviour (panel open for multi-select, closed for single-select), so a change in a future Material version fails CI rather than passing silently.

## Manual Testing Steps

Multi-select of dropdown (configurable-enum) and entity fields:

1. Open a record with a multi-select dropdown field (e.g. the "Team involved" field of a note) and click into the field.
2. Select several options in a row: the list of options stays open, each checkbox is ticked immediately and the chips are added, without having to click the field again.
3. Click an already selected option: it is unselected and the list still stays open.
4. Type a few letters to filter, then select two of the remaining options: the filter text stays in place, so no retyping is needed.
5. Click outside the field (or press Tab / Escape): the dropdown closes and the selection is applied to the record.
6. Save and reopen the record: the selected options are stored correctly.

Long lists and list filters:

7. Open a multi-select field with more than 100 options, scroll down and select an option: the list keeps its scroll position instead of jumping back to the top.
8. Open a list view with a multi-select filter (e.g. "Center" on the Children list), select several values, and use the "Select All" and "Clear" buttons in the dropdown: all checkboxes and the filtered table update while the dropdown stays open.

Creating a new option:

9. In a multi-select entity field, type a name that does not exist yet and select "Add new ...": the dropdown closes while the form for the new record is open. After saving it, the new entry is selected and the dropdown is open again with all options available.

Single select (unchanged):

10. Open a single-select dropdown field and select an option: the dropdown closes immediately as before.

## Summary by CodeRabbit

- **New Features**
  - Multi-select autocomplete panels now remain open after selecting options.
  - Selected options retain keyboard focus and display accurate checkbox states.
  - Creating a new option reopens the panel for continued selection.

- **Bug Fixes**
  - Improved filtered selection, displayed selection text, focus handling, and panel positioning.
  - Single-select autocomplete behavior remains unchanged, with the panel closing after selection.

### DIFF
--- a/src/app/core/common-components/basic-autocomplete/basic-autocomplete.component.html
+++ b/src/app/core/common-components/basic-autocomplete/basic-autocomplete.component.html
@@ -79,6 +79,7 @@
     matInput
     style="text-overflow: ellipsis"
     [matAutocomplete]="autoSuggestions"
+    [appKeepPanelOpenOnSelect]="!!multi()"
     [matAutocompleteConnectedTo]="autocompleteOrigin"
     (focusout)="onFocusOut($event)"
     [placeholder]="placeholder"
@@ -92,6 +93,7 @@ Autocomplete
   [disableRipple]="true"
   #autoSuggestions="matAutocomplete"
   (optionSelected)="select($event.option.value)"
+  (closed)="onDropdownClosed()"
   autoActiveFirstOption
   [hideSingleSelectionIndicator]="true"
   [panelWidth]="panelWidth()"
@@ -110,6 +112,7 @@ Autocomplete
       >
         <mat-option
           [value]="item"
+          (onSelectionChange)="onOptionSelectionChange($event)"
           *cdkVirtualFor="
             let item of autocompleteOptions();
             trackBy: trackByOptionValueFn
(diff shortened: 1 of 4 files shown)


Abhinegi2:
@coderabbit review

sleidig (approved):
damn, needs pretty deep hacking 😅 

But seems to work very well and looks good. Nice work 👍

PR #4230, opened 2026-08-06
fix(import): skip additional import actions with unregistered entity types

### LINKED ISSUE #4183: Entity actions context menu can crash entirely on a single misbehaving action factory / unguarded EntityRegistry lookup
## Summary

While investigating a CI failure on PR #4122 (unit test `entity-actions-menu.component.spec.ts > should emit actionTriggered Output upon action` threw `Error: Requested item is not registered in EntityRegistry. Key: Entity`), I found two related robustness gaps in the entity actions context menu. Both are the same class of bug as #4096 / fixed by #4102 ("entity context menu breaks on unknown entity type") — but in a **sibling code path that #4102 didn't cover**. A single misbehaving action-factory or a stale entity-type reference can still crash the *entire* context menu today.

I could not pin down the exact trigger that produced the original CI stack trace (see "Origin / non-reproduction" below), but these two gaps are verified directly from the code and are worth fixing on their own merits.

## Bug 1 — `EntityActionsMenuService.getActions()` has no error isolation for action factories

[`entity-actions-menu.service.ts:23-28`](https://github.com/Aam-Digital/ndb-core/blob/master/src/app/core/entity-details/entity-actions-menu/entity-actions-menu.service.ts#L23-L28):

```ts
getActions(entity?: Entity): EntityAction[] {
  return [
    ...this.actions,
    ...this.actionsFactories.flatMap((factory) => factory(entity)),
  ];
}
```

`getVisibleActions()` (line ~38) calls `this.getActions(entity)` *before* entering the per-action try/catch that #4102 added around `action.visible()`:

```ts
async getVisibleActions(...): Promise<EntityAction[]> {
  const entity = Array.isArray(entities) ? entities[0] : entities;
  const actions = this.getActions(entity).filter(...);   // <-- not protected
  ...
  for (const action of actions) {
    try {
      isVisible = action.visible ? await action.visible(entities) : true;
    } catch (err) {
      Logging.warn(...);   // <-- only covers visible(), not getActions()
      isVisible = false;
    }
    ...
  }
}
```

If **any** registered factory (`registerActionsFactories`) throws synchronously for a given entity, the exception propagates out of `getActions()`, out of `getVisibleActions()`, and rejects the whole `getActionsForSingle()` / `getActionsForBulk()` promise. Since `EntityActionsMenuComponent` consumes this via an Angular `resource()`, a rejection resets `actions` to `[]` — the menu renders **empty for that entity**, exactly the symptom described in #4096, just via a different call path (factories vs. `visible()`).

## Bug 2 — `ImportAdditionalService.createActionLabel()` calls `entityRegistry.get()` unguarded

[`import-additional.service.ts:322-334`](https://github.com/Aam-Digital/ndb-core/blob/master/src/app/core/import/additional-actions/import-additional.service.ts#L322-L334):

```ts
createActionLabel(
  importAction: AdditionalImportAction,
  forTargetType: boolean = false,
): string {
  const sourceType = this.entityRegistry.get(importAction.sourceType);
  const targetTypes = importAction["targetType"]
    ? asArray(importAction["targetType"]).map((type) =>
        this.entityRegistry.get(type),               // <-- throws if type isn't registered
      )
    : [];
  const relationshipType = importAction["relationshipEntityType"]
    ? this.entityRegistry.get(importAction["relationshipEntityType"])   // <-- same
    : null;
  ...
}
```

`Registry.get()` ([`dynamic-registry.ts:41`](https://github.com/Aam-Digital/ndb-core/blob/master/src/app/core/config/registry/dynamic-registry.ts#L41)) **throws** when the key isn't registered — it never returns `undefined`. None of the three lookups here are guarded with `.has()` first. This is the exact "dead guard" pattern #4102 already fixed in `public-forms.service.ts`'s `findMatchingLinkedFieldIds` — but `createActionLabel` wasn't touched by that PR.

`createActionLabel` is called from `ImportModule.getAdditionalImportActionsForTargetEntity()` ([`import.module.ts:40`](https://github.com/Aam-Digital/ndb-core/blob/master/src/app/core/import/import.module.ts#L40)), which is registered as one of the factories in Bug 1 — so **any** stale/renamed/removed entity-type reference inside `ImportAdditionalService`'s cached `linkableEntities` map will throw here, and (per Bug 1) take the whole context menu down with it, not just hide the one import action.

## How the two combine in production

`ImportAdditionalService.updateLinkableEntities()` computes `linkableEntities` once in the constructor and again whenever `entityConfigReady.setupCompleted$` fires — i.e. it's a **snapshot**, not a live view. `createActionLabel()` does a **live** `entityRegistry.get()` lookup. If an entity type referenced by an `AdditionalImportAction` (via a composite/indirect relationship field's `additional`) is renamed or removed from the config *after* the snapshot was taken but the stale action is still evaluated, `createActionLabel` throws, and because of Bug 1 that exception takes down the entire entity-details context menu — not just the one action.

## Suggested fix

- Wrap the `factory(entity)` call in `getActions()` (or in `getVisibleActions()`'s consumption of it) in a try/catch per factory, logging via `Logging.warn` and skipping only that factory's actions — mirroring the per-action isolation already added for `visible()` in #4102.
- Guard the three `entityRegistry.get()` calls in `createActionLabel` with `.has()` first (as `public-forms.service.ts` now does), skipping/filtering out actions that reference an unregistered type instead of throwing.

## Manual reproduction steps

**A. Minimal reproduction (isolated, to verify the fix):**

1. In a test/scratch context, get `EntityActionsMenuService` and `EntityRegistry` via DI.
2. `entityActionsMenuService.registerActionsFactories([() => { throw new Error("boom"); }]);`
3. Call `await entityActionsMenuService.getActionsForSingle(new Entity())`.
4. **Current behavior:** the promise rejects (uncaught `"boom"`), no actions are returned, and — in the real component — the whole action menu resource errors out.
5. **Expected after fix:** the call resolves with whatever actions the *other*, healthy factories/actions produced; the broken factory is logged via `Logging.warn` and skipped.

**B. Realistic app-level reproduction (via Bug 2 specifically):**

1. In Admin UI, create/import two entity types where one references the other through a composite relationship field with an additional linked entity-type field (the pattern used by `ImportAdditionalService.generateLinkActionsFor`'s indirect-action detection — e.g. similar to how `Note` links `Child` and `School`).
2. Rename or delete the *referenced* entity type in the entity config (so the string id used in the relationship field's `additional` config no longer matches any registered entity type), without also updating/removing the relationship field itself.
3. Open the entity-details view for an entity of the type that still has the now-dangling reference, and click the "⋮" context menu.
4. **Current behavior:** the context menu fails to open / renders with no actions at all (same visible symptom as #4096), because `getAdditionalImportActionsForTargetEntity` → `createActionLabel` throws on the unregistered type and the exception is not isolated (Bug 1).
5. **Expected after fix:** the menu opens normally with all *other* valid actions; the stale import-link action is silently omitted (optionally with a `Logging.warn` for diagnosability), matching how #4102 handled the analogous `PublicFormConfig` case.

## Origin / non-reproduction note

This was found while triaging a CI failure on PR #4122 ([failing run](https://github.com/Aam-Digital/ndb-core/actions/runs/30426816481/job/90534191844)): `entity-actions-menu.component.spec.ts`'s test using a raw `new Entity()` (base class, `getType() === "Entity"`) hit `EntityRegistry.get` failing with `Key: Entity`. That specific test/service code is **unchanged by PR #4122** — it's pre-existing.

I was not able to reproduce that exact failure locally (isolated run passes; a full local suite run reproduced a *different* flake instead — `json-editor-dialog.component.spec.ts`, a hook timeout — indicating general order-dependent flakiness in the vitest suite under load). I also could not construct a code path in the current codebase (checked all `@DatabaseEntity`-decorated classes and the `all-features` base config) where any `AdditionalImportAction` would have `targetType === "Entity"` literally, so the precise trigger for that specific CI run remains unexplained. The two bugs above are independent, verified findings from reading the code, not a confirmed fix for that specific flake — but they describe a real gap that would produce the same class of user-facing symptom (#4096) via a different path.

Abhinegi2:
- closes #4183

## PR Checklist

---

## Notes on the two issues

**The reachable one is the import wizard, not the context menu.**

An entity-datatype field's `additional` is a raw config value, so a renamed or removed entity type leaves a dangling reference behind. `findIndirectLinkActionsForField` copies that value straight into an action's `targetType`, and `createActionLabel` then looks it up in the `EntityRegistry`, which throws.

`getActionsLinkingFor` (used by the import wizard's "Advanced Import Actions" section) never filtered `targetType` against a live type, so the throw escaped through the option list and the "Link all imported records to" selector rendered **completely empty**. Every valid action was lost, not just the one with the stale reference.

The context-menu path (`getActionsLinkingTo`) turned out to be much harder to hit: it filters `a.targetType === entity.getType()`, which is always a registered type, and both `sourceType` and `relationshipEntityType` come from live registry iteration. I could not construct a config that breaks it. The guard is applied there too, for the config-reload race described in the issue.

**The factory isolation is hardening, not a live bug.** Only two action factories exist (`ImportModule`, `AttendanceModule`) and neither can currently throw, which is likely why the original CI trace could not be reproduced. The `try`/`catch` means the next factory anyone registers does not inherit the hole.

## Fix

- `ImportAdditionalService.hasAllReferencedTypes()` checks `entityRegistry.has()` for `sourceType`, every `targetType`, and `relationshipEntityType`; applied in both `getActionsLinkingFor` and `getActionsLinkingTo`, logging a warning and skipping the action.
- `EntityActionsMenuService.getActions()` wraps each factory in `try`/`catch`, mirroring the existing `visible()` isolation.

## Manual Testing Steps

The dangling reference has to be created first, since a config produced through the Admin UI cannot normally contain one.

1. Open the deployed app for this PR and create a demo system with the **[All Features]** use case.
2. Go to **Admin > Edit Application Configuration (JSON)**.
3. Find `entity:Note` > `attributes` > `schools` and change `"additional": "School"` to `"additional": "SchoolArchived"`. Save.
4. Go to **Import** and upload any small `.csv` (e.g. a `name` column with two rows).
5. On step 2 **Select Import Type(s)**, choose **Child**.
6. Expand **Advanced Import Actions [optional]** and open the **"Link all imported records to:"** dropdown.

Expected: the dropdown lists four options (School through School Enrollment, Event, Class Session, Recurring Activity). The dangling `SchoolArchived` action is absent, and the browser console shows a single `[warn] ImportAdditional: skipping action referencing an entity type that is not registered.`

Before this change the same dropdown opened **empty**, with `RegistryLookupError: Requested item is not registered in EntityRegistry. Key: SchoolArchived` in the console.

7. Undo the config change (the JSON editor offers an undo, or restore `"School"` manually) and confirm the dropdown still lists the same four options with no warning.

Also worth a quick check that the entity context menu is unaffected: open any Child record and confirm the "⋮" menu still shows its usual actions.

## Follow-up spotted while tracing (not fixed here)

`getActionsLinkingTo` filters with `a.targetType === targetEntityType`, but `targetType` can be an array (`field2.additional` allows one, and `createActionLabel` handles arrays via `asArray`). The strict equality never matches an array, so multi-target indirect actions silently never appear in the context menu. Happy to file this separately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Action menus now remain available when an individual action provider fails; other valid actions continue to appear.
  * Import actions referencing unavailable entity types are skipped instead of causing errors.
  * Improved logging identifies failed action providers and invalid import references.

### DIFF
--- a/src/app/core/entity-details/entity-actions-menu/entity-actions-menu.service.spec.ts
+++ b/src/app/core/entity-details/entity-actions-menu/entity-actions-menu.service.spec.ts
@@ -113,4 +113,39 @@ describe("EntityActionsMenuService", () => {
     expect(actions.map((action) => action.action)).toEqual(["healthy-action"]);
     expect(warnSpy).toHaveBeenCalled();
   });
+
+  it("should not break the whole menu when an action factory throws", async () => {
+    const entity = TestEntity.create("Record A");
+    const warnSpy = vi.spyOn(Logging, "warn").mockImplementation(() => {});
+
+    service.registerActions([
+      {
+        action: "static-action",
+        label: "Static",
+        icon: "check",
+        execute: async () => true,
+      },
+    ]);
+    service.registerActionsFactories([
+      () => {
+        throw new Error("Requested item is not registered in EntityRegistry.");
+      },
+      () => [
+        {
+          action: "healthy-factory-action",
+          label: "Healthy Factory",
+          icon: "check",
+          execute: async () => true,
+        },
+      ],
+    ]);
+
+    const actions = await service.getActionsForSingle(entity);
+
+    expect(actions.map((action) => action.action)).toEqual([
+      "static-action",
+      "healthy-factory-action",
+    ]);
+    expect(warnSpy).toHaveBeenCalled();
+  });
 });
--- a/src/app/core/entity-details/entity-actions-menu/entity-actions-menu.service.spec.ts
+++ b/src/app/core/entity-details/entity-actions-menu/entity-actions-menu.service.spec.ts
@@ -146,6 +146,9 @@ describe("EntityActionsMenuService", () => {
       "static-action",
       "healthy-factory-action",
     ]);
-    expect(warnSpy).toHaveBeenCalled();
+    expect(warnSpy).toHaveBeenCalledWith(
+      expect.stringContaining("action factory failed"),
+      expect.objectContaining({ error: expect.any(Error) }),
+    );
   });
 });
--- a/src/app/core/entity-details/entity-actions-menu/entity-actions-menu.service.ts
+++ b/src/app/core/entity-details/entity-actions-menu/entity-actions-menu.service.ts
@@ -21,10 +21,20 @@ export class EntityActionsMenuService {
   private readonly ability = inject(EntityAbility, { optional: true });
 
   getActions(entity?: Entity): EntityAction[] {
-    return [
-      ...this.actions,
-      ...this.actionsFactories.flatMap((factory) => factory(entity)),
-    ];
+    const factoryActions = this.actionsFactories.flatMap((factory) => {
+      try {
+        return factory(entity);
+      } catch (err) {
+        // a single misbehaving factory must not break the whole context menu
+        Logging.warn(
+          "EntityActionsMenu: action factory failed; skipping its actions.",
+          { error: err },
+        );
+        return [];
+      }
+    });
+
+    return [...this.actions, ...factoryActions];
   }
 
   /**
--- a/src/app/core/import/additional-actions/import-additional.service.spec.ts
+++ b/src/app/core/import/additional-actions/import-additional.service.spec.ts
@@ -354,7 +354,10 @@ describe("ImportAdditionalService", () => {
     expect(
       actions.some((a) => a["targetType"] === "RemovedEntity"),
     ).toBeFalsy();
-    expect(warnSpy).toHaveBeenCalled();
+    expect(warnSpy).toHaveBeenCalledWith(
+      expect.stringContaining("not registered"),
+      expect.objectContaining({ unregisteredType: "RemovedEntity" }),
+    );
     // building labels for the remaining actions must not hit an unregistered type
     expect(() =>
       actions.map((a) => service.createActionLabel(a)),
--- a/src/app/core/import/additional-actions/import-additional.service.spec.ts
+++ b/src/app/core/import/additional-actions/import-additional.service.spec.ts
@@ -10,9 +10,13 @@ import {
   expectEntitiesToMatch,
 } from "../../../utils/expect-entity-data.spec";
 import { CoreTestingModule } from "../../../utils/core-testing.module";
-import { DatabaseEntity } from "../../entity/database-entity.decorator";
+import {
+  DatabaseEntity,
+  EntityRegistry,
+} from "../../entity/database-entity.decorator";
 import { DatabaseField } from "../../entity/database-field.decorator";
 import { AttendanceItem } from "../../../features/attendance/model/attendance-item";
+import { Logging } from "../../logging/logging.service";
 
 describe("ImportAdditionalService", () => {
   let service: ImportAdditionalService;
@@ -42,6 +46,9 @@ describe("ImportAdditionalService", () => {
     participants: AttendanceItem[];
   }
 
+  @DatabaseEntity("GroupEntity")
+  class GroupEntity extends Entity {}
+
   @DatabaseEntity("RelationshipEntity")
   class RelationshipEntity extends Entity {
     @DatabaseField({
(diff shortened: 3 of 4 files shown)


Abhinegi2:
@coderabbit review

Abhinegi2:
@coderabbit review

Abhinegi2:
@coderabbit review

sleidig (approved):
Looks good to me. Thanks!

PR #4235, opened 2026-08-06
refactor: unify drag & drop in the admin editors

sleidig:
@Abhinegi2 , I have looked a bit into the multiple drag&drop code we have across components, while reviewing your SQL Admin PR. This is an attempt by claude to simplify and establish more code reuse. Please have a look at it :blush: 

-----

Two independent drag & drop cleanups, one commit each (plus a follow-up fix). Both touch admin editors but are otherwise unrelated — reviewable separately.

## 1. Shared flat-tree drag & drop — menu + report editors

New `src/app/utils/flat-tree/`: generic helpers to edit a tree as one flat, indented list in a single `cdkDropList`.

- The admin menu editor moves off nested, connected `cdkDropList`s. Those can't reliably move items across nesting levels, which forced a `structuredClone` of the whole tree on every drop and a `try/catch` swallowing invalid drops — dropping a parent into its own child is now structurally impossible. `allowSubMenu: false` (shortcuts) becomes `maxLevel: 0` instead of a special case.
- `AdminMenuItemComponent` no longer recurses or carries drag state; it renders one item.
- The SQL report editor uses the same helpers and loses its bespoke drop logic.
- Nesting is now chosen by horizontal drag distance, clamped between what the rows above **and below** allow — without the lower clamp, a shallow drop silently adopts the rows following it.
- Fixes along the way: `openEditDialog` passed the `allowEntityLinks` InputSignal itself rather than its value, so shortcut items offered entity links when edited.

## 2. Entity form editor ("configure data structure")

- Drop handlers no longer splice the arrays CDK hands over and then fake a signal change. Each drop list carries its identity as `cdkDropListData`, each field carries itself as `cdkDragData`, and a drop is a pure `moveFieldBetweenGroups(...)`.
- The toolbar's "drag & drop to / **from** here" now actually works both ways. It had no `id` and nothing connected to it. It also needed `cdkDragBoundary` removed: that clamps the hit-test pointer to `boundary.right - (previewWidth - grabOffset)`, which stops short of the toolbar — no boundary inside the page can satisfy it.

## Verified

Full suite green. New unit tests cover the tree logic, the template wiring (`cdkDropListData` / `cdkDragData` asserted on the rendered directives) and that the config input is never mutated. Drag gestures were exercised in a real browser: menu nest/un-nest and reorder; report nest/un-nest plus the correct refusal to lift a first child that has siblings; entity form group→group, group→toolbar and toolbar→group.

## Not tested

- **Shortcut dashboard widget** (`allowSubMenu=false`): unit-tested only, not clicked through in the app.
- **No e2e coverage for the admin menu editor** — there was none before either, and drag & drop has a long-standing `todo` in `admin.spec.ts`.

## Summary by CodeRabbit

* **New Features**
  * Improved drag-and-drop editing for admin form fields, including moving, reordering, removing, and adding fields to groups.
  * Added more intuitive menu editing with indentation, nesting, unnesting, and movement of complete menu sections.
  * Enhanced report-definition editing with easier query and group nesting, reordering, and removal.
  * Added clearer visual feedback during drag-and-drop interactions.
* **Bug Fixes**
  * Prevented invalid nesting, preserved child items during edits, and improved handling of drag-and-drop targets.

### DIFF
--- a/src/app/core/admin/admin-menu/admin-menu-item/admin-menu-item.component.spec.ts
+++ b/src/app/core/admin/admin-menu/admin-menu-item/admin-menu-item.component.spec.ts
@@ -61,19 +61,14 @@ describe("AdminMenuItemComponent", () => {
     } as MenuItemForAdminUi);
     expect(component.hasNoLinkWarning()).toBe(false);
 
-    const child = {
-      uniqueId: "2",
-      label: "Child",
-      icon: "user",
-      link: "/child",
-      subMenu: [],
-    } as MenuItemForAdminUi;
+    // sub-items are rendered as rows of their own, the editor reports them via `hasSubItems`
     fixture.componentRef.setInput("item", {
       uniqueId: "1",
       label: "Section",
       icon: "folder",
-      subMenu: [child],
+      subMenu: [],
     } as MenuItemForAdminUi);
+    fixture.componentRef.setInput("hasSubItems", true);
     expect(component.hasNoLinkWarning()).toBe(false);
   });
 });
--- a/src/app/core/admin/admin-menu/admin-menu-item/admin-menu-item.component.spec.ts
+++ b/src/app/core/admin/admin-menu/admin-menu-item/admin-menu-item.component.spec.ts
@@ -5,12 +5,19 @@ import { FontAwesomeTestingModule } from "@fortawesome/angular-fontawesome/testi
 import { MenuItemForAdminUi } from "../menu-item-for-admin-ui";
 import { provideRouter } from "@angular/router";
 import { Angulartics2Module } from "angulartics2";
+import { MatDialog } from "@angular/material/dialog";
+import { of } from "rxjs";
+import type { Mock } from "vitest";
 
 describe("AdminMenuItemComponent", () => {
   let component: AdminMenuItemComponent;
   let fixture: ComponentFixture<AdminMenuItemComponent>;
+  let mockDialog: { open: Mock };
 
   beforeEach(async () => {
+    mockDialog = {
+      open: vi.fn().mockReturnValue({ afterClosed: () => of(undefined) }),
+    };
     await TestBed.configureTestingModule({
       imports: [
         AdminMenuItemComponent,
@@ -22,6 +29,7 @@ describe("AdminMenuItemComponent", () => {
           provide: MenuService,
           useValue: { generateMenuItemForEntityType: () => [] },
         },
+        { provide: MatDialog, useValue: mockDialog },
         provideRouter([]),
       ],
     }).compileComponents();
@@ -70,5 +78,18 @@ describe("AdminMenuItemComponent", () => {
     } as MenuItemForAdminUi);
     fixture.componentRef.setInput("hasSubItems", true);
     expect(component.hasNoLinkWarning()).toBe(false);
+  });
+
+  it("should pass the allowEntityLinks value, not the input signal, to the edit dialog", async () => {
+    fixture.componentRef.setInput("allowEntityLinks", false);
+
+    await component.editMenuItem(component.item());
+
+    expect(mockDialog.open).toHaveBeenCalledWith(
+      expect.anything(),
+      expect.objectContaining({
+        data: expect.objectContaining({ allowEntityLinks: false }),
+      }),
+    );
   });
 });
--- a/src/app/core/admin/admin-menu/admin-menu-item/admin-menu-item.component.ts
+++ b/src/app/core/admin/admin-menu/admin-menu-item/admin-menu-item.component.ts
@@ -10,7 +10,6 @@ import {
 import { EntityMenuItem, MenuItem } from "app/core/ui/navigation/menu-item";
 import { MenuItemComponent } from "app/core/ui/navigation/menu-item/menu-item.component";
 import { FaIconComponent } from "@fortawesome/angular-fontawesome";
-import { CdkDragDrop, DragDropModule } from "@angular/cdk/drag-drop";
 import { MatFormFieldModule } from "@angular/material/form-field";
 import { FormsModule } from "@angular/forms";
 import { MatDialog } from "@angular/material/dialog";
@@ -18,7 +17,7 @@ import { MenuService } from "app/core/ui/navigation/menu.service";
 import { firstValueFrom } from "rxjs";
 import { AdminMenuItemDetailsComponent } from "../admin-menu-item-details/admin-menu-item-details.component";
 import {
-  hasNoLinkAndNoSubItems,
+  isManualItemWithoutLink,
   MenuItemForAdminUi,
 } from "../menu-item-for-admin-ui";
 import { MatNavList } from "@angular/material/list";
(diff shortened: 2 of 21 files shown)


sleidig:
@coderabbitai review

src/app/features/reporting/edit-report-definition/edit-report-definition.component.html
```diff
@@ -8,10 +8,10 @@
       cdkDropList
       (cdkDropListDropped)="onDrop($event)"
     >
-      @for (row of rows(); track row.uniqueId; let i = $index) {
+      @for (row of rows(); track row.id; let i = $index) {
         <div
           class="report-row"
-          [class.group-row]="row.isGroup"
+          [class.group-row]="isGroup(row.data)"
```
Abhinegi2: `row.isGroup` was a plain property before, now it is a method call, and the row template does it 5 times (lines 14, 28, 51, 80, 86). That re-runs on every change detection cycle for every row, and the `protected readonly isGroup = isGroupNode` alias in the .ts only exists to make it callable from the template.

Could the flag go back onto the row instead?

```ts
readonly rows = computed(() =>
  flatRows().map((r) => ({ ...r, isGroup: isGroupNode(r.data) })),
);
```

Then the template reads `row.isGroup` again.

src/app/core/ui/menu-item-list-editor/menu-item-list-editor.component.html
```diff
@@ -12,21 +12,24 @@
...
+  @for (row of rows(); track row.id; let i = $index) {
+    <div
+      class="drop-item"
+      [style.margin-left.px]="row.level * indentPerLevel"
+      cdkDrag
+      [cdkDragData]="row"
+    >
+      <app-admin-menu-item
+        [item]="row.data"
+        [hasSubItems]="hasSubItems(i)"
```
Abhinegi2: Same as in the report editor: `hasSubItems(i)` is a method call in the template, so it re-runs for every row on every change detection cycle. It only compares a row's level with the next one, so it could be folded into the `rows` computed as a `hasSubItems` field.

src/app/utils/flat-tree/flat-tree.spec.ts
```diff
@@ -0,0 +1,233 @@
...
+
+  it("measures a subtree as the row itself plus its descendants", () => {
+    const rows = flattenTree(tree, adapter);
+
+    expect(subtreeLength(rows, 0)).toBe(1); // a
+    expect(subtreeLength(rows, 1)).toBe(4); // G, b, Sub, c
+    expect(subtreeLength(rows, 3)).toBe(2); // Sub, c
+  });
+
+  describe("moveSubtree", () => {
```
Abhinegi2: Nested `describe` block. Convention in this repo is flat `it()` with setup inlined per test (432 of 452 spec files). Could the `moveSubtree` cases be hoisted to the top level?

src/app/core/ui/menu-item-list-editor/menu-item-list-editor.component.spec.ts
```diff
@@ -110,15 +111,112 @@ describe("MenuItemListEditorComponent", () => {
...
 
     // Assert
-    const items = component.items();
-    expect(items.length).toBe(1);
-    expect(items[0]).toBe(secondItem);
+    expect(component.items()).toEqual([secondItem]);
     expect(emittedValues).toEqual([[secondItem]]);
   });
 
+  describe("drag & drop of the flat, indented list", () => {
```
Abhinegi2: Same here: nested `describe`. Hoisting the drag & drop cases keeps the `it()` calls flat, the `menuItem` and `drop` helpers can stay at the outer level.

Abhinegi2 (commented):
Nice cleanup, the flat-tree util reads well and the branch coverage in its spec is good. A few things on the new code:

Abhinegi2:
@coderabbit review

src/app/core/ui/menu-item-list-editor/menu-item-list-editor.component.scss
```diff
@@ -1,3 +1,25 @@
...
+
+// CDK drag & drop feedback
+.cdk-drag-preview {
+  box-sizing: border-box;
+  border-radius: 4px;
+  box-shadow:
+    0 5px 5px -3px rgba(0, 0, 0, 0.2),
+    0 8px 10px 1px rgba(0, 0, 0, 0.14),
+    0 3px 14px 2px rgba(0, 0, 0, 0.12);
+}
```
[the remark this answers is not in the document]
Abhinegi2: Checked this and I don't think it applies. CDK builds the preview with deepCloneNode(this._rootElement), i.e. cloneNode(true), so the clone keeps the _ngcontent-* attribute that emulated encapsulation stamped on the cdkDrag element. The compiled selector .cdk-drag-preview[_ngcontent-x] therefore still matches even though the preview is appended to the body, since component styles are plain global stylesheets scoped by attribute rather than by DOM position.

cdkDragPreviewContainer="parent" would also be the wrong trade here: the drop list is flex-column gap-regular, and the CDK docs note that the parent container "is going to have an effect over selectors like :nth-child and some flexbox configurations", so the preview would push the rows around while dragging.

The block is unchanged, it just moved from admin-menu-item.component.scss along with the cdkDrag element it styles.

src/app/core/admin/admin-menu/admin-menu-item/admin-menu-item.component.ts
```diff
@@ -134,7 +110,7 @@ export class AdminMenuItemComponent {
       width: "600px",
       data: {
         item: item ? { ...item } : {},
-        allowEntityLinks: this.allowEntityLinks,
+        allowEntityLinks: this.allowEntityLinks(),
```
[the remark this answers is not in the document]
Abhinegi2: Leaving this in. allowEntityLinks is an input() signal and was being passed to the dialog as the signal function instead of its value, so shortcut items offered entity-type links when edited. The surrounding component is being rewritten in this PR anyway, so a separate PR would have to touch the same line and land in a fixed order.

Added a regression test in admin-menu-item.component.spec.ts instead, verified it fails when the () is removed again.

src/app/core/admin/admin-menu/admin-menu-item/admin-menu-item.component.spec.ts
```diff
@@ -5,12 +5,19 @@ import { FontAwesomeTestingModule } from "@fortawesome/angular-fontawesome/testi
...
 
 describe("AdminMenuItemComponent", () => {
   let component: AdminMenuItemComponent;
   let fixture: ComponentFixture<AdminMenuItemComponent>;
+  let mockDialog: { open: Mock };
 
   beforeEach(async () => {
+    mockDialog = {
+      open: vi.fn().mockReturnValue({ afterClosed: () => of(undefined) }),
+    };
```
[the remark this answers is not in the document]
Abhinegi2: not applicable; the repo moved off jasmine to vitest

Abhinegi2 (approved):

PR #4236, opened 2026-08-06
fix(sync): raise permission purge to error level when local edits may be lost

sleidig:
- relates to #4198

## What

The sync path that purges local documents after the server reports lost permissions logged at `warn` level unconditionally (`synced-pouch-database.ts`). Sentry issue [AAM-DIGITAL-74M](https://aam-digital.sentry.io/issues/AAM-DIGITAL-74M) has been `escalating` for eight consecutive daily monitoring runs — **177 lifetime events, 37 in the last 24h alone** (up from 16 the previous day) — and because every event is `level=warning`, it has never generated a single alert. A path that can destroy unsynced user data is the one thing that should not be silent.

Simply raising everything to `error` would be noisy and wrong. Monitoring across several runs established that **the document count does not measure data loss — the sync gap does**:

| observed purge | docs | gap since last completed sync | actual risk |
|---|---|---|---|
| largest today | 1958 | ~28 seconds | none — everything had just synced |
| earlier run | 3294 | ~27 seconds | none |
| earlier run | 531 | ~35.5 hours | real — device was a day and a half stale |

Triaging by volume ranks those exactly backwards.

## Change

Severity is now derived from the sync gap:

- gap within `PURGE_DATA_LOSS_RISK_THRESHOLD` (5 min, 10× `SYNC_INTERVAL`) → stays `Logging.warn`
- gap beyond it, **or a missing/unparseable last-sync time** → `Logging.error`, so it reaches alerting

`syncGapMs` and `riskOfDataLoss` are added to the structured context, so triage no longer means subtracting two timestamps by hand.

The purge itself is unchanged — this only affects how it is reported. The threshold is a single named constant and easy to re-tune once real error-level volume is visible.

Ordering note: `purgeDocsWithLostPermissions()` runs before `syncState` emits `COMPLETED`, which is what rewrites the last-sync key — so the value read here is genuinely the *previous* completed sync, which is the window in which local edits are at risk.

## Testing

`npx ng test --configuration=ci --include='src/app/core/database/pouchdb/synced-pouch-database.spec.ts'` → **22 passed**, including three new cases covering the just-synced (warn), stale (error), and unknown-last-sync (error) branches.

Manual verification: not reproducible locally without a server-driven `lostPermissions` response; the behaviour is covered by the unit tests above, and the change is confined to log severity and context.

## PR Checklist

- [x] manually tested (all) required functionality yourself and added comment with clear steps for manual testing — covered by unit tests; see note above

---
_Generated by [Claude Code](https://claude.ai/code/session_01NeFid5CygnguNrhyDLgoJR)_

### DIFF
--- a/src/app/core/database/pouchdb/synced-pouch-database.spec.ts
+++ b/src/app/core/database/pouchdb/synced-pouch-database.spec.ts
@@ -10,6 +10,7 @@ import { Subject } from "rxjs";
 import { SyncState } from "../../session/session-states/sync-state.enum";
 import { SyncedPouchDatabase } from "./synced-pouch-database";
 import { NotAvailableOfflineError } from "../../session/not-available-offline.error";
+import { Logging } from "../../logging/logging.service";
 
 describe("SyncedPouchDatabase", () => {
   let service: SyncedPouchDatabase;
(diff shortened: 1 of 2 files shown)


sleidig:
This is not a technical bug but just logged to help us monitor user-specific behavior that may potentially be dangerous (e.g. to be able to respond to user support requests).

We will _not_ change the log level for this.

Issue #1991, opened 2023-09-07
Import Module improvements


sleidig:
Based on initial user observations, there are a number of small-ish things to improve about our import UI:

- not intuitive what are columns, what are profile entity fields (visualization?)
- no warning indicator for date after selecting field, but format config is required
- popup warning for date value mapping about "unmapped values" would better be phrased as "unparseable values" or the like
- warning popup for boolean value mapping "unmapped values" is misleading if the user actually wants to map some values to be - imported as "unchecked" (i.e. false) but there is no way to distinguish unmapped and mapped to "false" currently
- if column headers are not in first row -> column mapping screen very confusing (!)
- boolean value mapping: more visual clarity that left value will turn into right value (checkbox is not selecting what value from left is imported but represents the actual value after import)
- dropdown options that are long and have a line break are hard to distinguish from the next option in the list
- imported data might contain multiple different date formats (with/without leading zero of month, ...)

- preview screen:
  - empty strings are shown as "null"
  - unmapped or empty values of enum props are shown as "[invalid option] undefined" in the preview (should just be empty)
  - no loading bar for preview (but very slow - maybe for address lookup)
  - sort order of rows to match the imported file? or maybe an extra column giving the row in imported file?
  - scrollbar horizontal not available in preview (some columns seem to be "missing" because they are out of the screen)

MalteHD:
@sleidig @george-neha :
I had two enhancement ideas and found a potential issue:

1) For date-fields (which need configuration mapping), pre-set the default mapping to YYYY-MM-DD. Reasoning: a) That is the format in which Excel stores date fields (if I'm not mistaken) so should already cover a significant % of imports. b) it gives the user a better idea as to what kind of input our system expects at this stage.

2) For dropdown type fields, I often have imported values that match 100% to the available dropdown value options. These will be pre-selected which is a very nice feature of ours. For me as a user, it would be nice if in case of such a "100% matching", that would be indicated to me somehow. Maybe even in the list with all the fields and maybe even set this mapping to "fine". So that I don't have to click on configure mapping and check myself and confirm. That would save me multiple clicks for each field.

[...]

PR #4247, opened 2026-08-10
fix(import): count all values of a column that have no value mapping

### LINKED ISSUE #4221: Import: warning count on a column does not match the number of unmapped values
**Describe the bug**

In the import column mapping step, the "Configure value mapping" button of a column shows a warning badge with the number of file values that have no mapping. That number does not match how many values the user actually left unmapped, so it cannot be used to decide whether a column is ready to import.

**To Reproduce**

1. Go to Import and upload a file with a column holding several distinct values for a dropdown field (e.g. a gender column with `female`, `male`, `other`).
2. Continue to "Map Columns" and open "Configure value mapping" for that column.
3. Assign a system value to every imported value and press "Save & Close".
4. Reopen the dialog, clear the assignment of **one** value, save. The badge shows `1`, which is correct.
5. Reopen the dialog, clear **two more** assignments, save. The badge does not show `3`.

**Expected behavior**

The badge always shows how many of the column's distinct file values currently have no assigned value, and disappears once all of them are mapped.

**Additional context**

Not browser specific, the count is computed in the app itself. Found while implementing the review flow for the value mapping dialogs (#4185, PR #4217).

Leads for whoever picks this up, none of them verified yet:

- The count is `Object.values(values).filter((v) => v === undefined).length` in `discrete-import-config.component.ts`, so it depends on unmapped entries surviving as keys whose value is `undefined`.
- `JSON.stringify` drops object keys with an `undefined` value. Any place where the column mapping is serialized and parsed back (for example the auto-mapping comparison in `import-column-mapping.component.ts`, or storing the config in `ImportMetadata`) silently loses exactly those entries, which would make the count too low after a round trip.
- Clearing a dropdown in the dialog may leave the form control at `null` rather than `undefined`, which is not counted at all.
- The same "`undefined` means unmapped" assumption drives the "Some values don't have a mapping" confirmation in the dialog's `save()`, so that prompt is likely wrong in the same cases. Note it uses a plain truthiness check, which additionally treats a deliberate `false` (unchecked) as unmapped, as already listed in #1991.

A fix should cover both the badge and that confirmation prompt, and come with a unit test per case (all mapped, one cleared, several cleared, reopened after a save).

Abhinegi2:
- closes #4221

The warning badge on a column's "Configure value mapping" button counted mapping entries whose value is empty. An empty value does not survive serializing the mapping (e.g. when it is stored with the import history), so those entries were silently dropped and the badge showed too low a count. Reopening the dialog then also auto-assigned those values again, undoing the user's decision to not import them.

Values without an assignment are now marked as `null` instead, so they survive and the count stays correct. A value that is deliberately mapped to an empty target value (an unchecked checkbox) counts as mapped, both for the badge and for the confirmation prompt on save. This also ticks the "warning popup for boolean value mapping is misleading" item of #1991.

Not covered here: reusing a previous mapping from the import history applies it to a new file, whose values are not necessarily keys of the stored mapping, so the badge cannot count them. Different trigger and repro, to be filed separately.

## PR Checklist

---

## Manual Testing Steps

Test on the PR deployment (demo mode is enough, the import runs entirely in the browser). The steps use the education base config: Participant/Child has a `Gender` dropdown, School has a `Private School` checkbox.

Save these two files locally first.

`gender.csv`

```csv
name,gender
Aisha,f
Ben,m
Chandra,nb
Dara,none
Ella,f
```

`school.csv`

```csv
name,private
Green Valley School,yes
Riverbank School,no
```

### 1. The badge counts every value that has no mapping

1. Import, upload `gender.csv`, select the record type Participant, continue.
2. Map the `gender` column to `Gender`. The value mapping dialog opens on its own.
3. All 4 values of the file (`f`, `m`, `nb`, `none`) are pre-filled with themselves. Assign each one properly: `f` to Female, `m` to Male, `nb` and `none` to Non-binary. Save & Close.
   - no badge on the button, the tooltip states that all values are mapped
4. Reopen, clear **one** assignment, Save, confirm the "Ignore values?" prompt.
   - badge shows `1`
5. Reopen, clear **two more** assignments, Save, confirm.
   - badge shows `3` (this showed a wrong, too low number before)
6. Reopen once more.
   - the 3 cleared rows are still empty (before, cleared values were auto-assigned again)
   - after Save & Close the badge still shows `3`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved discrete-value import mapping to correctly handle unmapped `null` and `undefined` values.
  * Preserved valid mappings for falsy values such as `false` and `0`.
  * Ensured unmapped values remain consistent when saved, serialized, and reopened.
  * Updated mapping counts, badges, and tooltips to accurately reflect unmapped entries.

### DIFF
--- a/src/app/core/basic-datatypes/configurable-enum/configurable-enum-datatype/configurable-enum.datatype.spec.ts
+++ b/src/app/core/basic-datatypes/configurable-enum/configurable-enum-datatype/configurable-enum.datatype.spec.ts
@@ -172,6 +172,21 @@ describe("Schema data type: configurable-enum", () => {
     expect(actualMapped).toBeUndefined();
   });
 
+  it("should return undefined for values that are marked as not to be imported", async () => {
+    enumService.getEnumValues.mockReturnValue(genders);
+
+    const actualMapped = await dataType.importMapFunction(
+      "skipped",
+      {
+        dataType: "configurable-enum",
+        additional: "genders",
+      },
+      { values: { MALEx: GENDER_MALE.id, skipped: null } },
+    );
+
+    expect(actualMapped).toBeUndefined();
+  });
+
   it("should create invalid option when mapping dialog is skipped", async () => {
     enumService.getEnumValues.mockReturnValue(genders);
 
--- a/src/app/core/basic-datatypes/discrete/discrete-import-config/discrete-import-config.component.spec.ts
+++ b/src/app/core/basic-datatypes/discrete/discrete-import-config/discrete-import-config.component.spec.ts
@@ -63,6 +63,17 @@ describe("DiscreteImportConfigComponent", () => {
     expect(component.tooltip()).toContain("All values");
   });
 
+  it("should count values that are marked as not to be imported", () => {
+    fixture.componentRef.setInput("col", {
+      column: "gender",
+      propertyName: "category",
+      additional: { values: { female: "F", male: null, other: null } },
+    });
+
+    expect(component.badge()).toBe("2");
+    expect(component.tooltip()).toContain("2");
+  });
+
   it("should open dialog and notify with the configured mapping", async () => {
     const onChangeFn = vi.fn();
     const rawData = [{ gender: "male" }, { gender: "female" }];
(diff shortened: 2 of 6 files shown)


Abhinegi2:
@coderabbit review

sleidig (approved):
Looks good to me 👍

PR #4264, opened 2026-08-13
feat(change-history): admin Change Log screen to view all changes across the system

### LINKED ISSUE #4039: Change Log: UI to view all changes across system
#4027 provides a UI to see the change history of a specific record. However, admins may want to see changes across the whole system. Also, deletions cannot be seen as history because then the details view to access that change history UI is not available anymore.

This could be linked from the Admin Overview and also from the List View of entities (with the entity type pre-selected as a filter)

Abhinegi2:
- closes #4039

First draft, scoped to the admin screen only. Lists every recorded change across the whole system: **when, which record, what kind of change, changed by, changed fields** — plus filters for record type, action, author, date range and related record.

Deliberately out of scope for this draft (agreed with the team): undo/restore, and a field-level before/after diff in the list itself. Clicking a row opens that record's full change-history dialog rather than its detail page.

The listing follows the same shape as the User Management screen (signals + `mat-table` + `mat-paginator`), and reuses the existing action badge, entity block and field-label components from the per-record change history.

### Notes for the reviewer

**Why a different query than the per-record history.** Audit `_id`s are keyed by entity (`AuditRecord:<entityId>:<ts>:<rev>`), so `_all_docs` orders by record, not by time, and cannot produce a system-wide newest-first list. This queries the audit database's proxied `_find` endpoint sorted on a `timestamp` index that is created on demand. No backend change was needed — both `_find` and `_index` are already proxied and permission-filtered.

**Paging.** Each page requests one record beyond its size and reports that as `hasMore` rather than returning it, so a page that exactly fills is distinguishable from the last page. Without this, "next" was offered on an exactly-full result and opened an empty page. Paging is positional (`skip`) rather than by cursor, since a cursor cannot answer "is there more" without consuming the record it would then have to skip.

**Changed-field names come from the audit document alone** — an update delta is already keyed by field name, so the names (unlike before/after values) need no state replay. That keeps a row renderable without loading a per-entity history.

**Permissions.** The screen is behind the `admin_app` role; the data itself is gated by `read` on the `AuditRecord` CASL subject, enforced server-side by the replication-backend proxy. Two things worth a decision:

- `AuditRecord` is not granted in any base config, so it currently only works via the "no rules defined = allow everything" fallback. Systems that define permission rules will show the no-access state until an `AuditRecord` read rule is added.
- `AuditRecord` is all-or-nothing — there is no per-entity-type filtering of audit data, so anyone holding it sees record names and changed-field names for records they may not be permitted to open. Is that the intended granularity?

**Known trade-offs**

- Only the date bounds use the index; record type and author are applied while CouchDB walks it. Results are correct, but a narrow filter over a long history means a longer walk to fill a page. A backend `entityType` field plus a composite index would fix this later.
- `skip` costs more than a cursor at depth. Fine at 10–100 per page; noticeable only if someone pages hundreds deep.
- The author filter options are sampled from the most recent 1000 records, since the audit database has no index of its authors.
- No free-text search: the audit records store only entity ids, and record names live in the app database, so a real search needs backend support.
- An edit that only touched the `updated` metadata renders as "no visible field changes" — the backend keeps that field in the diff on purpose and the UI hides it.
- Date presets use local-time boundaries, matching the reports screen.

`eslint.config.mjs` gained one entry: `subscriptSizing` is a Material structural attribute, but the i18n rule did not know it and `--fix` kept inserting an `i18n-subscriptSizing` marker for a non-translatable API value.

## PR Checklist

---

## Manual Testing Steps

This feature needs the replication-backend with change logging switched on, so it cannot be tested on the PR demo deployment (mock mode has no backend). Test on the dev site with this branch deployed.

**Setup**

1. Open the dev site and sign in with an account that has the app-admin role.
2. Make a few changes first so the log has content: edit a child's phone number, create a note, and delete a record.

**See the change log**

3. Go to **Admin → Admin Overview → Export and Backups → Change Log** (route: `/change-history`).
4. You should see a table, newest change at the top, with six columns: when the change happened (with "2 days ago" underneath), which record, its record type, what kind of change (a coloured "Updated" / "Created" / "Deleted" tag), who made it, and which fields changed as small chips.
5. Check the record you deleted still appears, marked "Deleted", with "record removed" instead of field chips. This is the main thing the per-record history cannot show.
6. Check the record names are real names, not ids. A record whose type configures no name shows "<Type> Record"; a record that no longer exists shows a greyed-out "not available".
7. Under each record name there is a small grey id. Click it: it should copy to the clipboard and confirm with "Record ID copied", and it must **not** open the row's dialog. Clicking anywhere else in the row does open that record's full change history.

**Entry points**

8. Open an entity list (e.g. Children), open its three-dot menu and choose **Change Log**. The screen opens with the record type already filtered to that list's type.
9. Open a record's details page, three-dot menu → **View change log**. The dialog shows that record's full history, with its internal id under the name (also click-to-copy).

**Filters**

10. Pick a record type, for example "Student". Only changes to students should remain, and the list should jump back to the first page.
11. Click the small **x** next to the filter. All record types come back, and the dropdown should not pop open when you click the x.
12. Pick "Deleted" under "Action". Only deletions remain.
13. Pick a name under "Changed by". Only that person's changes remain.
14. In "Date range", click the calendar icon and choose "Last month". Only changes from last month remain. You can also type a start and end date directly.
15. Paste a record id (copied in step 7) into "Related record". You should see the changes involving that record, including changes to other records that reference it. Record type, action and changed-by grey out with a tooltip explaining they are unavailable while this filter is set; the date range stays usable.

**Paging**

16. Set "Items per page" to 10 and page through with the arrows. Going forward and back should show the same rows each time.
17. Filter down to a type that has exactly 10 or 20 changes. When you reach the last page, the "next" arrow must be greyed out — you should never be able to reach a page that says "No changes recorded".

**Permissions and off states**

18. Sign in as a non-admin user and try to open `/change-history` directly. You should not get in.
19. On a system where change logging is switched off, the page shows "This feature is currently not enabled for your system" instead of the table.

## Summary by CodeRabbit

* **New Features**
  * Added an admin Change Log with filters for entity, action, author, related record, dates, and pagination.
  * Added record history access from entity lists and administration menus.
  * Added copyable record IDs and improved change-log record details.
  * Added actions for emptying records, resetting the system, and resetting the local device, with clearer warnings.
* **Improvements**
  * Improved entity display fallbacks, including optional IDs for unavailable records.
  * Improved labels when entity types have no configured name.


Abhinegi2:
@coderabbit review

george-neha:
1. Captured new details of an existing Child profile >Education details>ASER results.
In the change log under "Record"- dispkay Child name instead of ID(see below first record)
image

2. In the change log, add a filter for "Action" with values <Created/Updated/Deleted>

Abhinegi2:
> 1. Captured new details of an existing Child profile >Education details>ASER results.
>    In the change log under "Record"- dispkay Child name instead of ID(see below first record)
> 
regarding the ASER result:
1. since you linked an ASER result to a child, an ASER record is created. That’s why it is currently showing the ASER record instead of the child. Right now, the audit/changelog feature doesn’t show changes to related/linked records, which is why the list displays the ASER internal ID instead of the child.
2. But, I noticed that in the 'Record Type' column, we should also show the internal entity type. I’ll add that. (in above ss its missing)

> 2. In the change log, add a filter for "Action" with values <Created/Updated/Deleted>
3. Regarding the new filter, I think we can add it as well.

@sleidig, what do you think about points 1 and 3?

sleidig:
> > 1. Captured new details of an existing Child profile >Education details>ASER results.
> >    In the change log under "Record"- dispkay Child name instead of ID(see below first record)
> 
> 1. since you linked an ASER result to a child, an ASER record is created. That’s why it is currently showing the ASER record instead of the child. Right now, the audit/changelog feature doesn’t show changes to related/linked records, which is why the list displays the ASER internal ID instead of the child.

Good point for the UX. These raw IDs are super intimidating and impractical.

However, we cannot (technically, without significant extra work) and should not (for consistency) show the child name there. The change is about a different record (although it is closely related to the Child.
If we add an option for the `toStringAttributes` (defining how the title of a record looks), that can include the name of a related record (e.g. for ASER showing "Education Results (19.08.2026) for John Doe") that would be awesome and also show up here - but that is a different and somewhat complex topic.

For now, I would suggest we more clearly show that it is an "ASER" record. Maybe even our default displaying of entities that only show with their _id could read something like "<EntityType Label> Record" (and in small letters only the ID) everywhere? (i.e. it would show here as "ASER Record" instead of the raw ID).

Your thoughts, @Abhinegi2 , @george-neha ?


> > 2. In the change log, add a filter for "Action" with values <Created/Updated/Deleted>
> 
> 3. Regarding the new filter, I think we can add it as well.

Yes, sounds good (not 100% sure which place you are talking about, but let's offer that widely)

george-neha:
1. Child>Config data structure> Added a field (blood group) from the drag-and-drop column. The current data shown in the change log does not really give any information to the user(see below)
image
image

Abhinegi2:
@coderabbit review

Abhinegi2:
@coderabbit review

src/app/features/change-history/change-history-dialog/change-history-dialog.component.html
```diff
@@ -1,9 +1,19 @@
...
   <app-dialog-close></app-dialog-close>
 </h2>
+<!-- which record this history belongs to, so the dialog is identifiable at a
+     glance when it was opened from a list of many records -->
+<div class="record-subtitle text-secondary margin-bottom-small">
+  <app-entity-block
+    [entity]="entity"
+    [linkDisabled]="true"
+    [showIdEntityId]="true"
+  ></app-entity-block>
```
sleidig: Let's add a small, grey line here, displaying the raw entity ID. Otherwise it is really hard to get that for code-based troubleshooting or pasting it into the "related record" filter.

- show full ID (in a very light way that doesn't grab users attention, confusing them)
- click = copy to clipboard
- tooltip: This is the internal unique ID for this record. Copy this for advanced analysis or as context to help your technical administrators with troubleshooting.

src/app/features/change-history/change-history-list/change-history-list.component.html
```diff
@@ -0,0 +1,344 @@
...
+        </p>
+      } @else if (!isLoading() && entries().length === 0) {
+        <p class="padding-regular text-secondary" i18n="Change log empty">
+          No changes recorded for the selected filters.
+        </p>
+      }
+
+      <mat-paginator
+        class="table-footer"
+        [length]="pageLengthHint()"
```
sleidig, line 334: we should show the same style pagination as the normal entity lists with a `10+` indicating that there are more results on the next page.

@Abhinegi2 , could we actually reuse more of those components now that the paginated datasource was merged? But if that's complex, let's just manually align the UX for now.

src/app/core/admin/admin.routing.ts
```diff
@@ -155,6 +156,14 @@ export const adminRoutes: Routes = [
     },
     canActivate: [EntityPermissionGuard],
   },
+  {
+    path: "change-log",
```
sleidig: maybe just put this at /changelog (without the /admin/changelog nesting)?

Then we could maybe move the route to its own feature module also, instead of wiring /feature code into the /core admin.routing.ts?

src/app/core/basic-datatypes/entity/entity-block/entity-block.component.ts
```diff
@@ -56,6 +57,14 @@ export class EntityBlockComponent {
 
   linkDisabled = input(false);
 
+  /**
+   * Show the raw entity id instead of the generic "not available" when the
+   * referenced record cannot be loaded. Opt-in, for lists where the reader needs
+   * to identify *which* record is gone (e.g. an audit log of deletions) rather
+   * than only that one is missing.
+   */
+  showIdEntityId = input(false);
```
sleidig: should this rather be named "showEntityId"?

src/app/core/entity-list/entity-list/entity-list.component.ts
```diff
@@ -116,6 +119,18 @@ export class EntityListComponent<T extends Entity> implements OnInit {
...
   private readonly injector = inject(Injector);
+  private readonly sessionSubject = inject(SessionSubject);
+
+  /**
+   * The change log lives behind the admin route, so only offer the link to
+   * someone the route will actually let in.
+   */
+  readonly canViewChangeLog = toSignal(
+    this.sessionSubject.pipe(
+      map((session) => session?.roles?.includes("admin_app") ?? false),
```
sleidig: I think we have a constant to use for "admin_app" string somewhere? Or otherwise, we should introduce one

src/app/features/change-history/change-log/change-log.component.html
```diff
@@ -0,0 +1,344 @@
...
+          </th>
+          <td mat-cell *matCellDef="let entry">
+            <div class="flex-column">
+              <span>{{ entry.at | customDate: "short" }}</span>
+              <span class="text-secondary">{{
+                entry.at | notificationTime | async
+              }}</span>
+            </div>
+          </td>
+        </ng-container>
```
sleidig: Maybe we can tweak the layout for this, trying to give the column enough width for the whole date. And showing the duration in a bit smaller font, lighter color - to make it easier to separate?

Image

sleidig (commented):
first pass of review. Couldn't finish all code today yet ...

Overall looks pretty solid but the amount of code takes time to go through.

Abhinegi2:
@coderabbit review

sleidig:
Do we have e2e tests for the change history feature yet, @Abhinegi2  ? Let's add a scenario?

src/app/app.routing.ts
```diff
@@ -68,6 +69,11 @@ export const allRoutes: Routes = [
     canActivate: [AuthGuard],
     children: AttendanceModule.routes,
   },
+  {
+    path: "changelog",
+    canActivate: [AuthGuard],
+    children: ChangeHistoryModule.routes,
+  },
```
sleidig: we currently have a mix of naming: sometimes "changelog" sometimes "change history". I have no clear preference but let's make it consistent to only one term, that'll make code search easier in the future.

This applies to:
- route + module
- components, services, folders, etc.

src/app/features/change-history/change-history.service.spec.ts
```diff
@@ -1,34 +1,72 @@
...
+      {
+        provide: EntityAbility,
+        useValue: {
+          can: abilityCan,
+          on: (_event: string, listener: () => void) => {
+            abilityListeners.push(listener);
+            return () => undefined;
+          },
+        },
+      },
```
sleidig: do we really need this complex EntityAbility mock here, registering abilityListeners? Seems unused in the actual test expectations.

src/app/features/change-history/change-history.service.spec.ts
```diff
@@ -125,9 +163,187 @@ it("denies viewing history when no entity is given", () => {
...
+  expect(service.hasAuditPermission()).toBe(false);
+
+  abilityCan.mockReturnValue(true);
+  // the ability object is mutated in place by AbilityService, so only its
+  // "updated" event tells us the answer may have changed
+  abilityUpdated();
+
+  expect(service.hasAuditPermission()).toBe(true);
+});
+
```
sleidig: this seems over-engineered. Let's disregard changing permissions until the component is reloaded anyway. (or, if the functionality comes out-of-the-box, remove the test and just silently enjoy it)

src/app/core/permissions/permission-types.ts
```diff
@@ -43,6 +43,12 @@ export const PUBLIC_SECTION_KEY = "_public";
 export const LEGACY_DEFAULT_KEY = "default";
 export const LEGACY_PUBLIC_KEY = "public";
 
+/**
+ * The realm role that grants access to the administration features, as checked
+ * by the admin routes' {@link UserRoleGuard} configuration.
+ */
+export const ADMIN_APP_ROLE = "admin_app";
+
```
sleidig, line 54: refactor to replace literal `"admin_app"` across the code base with this
(in a separate PR)

src/app/features/change-history/change-history-list/change-history-list.component.html
sleidig, line 1: does it really make sense to implement a mat-table, including reinventing all the filter components, etc.? Or could we switch to our entities-table component / EntityList component?

I think we did consider it, but can you share the trade-offs based on the current state of code again, @Abhinegi2 ?

src/app/core/common-components/record-id-display/record-id-display.component.ts
```diff
@@ -0,0 +1,63 @@
...
+ * The internal id of a record, shown quietly and copyable in one click.
+ *
+ * A record's title identifies it only by convention: records generated from one
+ * template share theirs (every event of a recurring activity), so the id is what
+ * actually tells two apart. It is also what the change log's related-record
+ * filter takes, and what a technical administrator needs for troubleshooting.
+ *
+ * Deliberately understated: readers looking for a record by name should not have
+ * their attention pulled to a string that means nothing to them.
+ */
```
sleidig, line 22: isn't this overlapping quite a bit with the entity-block component now, if that also has an option to "showOnlyId"?

src/app/features/change-history/change-history-normalize.ts
```diff
@@ -41,6 +41,36 @@ function isHidden(field: string): boolean {
...
+ *
+ * This is what the system-wide change log needs: a top-level jsondiffpatch
+ * delta is already keyed by field name, so the *names* (unlike the before/after
+ * *values*, see {@link buildChangeEvents}) need no prior state. That makes a row
+ * renderable from its own audit doc alone, so the log can page over records of
+ * many different entities without loading a history per entity.
+ *
+ * A `delete` returns no fields on purpose: PouchDB pushes a tombstone stripped
+ * of its content, so the delta would list every field of the record as removed
+ * rather than the one thing that happened.
```
sleidig: Is referencing PouchDB really important here? Better if this feature is database-agnostic, generic. (It probably is, except for the elaborate code comment, I guess?)

src/app/core/permissions/permission-types.ts
sleidig: done here: https://github.com/Aam-Digital/ndb-core/pull/4347

src/app/features/change-history/change-history-list/change-history-list.component.html
Abhinegi2: blocker is the data source, not the table. `TableRow`, `EntitiesTableDataSource` and `PaginatedDataSource` are all typed `<T extends Entity>`, and loading goes through `entityMapper.findType(entityCtr, …)` — the app database. Audit records aren't entities: separate `<db>-audit` database behind the `/db` proxy, Mango `_find` plus a view for the related-record filter, permission-filtered server-side. And there's no seam to plug another source in: `resolveDataSource()` picks from a fixed map and forces `InMemoryDataSource` outside online-only mode.

also filters follow from that rather than being a separate problem — `FilterGeneratorService` builds options from an entity *schema*, and these aren't schema fields ("record type" is a list of entity types; "changed by" is sampled from the audit database, which has no author index).

So reusing it needs either modelling `AuditRecord` as a real `Entity` with a schema, or generalising `EntitiesTableDataSource` off `Entity` — both core refactors beyond this PR. 

so can create a separate issue for this @sleidig  what you think?

sleidig: We do have https://github.com/Aam-Digital/ndb-core/blob/526b1db67ffa308aea3e436306dc234ff1cf42c4/src/app/core/entity/model/entity.ts#L79

that allows to configure another couchDB that entities are coming from, and/or 
https://github.com/Aam-Digital/ndb-core/blob/526b1db67ffa308aea3e436306dc234ff1cf42c4/src/app/core/entity/entity-special-loader/entity-special-loader.service.ts#L9-L15

to define special loading logic.

Could we integrate the AuditRecords into that and then use one of the existing data sources?
Maybe briefly talk with @TheSlimvReal to get his thoughts, as he has implemented the data source.

We shouldn't invest endless days on this right now but on the other hand it'll pay of to reduce code duplication as much as possible here before merging.

src/app/features/change-history/change-history.service.ts
```diff
@@ -110,6 +154,157 @@ export class ChangeHistoryService {
...
+    filters: ChangeHistoryFilters,
+    pageSize: number,
+    pageIndex: number,
+  ): Promise<ChangeHistoryPage> {
+    await this.ensureReferenceIndex();
+    const response: ViewResponse = await this.getAuditDb().query(
+      AUDIT_REFERENCE_VIEW,
+      buildReferenceViewQuery(filters, pageSize, pageIndex),
+    );
+    // a row without a doc is one the backend's permission filter removed
```
TheSlimvReal, line 218: Are you sure about this? I have just looked at the [backend code](https://github.com/Aam-Digital/replication-backend/blob/2c134b915578b8ab164a0ce66777cf42415c96f7/src/restricted-endpoints/document/design-document.controller.ts#L131-L143) and I understand it filters out the rows where permissions are missing and does not leave empty values in the response.

src/app/features/change-history/change-history.service.ts
```diff
@@ -110,6 +154,157 @@ export class ChangeHistoryService {
...
+    const docs = (response.rows ?? [])
+      .map((row) => row.doc)
+      .filter((doc): doc is RawAuditDoc => !!doc);
+    return {
+      entries: docs.slice(0, pageSize).map(toChangeHistoryEntry),
+      hasMore: docs.length > pageSize,
```
TheSlimvReal, line 224: If my assumption in the previous comment is right and the whole row is dropped if permissions are missing, this approach does not work as less docs than requested can be returned while there still are more to fetch on the DB. If think this needs a different approach like I already implemented for the `_find` endpoint [here](https://github.com/Aam-Digital/replication-backend/blob/2c134b915578b8ab164a0ce66777cf42415c96f7/src/restricted-endpoints/replication/bulk-document/bulk-doc-endpoints.controller.ts#L187-L225). I am not 100% sure if the bookmark is also available with for queries. If not, I think with the returned `offset` this could also be achieved.

src/app/features/change-history/change-history-list/change-history-list.component.html
TheSlimvReal: I had a first look at the code and I also think that there is a lot of duplicate code for the whole management of the entities, the database and the table. It is right, that our current components for this require the docs to be entities. Is there a reason why `RawAuditDoc` is not an entity?

- The transformation of `RawAuditDoc` to `ChangeEvent` seems to be exactly what the `EntitySchemaService` does. Properties with special logic could become a new datatype.
- The `ChangeHistoryService` uses the same approach as the `EntityMapperService` or `DatabaseIndexingService` for interacting with the db
- The `ChangeHistoryComponent` mostly has features of the existing `EntityListComponent` (or at least `EntitiesTableComponent` if filters need to be created manually)
- The implemented pagination is exactly what we recently added to the generic data source and it would be a shame if we cannot make it work here

As far as I can tell, the changes would not require any data migration. It would definitely need many code changes, maybe also outside of this PRs new additions. But I think it would be worth the effort because it would remove most of the more complex code and only need some simple component that define the view, some new datatypes and maybe filters that easily plug into the existing system, as well as a simple service that creates the necessary views. All the complex stuff (db interaction, displaying of the results, pagination) would be taken care of).
I am curious to hear of your concerns (@Abhinegi2  @sleidig). We can also hop on a call to look at the code in detail for this.

Abhinegi2:
## Plan for the reuse refactor

Agreed on the direction: `AuditRecord` becomes a real entity and the list moves to `<app-entities-table>` over `PaginatedDataSource`. Below is the plan, and the findings behind a few decisions that aren't obvious from the outside.

### Three PRs

**1. ndb-core — remote-only databases per entity type.** Based on this branch. `Entity.DATABASE_REMOTE_ONLY` plus one branch and three skip guards in `DatabaseResolverService`, a small `remote-only-databases.ts`, and a bookmark-paginated `find()` fake for the testing module.

Separate because it touches `registerDatabase()`, which every login goes through — about 30 lines that are easy to check on their own and easy to skim inside a large refactor diff.

**2. ndb-core — the refactor.** This PR, as six commits:

1. drop the related-record filter and its CouchDB view
2. `supportsArchiving` input on `EntitiesTableComponent`, plus the `display-entity` / `record-id-display` fixes the list needs
3. the `AuditRecord` entity and three view components
4. the list on `<app-entities-table>` + `buildAuditFilter`; delete `change-history-query.ts` and the hand-rolled `_find` / index / paging code
5. permissions: grant `AuditRecord` read by name, dual route guard, CASL-aware menu item
6. revert the `eslint.config.mjs` entry and the five `subscriptSizing` attributes

**3. replication-backend — audit write guard and server-side index creation.** Different repo, so it isn't in the stack. It needs deploying, not just merging, before this goes live.

PRs 1 and 2 both target this branch and merge into it, so there's no rebasing against a moving master and no squash-collision.

### Why the reuse is now the shorter path

The list pagination is broken today, not on the next release. The backend does:

```
body.limit = body.limit ?? 25;
delete body.skip;   // "body.skip is discarded, only body.bookmark is supported"
```

That shipped on 2026-08-30. `buildChangeHistoryQuery` sends `skip: pageIndex * pageSize` to exactly that endpoint, so page 2 returns page 1. Nothing in CI catches it — `src/assets/config.json` is gitignored, so unit tests, e2e and every `pr-*` preview build fall back to `session_type: mock`, where `find()` throws anyway.

So the paging has to move to a bookmark cursor regardless, which is what `PaginatedDataSource` already implements.

Two things the generic path turns out to give us for free:

- `EntitiesTableComponent.recordsDataSource` is `input.required<EntitiesTableDataSource<T>>()`, and three components already hand it a hand-built data source — so this screen can construct `PaginatedDataSource` directly and never touch `resolveDataSource()` or its `session_type` short-circuit.
- `RemotePouchDatabase.find()` builds the `_id` prefix range and, when `sort.prop` is set, creates the sort index on demand with a `partial_filter_selector` over that range. That is exactly what `AUDIT_TIMESTAMP_INDEX` does by hand, so the index bootstrap goes away too.

Also worth noting: `pouchdb-find` **is** installed (`PouchDB.plugin(pouchdbFind)` runs at module load). `PouchDatabase.find()` throws only because PouchDB's local Mango engine has no bookmark cursor — the comment on this branch saying the plugin is missing is wrong and gets removed.

### The two hooks suggested in review

**`Entity.DATABASE` alone isn't enough.** `EntityMapperService` does honour it, but `DatabaseResolverService.registerDatabase()` calls `createDatabase()` and `changes().subscribe()` and never calls `init()` — only `app` and `notifications` are initialised anywhere. Every read awaits `firstValueFrom(this.databaseInitialized, …)` on a Subject that only `init()` completes, so the first `findType(AuditRecord)` never settles: an infinite spinner with a clean console, in every session type including `online`.

If it were initialised, synced mode is worse — the class is chosen purely from `environment.session_type`, so it builds a `SyncedPouchDatabase` and replicates an unbounded audit log into IndexedDB.

And that `changes()` subscription, on a remote handle, is a 10-second poll of `_changes?include_docs=true` forever, feeding the global changes feed — whose only consumer resets every open paginated list to page 1. The change log would jump back to page 1 on every audited write anywhere in the system.

Hence PR 1: route remote-only names through the existing `createRemoteDatabase()` (which calls `init()` itself), skip the `changes()` subscription, and skip the name in `resetDatabases` / `destroyDatabases` / `resetSync` — that last one because `destroy()` sends a real `DELETE` against the audit database, harmless today only because the backend defines no `DELETE /:db` route.

**`EntitySpecialLoaderService` can't carry this.** `loadData(loaderMethod): Promise<E[]>` takes no filter and no page argument and returns all documents (and only the `ChildrenService` branch is implemented), and `resolveDataSource()` forces `InMemoryDataSource` whenever a `loaderMethod` is set — "special loaders are not supported by the paginated data source". Routing audit records there would load the whole audit database into memory.

### Why `AuditRecord` declares no `entityId` field

The audit document's `entityId` key collides with `Entity`'s `private get/set entityId`. `@DatabaseField() entityId` creates no own property — the decorator's `target[prop] = undefined` fires the inherited setter, polluting `AuditRecord.prototype._id`, and reads return the audit id's tail instead of `Child:123`. `EntitySchemaField` has no db-name alias, so it can't be worked around on the frontend.

So the changed record is a plain non-schema getter over the audit doc's own `_id`, parsed with an anchored regex stripping `:<ISO timestamp>:<rev>` — which round-trips correctly for an entity id that itself contains colons. The record-type filter is a prefix range on the raw field in the selector, `{entityId: {$gte: "Child:", $lt: "Child:￰"}}`, which is legal in a Mango selector even though it can't be a schema field.

This is also what keeps "no data migration required" true: nothing new gets stored, and every pre-existing audit document is readable as-is. A flat `entityType` field would have silently excluded all of them from a type-filtered query.

### What stays hand-built, and why

The five filter controls, targeting `EntitiesTableComponent` rather than `EntityListComponent`. Two concrete blockers:

- `FilterGeneratorService` sources non-schema options from the in-memory `data` array, which `PaginatedDataSource` deliberately leaves empty — so the "changed by" dropdown would come back with zero options and then be silently dropped.
- `DateFilter.getFilter()` emits `YYYY-MM-DD` bounds, so `$lte: "2026-08-01"` sorts below every `2026-08-01T…` instant and drops the whole last day of a selected range. That's a latent bug on the DB path for any list, not specific to this screen, so I'd rather fix it in its own issue than inside this PR. Keeping `endOfDay()` here for now.

The query builder still collapses from 247 lines to a small pure `buildAuditFilter()`.

One thing to pin explicitly: `sortBy: {active: "timestamp", direction: "desc"}`. Without it the list orders by `_id`, which is grouped by record rather than newest-first — plausible-looking and wrong.

### The related-record filter comes out

Deleting `audit-reference-index.ts` and its spec, `queryChangesRelatedTo`, `ensureReferenceIndex`, `buildReferenceViewQuery`, the related-record input, and the filter mutual-exclusion behaviour it forced.

This is how I'd like to resolve both threads on `change-history.service.ts` — by removing the path rather than patching it. The short-page probe is genuinely unsound there: `queryView` does `result.rows = result.rows.filter(...)` with no refill loop, and views have no bookmark, so there's no cursor to continue from either. It can't be made sound client-side.

Worth separating from the other thread, though: on the `_find` path the refill loop *does* return each page filled to exactly the requested limit, so "a short page means the end" is sound there — which is the rule `PaginatedDataSource` uses.

Two follow-up issues:

- *replication-backend:* a cursor-paginated, always-permission-filtered view endpoint. The `offset` suggestion in the thread is the right direction — views expose `offset` and `total_rows`, and with `startkey_docid` that's how the cursor gets reconstructed. `reduce` / `group` would have to be rejected outright, since a reduce row has no doc to evaluate.
- *ndb-core:* the filter itself — and probably not by reviving the view. A flat `referencedIds: string[]` written server-side (the same syntactic id extraction the map function does client-side) turns it into `{referencedIds: {$elemMatch: {$eq: id}}}` on the ordinary `_find` path: one query, composable with the other filters, and the mutual-exclusion behaviour becomes unnecessary. The issue should note the clause isn't index-served, and that new-field-only writes exclude pre-backfill documents.

### No effect on the already-merged per-record dialog

Its data path doesn't change. This branch removes exactly one line from `change-history.service.ts` — an unused import — and `getHistory()` is untouched. Three things kept that way deliberately, each with a spec:

- `getHistory` does not go through `findType`. `loadType` hardcodes the prefix to `ENTITY_TYPE + ":"`, so it would return the whole audit database instead of one entity's slice — and `RemotePouchDatabase.find()` spreads its type range *after* `...query`, overwriting any caller-supplied `_id` selector, so the `AuditRecord:<entityId>:` sub-prefix can't be passed either.
- No schema-transformed entities reach `buildChangeEvents`. `transformDatabaseToEntityFormat` drops every key absent from the schema, and the replay needs the raw delta.
- `buildChangeEvents` stays exactly as merged. The schema-transform argument holds for the list row — `changedFieldsOf` reads one document — but not for the dialog, which is a chronological fold *across* documents: `state` is seeded from the create/baseline snapshot and each successive delta is patched onto it. No per-document datatype can produce that.

One optional tidy-up: move `getAuditDb()` off its private `createRemoteDatabase()` handle onto `dbResolver.getDatabase(AuditRecord.DATABASE)`. Same class either way, and it removes the duplicated database ownership raised in review. Happy to skip it and leave the merged feature bit-identical.

Two changes to the existing feature that are already on this branch and worth flagging: the i18n rename "Change History" → "Change Log" needs retranslation, and the `entity-block` change is app-wide — any entity whose type configures no display value now renders `"<Type> Record"` plus a small id instead of a raw uuid, everywhere, not just here.

### Permissions

`AuditRecord` gets granted read by name to `admin_app` in `basic/Config_Permissions.json`, mirroring the by-name `NotificationEvent` grant in `_default`.

One correction to this PR's own description while I'm here: on a base-config instance the screen works through the `manage` / `all` wildcard, so it isn't relying on the "no rules defined" fallback. Where it genuinely fails is any instance whose permission config *enumerates* subjects — the backend is fail-closed there, so the screen renders a permanently empty table, and those are exactly the instances most likely to want an audit trail. Only the `basic` base config ships a permissions document at all.

Keeping both route gates: dropping the role gate would expose the system-wide log to every `user_app` user on a base-config instance, and dropping the CASL gate would hide that empty-table failure mode.

Granularity stays all-or-nothing across record types. Per-record scoping isn't possible on the current document shape — real rules condition on the audited entity's own fields, and an update's audit record carries only a delta (a delete's is a content-free tombstone). Per-*type* narrowing becomes possible with CASL conditions once an `entityType` field exists.

Side benefit: registering `AuditRecord` as a real entity type makes it selectable in the admin role-permission matrix, which it isn't today.

### One thing to check against a real backend first

`RemotePouchDatabase.find()` creates a single-field **ascending** partial index and pins `use_index`. Whether CouchDB serves `sort: [{timestamp: "desc"}]` from it when the selector constrains nothing on `timestamp` isn't something any test here can answer, and the hand-built index on this branch suggests it came up once already.

If it refuses, the fix is small: merge `{[sort.prop]: {$gt: null}}` into the selector, or rely on the descending index the backend PR pre-creates. I'll verify against an instance with `AUDIT_ENABLED=true` before writing commit 4.

### Decisions I'd like settled first

1. **Does the audit database get routed through `Entity.DATABASE`** (PR 1), or stay on the feature's own handle? Routing it makes `AuditRecord.DATABASE` mean something and avoids any feature-local data-source subclass. If we'd rather not touch the login path right now, the alternative is extracting one protected `fetchPage()` seam from `PaginatedDataSource` and overriding it in a small subclass — which leaves `AuditRecord.DATABASE` as a value nothing resolves, i.e. a trap for whoever next calls `loadType(AuditRecord)`.

TheSlimvReal:
The point is correct, that currently any loader method using the `EntitySpecialLoader` does not support pagination. However, this is not a limitation, rather something that just hasn't been implemented yet because it wasn't needed. I already had some ideas how the pagination could be done here and as far as I could tell from this PR, it would be beneficial here. I am happy to help or even implement the backend part to make this possible.

Abhinegi2:
@coderabbit review

src/app/features/change-history/change-history-list/change-history-list.component.html
```diff
@@ -0,0 +1,84 @@
...
+        <mat-label i18n="Change log filter label">Related record</mat-label>
+        <input
+          matInput
+          #relatedRecordInput
+          type="text"
+          placeholder="e.g. User:1"
+          i18n-placeholder="Change log filter placeholder"
+          [value]="relatedEntityFilter() ?? ''"
+          (change)="setRelatedEntityFilter(relatedRecordInput.value)"
+        />
```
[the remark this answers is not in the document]
Abhinegi2: Applied the class to the `<mat-form-field>`. The rule was defined but matched nothing, so the width never took effect and long `Type:uuid` ids were cut off. Fixed in 4a27a215f.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

e2e/tests/entity.spec.ts
```diff
@@ -684,3 +684,41 @@ test("Closing an untouched new-entity dialog does not prompt discard-changes, bu
...
+
+  await page.getByRole("navigation").getByText("Children").click();
+  await page.getByRole("cell", { name: HISTORY_CHILD_NAME }).click();
+
+  // the entry point is offered to every user, not only those who may read the
+  // audit data - the dialog itself explains why there is nothing to show
+  await page
+    .locator("app-entity-actions-menu button[mat-icon-button]")
+    .first()
+    .click();
```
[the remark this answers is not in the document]
Abhinegi2: Switched to `getByRole`. The button had no accessible name at all, so the label had to come first — added `aria-label="More actions"` to the menu trigger. That was a real accessibility gap on every entity details view, not only a brittle locator. Fixed in 4a27a215f.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

src/app/core/common-components/entities-table/data-source/data-source-resolver.ts
```diff
@@ -51,8 +56,9 @@ function getDataSource(
     return availableDataSources[dataSource];
   }
 
-  if (loaderMethod) {
-    // special loaders are not supported by the paginated data source
+  if (loaderMethod && !loaderSupportsPagination(loaderMethod)) {
+    // a loader with no notion of a page loads everything it has, so the list
+    // pages it in memory
```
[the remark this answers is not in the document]
Abhinegi2: Not reachable as described: the change log builds its own `PaginatedDataSource` rather than going through `resolveDataSource`, and it is the only user of `AuditRecordsRelatedToEntity`. Routing that loader to `PaginatedDataSource` regardless of session type would not be right either, since that data source cannot query a local database.

The trap underneath was real though — `loadDataFor` returned nothing for a method it does not handle, which reaches the caller as a missing promise and fails far from the cause. It now throws naming the method. Fixed in 4a27a215f.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #4283, opened 2026-08-18
feat(i18n): implement multi-lingual configuration support

### LINKED ISSUE #3862: i18n: multi-lingual config (configure labels in multiple languages)
_As an admin I want to configure my system in more than one language in order to give my team members the app in their different local languages._

Make the platform truly multilingual, so that within one system different users can use different languages at the same time.

**Context**
We already support translation of all static texts of the app, this includes buttons and fixed features. However, the majority of the UI is actually custom configuration by admin users (e.g. entity type names, field labels, etc.). At the moment, all these texts can only be configured once, so there is no option to show field labels in a different language to different users of the same system.

#### Displaying multi-lingual config:
- extend the config format so that as an alternative to simple `string`, a map `{ "language-code": "text", "en": "Welcome" }` is stored.
   - keep supporting simply one string to avoid migrations or unnecessarily complex config
- In the ConfigService, when loading the config doc, dynamically parse this mutli-lingual config.
   - To all other code, only expose the existing simple format, where the texts are the ones for the current selected language
- some additional entity types need the same kind of parsing of mulit-lingual config. Probably we should implement a service that handles this and can be reused:
   - `PublicFormConfig`
   - `ConfigurableEnum`
   - maybe `TemplateExport`
   - `EmailTemplate`
   - `"Config:SetupWizard"`

#### Configuring multi-lingual config
- any config form field that hold translatable text displays a small [icon](https://fontawesome.com/icons/language?f=classic&s=solid) that opens a dialog to configure all translations for that text (similar to configuring enum options)
   - the form field itself can still be used as a simple text field to configure the text for the currently active language
   - the components store a multilingual map instead of simple string, if active
- (?) as an additional option, an Admin view lists all translatable texts in a global table, where columns for each language give a quick way to review and update translations without having to open an endless number of field dialogs

Image

---
## Implementation Steps
1. Start with manually defining a multi-lingual config file, only implement the Admin UI later, after the format has been tested
2. Figure out a way to detect which config parts can/are multi-lingual (need to be parsed) and which are static (keys, IDs but also if the config is only using one lanuage)
3. ...

sadaf895:
- closes #3862

## PR Checklist

---
## Manual Testing

Test A — the dialog works

1. Left nav → Children
2. Click the ⋮ at top right → Configure Data Structure
3. Click Details View & Fields
4. Hover the Name field → click the Edit Field button
5. Next to the Label input you should now see a language icon. Click it
6. The "Configure translations" dialog opens with one row per language. English should be pre-filled with Name
7. Type Vorname into the German row → click Save
8. Back in the field dialog, the Label input should still show Name — not [object Object]
9. Click Apply, then Save at the bottom → you should see "Configuration updated"

Test B — the translation actually persisted

1. Re-open the same field (hover Name → Edit Field) → click the language icon again
2. German should still read Vorname, English Name

Test C — the important one: other languages survive an unrelated edit

1. With the field dialog open, change the Label to Full Name 
2.. Also toggle something unrelated — e.g. type anything in Description
3.. Apply → Save
4. Re-open the field → language icon

Expected: English = Full Name (your edit), German = Vorname still intact.

## Summary by CodeRabbit

* **New Features**
  * Added multilingual editing for entity labels, field labels, descriptions, group headers, tabs, and menu items.
  * Added a translations dialog for entering and managing values across supported languages.
  * Added locale-aware display with fallback to the default language.
  * Preserved translations in other languages when updating configuration values.
  * Added localized English, German, and French content to the education configuration.

* **Tests**
  * Expanded unit and end-to-end coverage for multilingual configuration and localization flows.

### DIFF
--- a/src/app/core/admin/admin-entity-details/admin-entity-field/admin-entity-field.component.html
+++ b/src/app/core/admin/admin-entity-details/admin-entity-field/admin-entity-field.component.html
@@ -29,6 +29,16 @@
             <mat-form-field>
               <mat-label i18n>Label</mat-label>
               <input formControlName="label" matInput #formLabel />
+              <button
+                mat-icon-button
+                matSuffix
+                type="button"
+                (click)="openTranslations('label', $event)"
+                matTooltip="Configure this text in other languages"
+                i18n-matTooltip
+              >
+                <fa-icon icon="language"></fa-icon>
+              </button>
               @if (schemaFieldsForm.get("label").hasError("uniqueProperty")) {
                 <mat-error>
                   {{ schemaFieldsForm.get("label").getError("uniqueProperty") }}
@@ -59,6 +69,16 @@
                 matInput
                 [placeholder]="formLabel.value"
               />
+              <button
+                mat-icon-button
+                matSuffix
+                type="button"
+                (click)="openTranslations('labelShort', $event)"
+                matTooltip="Configure this text in other languages"
+                i18n-matTooltip
+              >
+                <fa-icon icon="language"></fa-icon>
+              </button>
             </mat-form-field>
 
             <mat-form-field>
@@ -75,6 +95,16 @@
                 matInput
                 rows="3"
               ></textarea>
+              <button
+                mat-icon-button
+                matSuffix
+                type="button"
+                (click)="openTranslations('description', $event)"
+                matTooltip="Configure this text in other languages"
+                i18n-matTooltip
+              >
+                <fa-icon icon="language"></fa-icon>
+              </button>
             </mat-form-field>
           </div>
 
--- a/src/app/core/admin/admin-entity-details/admin-entity-field/admin-entity-field.component.html
+++ b/src/app/core/admin/admin-entity-details/admin-entity-field/admin-entity-field.component.html
@@ -1,6 +1,4 @@
-<h2 mat-dialog-title i18n>
-  Configure Field "{{ resolvedLabel }}"
-</h2>
+<h2 mat-dialog-title i18n>Configure Field "{{ resolvedLabel }}"</h2>
 <app-dialog-close mat-dialog-close></app-dialog-close>
 
 <mat-dialog-content>
--- a/src/app/core/admin/admin-entity-details/admin-entity-field/admin-entity-field.component.html
+++ b/src/app/core/admin/admin-entity-details/admin-entity-field/admin-entity-field.component.html
@@ -1,5 +1,5 @@
 <h2 mat-dialog-title i18n>
-  Configure Field "{{ data.entitySchemaField.label }}"
+  Configure Field "{{ resolvedLabel }}"
 </h2>
 <app-dialog-close mat-dialog-close></app-dialog-close>
 
@@ -28,17 +28,9 @@
           <div class="entity-form-cell">
             <mat-form-field>
               <mat-label i18n>Label</mat-label>
-              <input formControlName="label" matInput #formLabel />
-              <button
-                mat-icon-button
-                matSuffix
-                type="button"
-                (click)="openTranslations('label', $event)"
-                matTooltip="Configure this text in other languages"
-                i18n-matTooltip
-              >
-                <fa-icon icon="language"></fa-icon>
-              </button>
+              <app-translatable-text-input
+                formControlName="label"
+              ></app-translatable-text-input>
               @if (schemaFieldsForm.get("label").hasError("uniqueProperty")) {
                 <mat-error>
                   {{ schemaFieldsForm.get("label").getError("uniqueProperty") }}
@@ -64,21 +56,10 @@
                   i18n-matTooltip
                 ></fa-icon>
               </mat-label>
-              <input
+              <app-translatable-text-input
                 formControlName="labelShort"
-                matInput
-                [placeholder]="formLabel.value"
-              />
-              <button
-                mat-icon-button
-                matSuffix
-                type="button"
-                (click)="openTranslations('labelShort', $event)"
-                matTooltip="Configure this text in other languages"
-                i18n-matTooltip
-              >
-                <fa-icon icon="language"></fa-icon>
-              </button>
+                [placeholder]="resolvedLabel"
+              ></app-translatable-text-input>
             </mat-form-field>
 
             <mat-form-field>
@@ -90,21 +71,11 @@
                   i18n-matTooltip
                 ></fa-icon>
               </mat-label>
-              <textarea
+              <app-translatable-text-input
                 formControlName="description"
-                matInput
-                rows="3"
-              ></textarea>
-              <button
-                mat-icon-button
-                matSuffix
-                type="button"
-                (click)="openTranslations('description', $event)"
-                matTooltip="Configure this text in other languages"
-                i18n-matTooltip
-              >
-                <fa-icon icon="language"></fa-icon>
-              </button>
+                [multiline]="true"
+                [rows]="3"
+              ></app-translatable-text-input>
             </mat-form-field>
           </div>
 
--- a/src/app/core/admin/admin-entity.service.ts
+++ b/src/app/core/admin/admin-entity.service.ts
@@ -1,4 +1,4 @@
-import { EventEmitter, inject, Injectable } from "@angular/core";
+import { EventEmitter, inject, Injectable, LOCALE_ID } from "@angular/core";
 import { EntityConstructor } from "../entity/model/entity";
 import { Config } from "../config/config";
 import { EntityConfig } from "../entity/entity-config";
@@ -8,6 +8,9 @@ import { EntityListConfig } from "../entity-list/EntityListConfig";
 import { EntityDetailsConfig } from "../entity-details/EntityDetailsConfig";
 import { DynamicComponentConfig } from "../config/dynamic-components/dynamic-component-config.interface";
 import { NoteDetailsConfig } from "#src/app/child-dev-project/notes/note-details/note-details-config.interface";
+import { mergeTranslatableValues } from "../config/multi-lingual-config";
+import { DEFAULT_LANGUAGE } from "../language/language-statics";
+import { availableLocales } from "../language/languages";
 
 /**
  * Simply service to centralize updates between various admin components in the form builder.
@@ -18,6 +21,9 @@ import { NoteDetailsConfig } from "#src/app/child-dev-project/notes/note-details
 export class AdminEntityService {
   public entitySchemaUpdated = new EventEmitter<void>();
   private entityMapper = inject(EntityMapperService);
+
+  private readonly locale = inject(LOCALE_ID);
+  private readonly validLocaleIds = availableLocales.values.map((v) => v.id);
 
   /**
    * Set a new schema field to the given entity and trigger update event for related admin components.
@@ -100,13 +106,38 @@ export class AdminEntityService {
       if (field.isInternalField) {
         continue;
       }
-      entitySchemaConfig.attributes[fieldId] = field;
+      // the runtime schema holds values resolved to the active language, so merge
+      // onto the raw config to keep translations of other languages (#3862)
+      entitySchemaConfig.attributes[fieldId] = this.mergeTranslations(
+        entitySchemaConfig.attributes[fieldId],
+        field,
+      );
     }
 
     // Add additional general settings if available
     if (configEntitySettings) {
-      Object.assign(entitySchemaConfig, configEntitySettings);
+      Object.assign(
+        entitySchemaConfig,
+        this.mergeTranslations(entitySchemaConfig, configEntitySettings),
+      );
     }
+  }
+
+  /**
+   * Merge an edited (language-resolved) config part onto its raw counterpart from
+   * the config document, so that translations for other languages are preserved.
+   */
+  private mergeTranslations<T extends Record<string, any>>(
+    raw: Record<string, any> | undefined,
+    edited: T,
+  ): T {
+    return mergeTranslatableValues(
+      raw,
+      edited,
+      this.locale,
+      DEFAULT_LANGUAGE,
+      this.validLocaleIds,
+    );
   }
 
   private getEntitySchemaFromConfig(
--- a/src/app/core/admin/admin-entity/admin-entity-general-settings/admin-entity-general-settings.component.html
+++ b/src/app/core/admin/admin-entity/admin-entity-general-settings/admin-entity-general-settings.component.html
@@ -12,7 +12,9 @@
         <div class="entity-form-cell">
           <mat-form-field>
             <mat-label i18n>Label</mat-label>
-            <input formControlName="label" matInput #formLabel />
+            <app-translatable-text-input
+              formControlName="label"
+            ></app-translatable-text-input>
           </mat-form-field>
 
           <mat-form-field floatLabel="always">
@@ -25,11 +27,10 @@
                 i18n-matTooltip
               ></fa-icon>
             </mat-label>
-            <input
+            <app-translatable-text-input
               formControlName="labelPlural"
-              matInput
-              [placeholder]="formLabel.value"
-            />
+              [placeholder]="resolvedLabel"
+            ></app-translatable-text-input>
           </mat-form-field>
 
           <app-icon-input [control]="iconControl"></app-icon-input>
--- a/src/app/core/admin/building-blocks/admin-tabs/admin-tabs.component.html
+++ b/src/app/core/admin/building-blocks/admin-tabs/admin-tabs.component.html
@@ -22,6 +22,7 @@
             ></fa-icon>
             @if (tabIndex === matTabGroup.selectedIndex) {
               <app-admin-section-header
+                [translatable]="translatableTitle()"
                 [(title)]="tab[tabTitleProperty()]"
                 (remove)="removeTab(tabIndex)"
               ></app-admin-section-header>
--- a/src/app/core/admin/building-blocks/admin-tabs/admin-tabs.component.html
+++ b/src/app/core/admin/building-blocks/admin-tabs/admin-tabs.component.html
@@ -28,7 +28,7 @@
               ></app-admin-section-header>
             } @else {
               <!-- only current tab can be renamed -->
-              {{ tab[tabTitleProperty()] }}
+              {{ tabTitle(tab) }}
             }
           </div>
         </div>
--- a/src/app/core/config/config.service.ts
+++ b/src/app/core/config/config.service.ts
@@ -1,4 +1,4 @@
-import { Injectable } from "@angular/core";
+import { inject, Injectable, LOCALE_ID } from "@angular/core";
 import { HttpStatusCode } from "@angular/common/http";
 import { shareReplay } from "rxjs/operators";
 import { addDefaultNoteDetailsConfig } from "../../child-dev-project/notes/add-default-note-views";
@@ -9,6 +9,14 @@ import { Logging } from "../logging/logging.service";
 import { Config } from "./config";
 import { ConfigMigration } from "./config-migration";
 import { applyConfigMigrations } from "./config-migrations";
+import {
+  CONFIG_ENTITY_ROUTE_PREFIX,
+  normalizeRoutePath,
+} from "./dynamic-routing/route-paths";
+import { PREFIX_VIEW_CONFIG } from "./dynamic-routing/view-config.interface";
+import { resolveTranslatableConfig } from "./multi-lingual-config";
+import { DEFAULT_LANGUAGE } from "../language/language-statics";
+import { availableLocales } from "../language/languages";
 
 /**
  * Access dynamic app configuration retrieved from the database
@@ -28,6 +36,17 @@ export class ConfigService extends LatestEntityLoader<Config> {
    * Subscribe to receive the current config and get notified whenever the config is updated.
    */
   private currentConfig: Config;
+
+  /**
+   * The current config with all multi-lingual values resolved to the active
+   * locale (see #3862). This is the display view returned by
+   * getConfig()/getAllConfigs(); currentConfig stays raw (translation maps
+   * intact) as the persisted source of truth.
+   */
+  private resolvedConfigData: Record<string, any>;
+
+  private readonly locale = inject(LOCALE_ID);
+  private readonly validLocaleIds = availableLocales.values.map((v) => v.id);
 
   configUpdates = this.entityUpdated.pipe(shareReplay(1));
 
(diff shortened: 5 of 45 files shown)


src/assets/base-configs/all-features/Config_CONFIG_ENTITY.json
```diff
@@ -1319,7 +1319,10 @@
         },
         "projectNumber": {
           "dataType": "string",
```
sleidig, line 1321: our src/assets/base-configs/basic configs have a /de and /fr version that otherwise should have the same structures.

Can you generate a single, multilingual config from these (and remove the other language versions?

src/app/core/admin/admin-entity-details/admin-entity-field/admin-entity-field.component.html
```diff
@@ -59,6 +69,16 @@ <h2 mat-dialog-title i18n>
                 matInput
                 [placeholder]="formLabel.value"
               />
+              <button
```
sleidig: maybe we can create a directive or an input component that includes this button and wiring (click to open, update value on save, etc.), so that we don't have to repeat it here (and in other places)

maybe such a component can even take a formControl that holds the full object with all translations and handles the display of the current value in the textbox internally?

sleidig (changes requested):
Nice, @sadaf895 ! This is gonna be an awesome feature 😍 

The core stuff looks pretty good to me. Thanks for sharing this version.

To finalize and get read for full test, I think we still need a few related functionalities. Can you look into that?
- [x] add UI for "normal" users to switch the language (only for their account). Maybe place that in the user profile screen?
- [x] also: the navigation menu (and other places, like the Entity Labels) can't be translated yet (at least in the Admin UI).

src/assets/base-configs/all-features/Config_CONFIG_ENTITY.json
sadaf895: Not done yet, because it mostly depends on Phase 2.  Of the 12 /de + /fr pairs, only education/*/Config_CONFIG_ENTITY.json can become a single multilingual file today — that's the one that goes through ConfigService. The other 11 pairs are separate DatabaseEntities that don't resolve translations yet: ConfigurableEnum, PublicFormConfig, EmailTemplate, ReportConfig, MarkdownContent, ProgressDashboardConfig, and Config:SetupWizard (which is its own Config document loaded via entityMapper.load, not part of CONFIG_ENTITY). Merging those now would render [object Object].

Also worth flagging: removing the education_de / education_fr entries from available-configs.json also removes them from the welcome-screen scenario picker, and breaks the i18n e2e test that selects "Bildungsprojekt". That's arguably the right end state — language comes from the language selector rather than the scenario — but it's a UX change worth deciding rather than sliding in.

My suggestion will be do the whole config merge in the Phase 2 PR once the remaining entity types resolve, so it lands in one consistent step. What is your call about this? @sleidig

sadaf895:
> Nice, @sadaf895 ! This is gonna be an awesome feature 😍
> 
> The core stuff looks pretty good to me. Thanks for sharing this version.
> 
> To finalize and get read for full test, I think we still need a few related functionalities. Can you look into that?
> 
> * [ ]  add UI for "normal" users to switch the language (only for their account). Maybe place that in the user profile screen?
> * [x]  also: the navigation menu (and other places, like the Entity Labels) can't be translated yet (at least in the Admin UI).

Regarding UI for normal users not done yet I'd suggest doing this as a follow-up PR, since it doesn't touch the config-translation code at all and because this PR is already large. Though most of it already exists: ProfileComponent is there, LanguageSelectComponent is there (currently only on the setup screen), and SiteSettings.displayLanguageSelect exists as an admin toggle that isn't used anywhere yet.
The missing piece is persistence: switchLocale() only writes to localStorage, so today the choice is per-browser, not per-account.
One question: where should the preference live? I'd suggest a per-user settings entity keyed by the user id, the same pattern as NotificationConfig in that screen — works for any account, unlike a field on the user's entity record. Fine to keep it simpler if you prefer. Also, right now LanguageService switches everyone's locale when an admin changes defaultLanguage, which would override a user's own choice — I'd make it apply only to users who haven't picked one. What are your thoughts @sleidig ?

src/assets/base-configs/all-features/Config_CONFIG_ENTITY.json
sleidig: ah, that's true. Forgot about the enums etc. :sweat_smile: 
And we will have to adapt the welcome screen a bit (I had that in the back of my mind).

Maybe you can already add the translations to the Englisch main config file, though (while keeping the others, which don't usually change)? Then this makes testing a bit easier

sleidig:
> Regarding UI for normal users not done yet I'd suggest doing this as a follow-up PR, since it doesn't touch the config-translation code at all and because this PR is already large. Though most of it already exists: ProfileComponent is there, LanguageSelectComponent is there (currently only on the setup screen), and SiteSettings.displayLanguageSelect exists as an admin toggle that isn't used anywhere yet. The missing piece is persistence: switchLocale() only writes to localStorage, so today the choice is per-browser, not per-account. One question: where should the preference live? I'd suggest a per-user settings entity keyed by the user id, the same pattern as NotificationConfig in that screen — works for any account, unlike a field on the user's entity record. Fine to keep it simpler if you prefer. Also, right now LanguageService switches everyone's locale when an admin changes defaultLanguage, which would override a user's own choice — I'd make it apply only to users who haven't picked one. 

@sadaf895 , I like the approach of a per-user Settings entity (and the global setting as fallback for those who haven't selected anything individually)!
I think it could be `SiteSettings:<user-account-id>` (although the logo and colors are probably something individuals shouldn't be able to overwrite). That would make loading and applying it easy to understand the connection.

Do you want to go ahead with this on a new PR that is based upon this one? Then we can review separately (thanks for that :blush: ) and decide in the end if we merge as two or one combined commit.

src/assets/base-configs/all-features/Config_CONFIG_ENTITY.json
sadaf895: Done add the de and fr into english config.  Deliberately not converted:
- columnGroups.groups[].name and the columnGroups.default/.mobile values referencing it - these are identifiers matched by value, and need a stable id plus a migration first
- entity:Child.toBlockDetailsAttributes.title - holds a field name ("name"), not display text
- dashboard widget periodLabel - no confirmed merge-on-save protection, so a map there could be flattened by an admin edit.

sadaf895:
@coderabbitai review

src/app/core/config/translatable-text-input/translatable-text-input.component.ts
```diff
@@ -0,0 +1,121 @@
...
+    isTranslatableText(this.valueSignal(), this.validLocaleIds),
+  );
+
+  /**
+   * Apply text typed by the user to the active language only,
+   * keeping any other languages of a multi-lingual value untouched.
+   */
+  onTextInput(text: string) {
+    const current = this.valueSignal();
+    this.value = isTranslatableText(current, this.validLocaleIds)
```
Abhinegi2, line 98: On a value that is already a per-language map, deleting the last character of the active language's text does not clear the field: `onTextInput("")` stores `{ ..., [locale]: "" }`, and `displayText` resolves that through the fallback chain, which skips empty strings. The field immediately refills with another language's text, so the admin cannot clear it.

Preferring an already-present slot over the resolved fallback in `displayText`, even when that slot is empty, would fix it.

src/app/core/config/configure-translations-popup/configure-translations-popup.component.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  /**
+   * Build the new raw value. Only the default language on its own collapses to a
+   * plain string: {@link initialTextFor} reads one back as that language's text.
+   */
+  private buildValue(): TranslatableText | undefined {
+    const filled = this.rows.filter((row) => !!row.text?.trim());
+
+    if (filled.length === 0) {
+      return undefined;
```
Abhinegi2, line 93: `buildValue()` returns `undefined` when every language row is emptied, and the caller treats `undefined` as "dialog cancelled" (`translatable-text-input.component.ts` line 113) and keeps the old value. So clearing all translations and pressing Save is a silent no-op.

Closing with a distinct value for "cleared", or distinguishing cancel by the dialog's close reason rather than by the payload, would separate the two cases.

src/app/core/admin/admin-entity-details/admin-entity-field/admin-entity-field.component.html
```diff
@@ -1,6 +1,4 @@
-<h2 mat-dialog-title i18n>
-  Configure Field "{{ data.entitySchemaField.label }}"
-</h2>
+<h2 mat-dialog-title i18n>Configure Field "{{ resolvedLabel }}"</h2>
```
Abhinegi2, line 1: `resolvedLabel` is a getter read from the template, so it re-runs `resolveTranslatableText` on every change-detection cycle (and again on line 59). A `computed` over `toSignal(schemaFieldsForm.get("label").valueChanges)` would resolve only when the label actually changes.

src/app/core/admin/admin-entity-details/admin-entity-field/admin-entity-field.component.html
```diff
@@ -54,11 +54,10 @@ <h2 mat-dialog-title i18n>
                   i18n-matTooltip
                 ></fa-icon>
               </mat-label>
-              <input
+              <app-translatable-text-input
                 formControlName="labelShort"
-                matInput
-                [placeholder]="formLabel.value"
-              />
+                [placeholder]="resolvedLabel"
```
Abhinegi2, line 59: Second read of the same getter - same fix as above.

src/app/core/admin/admin-entity/admin-entity-general-settings/admin-entity-general-settings.component.html
```diff
@@ -25,11 +27,10 @@ <h2 i18n>General Settings of "{{ entityConstructor().label }}" Records</h2>
                 i18n-matTooltip
               ></fa-icon>
             </mat-label>
-            <input
+            <app-translatable-text-input
               formControlName="labelPlural"
-              matInput
-              [placeholder]="formLabel.value"
-            />
+              [placeholder]="resolvedLabel"
```
Abhinegi2, line 32: Same as in the field dialog: `resolvedLabel` is a getter read from a binding, so it re-runs on every change-detection cycle. A `computed` over the form control's `valueChanges` avoids that.

src/app/core/admin/building-blocks/admin-tabs/admin-tabs.component.html
```diff
@@ -22,12 +22,13 @@
...
             @if (tabIndex === matTabGroup.selectedIndex) {
               <app-admin-section-header
+                [translatable]="translatableTitle()"
                 [(title)]="tab[tabTitleProperty()]"
                 (remove)="removeTab(tabIndex)"
               ></app-admin-section-header>
             } @else {
               <!-- only current tab can be renamed -->
-              {{ tab[tabTitleProperty()] }}
+              {{ tabTitle(tab) }}
```
Abhinegi2, line 31: `tabTitle(tab)` is a method call in a binding, so it re-runs on every change-detection cycle for every non-selected tab. `tabs` is a `model()`, so a `computed` mapping tabs to their resolved titles would avoid the repeated resolve.

Abhinegi2 (commented):
Review notes on the blocking items only. Two of them land in files this PR does not touch, so they could not be attached to a diff line:

**`src/app/core/dashboard/admin-dashboard/admin-dashboard.component.ts` line 71** - the dashboard editor reads its config through `configService.getConfig()`, which this PR changes to return language-resolved values, and `save()` writes that same object back via `exportConfig(true)` + `saveConfig()`. Just dragging a widget to reorder it and pressing Save replaces every widget's translation map with the editing admin's single language (shortcut labels, progress-widget titles, attendance period labels). The `getRawConfig()` fix currently sits in the follow-up PR, so master regresses for as long as this one is merged on its own.

**`src/app/features/matching-entities/admin-matching-entities/admin-matching-entities.component.ts` line 57** - same pattern: `getConfig("appConfig:matching-entities")` read, then written back through `exportConfig(true)` + `saveConfig()`. Saving that admin screen flattens `matchActionLabel`, any column `label`, and any `availableFilters` label that was configured per language.

Both want `getRawConfig()`, with resolution only for display.

src/app/core/admin/building-blocks/admin-tabs/admin-tabs.component.html
sadaf895: A computed over tabs() doesn't work here — the selected tab's title is written back by mutating the tab object ([(title)]="tab[tabTitleProperty()]"), so tabs.set/update is never called and the array reference never changes. The computed then serves a stale title: rename the selected tab, switch away, and its label keeps the old text. I have 
confirmed with your exact suggestion. I've added a unit test to record the constraint.

src/app/core/admin/admin-entity.service.ts
```diff
@@ -100,15 +106,40 @@ export class AdminEntityService {
       if (field.isInternalField) {
         continue;
       }
-      entitySchemaConfig.attributes[fieldId] = field;
+      // the runtime schema holds values resolved to the active language, so merge
+      // onto the raw config to keep translations of other languages (#3862)
```
Abhinegi2: Drop the issue number — comments here shouldn't reference ticket numbers.

```suggestion
      // onto the raw config to keep translations of other languages
```

src/app/core/config/config.service.ts
```diff
@@ -29,6 +32,17 @@ export class ConfigService extends LatestEntityLoader<Config> {
    */
   private currentConfig: Config;
 
+  /**
+   * The current config with all multi-lingual values resolved to the active
+   * locale (see #3862). This is the display view returned by
```
Abhinegi2: Drop the issue number — comments here shouldn't reference ticket numbers.

```suggestion
   * locale. This is the display view returned by
```

src/app/core/config/config.service.ts
```diff
@@ -138,25 +158,61 @@ export class ConfigService extends LatestEntityLoader<Config> {
     return rawObject ? JSON.parse(value) : value;
   }
 
+  /**
+   * Get a config item, with any multi-lingual values resolved to the active locale.
+   *
+   * Use getRawConfig() instead if you will edit the value and save it back to the
+   * config document, otherwise translations for other languages would be lost (#3862).
```
Abhinegi2: Drop the issue number — comments here shouldn't reference ticket numbers.

```suggestion
   * config document, otherwise translations for other languages would be lost.
```

src/app/core/config/config.service.ts
```diff
@@ -138,25 +158,61 @@ export class ConfigService extends LatestEntityLoader<Config> {
...
+   */
   public getConfig<T>(id: string): T | undefined {
+    return this.resolvedConfigData?.[id];
+  }
+
+  /**
+   * Get a config item with its raw values (multi-lingual maps left intact).
+   * Required for any read-modify-save flow (see #3862).
```
Abhinegi2: Drop the issue number — comments here shouldn't reference ticket numbers.

```suggestion
   * Required for any read-modify-save flow.
```

src/app/core/config/config.service.ts
```diff
@@ -138,25 +158,61 @@ export class ConfigService extends LatestEntityLoader<Config> {
...
-    if (!this.currentConfig?.data) {
+    return ConfigService.collectConfigsByPrefix<T>(
+      this.resolvedConfigData,
+      prefix,
+    );
+  }
+
+  /**
+   * Like getAllConfigs(), but with raw values (multi-lingual maps left intact).
+   * Required for any read-modify-save flow (see #3862).
```
Abhinegi2: Drop the issue number — comments here shouldn't reference ticket numbers.

```suggestion
   * Required for any read-modify-save flow.
```

src/app/core/config/multi-lingual-config.ts
```diff
@@ -0,0 +1,218 @@
...
+] as const;
+
+/**
+ * Merge an edited, locale-resolved config object back onto its raw counterpart,
+ * preserving the translations of all other languages.
+ *
+ * Admin UIs read config through the *resolved* view (plain strings) and write the
+ * whole object back. Without this merge, saving would replace a translation map
+ * with the single string of the editing admin's language and silently drop every
+ * other language (#3862).
```
Abhinegi2: Drop the issue number — comments here shouldn't reference ticket numbers.

```suggestion
 * other language.
```

src/app/core/entity/entity-config.service.ts
```diff
@@ -205,6 +205,18 @@ export class EntityConfigService {
     return this.configService.getConfig<EntityConfig>(configName);
   }
 
+  /**
+   * Like {@link getEntityConfig}, but with raw values: any multi-lingual texts
+   * are returned as their full per-language map rather than resolved to the
+   * active language. Required when editing config values that are saved back
+   * to the config document (see #3862).
```
Abhinegi2: Drop the issue number — comments here shouldn't reference ticket numbers.

```suggestion
   * to the config document.
```

src/assets/base-configs/available-configs.json
```diff
@@ -66,20 +66,20 @@
...
-      "education/fr/ProgressDashboardConfig_examples.json",
-      "education/fr/Config_CONFIG_ENTITY.json",
-      "education/fr/configurable-enums.json",
-      "basic/fr/ConfigurableEnum_attendance-status.json",
-      "basic/fr/ConfigurableEnum_warning-levels.json",
-      "basic/fr/ConfigurableEnum_genders.json",
-      "basic/fr/ConfigurableEnum_rating-answer.json",
-      "education/fr/Config_SetupWizard.json",
-      "basic/fr/MarkdownContent_help.json",
+      "education/ProgressDashboardConfig_examples.json",
```
sleidig, line 69: can we make these also as multilingual config strings?

In general, the initial system config selection UI doesn't yet work very well with the multilingual config files: If I select "German", I can't see the configs that include "de" translations, unless that config is separately listed as a German config here ...

Image


maybe `locale: "fr"` here can become an array `locale: ["en", "de", "fr"]` and the `name` and `description` run as multilingual fields?

src/app/features/reporting/reporting/reporting.component.ts
```diff
@@ -94,9 +95,19 @@ export class ReportingComponent {
...
-        .then((res) => res.sort((a, b) => a.title?.localeCompare(b.title))),
+      this.entityMapper.loadType(ReportEntity).then((res) =>
+        res
+          // resolve once here, so everything downstream works on plain strings.
+          .map((report) => {
+            const forDisplay = report.copy();
+            forDisplay.title = resolveActiveText(report.title);
+            forDisplay.description = resolveActiveText(report.description);
+            return forDisplay;
+          })
```
sleidig: shouldn't this be done as a transformation by `TranslatableTextDatatype`  in general?

src/app/core/language/active-locale.ts
```diff
@@ -0,0 +1,52 @@
...
+  if (locale) {
+    activeLocale = locale;
+  }
+  if (locales?.length) {
+    validLocaleIds = locales;
+  }
+}
+
+/** Values that are not translation maps are returned unchanged. */
+export function resolveActiveText(
```
sleidig: maybe `resolveLocaleText` or `resolveActiveLocaleText`? That makes the purpose clearer when reading its use somewhere that is not obviously related to translation

src/assets/base-configs/available-configs.json
sadaf895: Agreed. And one thing that the baseConfig is a public interface. Per setup/README.md, partners host their own available-configs.json at a baseUrl. So I kept locale as a union rather than replacing it, name and description are backward-compatible for free since resolveLocaleText returns a plain string unchanged. And existing partner descriptors keep working.

src/app/features/reporting/reporting/reporting.component.ts
sadaf895: Deliberately not as the datatype's transforms are identity because they're symmetric. If it resolved on load, the entity would hold just the active language's string, and the next entityMapper.save() would persist that and drop the others. Raw stays persisted, resolving happens only for display.
The datatype does resolve where there's no display component in the path , sortValue() and getExportColumns() — plus viewComponent = "DisplayTranslatableText" for normal rendering. Reporting skips all that: the title goes into a mat-select and gets sorted with localeCompare. It resolves onto a report.copy() and this component only loads, never saves, so nothing resolved can get written back. I also dropped the description line code  that field is plain LongTextDatatype, so resolving it did nothing.

sleidig: ah, I remember this challenge from some previous discussion. Sorry, forgot about that here.

Still, there might be an approach that keeps responsibility for this in the datatype, maybe?

Do you think the following implementation could work?

### keep a helper object of the full translate
1. database -> object:
  - transform into the plain string for current locale
  - AND store to original multi-lingual object somewhere temporarily
2. object -> database:
  - look up the temporary object (if it exists) and save that instead of the plain string
3. editing the full multi-lingual object could then access that "temporary" copy instead of the plain property. But it would limit the "knowledge" about multi-lingual data to these special places

The transform methods have access to the whole entity (`parent`) and could write a `_label` for a `label` field on it, for example:
https://github.com/Aam-Digital/ndb-core/blob/d870520029f564497bd7146c8d92d357d8da17c0/src/app/core/entity/default-datatype/default.datatype.ts#L182-L187

I like the idea of this because it could allow us to continue implementing components without ever thinking about the multilingual feature at all. And I think editing multilingual is probably more rare than reading it? Or would this cause even more code changes than the current approach?

sadaf895: Actually mechanically it works ,both transforms get parent, and transformEntityToDatabaseFormat only iterates schema keys, so a _title shadow wouldn't leak into the document.
Two catches are there:
1. On load, parent is the raw DB doc, not the entity, and only schema keys are copied onto it so the shadow needs a change in entity-schema.service to land anywhere.
2. A stale shadow silently discards edits: if anything sets entity.title = "New", _title still holds the old map and save writes that. Fixable by comparing against the resolved value before merging basically what mergeTranslatableValues() already does.

The issue is that it only 5 entity fields use the datatype. Most multilingual text lives in the Config data blob, enum options and entity statics — none are schema fields, so a datatype can't see them. We'd end up with two mechanisms for a subset. So I am thinking that let's keep explicit resolution here, and do the datatype shadow as a follow-up.

And you're right that reading is far more common than editing that's exactly why the current split works: reads go through resolveLocaleText, and only the admin edit screens touch the raw value.

PR #4284, opened 2026-08-18
fix: allow configuring which origins may embed the app in an iframe

### LINKED ISSUE #4286: embedded demo on www.aam-digital.com doesn't load anymore
**Describe the bug**

The demo system that is embedded via iframe on our website no longer loads. The browser blocks the frame:

> Refused to display 'https://demo.aam-digital.com/' in a frame because it set 'X-Frame-Options' to 'sameorigin'

**To Reproduce**

1. Open the page on www.aam-digital.com that embeds the demo system
2. The iframe stays blank; the browser console shows the error above

**Expected behavior**

The demo loads inside the iframe as before.

**Root cause**

This was not a deliberate change. Until #4094 the nginx template contained:

```nginx
add_header X-Frame-Options: SAMEORIGIN;
```

With the colon, nginx emitted a header whose *field name* was `X-Frame-Options:` — not a valid header token, so browsers never matched it as X-Frame-Options and framing worked by accident. #4094 (released in 3.86.0) corrected the syntax, the header became effective for the first time, and the embed broke when the demo instance was updated past that version.

`X-Frame-Options` cannot express "allow this one other origin" — `ALLOW-FROM` never shipped in Chrome/Safari and was removed from Firefox — so allowing the website to embed the demo requires CSP's `frame-ancestors`. Note that our existing CSP is sent **report-only**, and `frame-ancestors` is ignored in a report-only policy, so it has to be sent as a separate enforcing header.

**Fix**

- #4284 — replaces `X-Frame-Options` with an enforcing `Content-Security-Policy: frame-ancestors 'self' ${CSP_EXTRA_FRAME_ANCESTORS}`. The empty default is equivalent to the previous `SAMEORIGIN` behaviour, so no existing instance changes.
- Aam-Digital/ndb-setup#115 — passes the new variable through to the app container.

Remaining after both merge (manual steps, not covered by either PR):

- [ ] deploy an app image containing #4284 to the demo instance
- [ ] set `CSP_EXTRA_FRAME_ANCESTORS` in that instance's `.env` to the website origin(s) — the apex and `www` hosts are different origins and both need listing if both serve the embedding page
- [ ] verify with the browser cache disabled: the Angular service worker can replay a cached `index.html` together with its old headers, which looks exactly like the fix having failed

**Additional context**

This affects any instance that is embedded in another site. Only the demo system is known to be affected.

sleidig:
## Problem

Embedding an instance in an iframe on another site is currently impossible, and it broke silently rather than by decision.

Until #4094 the header was written as `add_header X-Frame-Options: SAMEORIGIN;` — with a colon, so nginx emitted a header whose *field name* was `X-Frame-Options:`. That is not a valid header token, so browsers never matched it as X-Frame-Options and framing worked by accident. #4094 (released in 3.86.0) corrected the syntax, the header became effective for the first time, and existing iframe embeds started failing with:

> Refused to display '…' in a frame because it set 'X-Frame-Options' to 'sameorigin'

`X-Frame-Options` has no way to express "allow this one other origin" — `ALLOW-FROM` never shipped in Chrome/Safari and was removed from Firefox — so this needs CSP's `frame-ancestors`.

## Change

- Replace `X-Frame-Options SAMEORIGIN` with `Content-Security-Policy: frame-ancestors 'self' ${CSP_EXTRA_FRAME_ANCESTORS}`.
- New docker env variable `CSP_EXTRA_FRAME_ANCESTORS`, **empty by default**, so every existing deployment keeps exactly today's behaviour without knowing about it.
- Document it in the security concepts doc.

### Why a separate, enforcing CSP header

`frame-ancestors` is ignored in a `Content-Security-Policy-Report-Only` policy, so it cannot simply be appended to the existing `$CSP` whitelist. Multiple CSP headers are enforced independently, and this one contains no other directive — it therefore restricts nothing except framing, and the existing whitelist stays report-only as before.

### Why `X-Frame-Options` is dropped rather than kept alongside

Where both are present, browsers must ignore XFO in favour of `frame-ancestors`, so keeping it would mean shipping two headers that deliberately contradict each other on any instance that allows embedding. Its only remaining function is for browsers predating CSP2 — which cannot run this app anyway. Happy to keep it as a one-line addition if reviewers prefer the extra header for security scanners.

## Verification

Built the image from this branch and ran it both ways:

**Default (variable unset)** — equivalent to the previous SAMEORIGIN behaviour:
```
Content-Security-Policy: frame-ancestors 'self'
Content-Security-Policy-Report-Only: default-src 'self' … (unchanged)
X-Content-Type-Options: nosniff
```

**With `CSP_EXTRA_FRAME_ANCESTORS="https://example.com https://www.example.com"`:**
```
Content-Security-Policy: frame-ancestors 'self' https://example.com https://www.example.com
```

nginx starts cleanly in both cases (no errors in the container log), and `$CSP` still substitutes correctly after adding the new name to the `envsubst` allowlist.

## Sequencing

Aam-Digital/ndb-setup#115 passes the variable through to the app container. It is inert until an image containing this change is deployed, and no instance becomes embeddable until its `.env` actually sets the variable.

---

Relates to #4286

## Summary by CodeRabbit

* **New Features**
  * Added configurable iframe embedding support for trusted origins.
  * Same-origin framing remains enabled by default, with additional origins configurable at deployment time.

* **Documentation**
  * Documented iframe embedding controls, supported origin formats, and configuration requirements.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -20,10 +20,15 @@ ENV CSP_REPORT_URI="https://o167951.ingest.sentry.io/api/1242399/security/"
 ENV CSP="default-src 'self' 'unsafe-eval' 'unsafe-inline' data: blob: https://*.tile.openstreetmap.org/ https://matomo.aam-digital.org https://fcmregistrations.googleapis.com/v1/projects/* https://www.gstatic.com/firebasejs/* https://*.aam-digital.com https://*.aam-digital.net https://*.aam-digital.app https://api.github.com/repos/Aam-Digital/ https://aam-digital.github.io https://sentry.io https://raw.githubusercontent.com $CSP_REPORT_URI; style-src 'self' 'unsafe-inline'"
 # 'unsafe-eval' required for pouchdb https://github.com/pouchdb/pouchdb/issues/7853#issuecomment-535020600
 
+# additional origins that are allowed to embed the app in an iframe (space-separated),
+# e.g. "https://example.com https://www.example.com" to embed the app in another website.
+# The app's own origin is always allowed; empty (default) means only that origin.
+ENV CSP_EXTRA_FRAME_ANCESTORS=""
+
 # variables are inserted into the nginx config
 # RESOLVER is all IPv4 nameservers from /etc/resolv.conf, joined by spaces, used so nginx can
 # re-resolve the Nominatim upstream at runtime with IPv6 disabled (see default.conf).
 CMD export RESOLVER=$(awk '$$1=="nameserver" && $$2 !~ /:/ {print $$2}' /etc/resolv.conf | xargs) &&\
     if [ -z "$RESOLVER" ]; then export RESOLVER=8.8.8.8; fi &&\
-    envsubst '$$PORT $$COUCHDB_URL $$QUERY_URL $$NOMINATIM_URL $$RESOLVER $$CSP $$CSP_REPORT_URI' < /etc/nginx/templates/default.conf > /etc/nginx/conf.d/default.conf &&\
+    envsubst '$$PORT $$COUCHDB_URL $$QUERY_URL $$NOMINATIM_URL $$RESOLVER $$CSP $$CSP_REPORT_URI $$CSP_EXTRA_FRAME_ANCESTORS' < /etc/nginx/templates/default.conf > /etc/nginx/conf.d/default.conf &&\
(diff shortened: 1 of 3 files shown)


sleidig:
@coderabbitai review

sleidig:
Tested successfully :heavy_check_mark:

PR #4290, opened 2026-08-19
docs(security): high-level security concept, CSP details to the build docs

sleidig:
- closes https://github.com/Aam-Digital/internal/issues/352

Part of moving the security concept developed in https://github.com/Aam-Digital/internal/issues/352 into the repos. Companion to https://github.com/Aam-Digital/aam-cloud-infrastructure/pull/221, which covers the cluster/operational side.

This repo's page covers **a single Aam Digital system**, conceptually — not the multi-instance platform, and not our internal threat model (which stays internal; there are deliberately no links to it from here).

## `doc/compodoc_sources/concepts/security.md` — rewritten

- **Browser-side**: Angular sanitization, the two CSP headers (whitelist report-only, framing enforced), and that the offline shell means fixes arrive on the next online session.
- **Authentication**: Keycloak owns password/MFA/session policy; **realm membership is what grants access to a system's data**.
- **Roles and permissions in the two deployment modes**: a comparison table plus short prose, condensed per review feedback. `database-only` explicitly does not support user roles or individual permissions — any valid login has full read/write access to all data, and access without a valid account is not possible at all (so public forms cannot work). A `Config:Permissions` doc is technically still possible to add there, but it is only partly enforced by the local device's UI and **must not be used as a security functionality**. The permission-backend section states the converse: there, a narrowed role actually restricts what can be fetched, not only what the interface offers.
- **Offline login checks no credential.** `LocalAuthService` only lists users from a previous online login and `offlineLogin()` opens the local database directly — no secret is stored or verified. Stated in both the authentication and the device section, because it is what makes the unencrypted local copy reachable by whoever holds the device.
- **The device copy**: unencrypted, bounded by what the user may sync, cleared only when a sync reports lost permissions; plus `session_type: online` and `session_type_choice`.
- **An explicit, unhedged out-of-scope section**: server and network security, encryption at rest, backups, operating Keycloak, and administrative server access — pointing at your IT department or an IT service provider, and naming a hosted (SaaS) system as usually the more reliable option. This matters because self-hosters will read this page as their guidance whether or not it was written for them.

## The other files

- **`build/README.md`** — the CSP details move here, next to the `Dockerfile` and `default.conf` that set them: the `CSP` whitelist and its report-only status, PouchDB's `'unsafe-eval'`, the `index.html` hash procedure, `CSP_EXTRA_FRAME_ANCESTORS`, and `X-Content-Type-Options`. Nothing dropped in the move; the report-only/enforced distinction from #4284 is now stated in both places. The file stays out of the published documentation — the concept page links to it on GitHub.
  - Also gained a `## Configuration` table documenting `PORT`, `COUCHDB_URL`, `QUERY_URL` and `NOMINATIM_URL`, carried over from #4297 (now closed as superseded by this PR). The CSP-related variables that table also listed stay documented in more detail further down in the existing CSP section — the table just points there instead of duplicating it.
- **`build/Dockerfile`** — the comment above the CSP variables pointed at the concept page; it now points at `build/README.md`.

## Note

Aam-Digital/internal#353 (the client-facing data security statement) is about the **hosted SaaS offer**, not about this open source product, and will live with the client-facing material rather than in a repo — see https://github.com/Aam-Digital/internal/issues/353#issuecomment-5346711952. This page must not be read as a claim about what we operate for SaaS clients.

### DIFF
--- a/doc/compodoc_sources/concepts/security.md
+++ b/doc/compodoc_sources/concepts/security.md
@@ -30,7 +30,7 @@ The directives, their defaults and how to change them are documented with the im
 Users authenticate against a **Keycloak** server (OpenID Connect). Aam Digital never sees or stores a password: the app receives a token and passes it on to the database.
 Password policy, multi-factor authentication, session lifetime, account lockout and email verification are therefore Keycloak configuration, not application code.
 
-Each system has its own Keycloak realm, and **membership of that realm is what grants access to that system's data**. Removing a person's access means disabling or removing their account in Keycloak — changing their roles inside the application is not sufficient on its own, for the reason described next.
+Each system has its own Keycloak realm, and **membership of that realm is what grants access to that system's data**. In a database-only deployment, removing a person's access means disabling or removing their account in Keycloak — changing their roles inside the application is not sufficient on its own; see "Database-only, and what it does not give you" below.
 
 Once a user has logged in online at least once, the app offers them an **offline login** on the same device. This is a profile selection, not an authentication step: no credential is checked, because none is stored and the authentication server cannot be reached. It opens the local copy of the data that is already on that device, and grants nothing beyond it — but it does mean that whoever holds the device can open that copy. See "The copy of data on each user's device" below.
 
@@ -52,24 +52,24 @@ Where those rules are actually _enforced_ depends on how the system is deployed,
 
 #### Database-only, and what it does not give you
 
-CouchDB is exposed directly and validates the user's token itself. Access is granted to a **single role that every user of the system holds**, so the boundary the server enforces is all-or-nothing: a user is either a member of this system's database, or not.
+This light-weight deployment does explicitly not support user roles and individual permissions.
+All users with a valid login have full read/write access to all data.
 
-The permission rules still run, but only in the browser. There they decide what the interface offers and block writes before they are sent — which is genuinely useful, because it keeps users out of areas that do not concern them and prevents accidental edits. It is a usability and safety layer. **It is not an access boundary**, and nothing about it survives a user who talks to the database API directly instead of using the app. Concretely, in this mode any user who can log in can:
+Access to the database without a valid user account is not possible.
+This also means _**Public Forms**_ cannot work for external data collection.
 
-- **read the entire database**, including entity types and records their role hides in the interface — and, since it is one database, that includes the configuration and the permission rules themselves;
-- **write or delete any document**, including those same permission rules, because the rule check that would have stopped it lives in the browser they are bypassing;
-- **keep reading after their roles change**, because a narrowed role restricts the interface but not what their token can fetch. Withdrawing access means disabling the account in Keycloak.
+> Adding a `Config:Permissions` doc (in preparation of introducing user roles) is technically still possible.
+> Those permissions are partly enforced by the local device's user interface
+> but this **does not** enforce actual data protection on the server.
+> DO NOT USE this as a security functionality.
 
-Two further consequences follow from the same design:
-
-- **There is no audit log.** Nothing records who changed what and when.
-- **Public forms cannot work.** They rely on the `_public` rules being applied to visitors without an account, and the database accepts only tokens carrying the role above — so an anonymous request never reaches it.
+Because CouchDB checks only realm membership and nothing about a user's role, narrowing someone's role never restricts what their existing token can fetch — withdrawing access means disabling their account in Keycloak, not just changing their role.
 
 This mode is therefore appropriate when everyone with an account in the system is trusted with all of its data — a small team working on one caseload — and not when the roles are meant to keep colleagues apart.
 
 #### With the permission backend
 
-The same rules are additionally applied server-side: reads are filtered as they are replicated, and writes are checked again before they are stored, with CouchDB itself no longer reachable from outside. This is what turns a role restriction into a real restriction, and it is the only configuration in which "this user may only see the records of their own project" is a statement about access rather than about the user interface. It is also the only one that can record who changed what.
+The same rules are additionally applied server-side: reads are filtered as they are replicated, and writes are checked again before they are stored, with CouchDB itself no longer reachable from outside. Unlike database-only mode, a narrowed role then restricts what the user can actually fetch, not only what the interface offers. This is what turns a role restriction into a real restriction, and it is the only configuration in which "this user may only see the records of their own project" is a statement about access rather than about the user interface. It is also the only one that can record who changed what.
 
 **So if different users of one system must not see each other's data, the permission backend is required.**
 Deploying either mode is described in [Aam-Digital/ndb-setup](https://github.com/Aam-Digital/ndb-setup#docker-compose-profiles).
@@ -79,7 +79,7 @@ Deploying either mode is described in [Aam-Digital/ndb-setup](https://github.com
 To work offline, the app stores a copy of the records a user may access in the browser's storage on that user's device. This is what makes Aam Digital usable without connectivity, and it has consequences worth stating plainly:
 
 - The copy is stored **unencrypted**. On a device without full-disk encryption, whoever holds the device can read it.
-- Its extent is whatever the user is allowed to sync — which, in database-only mode, is the whole database.
+- Its extent is whatever the user is allowed to sync.
 - Revoking permissions server-side does not reach a device that never connects again. Local data is cleared when the server reports lost permissions during a sync, and that requires the device to come online.
 - The offline login described above opens that copy without checking any credential, so the device itself is the only thing standing between a stranger and the data.
 
--- a/doc/compodoc_sources/concepts/security.md
+++ b/doc/compodoc_sources/concepts/security.md
@@ -63,7 +63,7 @@ The permission rules still run, but only in the browser. There they decide what 
 Two further consequences follow from the same design:
 
 - **There is no audit log.** Nothing records who changed what and when.
-- **Rules for anonymous visitors (`_public`) have no effect**, since the database only accepts tokens carrying the role above.
+- **Public forms cannot work.** They rely on the `_public` rules being applied to visitors without an account, and the database accepts only tokens carrying the role above — so an anonymous request never reaches it.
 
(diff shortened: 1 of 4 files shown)


doc/compodoc_sources/concepts/security.md
```diff
@@ -1,41 +1,124 @@
...
-Generating the hash by pasting the script into an online generator does not seem to work, probably because code is minified during the build process.
+Each system has its own Keycloak realm, and **membership of that realm is what grants access to that system's data**. Removing a person's access means disabling or removing their account in Keycloak — changing their roles inside the application is not sufficient on its own, for the reason described next.
+
+Once a user has logged in online at least once, the app offers them an **offline login** on the same device. This is a profile selection, not an authentication step: no credential is checked, because none is stored and the authentication server cannot be reached. It opens the local copy of the data that is already on that device, and grants nothing beyond it — but it does mean that whoever holds the device can open that copy. See "The copy of data on each user's device" below.
```
sleidig: Removing all roles should also force that account to purge all local data, so it should essentially work the same way.

```suggestion
Each system has its own Keycloak realm, and **membership of that realm is what grants access to that system's data**.

Once a user has logged in online at least once, the app offers them an **offline login** on the same device. This is a profile selection, not an authentication step: no credential is checked, because none is stored and the authentication server cannot be reached. It opens the local copy of the data that is already on that device, and grants nothing beyond it — but it does mean that whoever holds the device can open that copy. See "The copy of data on each user's device" below.
```

doc/compodoc_sources/concepts/security.md
```diff
@@ -1,41 +1,124 @@
...
+
+- **read the entire database**, including entity types and records their role hides in the interface — and, since it is one database, that includes the configuration and the permission rules themselves;
+- **write or delete any document**, including those same permission rules, because the rule check that would have stopped it lives in the browser they are bypassing;
+- **keep reading after their roles change**, because a narrowed role restricts the interface but not what their token can fetch. Withdrawing access means disabling the account in Keycloak.
+
+Two further consequences follow from the same design:
+
+- **There is no audit log.** Nothing records who changed what and when.
+- **Public forms cannot work.** They rely on the `_public` rules being applied to visitors without an account, and the database accepts only tokens carrying the role above — so an anonymous request never reaches it.
```
sleidig:
```suggestion
This light-weight deployment does explicitly not support user roles and individual permissions.
All users with a valid login have full read/write access to all data.

Access to the database without a valid user account is not possible.
This also means _**Public Forms**_ cannot work for external data collection.

> Adding a `Config:Permissions` doc (in preparation of introducing user roles) is technically still possible.
> Those permissions are partly enforced by the local device's user interface
> but this **does not** enforce actual data protection on the server.
> DO NOT USE this as a security functionality.
```

doc/compodoc_sources/concepts/security.md
```diff
@@ -1,41 +1,124 @@
...
+
+This mode is therefore appropriate when everyone with an account in the system is trusted with all of its data — a small team working on one caseload — and not when the roles are meant to keep colleagues apart.
+
+#### With the permission backend
+
+The same rules are additionally applied server-side: reads are filtered as they are replicated, and writes are checked again before they are stored, with CouchDB itself no longer reachable from outside. This is what turns a role restriction into a real restriction, and it is the only configuration in which "this user may only see the records of their own project" is a statement about access rather than about the user interface. It is also the only one that can record who changed what.
+
+**So if different users of one system must not see each other's data, the permission backend is required.**
+Deploying either mode is described in [Aam-Digital/ndb-setup](https://github.com/Aam-Digital/ndb-setup#docker-compose-profiles).
```
sleidig: rewrite this, after applying the shortened version for the section above, some of that content needs to move here, I think.

doc/compodoc_sources/concepts/security.md
```diff
@@ -1,41 +1,124 @@
...
+
+## The copy of data on each user's device
+
+To work offline, the app stores a copy of the records a user may access in the browser's storage on that user's device. This is what makes Aam Digital usable without connectivity, and it has consequences worth stating plainly:
+
+- The copy is stored **unencrypted**. On a device without full-disk encryption, whoever holds the device can read it.
+- Its extent is whatever the user is allowed to sync — which, in database-only mode, is the whole database.
```
sleidig:
```suggestion
- Its extent is whatever the user is allowed to sync.
```

simplify

doc/compodoc_sources/concepts/security.md
sleidig: Applied as suggested — replaced the CouchDB/token mechanism paragraph and the three read/write/audit-log bullets (already covered by the comparison table above) with this wording. The role-narrowing point ("keep reading after their roles change") wasn't dropped though — see the reply on the next thread for where it went.

doc/compodoc_sources/concepts/security.md
sleidig: Moved the role-narrowing point here: this section now says a narrowed role actually restricts what the user can fetch, not just what the interface offers — the converse of database-only. Deliberately didn't assert *when* that takes effect (e.g. per-request vs. cached) since that's replication-backend's behavior and I didn't have that repo open to verify it; a security doc shouldn't guess at enforcement timing. Also retargeted the Authentication section's forward-reference ("for the reason described next"), since it pointed at the bullet the suggestion above removed — it now points at the database-only section by name instead.

doc/compodoc_sources/concepts/security.md
sleidig: Applied, but note the range was start_line 81 through line 82, and the suggested single-line replacement would have deleted the "The copy is stored **unencrypted**..." bullet on line 81 along with it. That seemed like an unintended casualty (the PR description leans on that bullet), so I kept line 81 as-is and only replaced line 82 with your suggested text. Flag if you actually meant to cut the unencrypted bullet too.

doc/compodoc_sources/concepts/security.md
```diff
@@ -1,41 +1,124 @@
...
-We have made both technical and design choices to develop a secure platform.
-The Angular framework itself has some in-built protection against common security issues like cross-site scripting attacks (see [Angular > Security](https://angular.io/guide/security)).
-Beyond this, the following measures are implemented:
+A conceptual overview of how a single Aam Digital system protects the data it holds:
+what the application does, what it leaves to the deployment, and what it does not do at all.
 
-- If deployed including our ["replication-backend"](https://github.com/Aam-Digital/replication-backend), that server-side API also ensures authenticated users can only access and sync data their account has permissions for.
-- Password policy enforces users to set a strong password including special characters (either via Keycloak or the platforms user profile form)
-- Content Security Policy (CSP) headers restrict connections to and execution of code from sources that are not whitelisted.
+This is about **one deployed system** — the app, its database and the accounts that reach it.
```
sleidig:
```suggestion
This is about **one deployed system** (the app, its database and the accounts that reach it).
```

doc/compodoc_sources/concepts/security.md
```diff
@@ -1,41 +1,124 @@
...
+| ------------------ | --------------------------------------- | --------------------------------------------------------------------------------------------------------- |
+| What is exposed    | CouchDB directly                        | [replication-backend](https://github.com/Aam-Digital/replication-backend) in front of an internal CouchDB |
+| Who checks a token | CouchDB, against the realm's public key | the backend, against the realm's public key                                                               |
+| Read access        | the whole database of that system       | filtered per user, rule by rule                                                                           |
+| Write access       | any document                            | validated against the same rules                                                                          |
+| Audit log          | none                                    | optional (`AUDIT_ENABLED`): every write recorded with user and time                                       |
+
+#### Database-only, and what it does not give you
```
sleidig:
```suggestion
#### Database-only (no user roles)
```

doc/compodoc_sources/concepts/security.md
```diff
@@ -1,41 +1,124 @@
...
+> DO NOT USE this as a security functionality.
+
+Because CouchDB checks only realm membership and nothing about a user's role, narrowing someone's role never restricts what their existing token can fetch — withdrawing access means disabling their account in Keycloak, not just changing their role.
+
+This mode is therefore appropriate when everyone with an account in the system is trusted with all of its data — a small team working on one caseload — and not when the roles are meant to keep colleagues apart.
+
+#### With the permission backend
+
+The same rules are additionally applied server-side: reads are filtered as they are replicated, and writes are checked again before they are stored, with CouchDB itself no longer reachable from outside. Unlike database-only mode, a narrowed role then restricts what the user can actually fetch, not only what the interface offers. This is what turns a role restriction into a real restriction, and it is the only configuration in which "this user may only see the records of their own project" is a statement about access rather than about the user interface. It is also the only one that can record who changed what.
```
sleidig:
```suggestion
#### With permission backend

Permissions of user roles are enforced server-side (in addition to local UI checks on the device):
reads are filtered while data is synced or requested from the server,
and writes are checked again before they are stored, with the database itself no longer reachable from outside.
Unlike database-only mode, a narrowed role then restricts what the user can actually fetch, not only what the interface offers.

The permission backend also includes functionality to record a detailed change history of who changed what.
```

doc/compodoc_sources/concepts/security.md
```diff
@@ -1,41 +1,124 @@
...
+Deploying either mode is described in [Aam-Digital/ndb-setup](https://github.com/Aam-Digital/ndb-setup#docker-compose-profiles).
+
+## The copy of data on each user's device
+
+To work offline, the app stores a copy of the records a user may access in the browser's storage on that user's device. This is what makes Aam Digital usable without connectivity, and it has consequences worth stating plainly:
+
+- The copy is stored **unencrypted**. On a device without full-disk encryption, whoever holds the device can read it.
+- Its extent is whatever the user is allowed to sync.
+- Revoking permissions server-side does not reach a device that never connects again. Local data is cleared when the server reports lost permissions during a sync, and that requires the device to come online.
+- The offline login described above opens that copy without checking any credential, so the device itself is the only thing standing between a stranger and the data.
```
sleidig:
```suggestion
- The copy includes whatever that user is allowed to sync.
- Revoking permissions server-side does not reach a device that never connects again. Local data is cleared when the server reports lost permissions during a sync, and that requires the device to come online.
```

PR #4296, opened 2026-08-21
feat(i18n): implement user-specific language settings and preferences

sadaf895:
- closes https://github.com/Aam-Digital/ndb-core/issues/3862

## PR Checklist

## Manual Testing

1. The selector appears:

   Profile (bottom of the sidebar) → User Account tab

- Below your account details there's a "Choose your language" dropdown, with the note "Choose the language for your own account. Other users are not affected."

2. It applies:

- Pick Deutsch / German → the app reloads into German
- Go to Children → the first column reads "Projektnummer" (translated config label)

3.  The admin override guard — the real new behaviour:

- With your language set to German, go to Admin → Admin Overview → Site Settings
- Set Default Language to Français / French → Save
- You should stay in German, and the app should not reload

Previously that admin change switched everyone's language immediately. Now it only applies to users who haven't chosen their own.

## Summary by CodeRabbit

- **New Features**
  - Users can choose and save a personal language from their profile settings.
  - Personal language preferences take priority over the site’s default language.
  - Administrators can control whether the language selector appears.
  - Language changes are saved before being applied, with an error notification if saving fails.
  - Language selectors can display contextual hints and support deferred application.

- **Bug Fixes**
  - Site language updates no longer override a user’s saved language preference.

### DIFF
--- a/src/app/core/language/language.service.ts
+++ b/src/app/core/language/language.service.ts
@@ -5,6 +5,7 @@ import { WINDOW_TOKEN, LOCAL_STORAGE_TOKEN } from "../../utils/di-tokens";
 import { SiteSettings } from "../site-settings/site-settings";
 import { EntityMapperService } from "../entity/entity-mapper/entity-mapper.service";
 import { SiteSettingsService } from "../site-settings/site-settings.service";
+import { UserSettingsService } from "../site-settings/user-settings.service";
 import { filter } from "rxjs";
 import { UpdatedEntity } from "#src/app/core/entity/model/entity-update";
 
@@ -21,6 +22,7 @@ export class LanguageService {
   private window = inject<Window>(WINDOW_TOKEN);
   private siteSettings = inject(SiteSettingsService);
   private readonly entityMapper = inject(EntityMapperService);
+  private readonly userSettings = inject(UserSettingsService);
 
   constructor() {
     this.switchLocaleOnSiteSettingsUpdate();
@@ -39,8 +41,14 @@ export class LanguageService {
         untilDestroyed(this),
         filter((u) => u?.entity.getId(true) === SiteSettings.ENTITY_ID),
       )
-      .subscribe((updatedSettings: UpdatedEntity<SiteSettings>) => {
+      .subscribe(async (updatedSettings: UpdatedEntity<SiteSettings>) => {
         const updatedLanguage = updatedSettings?.entity?.defaultLanguage?.id;
+        if (!updatedLanguage) return;
+
+        // users who picked a language for their own account keep it
+        const ownLanguage = await this.userSettings.getLanguage();
+        if (ownLanguage) return;
+
         this.switchLocale(updatedLanguage);
       });
   }
@@ -57,16 +65,22 @@ export class LanguageService {
     this.window.location.reload();
   }
 
-  initDefaultLanguage(): void {
+  async initDefaultLanguage(): Promise<void> {
     const languageSelected = this.localStorage.getItem(
       LANGUAGE_LOCAL_STORAGE_KEY,
     );
+    if (languageSelected) return;
 
-    if (!languageSelected) {
-      this.siteSettings.defaultLanguage.subscribe(({ id }) => {
-        this.switchLocale(id);
-      });
+    const ownLanguage = await this.userSettings.getLanguage();
+    if (ownLanguage) {
+      this.switchLocale(ownLanguage);
+      return;
     }
+
+    // nothing chosen for this account: fall back to the system default
+    this.siteSettings.defaultLanguage.subscribe(({ id }) => {
+      this.switchLocale(id);
+    });
   }
 
   /**
--- a/src/app/core/site-settings/user-settings.service.ts
+++ b/src/app/core/site-settings/user-settings.service.ts
@@ -1,0 +1,75 @@
+import { inject, Injectable } from "@angular/core";
+import { ConfigurableEnumValue } from "../basic-datatypes/configurable-enum/configurable-enum.types";
+import { EntityMapperService } from "../entity/entity-mapper/entity-mapper.service";
+import { Logging } from "../logging/logging.service";
+import { SessionSubject } from "../session/auth/session-info";
+import { SiteSettings } from "./site-settings";
+
+/**
+ * Settings a user has chosen for their own account, stored as a `SiteSettings`
+ * document under their user account id (e.g. `SiteSettings:1234-abcd`).
+ *
+ * The global `SiteSettings:global` document remains the fallback for anyone who
+ * has not chosen a setting themselves.
+ *
+ * Deliberately separate from {@link SiteSettingsService}: that service also
+ * applies the site branding (title, colors, font, favicon) and caches it, and
+ * individual users must not be able to override those. Only the properties
+ * listed in {@link USER_OVERRIDABLE_SETTINGS} are personal.
+ */
+@Injectable({ providedIn: "root" })
+export class UserSettingsService {
+  /**
+   * The settings a user may choose for their own account.
+   * Everything else stays global and admin-controlled.
+   */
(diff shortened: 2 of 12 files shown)


sadaf895:
@coderabbitai review

src/app/core/language/language.service.ts
```diff
@@ -57,16 +65,22 @@ export class LanguageService {
...
     const languageSelected = this.localStorage.getItem(
       LANGUAGE_LOCAL_STORAGE_KEY,
     );
+    if (languageSelected) return;
 
-    if (!languageSelected) {
-      this.siteSettings.defaultLanguage.subscribe(({ id }) => {
-        this.switchLocale(id);
-      });
+    const ownLanguage = await this.userSettings.getLanguage();
```
Abhinegi2, line 74: `initDefaultLanguage()` is called from the `LanguageModule` constructor at bootstrap, before anyone logs in, so `userSettings.getLanguage()` always sees an empty `SessionSubject` and returns `undefined`.

So on any device where the user hasn't picked the language locally (new browser, second device, cleared storage), their saved DB preference is never applied — the site default gets written to `locale` instead, and from then on the early return above short-circuits every later run.

Could we re-resolve the language after login (when the session emits) rather than only at bootstrap? And a test that logs in *after* init would cover this.

Please double-check my reading of the bootstrap order here — I may be wrong.

src/app/core/language/language.service.ts
```diff
@@ -57,16 +65,22 @@ export class LanguageService {
     this.window.location.reload();
   }
 
-  initDefaultLanguage(): void {
+  async initDefaultLanguage(): Promise<void> {
     const languageSelected = this.localStorage.getItem(
       LANGUAGE_LOCAL_STORAGE_KEY,
     );
+    if (languageSelected) return;
```
Abhinegi2, line 72: Related to the comment below: nothing clears the `locale` localStorage key on logout, and init never re-checks per user.

So on a shared browser: user A picks German and logs out, user B logs in and gets German, and B's own saved language never applies because `languageSelected` is already set.

Suggest clearing or re-resolving the locale on session change.

src/app/core/site-settings/user-settings.service.ts
```diff
@@ -0,0 +1,67 @@
...
+    const userId = this.userId;
+    if (!userId) {
+      throw new Error("Cannot save user settings without a logged-in user");
+    }
+
+    const settings =
+      (await this.loadUserSettings(userId)) ?? new SiteSettings(userId);
+    settings.defaultLanguage = locale;
+
+    await this.entityMapper.save(settings);
```
Abhinegi2, line 65: Permissions are granted per entity type, so `SiteSettings:<userId>` and `SiteSettings:global` are the same subject for CASL.

With the shipped `Config_Permissions.json`, `_default` has SiteSettings **read only** — those users' saves will reject and they just get the error snackbar. And granting them write to make this feature work also lets them overwrite the global branding doc, which is exactly what the class comment above says must not happen.

Would a separate entity type (e.g. `UserSettings`) instead of a `SiteSettings` doc solve this more cleanly?

Not fully sure how this plays out on systems with custom permission configs — please verify.

src/app/core/site-settings/user-settings.service.ts
```diff
@@ -0,0 +1,67 @@
...
+  async setLanguage(locale: ConfigurableEnumValue): Promise<void> {
+    // read once: a session change during the load must not write this language
+    // onto whoever is logged in by then
+    const userId = this.userId;
+    if (!userId) {
+      throw new Error("Cannot save user settings without a logged-in user");
+    }
+
+    const settings =
+      (await this.loadUserSettings(userId)) ?? new SiteSettings(userId);
```
Abhinegi2, line 62: `new SiteSettings(userId)` carries all the class defaults, so the personal doc gets saved with `siteName: "Aam Digital - Demo"`, `dateFormat`, `hideSiteNameInToolbar` and `displayLanguageSelect` in it, not just the language.

Can we build the doc from only the whitelisted fields?

src/app/core/site-settings/user-settings.service.ts
```diff
@@ -0,0 +1,67 @@
...
+ * Settings a user has chosen for their own account, stored as a `SiteSettings`
+ * document under their account id, with `SiteSettings:global` as the fallback.
+ *
+ * Deliberately separate from {@link SiteSettingsService}, which also applies the
+ * site branding that individual users must not be able to override.
+ */
+@Injectable({ providedIn: "root" })
+export class UserSettingsService {
+  /** everything not listed here stays global and admin-controlled */
+  static readonly USER_OVERRIDABLE_SETTINGS = ["defaultLanguage"] as const;
```
Abhinegi2, line 18: `USER_OVERRIDABLE_SETTINGS` is declared but never read anywhere, so nothing actually enforces the "everything not listed here stays global" rule (see the save below writing the whole entity).

Either use it to filter what gets written, or drop it so it doesn't read as a guarantee that isn't there.

src/app/core/user/user-language-settings/user-language-settings.component.ts
```diff
@@ -0,0 +1,76 @@
...
+
+  private readonly languageSelect = viewChild(LanguageSelectComponent);
+
+  /**
+   * Save the language for this user and only then apply it,
+   * because applying it reloads the app.
+   */
+  async onLanguageSelected(localeId: string) {
+    const locale = this.availableLocales.find((l) => l.id === localeId);
+    if (!locale || this.saving()) {
```
Abhinegi2, line 56: The `saving()` guard silently drops a second pick while the first save is still in flight, but the dropdown has already moved to the new value.

So on a slow connection: pick German, then quickly pick French → French is ignored, and the app reloads into German while the select was showing French.

The `saving` signal isn't bound to anything in the template — could we disable the select while saving instead of dropping the event?

src/app/core/user/user-language-settings/user-language-settings.component.ts
```diff
@@ -0,0 +1,76 @@
...
+    this.saving.set(true);
+    try {
+      await this.userSettings.setLanguage(locale);
+      this.languageService.switchLocale(localeId);
+    } catch (err) {
+      Logging.error(
+        new Error("Failed to save the user's language", { cause: err }),
+      );
+      this.languageSelect()?.resetToCurrent();
+      this.snackBar.open(
```
Abhinegi2, line 69: `snackBar.open(msg)` with no duration and no action button leaves the error bar on screen indefinitely.

Elsewhere we pass a duration or a dismiss action (`alert.service.ts:75`, `admin-overview.component.ts:147`) — could we do the same here?

Abhinegi2 (commented):
Went through the diff against issue #3862. Below are the points I'd like to fix before merge, as inline comments.

Please check each one yourself — I may be wrong on some of them, especially where it depends on how a system's permission config is set up. If you disagree with any point, just reply on the comment and we can discuss.

One more that I couldn't put inline because the file isn't in this diff:

- `src/app/core/admin/admin-ai-agent/admin-ai-agent.component.ts:54` — `loadType(SiteSettings)` will now also return every user's personal settings doc, so the AI-context download grows by one doc per user and includes their account ids. Probably worth filtering to `SiteSettings.ENTITY_ID` there.

src/app/core/user/user-language-settings/user-language-settings.component.ts
sadaf895: changeLocale() sets currentLocale before emitting, so the dropdown commits before the parent decides, and saving wasn't bound to anything. Added a disabled input to app-language-select and bound it to saving(), so the pick is prevented rather than dropped. Kept the saving() guard as well: it still covers the window between setting the signal and change detection applying the binding. The failure path was already fine via resetToCurrent(), so the in-flight window was the only gap.

PR #4297, opened 2026-08-22
docs(build): document image configuration environment variables

sleidig:
## Summary
- The deployed image's `Dockerfile` and `default.conf` define several `ENV` variables (`PORT`, `COUCHDB_URL`, `QUERY_URL`, `NOMINATIM_URL`, `CSP`, `CSP_REPORT_URI`, `CSP_EXTRA_FRAME_ANCESTORS`) that operators can override to configure a deployment, but they were only documented as inline Dockerfile comments.
- Adds a "Configuration" section to `build/README.md` listing each variable, its default, and what it does, so downstream deployments (docker-compose, Helm, etc.) have a single reference instead of reading the Dockerfile.

## Test plan
- [x] `npx prettier --write build/README.md` to match repo formatting
- [ ] Docs-only change — no functional testing needed

### DIFF
--- a/build/README.md
+++ b/build/README.md
@@ -21,6 +21,29 @@ docker run -p=80:80 --name aam-digital aam/digital:latest
 docker stop aam-digital
 ```
 
+## Configuration
+
+The image is configured through environment variables, declared with their
+defaults as `ENV` values in [`Dockerfile`](./Dockerfile) and substituted into
+the nginx config at container start (see [`default.conf`](./default.conf)).
+Override any of them (e.g. via `docker run -e`, a `docker-compose.yml`
+`environment:` block, or a Helm chart's pod spec) to change that behavior;
+anything left unset keeps the default below.
+
+| Variable                    | Default                                | Purpose                                                                                                                                                                                                                                                                                                                        |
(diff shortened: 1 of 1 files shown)


sleidig:
Superseded by #4290, which touches this same file (adding the Configuration section right where #4290's own CSP documentation lives). Closing this to avoid a conflicting merge; here's exactly what carried over and what didn't:

- **Carried over**: `PORT`, `COUCHDB_URL`, `QUERY_URL`, `NOMINATIM_URL` — added to a `## Configuration` table in `build/README.md`, same as here.
- **Not carried over as separate rows**: this PR's `CSP`, `CSP_REPORT_URI` and `CSP_EXTRA_FRAME_ANCESTORS` rows. #4290 already documents those three in more detail in a dedicated CSP section further down the same file, so the merged table has one pointer row into that section instead of duplicating it.
- **Carried over from the CSP row's prose**: `CSP_REPORT_URI`'s default (aam-digital's Sentry security endpoint) — folded into the existing CSP section since this table didn't have that detail before.

Thanks for writing this up — closing in favor of the merged version.

PR #4271, opened 2026-08-15
chore(build): migrate to esbuild application builder

Abhinegi2:
Migrates the app build from the deprecated webpack-based builder to the esbuild-based application builder for much faster builds (local production build: ~2m 15s webpack vs ~30-40s esbuild, identical total bundle size).

- The `build` target now uses `@angular/build:application` instead of `@angular-devkit/build-angular:browser`. The separate `build-test` target (added for the vitest setup) is merged into it, so unit tests now build through `ndb-core:build:testing` and the duplicated config is gone.
- Output stays flat in `dist/` via `"outputPath": {"base": "dist", "browser": ""}`, so the Dockerfile, service worker regeneration, sourcemap handling and the e2e static server all keep working unchanged.
- `serve` and `extract-i18n` switch to their `@angular/build` equivalents.
- Two small workarounds for PouchDB under the new bundler:
  - a scoped npm override lifting `sublevel-pouchdb`'s pinned `readable-stream@1.x` (which requires the Node `stream` builtin that esbuild does not polyfill) to v4, which is browser-safe; other packages keep their own `readable-stream` versions.
  - a `setimmediate` polyfill for `memdown`, needed when the dev server's dependency prebundling skips its package.json "browser" field remapping.
- `allowedCommonJsDependencies` now lists the known CommonJS packages so the build stays warning-free.

Verified locally: production bundle and `ng serve` both boot demo mode, generate demo data, build indexes, render lists and export XLSX without errors. Storybook also starts and renders stories as before (its one console error exists identically on master).

- closes #2298

## PR Checklist

_Before you request a manual PR review from someone, please go through all of this list:_

- [ ] manually tested (all) required functionality yourself and added comment with clear steps for manual testing
- [ ] reviewed the "Files changed" section of the PR briefly
- [ ] added label **"Final Review"** to trigger UI regression testing (Argos visual review)
- [ ] triggered CodeRabbit AI review by commenting **"@coderabbitai review"** (e2e tests run automatically on every push)
- [ ] marked each code review comment as resolved (or commented on it with a question, if unsure)

--> then mark the PR "Ready for Review"

---

## Manual Testing Steps

1. Open the demo app, choose a use case with "Generate demo data" enabled and create the system. The dashboard should fill with sample data as usual.
2. Open the Children list, filter and sort a bit, and open one record's details.
3. In the list menu, use "Download" once as CSV and once as XLSX (Excel). Both files should download and contain the displayed records.
4. Reload the page once. The app should start up again into the demo without errors.
[...]

PR #4302, opened 2026-08-23
fix(map): restore marker clustering broken by the esbuild builder migration

### LINKED ISSUE #4301: 3.89.0 regression: marker clustering throws on every map screen — the esbuild builder migration breaks the leaflet.markercluster side-effect import
## Summary

Since `3.89.0`, every screen that renders a map throws immediately and the map never initialises.

[AAM-DIGITAL-76W](https://aam-digital.sentry.io/issues/AAM-DIGITAL-76W) — `TypeError: g.markerClusterGroup is not a function`

- **28 events**, first `2026-08-21T13:59:17Z`, last `2026-08-23T22:03:26Z`, state *escalating*
- **All 28 are on `ndb-core@3.89.0`. Zero on `3.88.0`.** `3.89.0` was tagged `2026-08-21T13:16:14Z`; the first event is 43 minutes later.
- **5 production tenants** (anonymised — real names and the mapping are in the [Slack post in #tech-monitoring-alerts](https://aam-digital.slack.com/archives/CKJFEJ0CD/p1787527589501609)): Tenant A1 ×17, Tenant A2 ×7, Tenant A3 ×2, Tenant A4 ×1, Tenant A5 ×1
- Affected routes: `/matching`, `/c/mentor/:id`, `/c/mentee/:id`, and **`/public-form/form/:id`** — the public intake form, so unauthenticated visitors hit it too.
- Throws from `ngAfterViewInit` → `MapComponent.initializeMap` at `src/app/features/location/map/map.component.ts:154`.

## Cause

`3.89.0` contains #4271 `chore(build): migrate to esbuild application builder`, which swapped `@angular-devkit/build-angular:browser` (webpack) for `@angular/build:application` (esbuild). It is the only build-affecting change in the release.

`map.component.ts` imports leaflet as a namespace and pulls the clustering plugin in for its side effects:

```ts
import * as L from "leaflet";
import "leaflet.markercluster";
```

`leaflet.markercluster` has no ESM build and imports nothing. Its UMD factory assigns onto the **global** `L`, which leaflet's own UMD sets to its live CommonJS exports object (`window.L = exports`, `leaflet/dist/leaflet-src.js`):

```js
var MarkerClusterGroup = L.MarkerClusterGroup = L.FeatureGroup.extend({ ... });
```

Under webpack, `import * as L` aliased that same exports object, so the plugin's additions were visible on `L`. esbuild instead synthesises the namespace with `__toESM(require_leaflet(), 1)`, which **copies** the exports onto a fresh object. The plugin still patches the live object (and `window.L`), but the copy the component holds never sees it.

Reproduced with the repo's own esbuild against the current lockfile:

```
namespace typeof L.markerClusterGroup: undefined     // import * as L from "leaflet"
global    typeof L.markerClusterGroup: function      // window.L
namespace === global L: false
```

## Fix

Bind `L` to the live exports object by using the default import — for a CommonJS module `__toESM` sets `default` to the raw `module.exports`:

```ts
import L from "leaflet";
import "leaflet.markercluster";
```

With that change the same bundle reports `typeof L.markerClusterGroup === "function"` and `L === window.L`, and the production bundle emits `P.default.markerClusterGroup({...})` in place of `g.markerClusterGroup({...})`.

`esModuleInterop` is already enabled in `tsconfig.json`, so the default import typechecks; `tsc --noEmit`, eslint, prettier, `ng build --configuration production` and the 11 `map.component.spec.ts` tests all pass.

Note that **a unit test cannot catch this class of bug**: vitest resolves the CommonJS interop differently from the production builder, so `L.markerClusterGroup` is defined under test either way. Any other side-effect plugin that patches a CommonJS module's exports has the same exposure under the new builder — `leaflet.markercluster` is the only one in the repo today.

Fix proposed in #4302.

sleidig:
- closes #4301

Since `3.89.0` every map screen throws `L.markerClusterGroup is not a function` and the map never initialises — 28 Sentry events across 5 production tenants, still firing, including the public intake form at `/public-form/form/:id`. All 28 events are on `3.89.0`; none on `3.88.0`.

`leaflet.markercluster` is a side-effect plugin: it imports nothing and assigns `MarkerClusterGroup` / `markerClusterGroup` onto the global `L`, which is leaflet's live CommonJS exports object. Under the previous webpack builder `import * as L from "leaflet"` aliased that object, so the additions were visible. The esbuild application builder adopted in #4271 synthesises the namespace with `__toESM(require_leaflet(), 1)`, which *copies* the exports, so the later mutation is never reflected.

The default import binds `L` to the live exports object again (`__toESM` sets `default` to the raw `module.exports` for a CommonJS module). One-line change plus a comment explaining why it must not be reverted to `import * as`.

## PR Checklist

---

### How this was verified

Bundling the same import pair with the repo's own esbuild and lockfile:

| import form | `typeof L.markerClusterGroup` | `L === window.L` |
| --- | --- | --- |
| `import * as L from "leaflet"` (before) | `undefined` | `false` |
| `import L from "leaflet"` (after) | `function` | `true` |

In the built production bundle the call site changes from `g.markerClusterGroup({...})` — the shape that appears in the deployed 3.89.0 stack trace — to `P.default.markerClusterGroup({...})`.

Also run clean: `tsc --noEmit`, `eslint`, `prettier --check`, `ng build --configuration production`, and the 11 tests in `map.component.spec.ts`.

**Manual test steps:** open any entity list with a map (`/matching`, or a mentor/mentee detail page with an address) on a production-mode build and confirm the map renders with clustered pins instead of an empty container; check the browser console is free of `markerClusterGroup is not a function`.

**No unit test is added,** deliberately: vitest resolves the CommonJS interop differently from the production builder, so `L.markerClusterGroup` is defined under test with *either* import form. A test would pass on the broken code and give false assurance.

https://claude.ai/code/session_01YN11fEb2uzGKavSrtvYvth

---
_Generated by [Claude Code](https://claude.ai/code/session_01YN11fEb2uzGKavSrtvYvth)_

### DIFF
--- a/e2e/tests/matching-entities.spec.ts
+++ b/e2e/tests/matching-entities.spec.ts
@@ -55,3 +55,25 @@ test("Matching entities widget renders and supports selecting a candidate", asyn
   // Once a candidate is picked, the create-match action enables.
   await expect(createMatchButton).toBeEnabled();
 });
+
+test("Matching view renders the map with clustered location markers", async ({
+  page,
+}) => {
+  // Full demo data: many children with addresses close to each other,
+  // which guarantees the map groups them into cluster icons.
+  await loadApp(page);
+
+  await page.getByRole("navigation").getByText("Schools").click();
+  await page.getByRole("navigation").getByText("Matching View").click();
+
+  await expect(page.locator(".leaflet-container")).toBeVisible();
+
+  // Cluster icons are created by the leaflet.markercluster side-effect plugin,
+  // which patches leaflet's live exports object. This breaks silently if the
+  // plugin's additions are not visible on the imported leaflet object (the map
+  // then throws during initialization and renders no markers), so assert the
+  // plugin-generated cluster element itself.
+  await expect(page.locator(".marker-cluster").first()).toBeVisible({
+    timeout: 10_000,
+  });
+});
--- a/e2e/tests/matching-entities.spec.ts
+++ b/e2e/tests/matching-entities.spec.ts
@@ -12,10 +12,21 @@ test("Matching entities widget renders and supports selecting a candidate", asyn
   const child = generateChild({ name: CHILD_NAME });
 
   // Create two schools so the right-side table has selectable rows.
+  // Both get near-identical coordinates so the map groups them into a cluster.
   const school1 = createEntityOfType("School", "match-school-1");
   school1["name"] = "Match Test School One";
+  school1["locationField"] = {
+    lat: 52.4791,
+    lon: 13.432,
+    display_name: "Match Test Street 1",
+  };
   const school2 = createEntityOfType("School", "match-school-2");
   school2["name"] = "Match Test School Two";
+  school2["locationField"] = {
+    lat: 52.4792,
+    lon: 13.4321,
+    display_name: "Match Test Street 2",
+  };
 
   await loadApp(page, [...users, child, school1, school2]);
 
@@ -54,25 +65,14 @@ test("Matching entities widget renders and supports selecting a candidate", asyn
 
   // Once a candidate is picked, the create-match action enables.
   await expect(createMatchButton).toBeEnabled();
-});
 
-test("Matching view renders the map with clustered location markers", async ({
-  page,
-}) => {
-  // Full demo data: many children with addresses close to each other,
-  // which guarantees the map groups them into cluster icons.
-  await loadApp(page);
-
-  await page.getByRole("navigation").getByText("Schools").click();
-  await page.getByRole("navigation").getByText("Matching View").click();
-
+  // The matching config includes location columns, so the widget also renders
+  // the map. Cluster icons are created by the leaflet.markercluster
+  // side-effect plugin, which patches leaflet's live exports object. This
+  // breaks silently if the plugin's additions are not visible on the imported
+  // leaflet object (the map then throws during initialization and renders no
+  // markers), so assert the plugin-generated cluster element itself.
   await expect(page.locator(".leaflet-container")).toBeVisible();
-
-  // Cluster icons are created by the leaflet.markercluster side-effect plugin,
-  // which patches leaflet's live exports object. This breaks silently if the
-  // plugin's additions are not visible on the imported leaflet object (the map
-  // then throws during initialization and renders no markers), so assert the
-  // plugin-generated cluster element itself.
   await expect(page.locator(".marker-cluster").first()).toBeVisible({
     timeout: 10_000,
   });
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -231,6 +231,8 @@ Practical rules:
 ### End-to-End Testing (Playwright)
 
 - Run tests: `npm run e2e`
+- Keep the number of stand-alone e2e tests to a minimum: the initial `loadApp()` is expensive,
+  so prefer integrating a new check as an additional step into an existing scenario.
 - See [`.github/instructions/e2e-tests.instructions.md`](.github/instructions/e2e-tests.instructions.md) for detailed patterns and examples
 
 ### Working with Test Results
--- a/src/app/features/location/map/map.component.ts
+++ b/src/app/features/location/map/map.component.ts
@@ -13,7 +13,11 @@ import {
 } from "@angular/core";
 import { MatDialog } from "@angular/material/dialog";
 import { MatButtonModule } from "@angular/material/button";
-import * as L from "leaflet";
+// default import, not `import * as L`: leaflet.markercluster is a side-effect plugin that adds
+// markerClusterGroup to leaflet's live CommonJS exports object. The esbuild application builder
+// synthesises a *copied* namespace object for `import * as`, so the plugin's additions would not
+// be visible on it and L.markerClusterGroup would be undefined at runtime.
+import L from "leaflet";
 import "leaflet.markercluster";
 import { BehaviorSubject, Subject } from "rxjs";
 import { ConfigService } from "../../../core/config/config.service";


Abhinegi2:
@coderabbit review

sleidig:
@Abhinegi2 thanks for the e2e test — that covers the gap I flagged in the description properly. **`qa-e2e / test-e2e` passed on `d25eae8d`**, which is better evidence than what I had: I could only show the bundle emits `P.default.markerClusterGroup(...)`, whereas your test proves a cluster icon actually renders in a browser against a production build. Worth noting it's deterministic rather than flaky — `faker.geoAddress()` uses `nearbyGPSCoordinate` around a fixed origin under `seed: 1`, so the markers land in the same places every run and clustering at `maxClusterRadius: 20` is stable.

**One blocker I can't diagnose from here, and I'd rather flag it than guess: `qlty check` is red with "25 blocking issues".**

It went red with your commit — green on `4ad6d7c` at 23:35Z, red on `d25eae8d` at 07:57Z:

| check on `d25eae8d` | result |
|---|---|
| qlty check | ❌ **25 blocking issues** |
| qlty fmt / coverage / coverage diff | ✅ |
| qa / code-quality, qa / test-unit | ✅ |
| qa-e2e / test-e2e | ✅ |
| build / build-webapp, build-image, deploy | ✅ |
| SonarCloud, snyk | ✅ |

What I checked locally, all clean: `eslint` and `prettier --check` on both changed files (the two qlty plugins that overlap this repo's own tooling), plus `tsc --noEmit`, `ng build --configuration production`, and the map component's unit tests. So whatever the 25 are, they are not eslint or prettier findings on the diff, and SonarCloud is green so they are probably not radarlint-js ones either.

What I can't do: **qlty.sh is unreachable from my environment** — the network proxy blocks the domain, and the qlty CLI installs from the same host, so I can neither open the issue list nor reproduce the run. qlty posts a commit status rather than a check run, so there are no annotations on the commit to read either. That leaves me unable to root-cause it, and I don't want to push a speculative change at a failure I can't see.

Could you open the [issue list](https://qlty.sh/gh/Aam-Digital/projects/ndb-core/pull/4302/issues) and paste what the 25 are? One specific thing worth ruling out first: qlty runs `osv-scanner` and `trivy fs-vuln`, and `npm audit` on this branch reports **21 pre-existing advisories** in transitive dev dependencies — `undici` ×5, `postcss` ×2, `webpack-dev-server` ×2, `image-size` ×2, `nanoid`, `uuid`. `package-lock.json` is untouched by both commits on this PR, so if that's what they are, this is a lockfile/CVE-feed issue that happens to have landed on our PR rather than something either commit introduced — and it should be fixed separately rather than here. If instead they're genuinely on `matching-entities.spec.ts`, tell me which rules and I'll fix them.

---
_Generated by [Claude Code](https://claude.ai/code)_

Abhinegi2:
@sleidig we can approve the qlty coverage, as it always marks e2e file changes as blocked. But you can check it one more time just to be sure.

sleidig:
Checked that, and the precedent backs it up — @sleidig, this is the double-check you were asked for.

**#4271** also touched e2e files (`e2e/fixtures.ts`, `e2e/tests/attendance.spec.ts`), and all four of its qlty statuses read **"Approved by @Abhinegi2"** rather than "No blocking issues". So qlty did flag that PR the same way and the established practice here is a human approval, not a code change. That matches what @Abhinegi2 says and, combined with the timeline on this PR — green on the commit that touched only `map.component.ts`, red the moment a `.spec.ts` under `e2e/` was added — I'm satisfied the 25 issues are not this PR's content.

That also **retracts my earlier guess**: I floated the pre-existing `npm audit` advisories (`undici`, `postcss`, `webpack-dev-server`, …) as the likely cause. That looks wrong — the e2e-file rule explains the timing far better, and it needs no lockfile work. Worth ignoring that lead rather than chasing it.

One small correction for whoever clicks the approval: the red status is **`qlty check`**, not `qlty coverage`. `qlty coverage` and `qlty coverage diff` both passed on `d25eae8d` (84.1%, "no executable lines in diff"), as did `qlty fmt`. So it's only the one status that needs approving.

I can't do that myself — I have no qlty access, and overriding a quality gate is a maintainer's call regardless. Everything else on `d25eae8d` is green, including `qa-e2e / test-e2e`, so once that status is approved this is ready from CI's point of view.

---
_Generated by [Claude Code](https://claude.ai/code)_

Abhinegi2 (approved):

e2e/tests/matching-entities.spec.ts
```diff
@@ -55,3 +55,25 @@ test("Matching entities widget renders and supports selecting a candidate", asyn
   // Once a candidate is picked, the create-match action enables.
   await expect(createMatchButton).toBeEnabled();
 });
+
+test("Matching view renders the map with clustered location markers", async ({
+  page,
+}) => {
```
sleidig: @Abhinegi2 , can we integrate this e2e test as a step into an existing scenario (guideline: reduce number of stand-alone e2e tests to a minimum because initial loadApp takes a lot of time --> may want to add that to Agents.md also)

sleidig: Agreed with the guideline. I tried to apply it and hit a specific obstacle worth knowing before someone spends time on it — @Abhinegi2, this is context for your call, not me pre-empting it.

**The sibling test in this same file can't host it.** `loadApp` is either/or, not additive — `DemoDataInitializerService.generateDemoData()` does:

```ts
if (window["e2eDemoData"]) {
  await db.putAll(window["e2eDemoData"]);   // fixtures only
} else {
  await this.demoDataService.publishDemoData();   // full demo data
}
```

So passing entities **replaces** the demo data rather than adding to it. The existing matching test loads 1 child + 2 schools, and the schools are created via `createEntityOfType("School", …)` with only a name — no address. Folding the map assertion in there would leave at most one marker and nothing to cluster. Adding addresses to those two schools wouldn't fix it reliably either: `faker.geoAddress()` scatters over a ~10km radius and `maxClusterRadius` is 20px, so whether two markers merge depends on the fitted zoom. That trades a slow test for a flaky one, which I think is the worse deal — the dense default demo data is exactly what makes this assertion deterministic.

**The tests that already load full demo data** (`loadApp(page)` with no args) are: `attendance.spec.ts` ×3 ("View and download attendance report", "Children list displays monthly attendance percentage", "Recurring activities list…"), `notifications.spec.ts`, and one in `admin.spec.ts`. Appending a `/matching` navigation + cluster assertion to the end of any of those costs no extra `loadApp` and stays deterministic — but it puts map coverage in a thematically unrelated spec, and none of them touches maps today.

So it's a real trade-off rather than a clear win:

| option | extra `loadApp` | deterministic | coherent |
|---|---|---|---|
| fold into the sibling matching test | none | ❌ | ✅ |
| append to an existing full-demo-data test | none | ✅ | ❌ |
| keep standalone (current) | +1 | ✅ | ✅ |

My inclination is option 2 with a comment explaining why the map check lives there, since the guideline is about wall-clock cost and that's the only option that pays it without introducing flake. Happy to push that — just say which host you'd prefer.

One caveat on me doing it: **I can't validate an e2e restructure in my environment.** The repo pins a Playwright build expecting `chromium_headless_shell-1234` and only `-1194` is available here; with that binary even the *pre-existing* test in this file fails at `loadApp`. So I'd be relying on CI to confirm rather than proving it locally first, and I'd rather flag that than push blind at a green PR.

**On the AGENTS.md note** — happy to add it. Suggested wording for the "End-to-End Testing (Playwright)" section:

> - Prefer adding a step to an existing scenario over writing a new stand-alone test: each test pays the full `loadApp` cost, which dominates the suite's runtime. Add a new top-level `test()` only when the scenario needs different fixture data — `loadApp(page, entities)` *replaces* the demo data rather than extending it.

Say the word and I'll put it in a separate PR rather than widening this one.

---
_Generated by [Claude Code](https://claude.ai/code)_

sleidig:
@Abhinegi2 this is a better solution than the one I proposed — pinning the two schools ~11m apart makes the cluster deterministic by construction rather than relying on the demo data's seeded scatter, and it keeps the check in the matching spec instead of parking it in an unrelated one. It also sidesteps the `loadApp` either/or problem entirely, which I'd assumed was a dead end.

I checked the wiring against the config rather than assume it, since the whole test hinges on it:

- `entity:School` does declare **`locationField`** as its `dataType: "location"` attribute — the field name is right.
- `appConfig:matching-entities` pairs the columns as `["address", "locationField"]`, so the left side reads `Child.address` and the right side reads `School.locationField`. The widget really does plot the schools, so the map has three markers: the child's `generateChild()` address plus your two schools.
- At 52.4791/13.4320 and 52.4792/13.4321 the two schools are ~11m apart against `maxClusterRadius: 20`px, so they cluster at any sane zoom. No flake window.

**One consequence worth catching before it surprises someone:** this changes the Argos baseline for `matching-entities-loaded`. The screenshot is taken after `loadApp`, and the widget's map now renders three markers with a cluster icon where it previously had one lone marker. Argos hasn't run on this PR (no `argos` status, and the "Final Review" label is still unchecked in the description), so it won't fail CI here — but the diff will show up the next time visual review runs. Worth either adding the label now to re-baseline deliberately, or knowing to accept that diff later.

Also noting you took the AGENTS.md line — the wording you used covers it well.

CI is re-running on `148e8532`; `qa-e2e / test-e2e` is the one that matters and it's in progress. I'll flag it here if it goes red.

---
_Generated by [Claude Code](https://claude.ai/code)_

sleidig:
Argos is up: **2 changed, 1 ignored — waiting for your decision**. That's the baseline shift I flagged, and both diffs should be the fix working rather than regressions. I under-called it though — I predicted one changed screenshot and there are two, so here's the reasoning for both.

I can't open the Argos build from my environment (`app.argos-ci.com` is blocked by the same egress proxy as qlty.sh), so the identification below is inferred from the code, not observed. Worth checking against what the build actually shows.

**Why a map screenshot changes at all:** before this fix `MapComponent.initializeMap()` threw at `map.component.ts:154`, so `ngAfterViewInit` aborted and *no map rendered anywhere*. After it, maps render. Any screenshot containing a map therefore goes from "empty container" to "map with markers" — that diff is the visual proof the fix works.

Expected diffs:

1. **`matching-entities-loaded`** (`matching-entities.spec.ts`) — the matching widget's map now renders, and with the two schools pinned ~11m apart it shows a cluster icon plus the child's own address marker. Changed by both the fix and the new fixture coordinates.
2. **`child-school-relation-new`** (`entity.spec.ts:480`) — taken on the child's **Education tab**, which per the comment at line 452 hosts "multiple RelatedEntities tables (School History, ASER, **matching**)". The screenshot is `fullPage` with the add-dialog open, so the matching widget behind the dialog is captured. Its data is unchanged, so this one should be purely the fix: a map appearing where the container was previously blank, with one marker for the child's `generateChild()` address.

Nothing else in the suite should move: `ViewLocation` / `EditLocation` are the `location` datatype's components and neither embeds `app-map` — the map only appears inline in the matching widget, and otherwise in the `MapPopup` dialog. So the other 38 screenshots have no map to change. `public-form-filled` in particular configures only the `name` field, so despite the production regression hitting `/public-form/form/:id`, that screenshot has no map in it.

**So both diffs should be accepted, not rejected** — if either shows something other than "blank container → map with markers", that's worth a closer look and I'd want to know.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #4308, opened 2026-08-24
fix(logging): improve sentry log message grouping

sleidig:
## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error messages and identification for stalled synchronization.
  - Improved handling and reporting of connectivity failures, including interrupted or unreachable requests.
  - Standardized error grouping to reduce duplicate reports and improve troubleshooting.
  - Preserved underlying error details and context in reported failures.

- **Documentation**
  - Updated logging guidance with recommended error-context handling and expanded grouping behavior.

PR #4311, opened 2026-08-25
fix: EntityPermissionError Sentry grouping + silent failure in config editor

### LINKED ISSUE #4309: Admin config save fails with EntityPermissionError on a production tenant — and each failure opens a brand-new Sentry issue because the message interpolates a timestamped entity id
Found by the monitoring health check, window 23 Aug 23:15Z → 24 Aug 19:00Z. Tenant labels are anonymised; the mapping is in the `#tech-monitoring-alerts` post for this run ([permalink](https://aam-digital.slack.com/archives/CKJFEJ0CD/p1787613891608239)).

There are two problems here. The first is user-facing and needs a judgement call I can't make from monitoring. The second is a grouping defect with a concrete proposed fix.

## 1. An admin could not save configuration (tenant A8, production)

One admin session, one tenant, all on `ndb-core@3.89.0`:

| time (UTC) | Sentry issue | message | route |
|---|---|---|---|
| 15:39:45 | [AAM-DIGITAL-775](https://aam-digital.sentry.io/issues/AAM-DIGITAL-775) | `error during putAll` | `/admin` |
| 15:39:46 | [AAM-DIGITAL-776](https://aam-digital.sentry.io/issues/AAM-DIGITAL-776) | `Error: Failed to upload configuration` | `/admin` |
| 15:52:53 | [AAM-DIGITAL-777](https://aam-digital.sentry.io/issues/AAM-DIGITAL-777) | `not permitted to "create" entity "Config:CONFIG_ENTITY"` | `/admin` |
| 15:53:46 | [AAM-DIGITAL-778](https://aam-digital.sentry.io/issues/AAM-DIGITAL-778) | `not permitted to "create" entity "Config:Permissions:2026-08-24_17-53-46"` | `/admin/user-roles` |

The visible effect is that configuration changes fail to save. The first two events are one second apart, so `putAll` failing and the config upload failing are the same attempt.

**One tenant, not many**, so by the usual split this is a config or permissions problem on A8 rather than shared infrastructure. What I can't tell from here is whether the acting user genuinely lacks `create` on `Config`, or whether the admin UI is evaluating the ability incorrectly — inspecting A8's `Config:Permissions` document against that user's roles would answer it. Filing rather than fixing for that reason.

Two related issues, [AAM-DIGITAL-773](https://aam-digital.sentry.io/issues/AAM-DIGITAL-773) and [AAM-DIGITAL-774](https://aam-digital.sentry.io/issues/AAM-DIGITAL-774), are the same error class but on the dev instance under the `pr-4160` and `pr-4264` preview builds — both open PRs actively reworking the role-permission admin UI. **That half is expected development noise and is not evidence for this issue**; only the four A8 rows above are production.

## 2. The grouping defect: this error can never accumulate, escalate, or be archived

`Config:Permissions:2026-08-24_17-53-46` interpolates **a timestamp** into the error message. Because `EntityPermissionError` carries a stack trace and is not in the cause-grouped list, Sentry groups it by stack, and the varying message keeps producing fresh issues. Every future occurrence opens a **brand-new issue ID**.

That has the second-order effect worth naming explicitly: an issue that never accumulates events can never cross an escalation threshold, so an `archived_until_escalating` rule would silence it permanently, and archiving any one instance does nothing about the next.

### Proposed fix

`EntityPermissionError` fits the documented criterion in `src/app/core/logging/logging.service.ts` almost exactly:

> Only add error types whose name and message already identify the problem on their own, so that the stack adds nothing but noise.

It is thrown from a single central place — `entity-mapper.service.ts`, `if (this.ability.cannot(checkedAction, entity)) throw new EntityPermissionError(...)` — while the stack differs per caller and per build. That is precisely the case `CAUSE_GROUPED_ERROR_TYPES` exists for.

```diff
 const CAUSE_GROUPED_ERROR_TYPES = [
   "DatabaseException",
   "SyncStalledError",
   "ConfigLoadError",
   "PermissionRulesLoadError",
   "SiteSettingsLoadError",
   "RegistryLookupError",
   "RegistryDuplicateError",
+  "EntityPermissionError",
 ];
```

Routing it through `groupByErrorChain` fingerprints on `[type, normalizedValue]`, and the existing `VOLATILE_VALUE_PATTERNS` already handle the volatile part: `Config:Permissions:2026-08-24_17-53-46` normalises via the `<entityId>` pattern and then `\d+` → `<n>`, collapsing every timestamped variant onto one key. The four rows above would become two issues (`create` and `update`) instead of four-and-growing.

**No information is lost.** `enrichSentryEvent` already lifts `entityId` out of the error object into structured extra data, and `EntityPermissionError` exposes `action`, `entityId` and `entityType` as readonly properties. The id is therefore *already* available as context — interpolating it into the message is redundant as well as harmful to grouping. The route stays available as the `transaction` tag.

### One caveat on timing

3.90.0 shipped #4308 (`logging: improve sentry log message grouping`) earlier today, so this touches an area that just changed. The four events above predate that release, and I have not been able to observe an `EntityPermissionError` under 3.90.0 yet — so it is worth confirming #4308 didn't already change this path before applying the diff. Deliberately proposing rather than pushing a second grouping change on the same day.

sleidig:
closes #4309

## (2) Grouping defect (the concrete proposed fix from the issue)

`EntityPermissionError` interpolated the entity id into its message. Since the id is per-record (and for new entities even timestamp-based), and the error carries a stack trace that differs per call site/build, every occurrence opened a brand-new Sentry issue.

- Removed the entity id from the message (kept as a `readonly` property, unchanged).
- Added `EntityPermissionError` to `CAUSE_GROUPED_ERROR_TYPES` in `logging.service.ts`, exactly as proposed in the issue body — it's thrown from one central place (`entity-mapper.service.ts`) and reached from many call sites, which is what that list exists for.
- Verified against the post-#4308 `logging.service.ts` (the caveat raised in the issue) — the new behavior composes cleanly with that change; no conflict.

One trade-off worth flagging for review: when the error is caught explicitly and re-logged via `Logging.error("some static message", err)` (as the admin-overview catch blocks below now do), the entity id no longer appears in `enrichSentryEvent`'s top-level `extra.entityId` (that path only inspects `hint.originalException`, which is the `LoggedError` wrapper, not the wrapped cause). It should still be visible nested under `extra.context` (the raw error object gets attached there), but I have not confirmed this against a real Sentry event. For the *uncaught* case described in the issue (AAM-DIGITAL-777/778), it is captured directly and `enrichSentryEvent` picks up `entityId` correctly.

## (1) Try-catch on config upload/edit

`uploadConfigFile` (the "Upload Application Configuration (JSON)" button — the one named in the issue) already had a try-catch + snackbar since #3999; `AAM-DIGITAL-776` ("Error: Failed to upload configuration") is that handler's own log line firing as designed, not a gap.

The actual unguarded path is `editConfig()` (the "Edit Configuration" JSON-editor flow) and its "Undo" callback — both had **no error handling at all**. A failed save (e.g. a permission error) became a silently-dropped unhandled promise rejection: no snackbar, and the admin has no idea anything went wrong. A failed revert also left the progress dialog stuck open forever. Fixed both with the same try-catch/snackbar pattern already used by `uploadConfigFile`, plus a `finally` so the progress dialog always closes.

If the actual ask was different (e.g. clearer feedback in the existing upload snackbar), happy to adjust.

## Testing

- `entity-mapper.service.spec.ts`: locks in that the id stays off the message but remains on `error.entityId`.
- `logging.service.spec.ts`: new grouping-fingerprint test, asserting the exact fingerprint value (not just equality between two events) so it actually fails without the `CAUSE_GROUPED_ERROR_TYPES` change — verified by reverting that one line locally and confirming the test fails.
- `admin-overview.component.spec.ts`: new tests for `editConfig()` happy path, cancel, save failure (snackbar shown), and revert failure (progress dialog still closes).

All new/affected specs pass locally (74 tests across the three touched spec files).

### DIFF
--- a/src/app/core/entity/entity-mapper/entity-mapper.service.spec.ts
+++ b/src/app/core/entity/entity-mapper/entity-mapper.service.spec.ts
@@ -368,6 +368,16 @@ describe("EntityMapperService permission checks", () => {
     expect(mockAbility.cannot).toHaveBeenCalledWith("create", entity);
   });
 
+  it("should keep the entity id out of the EntityPermissionError message, so Sentry can group by it", async () => {
+    mockAbility.cannot.mockReturnValue(true);
+    const entity = new Entity("test-no-create");
+
+    const error = await entityMapper.save(entity).catch((e) => e);
+
+    expect(error.message).not.toContain(entity.getId());
+    expect(error.entityId).toBe(entity.getId());
+  });
+
   it("should throw EntityPermissionError when saving an existing entity without update permission", async () => {
     mockAbility.cannot.mockImplementation(
       (action: string) => action === "update",
--- a/src/app/core/entity/entity-mapper/entity-mapper.service.spec.ts
+++ b/src/app/core/entity/entity-mapper/entity-mapper.service.spec.ts
@@ -368,16 +368,6 @@ describe("EntityMapperService permission checks", () => {
     expect(mockAbility.cannot).toHaveBeenCalledWith("create", entity);
   });
 
-  it("should keep the entity id out of the EntityPermissionError message, so Sentry can group by it", async () => {
-    mockAbility.cannot.mockReturnValue(true);
-    const entity = new Entity("test-no-create");
-
-    const error = await entityMapper.save(entity).catch((e) => e);
-
-    expect(error.message).not.toContain(entity.getId());
-    expect(error.entityId).toBe(entity.getId());
-  });
-
   it("should throw EntityPermissionError when saving an existing entity without update permission", async () => {
     mockAbility.cannot.mockImplementation(
       (action: string) => action === "update",
--- a/src/app/core/entity/entity-mapper/entity-permission-error.ts
+++ b/src/app/core/entity/entity-mapper/entity-permission-error.ts
@@ -4,13 +4,13 @@ export class EntityPermissionError extends Error {
     public readonly entityId: string,
     public readonly entityType: string,
   ) {
-    // entityId is deliberately kept out of the message: it is per-record (and
-    // for new entities even timestamp-based), so interpolating it would defeat
-    // Sentry grouping. It stays available as a property and is picked up as
-    // structured context by enrichSentryEvent in logging.service.ts.
-    super(
-      `Current user is not permitted to "${action}" entity of type "${entityType}"`,
-    );
+    // action, entityId and entityType are deliberately kept out of the message
+    // (entityId is per-record and, for new entities, even timestamp-based; all
+    // three would otherwise fragment Sentry grouping). They stay available as
+    // properties and are picked up as structured context by enrichSentryEvent
+    // in logging.service.ts, so they remain visible in Sentry's "Additional
+    // Data" without being part of the message or name.
+    super("Current user is not permitted this action");
     this.name = "EntityPermissionError";
   }
 }
--- a/src/app/core/entity/entity-mapper/entity-permission-error.ts
+++ b/src/app/core/entity/entity-mapper/entity-permission-error.ts
@@ -4,8 +4,12 @@ export class EntityPermissionError extends Error {
     public readonly entityId: string,
     public readonly entityType: string,
   ) {
+    // entityId is deliberately kept out of the message: it is per-record (and
+    // for new entities even timestamp-based), so interpolating it would defeat
+    // Sentry grouping. It stays available as a property and is picked up as
+    // structured context by enrichSentryEvent in logging.service.ts.
     super(
-      `Current user is not permitted to "${action}" entity "${entityId}" (${entityType})`,
+      `Current user is not permitted to "${action}" entity of type "${entityType}"`,
     );
     this.name = "EntityPermissionError";
   }
--- a/src/app/core/logging/logging.service.ts
+++ b/src/app/core/logging/logging.service.ts
@@ -452,7 +452,6 @@ const CAUSE_GROUPED_ERROR_TYPES = [
   "SiteSettingsLoadError",
   "RegistryLookupError",
   "RegistryDuplicateError",
-  "EntityPermissionError",
 ];
 
 /**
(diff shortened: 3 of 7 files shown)


src/app/core/entity/entity-mapper/entity-mapper.service.spec.ts
```diff
@@ -368,6 +368,16 @@ describe("EntityMapperService permission checks", () => {
...
+  it("should keep the entity id out of the EntityPermissionError message, so Sentry can group by it", async () => {
+    mockAbility.cannot.mockReturnValue(true);
+    const entity = new Entity("test-no-create");
+
+    const error = await entityMapper.save(entity).catch((e) => e);
+
+    expect(error.message).not.toContain(entity.getId());
+    expect(error.entityId).toBe(entity.getId());
+  });
+
```
sleidig:
```suggestion
```

src/app/core/entity/entity-mapper/entity-permission-error.ts
```diff
@@ -4,8 +4,12 @@ export class EntityPermissionError extends Error {
     public readonly entityId: string,
     public readonly entityType: string,
   ) {
+    // entityId is deliberately kept out of the message: it is per-record (and
+    // for new entities even timestamp-based), so interpolating it would defeat
+    // Sentry grouping. It stays available as a property and is picked up as
+    // structured context by enrichSentryEvent in logging.service.ts.
     super(
-      `Current user is not permitted to "${action}" entity "${entityId}" (${entityType})`,
+      `Current user is not permitted to "${action}" entity of type "${entityType}"`,
```
sleidig: Does this error object still allow to find those details in the logs somewhere, without being in the name?

src/app/core/entity/entity-mapper/entity-permission-error.ts
```diff
@@ -4,8 +4,12 @@ export class EntityPermissionError extends Error {
...
     );
```
sleidig:
```suggestion
super("Current user is not permitted this action");
```

src/app/core/logging/logging.service.ts
```diff
@@ -452,6 +452,7 @@ const CAUSE_GROUPED_ERROR_TYPES = [
   "SiteSettingsLoadError",
   "RegistryLookupError",
   "RegistryDuplicateError",
+  "EntityPermissionError",
```
sleidig: do we still need this, if the error message is generalized? Prefer to fix at root and keep the logging grouping to a minimum

src/app/core/entity/entity-mapper/entity-permission-error.ts
sleidig: Good catch — with the message now fully static, nothing was left to surface `action`/`entityType` either. Extended `enrichSentryEvent`'s existing extras mechanism (previously only `entityId`, `status`, `reason`, `name`) to also pick up `action` and `entityType`, so all three stay visible in Sentry's Additional Data without touching the message or name. Added a test for that in 593128e.

src/app/core/logging/logging.service.ts
sleidig: Removed it. Worth noting the trade-off though: Sentry's default grouping is stack-based whenever a stack trace is present (which a normally-thrown Error always has), not message-based - so even with a fully static message, two occurrences from different call sites (e.g. the admin overview vs. the user-roles page) will still land in separate issues, each keyed off its own stack. What the static message *does* fix on its own is the original failure mode: a single call site no longer spawns a new issue per timestamp, since repeated occurrences from the same place now hash to the same stack + same message. So this is a real improvement (unbounded growth -> stable per call site) even without full cross-call-site consolidation. Went with your preference to keep the grouping list minimal rather than chase full consolidation here.

Issue #4316, opened 2026-08-25
Anonymizing a linked record through "retain-anonymized" cannot be undone


sleidig:
**Describe the bug**

An entity-reference field (`dataType: "entity"`) configured with `anonymize: "retain-anonymized"` ("Partially Anonymize" in the Admin UI) does not anonymize the *value* — it runs the **full anonymize action on the referenced record**:

https://github.com/Aam-Digital/ndb-core/blob/master/src/app/core/basic-datatypes/entity/entity.datatype.ts#L339-L357

```ts
override async anonymize(value, schemaField, parent): Promise<string> {
  const referencedEntity = await this.entityMapper.load(schemaField.additional, value);
  if (!referencedEntity) { return value; }
  await this.removeService.anonymize(referencedEntity);
  return value;
}
```

The result of that nested action is discarded: `EntityActionsService.anonymize()` returns a `boolean` and `EntityDatatype.anonymize()` returns `value` unchanged. So the referenced record's state *before* the change never reaches the `CascadingActionResult.originalEntitiesBeforeChange` of the outer action (`EntityAnonymizeService.anonymizeProperty()`).

Consequences:

1. The undo snackbar of the outer anonymize action does not restore the referenced record.
2. The nested call raises its own undo snackbar, but it is immediately superseded by the outer action's snackbar — so that undo is not reachable either.
3. As anonymization is irreversible by design ("Details that are removed during anonymization cannot be recovered"), the referenced record is permanently anonymized with no way back.

Side effect of going through `EntityActionsService.anonymize()` for the nested record: it opens its own confirmation dialog while the outer action's progress dialog is still open.

**To Reproduce**

1. Configure entity type A with a field of `dataType: "entity"` referencing entity type B (with `hasPII: true`), and set that field's anonymize mode to "Partially Anonymize" (`anonymize: "retain-anonymized"`).
2. Open a record of type A that links to a record of type B, trigger "Anonymize" and confirm.
3. A second "Anonymize?" dialog appears, naming the linked B record — confirm it as well.
4. Click "Undo" in the snackbar.
5. The A record is restored, the linked B record stays anonymized.

**Expected behavior**

The nested anonymization is part of the same user action and should be part of its undo: `EntityDatatype.anonymize()` should propagate the `CascadingActionResult` of the referenced record so the outer action's `originalEntitiesBeforeChange` includes it. The overall action should also show only one confirmation dialog and one snackbar.

[...]

PR #4319, opened 2026-08-25
fix(anonymize): never anonymize a record that is only linked from another

### LINKED ISSUE #4320: Anonymize propagates into linked records based on the anonymize mode, not an explicit relationship
**Describe the bug**

Anonymizing a record can also anonymize a completely separate record that is only linked from it — e.g. a case record's "responsible staff member" link wipes that user's profile.

This happens whenever the linking field is set to "Partially Anonymize" (`anonymize: "retain-anonymized"`). For `dataType: "entity"` that mode does not anonymize the *value*: it loads the referenced record and runs the full anonymize action on it (`EntityDatatype.anonymize()`).

The mode is a per-field data retention setting, but for entity references it silently decides a relationship question. An admin picking it from the dropdown has no way to know it reaches into other records.

**To Reproduce**

1. Set an entity-reference field (e.g. a link to a User) to "Partially Anonymize".
2. Anonymize a record that has that link set.
3. A second confirmation dialog appears; the linked record is anonymized too.

Note that step 2 also applies to records reached by the cascade, not just the one the action was triggered on.

**Expected behavior**

Propagating into a linked record should require the relationship to be tagged explicitly (as `entityReferenceRole` already does for the opposite direction), not follow from a retention mode. A user profile is never "part of" a case record; a survey response may well be.

Without such a tag, "Partially Anonymize" on an entity reference should do what its tooltip promises for datatypes without a partial form, and remove the value.

**Additional context**

Two related asymmetries with delete, which uses the same cascade:

- Delete never propagates along a record's own references; only anonymize does.
- `anonymizeEntity()` does not check `entity.anonymized`, so a composite reference cycle recurses indefinitely. Delete terminates naturally because the record is gone.
- `EntityAnonymizeService` ignores `enableUserAccounts`: anonymizing a staff profile leaves a working login attached to the emptied record, and there is no guard against doing it to your own account. `EntityDeleteService` has both.

See also #4316 (this cascade cannot be undone, and the "Partially Anonymize" description is wrong for entity references).

sleidig:
Fixes #4320 — anonymizing a record could also anonymize a completely separate record that was only linked from it, e.g. a case record's "responsible staff member" link wiping that user's profile.

## The problem

For `dataType: "entity"`, `anonymize: "retain-anonymized"` ("Partially Anonymize") did not anonymize the *value*: `EntityDatatype.anonymize()` loaded the referenced record and ran the full anonymize action on it. So whether anonymization reached into another record was decided by a per-field **data retention** setting, while the question it answers is one about the **relationship**. An admin picking that option from the dropdown had no way to know it touched other records — and it applied to every record the cascade reached, not just the one the action was triggered on.

## The fix

A referenced record is now always treated as an independent record: the reference is removed, as for any other datatype that does not support partial anonymization — which is what the option's own description in the Admin UI promises. `EntityDatatype` no longer depends on `EntityActionsService` at all.

There is a legitimate case for the opposite behaviour (a survey response that is genuinely only a part of the record linking to it), but expressing it belongs on the relationship, not on the anonymize mode — most likely as an explicit counterpart to the existing `entityReferenceRole: "composite"`. That is deliberately left out until someone actually asks for it.

No shipped base config sets `retain-anonymized` on an entity field, so the only configurations affected are those where an admin picked "Partially Anonymize" expecting value masking — which is the bug.

## Also: a recursion guard for the cascade

`anonymizeEntity()` now returns early for a record that is already anonymized. Without it the cascade could revisit a record it had already processed, and a cycle of records referencing each other as `"composite"` recursed until the heap was exhausted (verified: removing the guard makes the new unit test run for ~140s and die with an out-of-memory error). Deletion terminates naturally in that situation because the record is gone by the time the cascade comes back around; an anonymized record stays in place with its retained references intact.

## Tests

- **Unit** (`entity.datatype_advanced.spec.ts`): anonymizing an entity reference removes the value and leaves the referenced record untouched.
- **Unit** (`entity-anonymize.service.spec.ts`): two records referencing each other as `"composite"` are both anonymized and the cascade terminates; an already anonymized record is not saved again. Note that the cycle test fails by hanging if the guard is removed, as termination tests do.
- **E2E** (`anonymize.spec.ts`): one scenario walking the whole flow through the UI — configure the note's user reference via the Admin UI, anonymize a child with two notes, assert the note belonging to that child only is cascaded into while the note shared with a second child is kept and reported for manual review, then anonymize the remaining note and assert its user reference is cleared and the linked user profile still has all its data. Both modes such a link is realistically configured with are exercised: "Remove" (`anonymize: ""`) and "Partially Anonymize".

Verified by mutation, so the assertions do not pass vacuously: restoring the old recursion makes the e2e fail (the reference survives and the user profile is anonymized).

## Notes for review

- Delete is not touched: it has no outgoing propagation at all, so nothing changes there.
- One gap from #4320 is not addressed here: `EntityAnonymizeService` still ignores `enableUserAccounts`, so anonymizing a staff profile leaves a working login attached to the emptied record, and there is no guard against doing it to your own account. `EntityDeleteService` has both.
- No Argos screenshot is added — the flow shows no UI that existing baselines do not already cover.

Related: #4316 (this cascade cannot be undone, and the "Partially Anonymize" description is wrong for entity references — the latter becomes accurate with this change).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved anonymization of linked records: composite-linked records are anonymized and archived, while shared records remain unchanged for manual review.
  * References to anonymized entities are removed without altering the referenced profile’s data.
  * Prevented empty entries from appearing in anonymized list fields.
  * Repeated anonymization now safely skips records already anonymized, including cyclic relationships.
* **Documentation**
  * Clarified cascading deletion and anonymization behavior for composite and referenced relationships.

### DIFF
--- a/e2e/tests/anonymize.spec.ts
+++ b/e2e/tests/anonymize.spec.ts
@@ -1,0 +1,172 @@
+import { E2E_REF_DATE, expect, loadApp, Page, test } from "#e2e/fixtures.js";
+import { generateChild } from "#src/app/child-dev-project/children/demo-data-generators/demo-child-generator.service.js";
+import { generateNote } from "#src/app/child-dev-project/notes/demo-data/demo-note-generator.service.js";
+import { generateUsers } from "#src/app/core/user/demo-user-generator.service.js";
+import { createEntityOfType } from "#src/app/core/demo-data/create-entity-of-type.js";
+
+const CHILD_NAME = "<ANONYMIZE TEST CHILD>";
+const OTHER_CHILD_NAME = "<ANONYMIZE TEST OTHER CHILD>";
+const SUBJECT_SOLE_LINK = "<ANONYMIZE TEST RECORD>";
+const SUBJECT_SHARED_LINK = "<ANONYMIZE TEST SHARED RECORD>";
+const LINKED_USER_NAME = "<ANONYMIZE TEST USER>";
+const LINKED_USER_PHONE = "555-0100";
+
+/**
+ * Set the anonymize mode of the note's user reference ("Team involved")
+ * through the Admin UI.
+ */
+async function setNoteAuthorsAnonymizeMode(page: Page, mode: string) {
+  await page.getByRole("navigation").getByText("Notes").click();
+  await page
+    .locator("button[mat-icon-button][color='primary']")
+    .first()
+    .click();
+  await page.getByText("Configure Data Structure").click();
+  await page.getByText("Details View & Fields").click();
+
+  const authorsField = page
+    .locator(".admin-form-field")
+    .filter({ hasText: "Team involved" })
+    .first();
+  await authorsField.scrollIntoViewIfNeeded();
+  await authorsField.hover();
+  await authorsField.getByRole("button", { name: "Edit Field" }).click();
+
+  const fieldDialog = page.locator("mat-dialog-container");
+  await fieldDialog.getByRole("tab", { name: "Advanced Options" }).click();
+  await fieldDialog
+    .locator("app-anonymize-options")
+    .getByRole("combobox")
+    .click();
+  await page.getByRole("option", { name: mode, exact: true }).click();
+  // move the pointer away so the option's tooltip does not cover the button
+  await page.mouse.move(0, 0);
+  await fieldDialog.getByRole("button", { name: "Apply", exact: true }).click();
+  await expect(fieldDialog).not.toBeVisible();
+
+  await page.getByRole("button", { name: "Save" }).first().click();
+  await expect(page.getByText("Configuration updated")).toBeVisible();
+}
+
+/**
+ * Anonymization must never spread to a record that is only linked *from* an
+ * anonymized record. It cascades to records linking *to* it through a
+ * "composite" relation, and clears the record's own references according to
+ * their configured anonymize mode.
+ *
+ * This flow covers all of that in one pass:
+ * - the note belonging to the anonymized child only is cascaded into,
+ * - the note shared with another child is kept and reported for manual review,
+ * - both anonymize modes a user reference is realistically configured with
+ *   clear the reference and leave the referenced record alone,
+ * - and the linked user profile keeps all of its data throughout.
+ */
+test("Anonymize cascades to a record's own notes but never to a linked user profile", async ({
+  page,
+}) => {
+  const users = generateUsers();
+  const linkedUser = createEntityOfType("User", "anonymize-e2e-user");
+  linkedUser.name = LINKED_USER_NAME;
+  linkedUser.phone = LINKED_USER_PHONE;
+
+  const child = generateChild({ name: CHILD_NAME });
+  const otherChild = generateChild({ name: OTHER_CHILD_NAME });
+
+  const note = generateNote({
+    child,
+    author: linkedUser,
+    date: new Date(E2E_REF_DATE),
+  });
+  note.subject = SUBJECT_SOLE_LINK;
+
+  // a note linked to another child as well is kept, so the assertions below
+  // distinguish "the cascade anonymized this note" from "no note is listed"
+  const sharedNote = generateNote({
+    child,
+    author: linkedUser,
+    date: new Date(E2E_REF_DATE),
+  });
+  sharedNote.subject = SUBJECT_SHARED_LINK;
+  // qlty-ignore: radarlint-js:typescript:S1874 - the demo config still links notes to children through this deprecated field
+  sharedNote.children.push(otherChild.getId());
+
+  await loadApp(page, [
+    ...users,
+    linkedUser,
+    child,
+    otherChild,
+    note,
+    sharedNote,
+  ]);
+
+  // "Remove" is what a link to a user profile is usually configured with,
+  // and is stored as `anonymize: ""`
+  await setNoteAuthorsAnonymizeMode(page, "Remove");
+
+  // both notes are listed while the child they belong to is still active
+  await page.getByRole("navigation").getByText("Notes").click();
+  await expect(
+    page.getByRole("cell", { name: SUBJECT_SOLE_LINK }),
+  ).toBeVisible();
+  await expect(
+    page.getByRole("cell", { name: SUBJECT_SHARED_LINK }),
+  ).toBeVisible();
+
+  // anonymize the child
+  await page.getByRole("navigation").getByText("Children").click();
+  await page.getByRole("cell", { name: CHILD_NAME }).click();
+  await page.locator("app-entity-actions-menu").getByRole("button").click();
+  await page.getByRole("menuitem", { name: "Anonymize" }).click();
+  await page.getByRole("dialog").getByRole("button", { name: "Yes" }).click();
+
+  // the shared note is linked to another child as well, so it is left
+  // unchanged and reported for manual review
+  await page.getByRole("dialog").getByRole("button", { name: "OK" }).click();
+
+  await expect(page.getByText("Anonymized & Archived")).toBeVisible();
+
+  // the cascade reached the note that is linked to this child only: it is
+  // archived along with being anonymized and drops out of the list, while
+  // the note shared with another child stays
+  await page.getByRole("navigation").getByText("Notes").click();
+  await expect(
+    page.getByRole("cell", { name: SUBJECT_SHARED_LINK }),
+  ).toBeVisible();
+  await expect(
+    page.getByRole("cell", { name: SUBJECT_SOLE_LINK }),
+  ).toBeHidden();
+
+  // "Partially Anonymize" does not mark the referenced record as a part of the
+  // note, so it must not reach into the user profile either - the reference is
+  // removed instead
+  await setNoteAuthorsAnonymizeMode(page, "Partially Anonymize");
+
+  await page.getByRole("navigation").getByText("Notes").click();
+  await page.getByRole("cell", { name: SUBJECT_SHARED_LINK }).click();
+  const noteDialog = page.getByRole("dialog");
+  await expect(noteDialog.getByText(LINKED_USER_NAME)).toBeVisible();
+
+  await noteDialog
+    .locator("app-entity-actions-menu")
+    .getByRole("button")
+    .click();
+  await page.getByRole("menuitem", { name: "Anonymize" }).click();
+  await page.getByRole("dialog").getByRole("button", { name: "Yes" }).click();
+
+  await expect(noteDialog.getByText(LINKED_USER_NAME)).toBeHidden();
+  await noteDialog.getByRole("button", { name: "Cancel" }).click();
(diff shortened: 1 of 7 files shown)


sleidig:
@Abhinegi2 , can you take over this PR to do a final testing? And validate if the change makes sense conceptually :blush:

Abhinegi2:
> @Abhinegi2 , can you take over this PR to do a final testing? And validate if the change makes sense conceptually 😊

Yes, makes sense to me conceptually.  User profile is its own record, so anonymizing a case record should never wipe it. I’ll take over the pr now and do a final code review/ functional testing

src/app/core/basic-datatypes/entity/entity.datatype.ts
```diff
@@ -340,19 +343,8 @@ export class EntityDatatype extends StringDatatype {
...
-
-    if (!referencedEntity) {
-      // TODO: remove broken references?
-      return value;
-    }
-
-    await this.removeService.anonymize(referencedEntity);
-    return value;
+  ): Promise<string | undefined> {
+    return undefined;
```
Abhinegi2, line 347: Array entity-reference fields do not get the reference removed, they end up holding `[null, null]`.

`EntityAnonymizeService.anonymizeProperty` maps each element for an `isArray` field and assigns the resulting array:

```ts
anonymizedValue = await Promise.all(
  asArray(entity[key]).map((v) => dataType.anonymize(v, ...)),
);
```

With `anonymize()` returning `undefined` per element that becomes `[undefined, undefined]`, `StringDatatype.transformToDatabaseFormat(undefined)` passes each one through, and the saved document gets `"refs": [null, null]`.

Verified locally with a throwaway spec, an entity with a `retain-anonymized` reference field as `isArray: true` next to a single-value one:

```
refs:      [ undefined, undefined ]
singleRef: undefined
db:        { refs: [undefined, undefined], _id: ..., inactive: true, anonymized: true }
```

The single-value case is clean, the key drops out of the document entirely. The array case is not, and array is the common shape for `dataType: "entity"`: `Note.authors`, `Note.children` and `relatedEntities` are all `isArray: true`. The e2e added in this PR configures `authors`, so it walks this exact path, but it only asserts that the user's name disappeared from the note dialog, which is equally true for `[null]`.

Two ways out:

```ts
anonymizedValue = (
  await Promise.all(asArray(entity[key]).map((v) => dataType.anonymize(...)))
).filter((v) => v !== undefined);
```

or route the field through `removeProperty` when the datatype has no partial anonymization, which is what the new README paragraph promises ("simply removes the reference").

src/app/core/basic-datatypes/entity/entity.datatype_advanced.spec.ts
```diff
@@ -88,28 +88,21 @@ describe("Schema data type: entity (advanced functionality)", () => {
     ).resolves.toBe(undefined);
   });
 
-  it("should anonymize entity recursively", async () => {
+  it("should remove the reference without anonymizing the referenced entity", async () => {
```
Abhinegi2, line 81: Leftover test setup: the mock `EntityActionsService` provider at line 34 and its import at line 6 can go now that `EntityDatatype` no longer injects it.

src/app/core/basic-datatypes/entity/entity.datatype_advanced.spec.ts
```diff
@@ -88,28 +88,21 @@ describe("Schema data type: entity (advanced functionality)", () => {
...
+      referencedEntity.getId(),
+      { additional: TestEntity.ENTITY_TYPE, dataType: "entity" },
       null,
     );
 
-    expect(anonymizedValue).toEqual(testValue);
-    expect(removeService.anonymize).toHaveBeenCalledWith(referencedEntity);
+    expect(anonymizedValue).toBeUndefined();
+    expect(
+      entityMapper.get(TestEntity.ENTITY_TYPE, referencedEntity.getId()),
```
Abhinegi2: This assertion cannot fail. `MockEntityMapperService` stores the entity instance itself (`mock-entity-mapper-service.ts:70`) and `get` hands the same object back (`:103`), so this compares `referencedEntity` to itself. It would still pass if the referenced record had been anonymized in place, which is exactly what the test is meant to rule out.

Either compare a concrete field:

```ts
const stored = entityMapper.get(
  TestEntity.ENTITY_TYPE,
  referencedEntity.getId(),
) as TestEntity;
expect(stored.name).toBe("test");
```

or assert the referenced record was never written, matching the style of the new `EntityAnonymizeService` specs:

```ts
vi.spyOn(entityMapper, "save");
// ...
expect(entityMapper.save).not.toHaveBeenCalled();
```

Abhinegi2 (commented):
Pre-review pass. Three things I would fix before merge, the first one is a behaviour bug I could reproduce; the rest of what I found is minor and I can raise it separately.

src/app/core/basic-datatypes/entity/entity.datatype.ts
Abhinegi2: Fixed in 78ef58e83. `anonymizeProperty` now filters `undefined` out of the mapped array, so such a field ends up `[]` rather than `[null, null]`. The single-value branch is untouched, so the `undefined` you chose there still stands.

Added a unit test on `AnonymizableEntity.referencesToRetainAnonymized`, a field that already existed in the spec but was never set by any test. It fails without the fix (`[undefined, undefined]` vs `[]`).

One thing worth knowing about the shape I picked: the two modes now write different empty values for the same field. "Remove" sets an array field to `null` through `removeProperty`, "Partially Anonymize" sets `[]`. I went with `[]` because it keeps the declared `string[]` type and needs no null guard on readers. If you would rather have one representation, the alternative is routing a field through `removeProperty` when its datatype has no partial anonymization, but that also turns the single-value case into `null`. Happy to switch.

src/app/core/basic-datatypes/entity/entity.datatype_advanced.spec.ts
Abhinegi2: Removed in b42bc5855, along with the now-unused import.

src/app/core/basic-datatypes/entity/entity.datatype_advanced.spec.ts
Abhinegi2: Fixed in b42bc5855. It now checks `storedEntity.name` and `storedEntity.anonymized` against literals, plus `expect(entityMapper.save).not.toHaveBeenCalled()`, so the assertion fails if the referenced record is touched.

Abhinegi2:
@coderabbit review

e2e/tests/anonymize.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+  await page.getByRole("cell", { name: LINKED_USER_NAME }).click();
+
+  await expect(
+    page.getByRole("heading", { name: LINKED_USER_NAME }),
+  ).toBeVisible();
+  await expect(
+    page.locator("#entity-field__phone").getByRole("textbox"),
+  ).toHaveValue(LINKED_USER_PHONE);
+  await expect(page.getByText("Anonymized & Archived")).toBeHidden();
+});
```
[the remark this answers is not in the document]
Abhinegi2: Added in a157421ce. Two screenshots, which is in line with what the other specs carry for a non-visual flow (`notes` has 2, `file-upload`, `matching-entities` and `public-form` have 1 each; only the admin config specs go higher).

Picked the two states this flow produces that no existing baseline covers:

- `anonymize-related-records-review`, the cascade's "Related records may still contain personal data" dialog. Nothing else in the suite captures it. The test was clicking OK on that dialog without asserting anything, so it now checks the heading first.
- `note-anonymized-reference-cleared`, the anonymized note with its user reference gone, so a reference that is left behind instead of removed shows up as a visual diff.

Left out the user profile after anonymizing, since the test already asserts its name and phone field by field, so a snapshot adds nothing there.

Worth noting for anyone adding screenshots near a save action: both of these points still had the confirmation snackbar of a preceding action on screen, and snackbars dismiss themselves after 8s, so whether they got captured depended on how fast the run reached that step. Hidden now through the `argosCSS` option, the same mechanism the fixture already uses for tooltips.

Verified with `SCREENSHOT=1`, both frames render clean.

src/app/core/basic-datatypes/entity/entity.datatype.ts
sleidig: We did intentionally set anonymized values to `null` so that the UI and troubleshooting can distinguish it from something that was never there initially. Whether an array with multiple null values makes sense is a good question though (and if our components handle it seamlessly) :thinking:

Abhinegi2:
@coderabbit review

Abhinegi2 (approved):
looks good to me now

src/app/core/entity/entity-actions/entity-anonymize.service.ts
```diff
@@ -74,11 +84,15 @@ export class EntityAnonymizeService extends CascadingEntityAction {
...
-      anonymizedValue = await Promise.all(
+      const anonymizedValues = await Promise.all(
         asArray(entity[key]).map((v) =>
           dataType.anonymize(v, entity.getSchema().get(key), entity),
         ),
       );
+      // a datatype without partial anonymization returns nothing for a value.
+      // drop those entries instead of leaving empty slots in the array,
+      // which would be stored as `null` values in the database.
+      anonymizedValue = anonymizedValues.filter((v) => v !== undefined);
```
sleidig, line 95: @Abhinegi2 , did these `null` values cause any UI or log issues?

Abhinegi2: No log issues, entity-block returns early when the id is empty so nothing ever reaches the loader.

UI does change a bit though. Say a note has Team involved = Anandamayi, Charumati:

| | stored | field shows |
|---|---|---|
| before | `authors: [null, null]` | `-  -` (one block per null, tooltip "no record defined") |
| now | `authors: []` | empty |

And if that note had 5 people instead of 2, and renders "5 in total", so you can still see how many references were removed from an anonymized record.

sleidig: Thanks for clarifying!

hmm ... can't quite make up my mind which is the better UX. Knowing that there had been 5 authors, but their details have been anonymized? Or removing somewhat useless placeholders from the UI?

sleidig: We discussed the UX for this a bit in standup today:

We are aiming for an indicator on every field that was touched by "anonymize", similar to the one for "partially anonymized" fields:
image

This is it's own issue, though. Keeping the code as is, for now. But we might add back the `null` values.

sleidig: see https://github.com/Aam-Digital/ndb-core/issues/4326

PR #4327, opened 2026-08-27
fix(core): allow to configure a default_data_source in config.json

sleidig:


src/app/core/common-components/entities-table/data-source/data-source-resolver.ts
```diff
@@ -25,15 +25,23 @@ function getDataSource(
...
-  if (environment.session_type === SessionType.online) {
-    if (!dataSource && loaderMethod) {
-      // special loaders are not supported by the paginated data source
-      return InMemoryDataSource;
-    }
-    return dataSource && availableDataSources[dataSource]
-      ? availableDataSources[dataSource]
-      : availableDataSources.paginated;
-  } else {
+  if (dataSource && availableDataSources[dataSource]) {
```
TheSlimvReal, line 33: I think you should keep the check for the online session. The paginated data source does currently not work against the local DB and also I don't see much reason for using paginated in a local setup.

src/app/core/common-components/entities-table/data-source/data-source-resolver.ts
```diff
@@ -25,15 +25,23 @@ function getDataSource(
...
+    return availableDataSources[dataSource];
+  }
+
+  if (loaderMethod) {
+    // special loaders are not supported by the paginated data source
     return InMemoryDataSource;
   }
+
+  const systemDefault = availableDataSources[environment.default_data_source];
+  if (systemDefault) {
```
TheSlimvReal: Same here, should not pick up paginated data source if it is not in online mode.

TheSlimvReal (changes requested):
Looks good but the paginated data source should only be available in the online-only mode

src/app/core/common-components/entities-table/data-source/data-source-resolver.ts
sleidig: Ah, I thought although it might not add much value it would still be functional. Okay, will disable it for that completely

TheSlimvReal (approved):

PR #4340, opened 2026-08-31
fix(logging): report DB responses by root cause, count conflicts instead of alerting

### LINKED ISSUE #4332: AAM-DIGITAL-77H is CouchDB 409 conflicts, not permission denials — reported as "Failed to fetch", at warning level, with no way to tell a lost save from replication noise
Filed by the automated monitoring routine. Window: **26 Aug 23:23Z → 27 Aug 23:24Z**.

Tenant identifiers are anonymised per the standing rule that client system URLs and IDs stay out of GitHub. The mapping is in the corresponding `#tech-monitoring-alerts` post.

## What changed: #4325 shipped this afternoon and immediately falsified the previous diagnosis

[AAM-DIGITAL-77H](https://aam-digital.sentry.io/issues/AAM-DIGITAL-77H) `Failed to fetch from DB with 40X error` — 41 lifetime occurrences, **escalating**, first seen 25 Aug 06:24Z.

My previous run read this issue as *"40X sync denials"* — i.e. permission denials — and built an argument on it about the server half being invisible. **That was wrong.** #4325 (`logging: report the HTTP status of unexpected 4XX DB responses`) shipped in `3.90.1` at 15:34Z today, and both events that have landed on `3.90.1` since carry the status:

| Time (UTC) | Tenant | Requested URL | Status |
|---|---|---|---|
| 15:54:24 | Tenant B5 | `/db/app/Note%3A<uuid>` | **409** |
| 20:34:10 | Tenant B2 | `/db/app/HomeVisit%3A<uuid>` | **409** |

Two different tenants, two different entity types, two different routes (`/c/mentee/:id/`, `/c/home-visit/:id/`), two different users and countries. Not permission denials — **document update conflicts**.

## The code says the same thing, and rules the permission reading out entirely

`src/app/core/database/pouchdb/remote-pouch-database.ts:21-25` and `:203-220`

```ts
const EXPECTED_4XX_STATUSES: number[] = [
  HttpStatusCode.Unauthorized,   // 401
  HttpStatusCode.Forbidden,      // 403
  HttpStatusCode.NotFound,       // 404
];
```

```ts
} else if (EXPECTED_4XX_STATUSES.includes(result.status)) {
  // expired session (401), permission-filtered doc (403) and missing doc (404)
  // are part of normal operation and handled by callers
  Logging.debug("Failed to fetch from DB with 40X error", describeResponse(result));
} else {
  Logging.warn("Failed to fetch from DB with 40X error", {
    ...describeResponse(result),
    requestedUrl: remoteUrl,
  });
}
```

401 and 403 are routed to `debug` and never reach Sentry. So **by construction** an event in 77H can never be a permission denial. Everything in that issue is an *unexpected* 4xx, and on the evidence available that means 409.

## Why this needs a decision rather than a patch

A 409 on a single-document URL is a rejected write, reported by a code path named "fetch". What I cannot determine from the event is **whether a user lost an edit**:

- If these are PouchDB's own revision races during replication, they are handled internally and this is log noise that should be demoted to `debug` alongside 401/403/404.
- If these are `db.put()` calls from a user save that lost a revision race, then a user pressed Save and the write was rejected — at `level=warning`, so nobody is alerted, and 41 occurrences and rising.

`describeResponse()` does not capture the **request method**, so the event cannot distinguish the two, and that is exactly the fact that decides which fix is correct. This is a data-integrity judgement call, so it is a tracked issue rather than a PR.

## Proposed

1. **Include the request method** (and, for a write, whether a conflict handler subsequently resolved it) in `describeResponse()` / the `Logging.warn` context. This is the minimum needed to classify the remaining events, and it is the same shape of change as #4325 — which has already proved its worth here within hours of shipping.
2. **Handle 409 explicitly** rather than letting it fall into the generic "unexpected 4xx" bucket: either add it to `EXPECTED_4XX_STATUSES` (if step 1 shows it is replication noise) or raise it as its own named error so it groups, alerts and reads correctly.
3. **Rename the message.** "Failed to fetch from DB" for a rejected write is actively misleading — it is what sent the previous analysis down the wrong path. Something like "Unexpected DB response" would cover both directions.

## Evidence

- Sentry: [AAM-DIGITAL-77H](https://aam-digital.sentry.io/issues/AAM-DIGITAL-77H) — 41 lifetime, 4 events in this 24h window, `level=warning`, escalating
- Releases in-window: 2 events on `ndb-core@3.90.1` (status visible, both 409), 2 on `3.90.0` (status not recorded — these cannot be retroactively classified)
- Tenants seen in this window: B2, B5

## Related

- #4325 — shipped the status reporting that made this visible. Working as intended; this issue exists because of it.
- #4174 — "log only the document ID when a save conflict cannot be resolved" — the existing conflict-handling path that step 1 should correlate against.

sleidig:
Fixes #4332

## The issue's diagnosis was right about the mechanism, wrong about the population

#4332 read AAM-DIGITAL-77H as CouchDB 409 conflicts. Sampling the events across **every** transaction cluster before changing anything showed 409 is the *minority*. Of the events that can be classified at all (release ≥ 3.90.1, the first to record the status):

| Status | Count | Requests |
|---|---|---|
| 409 | 3 | all `PUT` on single-document URLs — conflicts, as the issue said |
| **400** | 4 | all `GET`: `_all_docs?startkey=…`, a plain `<Type>:<uuid>`, a `User:` id **containing a space**, and a `_local/<base64>` replication checkpoint read |

So one issue was holding at least two unrelated root causes, and the larger one had never been noticed. Everything on 3.90.0 and earlier carries `context: [{}]` and is permanently unclassifiable — which is the concrete argument for proposal 1 below, not a hypothetical one.

The space-in-id 400 turned out to be a proxy bug that drops URL encoding, unrelated to this repo — tracked and fixed in Aam-Digital/ndb-setup#116 / Aam-Digital/ndb-setup#117. **This PR does not diagnose the remaining 400s**; it makes them separable so they can be.

## Changes

**Report by root cause** (`remote-pouch-database.ts`)

- Record the **request method** alongside the response. The same status means different things for a read and a write, so without it an event cannot be classified — proposal 1.
- **Name the status in the message** (`Unexpected DB response: bad request`), so each root cause becomes its own issue with a title that says what happened. A numeric status could not be used: the grouping normalization masks digits, which is precisely what merged these in the first place.
- Drop "fetch" from the wording — calling a rejected write a failed fetch is what sent the first analysis down the wrong path — proposal 3.

**Route 409 to debug** — proposal 2. The fetch layer cannot know whether a conflict cost anyone anything; whoever issued the write always can:
- remote-only session: the write came from `PouchDatabase.put()`, and `resolveConflict()` decides between merging, overwriting and failing;
- synced session: writes go to the local database, so the only 409s crossing this layer are replication's own `PUT /_local/<checkpoint>` races between concurrent tabs, which PouchDB retries internally.

**Count conflicts instead of alerting on them.** Two users editing one record is normal operation for an offline-first app, and the user whose save was rejected is already told so by the form. So conflicts now go to usage analytics (entity type and outcome only — never the document id, cf. #4174) via a `conflictReporter` hook, wired in `DatabaseFactoryService` with the lazy-`Injector` pattern `IndexeddbMigrationService` already uses to avoid the `AnalyticsService → ConfigService → EntityMapperService → DatabaseResolver → DatabaseFactoryService` cycle.

**Stop the error-level conflict noise, which the issue didn't mention.** Demoting the fetch-layer warning alone would not have met the goal: `resolveConflict` throws a `DatabaseException`, and any `entityMapper.save()` whose rejection isn't caught by a form reaches Sentry's `ErrorHandler` as an **error** — louder than the warning. That is four separate issues for one problem (one with 27 events, one still active). So conflicts are now dropped in the `beforeSend` hook, alongside the existing precedent for offline network failures. `putAll` also stopped warning on unresolvable conflicts, and stopped attaching the whole batch's document ids to the report — a repeat of the #4174 leak.

## Trade-offs and follow-ups

- **Conflict frequency is now visible only in usage analytics, not in error monitoring.** Deliberate, and stated in the code.
- Matomo is gated on `environment.production` **and** a `usage_analytics` config doc, so the count is blind on instances without analytics configured. If the number is needed for every instance, this does not deliver it.
- `remove()` sends a 409 delete straight to `DatabaseException` without going through `resolveConflict`. A behaviour change, so per AGENTS.md it belongs in its own PR.
- The renamed messages open new monitoring issues by design; AAM-DIGITAL-77H should be archived once this ships.

## Testing

Full unit suite: 2782 passed, 0 failed. `tsc`, ESLint, Prettier and `qlty check --upstream origin/master` clean.

New behaviour tests (not wording tests): 409 → debug, status named per root cause, method recorded, conflicts dropped from remote monitoring by status *and* by message, other database failures still reported, conflicts counted by entity type and outcome, and a save that still succeeds when counting the conflict throws.

Two existing tests used a conflict message as a generic example for punctuation/id normalization; they are retargeted to non-conflict messages, since conflict events no longer reach the grouping code.

## Summary by CodeRabbit

* **New Features**
  * Added usage analytics for database document-update conflicts, including merged, overwritten, and unresolved outcomes.
  * Improved HTTP error reporting with clearer status descriptions and request-method context.

* **Bug Fixes**
  * Reduced noisy warnings for expected database conflicts and handled them at debug level.
  * Prevented document identifiers from appearing in conflict and error logs.
  * Excluded expected document-update conflicts from error monitoring while continuing to report other database failures.

### DIFF
--- a/src/app/core/database/database-factory.service.spec.ts
+++ b/src/app/core/database/database-factory.service.spec.ts
@@ -1,3 +1,4 @@
+import type { Mock } from "vitest";
 import { TestBed } from "@angular/core/testing";
 import { environment } from "../../../environments/environment";
 import {
@@ -58,5 +59,82 @@ describe("DatabaseFactoryService", () => {
     environment.session_type = SessionType.synced;
     const db = service.createDatabase("test-db") as SyncedPouchDatabase;
     expect(db.adapter).toBe("indexeddb");
+  });
+
+  describe("conflict tracking", () => {
+    let eventTrack: Mock;
+
+    beforeEach(() => {
+      eventTrack = vi.fn();
+      // stubbed rather than provided: AnalyticsService is resolved lazily to
+      // break a DI cycle, and importing it here would defeat that
+      vi.spyOn<any, any>(service, "getAnalyticsService").mockResolvedValue({
+        eventTrack,
+      });
+    });
+
+    it("should count a conflict as a usage event, by entity type and outcome", async () => {
+      environment.session_type = SessionType.mock;
+      const db = service.createDatabase("test-db") as PouchDatabase;
+
+      db.conflictReporter("unresolved", "Child");
+
+      await vi.waitFor(() =>
+        expect(eventTrack).toHaveBeenCalledWith("unresolved", {
+          category: "document_update_conflict",
+          label: "Child",
+        }),
+      );
+    });
+
+    it("should also track conflicts on a remote database", async () => {
+      const db = service.createRemoteDatabase(
+        "test-remote-db",
+      ) as PouchDatabase;
+
+      expect(db.conflictReporter).toBeDefined();
+      db.conflictReporter("overwritten", "Note");
+
+      await vi.waitFor(() =>
+        expect(eventTrack).toHaveBeenCalledWith("overwritten", {
+          category: "document_update_conflict",
+          label: "Note",
+        }),
+      );
+    });
+
+    it("should not fail a save when analytics is unavailable", async () => {
+      vi.spyOn<any, any>(service, "getAnalyticsService").mockResolvedValue(
+        null,
+      );
+      environment.session_type = SessionType.mock;
+      const db = service.createDatabase("test-db") as PouchDatabase;
+
+      expect(() => db.conflictReporter("unresolved", "Child")).not.toThrow();
+      await expect(
+        (service as any).trackConflict("unresolved", "Child"),
+      ).resolves.toBeUndefined();
+    });
+
+    it("should swallow an analytics failure, so counting never breaks the save", async () => {
+      eventTrack.mockImplementation(() => {
+        throw new Error("matomo unreachable");
+      });
+
+      await expect(
+        (service as any).trackConflict("unresolved", "Child"),
+      ).resolves.toBeUndefined();
+    });
+  });
+
+  it("should resolve analytics lazily to null when it is not available, and only once", async () => {
+    // the lazy resolution is what keeps the AnalyticsService -> ConfigService ->
+    // EntityMapperService -> DatabaseResolver -> DatabaseFactoryService cycle
+    // from being closed during bootstrap
+    const first = (service as any).getAnalyticsService();
+    const second = (service as any).getAnalyticsService();
+
+    expect(second).toBe(first);
+    await expect(first).resolves.toBeNull();
   });
 });
--- a/src/app/core/database/database-factory.service.spec.ts
+++ b/src/app/core/database/database-factory.service.spec.ts
@@ -61,69 +61,25 @@ describe("DatabaseFactoryService", () => {
     expect(db.adapter).toBe("indexeddb");
   });
 
-  describe("conflict tracking", () => {
-    let eventTrack: Mock;
-
-    beforeEach(() => {
-      eventTrack = vi.fn();
+  describe("analytics access", () => {
+    // what a database reports through it is covered by the pouch-database spec;
+    // here only the wiring matters, as a database cannot inject the service
+    it("should give a database from either entry point an analytics accessor", async () => {
+      const analytics = { eventTrack: vi.fn() };
       // stubbed rather than provided: AnalyticsService is resolved lazily to
       // break a DI cycle, and importing it here would defeat that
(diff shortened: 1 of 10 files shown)


sleidig:
@coderabbitai review

sleidig:
@coderabbitai review

PR #4350, opened 2026-09-03
feat(i18n): multi-lingual texts for config-like entities

sadaf895:
- closes  #3862

## PR Checklist

_Before you request a manual PR review from someone, please go through all of this list:_

- [ ] manually tested (all) required functionality yourself and added comment with clear steps for manual testing
- [ ] reviewed the "Files changed" section of the PR briefly
- [ ] added label **"Final Review"** to trigger UI regression testing (Argos visual review)
- [ ] triggered CodeRabbit AI review by commenting **"@coderabbitai review"** (e2e tests run automatically on every push)
- [ ] marked each code review comment as resolved (or commented on it with a question, if unsure)

--> then mark the PR "Ready for Review"

---
## Manual Testing

1. - [ ] Open any record with a dropdown (e.g. a Child's **Gender**, or a Note's **Category**)
- [x] Click the wrench icon on the dropdown to open **Edit dropdown options**
- [x] Click language icon next to an option's label → enter a German text → Save → Save the dialog
- [x] Reopen language dialog on that option → **both languages are still there**
- [x] Now change something unrelated on the same dropdown (e.g. an option's **colour**) and save
- [x] Reopen dialog again → **both languages are still there**

If German disappears at either point, the save flattened it — that's a blocker.

2. Dropdown options appear in the user's language

- [x] Switch the language to **Deutsch**
- [x] The option you translated shows its **German** label
- [x] Options you did *not* translate still show English (fallback works)
- [x] Switch back to English → everything reverts

3. Reports

- [x] Admin → **Reports** → open a report → edit **Title** → language dialog → add a German text → save
- [x] Reports page → the report dropdown shows the German title
- [ ] Run the report → **Export** → the file name uses the German title, not `[object Object]`

4. Progress widget

- [x] Dashboard → progress widget →  (wrench)
- [x] The **title** and each **task label** have a language icon button
- [x] Add a German text, save, and confirm the widget shows it in German

5. The dashboard editor must not destroy translations

- [x] In **English**, go to Admin → **Edit Dashboard Widgets**
- [x] Just **drag a widget to reorder it** (change nothing else) → **Save**
- [x] Switch the language to **Deutsch**
- [x] The dashboard shortcut labels are still German.
  If they show English, the editor flattened the translations on save — that's a blocker.

6.  Admin tabs

- [x] Admin → **Configure Data Structure** → **Child** → **Details View**
- [x] Every tab shows a readable name — **no `[object Object]`** on the inactive tabs
[...]

PR #4352, opened 2026-09-03
fix: render complex field values of change history instead of [object Object]

### LINKED ISSUE #4258: Enhance change history (object object)
See screenshot below

Image

sadaf895:
- closes https://github.com/Aam-Digital/ndb-core/issues/4258

## PR Checklist

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved text display for object and array values by showing their JSON representation instead of the unhelpful `[object Object]` text.
  - Preserved existing display behavior for other value types, including null values.

- **Documentation**
  - Clarified the component’s fallback use when displaying fields without a defined data type.


sadaf895:
@coderabbitai review

Abhinegi2:
@sadaf895 did a code review + manual/functional testing on this - 

Works as intended for the normal case.

still shows [object Object] in Change History  -  for `display-configurable-enum.component.html's` 
image
use thus setting
```
"values": [
    {
      "id": "F",
      "label": {
        "en-US": "female",
        "de": "weiblich",
        "fr": "féminin"
      },
      "color": "#f42525"
    },
    {
      "id": "X",
      "label": {
        "en-US": "Non-binary/third gender",
        "de": "neutral",
        "fr": "neutre"
      },
      "color": "#1ec859"
    }
  ],
```
like the direction of something like DisplayDescriptionOnly's approach for this. @sadaf895 not fully sure how that would interact with #4264 though (to showcase conifg doc history)

sadaf895:
> @sadaf895 did a code review + manual/functional testing on this -
> 
> Works as intended for the normal case.
> 
> still shows [object Object] in Change History - for `display-configurable-enum.component.html's` image use thus setting
> 
> ```
> "values": [
>     {
>       "id": "F",
>       "label": {
>         "en-US": "female",
>         "de": "weiblich",
>         "fr": "féminin"
>       },
>       "color": "#f42525"
>     },
>     {
>       "id": "X",
>       "label": {
>         "en-US": "Non-binary/third gender",
>         "de": "neutral",
>         "fr": "neutre"
>       },
>       "color": "#1ec859"
>     }
>   ],
> ```
> 
> like the direction of something like DisplayDescriptionOnly's approach for this. @sadaf895 not fully sure how that would interact with #4264 though (to showcase conifg doc history)

Good catch, thanks for actually configuring it. But I think that one's a different path from what this PR touches. Enum fields have their own viewComponent (DisplayConfigurableEnum), so they never fall through to DisplayText, which is the only thing this fix hardens — the [object Object] is coming from {{ val.label }} in display-configurable-enum.component.html.

The underlying reason is that getEnumValues() hands out the raw values, so a per-language label map goes straight into the template. It's not really Change History specific either — with that config the details page and the list table show the same thing. I have already come across this before while working on my i18n multilingual  and have already done this bit on the Phase 2 branch for #4350, resolving inside ConfigurableEnumService.getEnumValues() rather than in the template. 

And On #4264 — I had a look and I don't think it's affected yet. The Change Log only renders field names (changedFields is a string[], rendered via app-entity-field-label), not the before/after values, so nothing reaches DisplayText there. It'd only become a problem if field-level diffs get added to that screen later.

Also, could you expand a little on the DisplayDescriptionOnly idea please?

sleidig:
@sadaf895 , can you clarify here what the `enum` changelog issue is affecting:

- normal configurable-enum values (mono-lingual, as supported on current master) are displayed properly or as [object object]?
  - those must work and should also be fixed here in this PR, if necessary
- or is the issue only appearing with the new multi-lingual config that is not released yet?
  - that's fine, ignore here for the fix of the actual production problem this issue is about. That problem should be fixed in the final branch of the i18n work only, so that there are no regressions for multi-lingual field there. Maybe make a note for the i18n PR

sadaf895:
> @sadaf895 , can you clarify here what the `enum` changelog issue is affecting:
> 
> * normal configurable-enum values (mono-lingual, as supported on current master) are displayed properly or as [object object]?
>   
>   * those must work and should also be fixed here in this PR, if necessary
> * or is the issue only appearing with the new multi-lingual config that is not released yet?
>   
>   * that's fine, ignore here for the fix of the actual production problem this issue is about. That problem should be fixed in the final branch of the i18n work only, so that there are no regressions for multi-lingual field there. Maybe make a note for the i18n PR

It's the second one — only with the new multi-lingual config. Mono-lingual enums display fine on master. So nothing to fix here — and it's already handled in #4350.

Abhinegi2:
While testing this, I downloaded the **Download Backup (.json)** from dev site and restored that backup on a fresh local setup. That’s when I encountered the `[object Object]` issue.

So, @sadaf895  if this case is already handled in another PR, then everything looks good to me

Abhinegi2 (approved):

PR #4354, opened 2026-09-04
chore(permissions): remove the legacy permission section keys

### LINKED ISSUE #295: Remove backwards compatibility for legacy permission section keys
The reserved sections of the `Config:Permissions` document were renamed `default` -> `_default` and `public` -> `_public`, so that they can no longer be shadowed by a realm role of the same name.

Backend and frontend still read the legacy spellings, and the backend moves a legacy `default` section onto `_default` when it writes the managed defaults back. That backwards compatibility exists only so code can ship before every document has been migrated, and it should be removed in the next major version.

### To remove

replication-backend:
- https://github.com/Aam-Digital/replication-backend/pull/289
- `LEGACY_DEFAULT_KEY`, `LEGACY_PUBLIC_KEY` and `LEGACY_SECTION_KEYS` in `src/permissions/rules/permission.ts`
- the deprecated `public` / `default` members of `RulesConfig`, and the legacy entries in `RESERVED_RULE_CONFIG_KEYS`
- `RulesService.normalizeSectionKeys` and its call in `setPermission`
- the legacy branch in `RulesService.writeManagedDefaults` (read `_default` only)

ndb-core:
- the legacy key fallbacks added in Aam-Digital/ndb-core#4168

### Precondition

Every instance has to run the one-off config migration that renames the keys in its permission document first, otherwise its `default` / `public` rules stop being applied.

Abhinegi2:
- closes https://github.com/Aam-Digital/replication-backend/issues/295

- part of Aam-Digital/replication-backend#295 (frontend half; backend half is Aam-Digital/replication-backend#340)

## PR Checklist

---

## What

The reserved sections of `Config:Permissions` were renamed `default` -> `_default` and `public` -> `_public` in #4168. The non-prefixed spellings stayed readable only so code could ship before every document had been migrated. All instances now have `_default` / `_public` in their CouchDB, so the compatibility layer can go.

Removed:

- `LEGACY_DEFAULT_KEY` / `LEGACY_PUBLIC_KEY`, and their entries in `RESERVED_RULE_CONFIG_KEYS` and `inheritsDefaultRules`
- the deprecated `public` / `default` members of `DatabaseRules`
- `permissions-config-migration.ts` (`migrateLegacySectionKeys`) and its spec
- its callers in `AbilityService`, `RolePermissionsService` and `PublicFormPermissionService` — the last of these no longer rewrites the permissions document just to move a legacy section across

The one-off migrations in `cli/migration/` **stay**: `oneoff-20260724-permissions-key-rename` is what brings a stored document to the underscore form, and `oneoff-20260804-permissions-key-legacy-cleanup` drops the leftovers.

Behaviour on an already-migrated document is unchanged; only the fallback path is gone.

## Precondition

**An instance that has not run the key-rename migration loses its shared `default` / `public` rules with this change** — they are simply no longer read, which means every logged-in user loses the baseline permissions.

Draft until that is confirmed for all deployments. Note that ndb-core registers a service worker, so a client can be running an older cached build than what is deployed — worth checking version telemetry rather than deployment config alone.

The frontend started reading `_default` / `_public` in **3.87.2**; anything below that is the cut-off.

This should be released together with Aam-Digital/replication-backend#340 so no deployed frontend is left reading a spelling the backend no longer writes.

## Manual Testing Steps

1. On an instance whose `Config:Permissions` already uses `_default` / `_public` (i.e. any current system), log in as a normal user: the shared baseline rules still apply — check that a record type only granted through `_default` is still accessible.
2. Open **Admin > Users > User Roles & Permissions**: `_default` and `_public` are listed first, with their rule counts unchanged from before this PR.
3. Open a role and save a change through the matrix; reopen it and confirm the change persisted and no `default` / `public` section appeared in **Edit JSON**.
4. Open a public form in a logged-out browser and submit it: anonymous access still works through `_public`.
5. In **Edit JSON**, temporarily rename `_public` to `public` and save. Reload: the public form now correctly stops working, and the section is *not* silently migrated back — this is the behaviour change, and the reason the migration is a precondition. Restore `_public` afterwards.

## Testing

- Affected suites green locally: `core/permissions`, `features/public-form`, `core/admin/admin-role-permissions` — 207 passed
- Tests that only covered the legacy-to-current fallback were removed; the ones that used `default:` / `public:` merely as a key name now use the prefixed spelling


src/app/core/permissions/ability/ability.service.spec.ts
```diff
@@ -402,7 +402,7 @@ describe("AbilityService", () => {
       ];
       const config = new Config<DatabaseRules>(
         Config.PERMISSION_KEY,
-        Object.assign({ default: defaultRules } as DatabaseRules, rules),
+        Object.assign({ _default: defaultRules } as DatabaseRules, rules),
```
sleidig: we should use the const instead of the literal `_default` everywhere in code, I think.

src/app/core/permissions/README.md
```diff
@@ -136,7 +136,7 @@ Permissions use JSON format with a role → rules mapping:
...
+- **Reserved section keys**: `_default` and `_public` are reserved section keys, not roles. The leading underscore keeps them from colliding with a realm role of the same name, and any user role that starts with `_` is ignored when resolving rules. The non-prefixed `default` / `public` spellings of older documents are no longer read: a stored document is brought to the underscore form by the `oneoff-20260724-permissions-key-rename` migration, and the leftover legacy sections are dropped by `oneoff-20260804-permissions-key-legacy-cleanup` (both in `cli/migration/`). An instance that has not run those migrations loses its shared rules, so they are a precondition for this version. `_default` and `_public` are the only allowed underscore-prefixed keys; do not create realm roles, or any other rule section, whose name starts with `_`.
```
sleidig: already verified in production. Documentation of removed legacy code is left to GitHub only, outside of the code.

```suggestion
- **Reserved section keys**: `_default` and `_public` are reserved section keys, not roles. The leading underscore keeps them from colliding with a realm role of the same name, and any user role that starts with `_` is ignored when resolving rules.
```

sleidig (approved):
Code looks good. Go ahead and merge it.

Does https://github.com/Aam-Digital/ndb-core/pull/4167 already have all references to legacy roles removed? Do resolve conflicts once after merge, @Abhinegi2

Issue #3087, opened 2024-06-14
Replace creation of User entity with a more dynamic, configurable setup step


sleidig:
The setup script currently creates a `User` entity in a very fixed way.
Since we have moved on and allow a user account to link with any entity type, this needs an update to allow such handling.

See:
https://github.com/Aam-Digital/ndb-setup/blob/13922c82dc07fdfe23186ef274e2de4d5d746094/interactive_setup.sh#L164-L172

Possible approaches:
- don't create any entity for the user during setup (and maybe accept a uuid to link to a separately created entity? or how would a user be able to link their initial account with an entity they create later?)
- make this more flexible / configurable, what entity is created
- add a UI to the frontend that asks a user to create an entity for their account after login, if the account is not linked to any yet

sleidig:
If implemented, this can also remove the initial user doc creation in the setup scripts: https://github.com/Aam-Digital/aam-cloud-infrastructure infra/aam-digital-instances/bin/couchdb-setup.ts

sleidig:
Came at this from the deployment side while planning [aam-cloud-infrastructure#34](https://github.com/Aam-Digital/aam-cloud-infrastructure/issues/34) (Helm chart for the full stack). We're now assuming **approach 3** — the frontend asks an unlinked account to create its entity after login — and treating it as a prerequisite for the chart work. Four things that aren't captured above:

### The coupling is worse than a redundant document

The comment above points at `couchdb-setup.ts`, but the cost is not just that it creates a document. Two separate systems have to agree on the ID: `bin/realm-resource.ts` sets the Keycloak attribute `exact_username: "User:<email>"`, and `bin/couchdb-setup.ts` writes `{_id: "User:<email>"}`. Neither can see the other.

That agreement breaks for any instance whose realm was **adopted** rather than created — a system migrated in, where the owner has been logging in for years under whatever `exact_username` their account already carries. Creating `User:<email>` for them does not duplicate anything; it **orphans them from their real entity**, because login then resolves to the fresh empty document.

The current guard for that is a marker string parsed out of one script's stdout and fed into the other, and it is fragile in a specific way its own comment records: none of the Pulumi Command's inputs change when the script gains a line of output, so for every instance provisioned before the marker existed it is absent and unrecoverable. Approach 3 deletes this whole mechanism — the frontend knows which account is logged in and what it is linked to, so nothing has to be inferred across two systems.

### ndb-core already owns both halves

[...]

PR #4374, opened 2026-09-09
fix: enable editing the linked user entity of a user account

### LINKED ISSUE #4373: enable editing the linked user entity of user account
Currently the linked entity of a user account is read-only and cannot be changed after initial creation. This should be enabled now.

sleidig:
## Summary

Fixes #4373. The profile record linked to a login account was fixed at creation time and read-only afterwards, with no way to fix a mislinked account short of deleting and recreating it.

- **Write-path fix (own commit):** `KeycloakUserDto.fromUserAccount` always assigned `email`/`enabled`/`emailVerified`, even as explicit `undefined` when a partial update omitted them. Spreading that onto the current user in `updateKeycloakUser` overwrote real values with `undefined`, and `attributes` was rebuilt from scratch instead of merged — so any update carrying `userEntityId` would have wiped every other Keycloak attribute (and email/enabled with it). No current caller trips this today (only `email`/`roles` are ever sent), but it made it unsafe to ever update just the linked profile.
- **Feature:** the "Profile" field in the user account admin UI is no longer disabled for existing accounts. Saving now diffs the linked profile prefix-tolerantly (so a legacy `exact_username` without a `Type:` prefix doesn't look changed on every save), enforces the one-account-per-profile rule (comparing by account id, not entity id, so re-saving an account's own current profile isn't refused), confirms before re-linking, resets sync state on a profile change, and warns an admin when they re-link their own account (their session's `entityId` won't update until next login).
- The newly-enabled field also exposes the entity picker's clear ("×") action, which could otherwise silently no-op a save; the field is now required once an account already has a linked profile, while accounts with no profile at all (a valid, intentional state) stay optional to edit.
- Also includes a small unrelated docs commit (`src/app/core/setup/README.md`) documenting the existing system-init flow, carried over from a parallel branch since it's used as research context for a follow-up PR (#3087) that depends on this one.

## Test plan

- [x] `npm run test -- --watch=false --include='**/keycloak-admin.service.spec.ts' --include='**/user-details.component.spec.ts' --include='**/user-account-action-guard.service.spec.ts'`
- [x] `npm run lint`
- [x] `npm run build`
- [x] Manual, against real Keycloak: user administration → open an account → change the linked profile → confirm; check in Keycloak that `exact_username` is the new id and email/enabled/other attributes are untouched
- [ ] Manual: verify `EntityUserComponent` re-runs its `getUser` lookup for the now-orphaned old profile record rather than showing stale account data (flagged as a follow-up decision, not changed in this PR)

## Summary by CodeRabbit

- **New Features**
  - Existing accounts can now change their linked profile.
  - Profile changes require confirmation, detect conflicts, and warn when relinking the current user’s account.
  - Changing a linked profile triggers a synchronization reset.

- **Bug Fixes**
  - Linked profiles are correctly validated, while unlinked accounts may remain without one.
  - Failed profile conflict checks now stop relinking and display an alert.
  - Updating user details preserves unrelated profile information and explicitly handles disabled accounts.
  - Linked-profile guidance and required-field messages have been updated for clarity.

### DIFF
--- a/src/app/core/user/user-admin-service/keycloak-admin.service.spec.ts
+++ b/src/app/core/user/user-admin-service/keycloak-admin.service.spec.ts
@@ -194,6 +194,66 @@ describe("KeycloakAdminService", () => {
       .flush({});
   });
 
+  it("should preserve unrelated fields and merge attributes when updating only the linked profile", async () => {
+    // given
+    const mockUser = {
+      id: "test-id",
+      email: "existing@example.com",
+      enabled: true,
+      emailVerified: true,
+      attributes: { locale: ["en"] },
+    };
+
+    // when
+    service
+      .updateUser("test-id", { userEntityId: "User:new-entity-id" })
+      .subscribe((result) => {
+        expect(result).toEqual({ userUpdated: true });
+      });
+
+    // then
+    const reqGet = httpTestingController.expectOne(`${BASE_URL}/users/test-id`);
+    expect(reqGet.request.method).toEqual("GET");
+    reqGet.flush(mockUser);
+
+    const reqPut = httpTestingController.expectOne(`${BASE_URL}/users/test-id`);
+    expect(reqPut.request.method).toEqual("PUT");
+    expect(reqPut.request.body).toEqual({
+      id: "test-id",
+      email: "existing@example.com",
+      enabled: true,
+      emailVerified: true,
+      attributes: {
+        locale: ["en"],
+        exact_username: ["User:new-entity-id"],
+      },
+    });
+    reqPut.flush({});
+  });
+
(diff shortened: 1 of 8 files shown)


sleidig:
@Abhinegi2 , I had to do this as a prerequisite for #3087, which I am actually working on. Could you take this up to test and finalize the code if necessary (so far untested, only did a sanity check of the code, which seems valid)

Abhinegi2:
@coderabbit review

Abhinegi2:
@coderabbit review

Abhinegi2:
functionality seems to be working as expected. wasn’t able to break anything, doing a code review and will also go through the coderabbit comments

Abhinegi2:
@coderabbit review

Abhinegi2 (approved):

PR #4375, opened 2026-09-09
feat: link the initial admin account to an entity during system setup

### LINKED ISSUE #3087: Replace creation of User entity with a more dynamic, configurable setup step

sleidig:
## Summary
- Adds a step to the system-init assistant, shown after a base config is imported, that lets the first user create their own profile record and links their Keycloak account to it.
- `UserEntityLinkService` resolves which entity types accept accounts (offering a picker only when there are several), reuses the type's own details-view form config, and writes the link via the account-update path fixed in #4374.
- A successful link reloads the app so the next login issues a token carrying the new `username` claim; a failed link leaves the created entity in place and tells the user an admin can link it later.
- Documents the new step in `src/app/core/setup/README.md`.

Depends on #4374 (merged).

Closes #3087.

## Test plan
- [x] Unit tests: `user-profile-setup.component.spec.ts` (15 tests), `system-init-assistant.component.spec.ts` (3 tests) — type resolution (none/one/several), the profile reaching the server before the account is linked, reload only when the link was actually written, skip/notice/warning paths.
- [x] `npm run lint`
- [x] `npm run build` (dev + prod)
- [x] e2e: ran `app.spec.ts` and `admin.spec.ts` locally against a built app — all scenarios that boot via `?useCase=all-features` pass, confirming the new step correctly skips itself on the e2e mock session and doesn't block or trigger a reload.
- [x] Manual verification against a real Keycloak instance, on a local full stack (CouchDB + Keycloak + replication-backend behind the reverse proxy), with an account holding `admin_app` + `account_manager` and no `exact_username`:
  - the step offers the single account-enabled type without a picker and builds the form from that type's own details-view config
  - after saving, the entity is in CouchDB **and** `exact_username` is set on the account, the app reloads, and the User Account screen resolves the new profile
  - with the database unreachable, the account is left unlinked: no `exact_username` is written, there is no reload, the warning is shown as a toast and kept on the final screen, and the profile syncs up by itself once the server is reachable again

## Summary by CodeRabbit

* **New Features**
  * Added an optional first-time setup step for creating and linking an administrator profile.
  * Users can select a profile type, complete its form, save it, or skip setup for later.
  * Added setup status messages, including workspace preparation, successful completion, and profile-linking notices.
  * Demo mode continues to show sample-data guidance, while standard setups display a system-ready message.

* **Documentation**
  * Added documentation covering initial system setup, profile creation, configuration sources, reset options, and demo-mode behavior.

### DIFF
--- a/src/app/core/setup/system-init-assistant/system-init-assistant.component.html
+++ b/src/app/core/setup/system-init-assistant/system-init-assistant.component.html
@@ -51,18 +51,87 @@
     </div>
   } @else if (generatingData()) {
     <p i18n>Generating sample data for this demo ...</p>
-  } @else {
-    <p i18n>
-      We have loaded some sample data into this demo system. Explore and play
-      around – you can’t break anything!<br />
-      Changes you make in this demo will <em>not</em> be saved anywhere. The
-      system is reset after you close or reload the page.
-    </p>
+  } @else if (!profileStepReady()) {
+    <p i18n>Preparing your workspace ...</p>
+  } @else if (showProfileStep()) {
+    @if (profileEntityTypes().length > 1 && !selectedProfileType()) {
+      <p i18n>
+        Choose what kind of record represents you, so that your login account
+        can be linked to it.
+      </p>
 
-    <p i18n>
-      You can get assistance by clicking the "Assistant" button on the top right
-      of the toolbar. From there you can also switch to other demo scenarios.
-    </p>
+      <mat-form-field>
+        <mat-label i18n>Profile type</mat-label>
+        <mat-select (selectionChange)="onProfileTypeSelected($event.value)">
+          @for (type of profileEntityTypes(); track type.ENTITY_TYPE) {
+            <mat-option [value]="type">{{ type.label }}</mat-option>
+          }
+        </mat-select>
+      </mat-form-field>
+    }
+
+    @if (selectedProfileType() && profileForm()) {
+      <p i18n>
+        Create your own "{{ selectedProfileType().label }}" record, so that your
+        login account is linked to it.
+      </p>
+
+      <app-entity-form
+        [entity]="profileEntity()"
+        [fieldGroups]="profileFieldGroups()"
+        [form]="profileForm()"
+      ></app-entity-form>
+    }
+
+    <div class="demo-actions">
+      @if (selectedProfileType() && profileForm()) {
+        <button
+          mat-raised-button
+          class="full-width"
+          color="accent"
+          (click)="saveProfile()"
+          [disabled]="savingProfile()"
+          i18n
+        >
+          Create profile
+        </button>
+      }
+      <button
+        mat-stroked-button
+        class="full-width"
+        (click)="skipProfileStep()"
+        [disabled]="savingProfile()"
+        i18n
+      >
+        Skip for now
+      </button>
+    </div>
+  } @else {
+    @if (environment.demo_mode) {
+      <p i18n>
+        We have loaded some sample data into this demo system. Explore and play
+        around – you can’t break anything!<br />
+        Changes you make in this demo will <em>not</em> be saved anywhere. The
+        system is reset after you close or reload the page.
+      </p>
+
+      <p i18n>
+        You can get assistance by clicking the "Assistant" button on the top
+        right of the toolbar. From there you can also switch to other demo
+        scenarios.
+      </p>
+    } @else {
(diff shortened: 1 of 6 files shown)


sleidig:
@coderabbitai review

src/app/core/setup/system-init-assistant/system-init-assistant.component.html
```diff
@@ -51,18 +51,87 @@ <h1 i18n>Welcome to Aam Digital!</h1>
...
+      </p>
+
+      <app-entity-form
+        [entity]="profileEntity()"
+        [fieldGroups]="profileFieldGroups()"
+        [form]="profileForm()"
+      ></app-entity-form>
+    }
+
+    <div class="demo-actions">
```
sleidig, line 86: needs some margin between the buttons.

Maybe that works:
```suggestion
    <div class="demo-actions gap-regular">
```

src/app/core/setup/system-init-assistant/system-init-assistant.component.ts
```diff
@@ -54,6 +75,19 @@ export class SystemInitAssistantComponent implements OnInit {
   generatingData = signal<boolean>(false);
   availableLocales = signal<ConfigurableEnumValue[]>([]);
 
+  /** Whether the "create your profile" step has been evaluated and, if applicable, is shown. */
+  profileStepReady = signal<boolean>(false);
```
sleidig, line 79: we should extract all the new code/template into its own sub-component, I think

src/app/core/setup/user-entity-link.service.ts
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * Non-UI logic for the initial-setup step that links the first admin's login account
+ * to a "profile" entity they create during setup.
+ *
+ * @see /src/app/core/setup/README.md
+ */
+@Injectable({
+  providedIn: "root",
+})
+export class UserEntityLinkService {
```
sleidig, line 20: If the user-profile init has its own sub-component (see above), maybe merging these few functions back into the component.ts for easier readability is worth it.

sleidig (commented):
A few code cleanups but overall the flow and UX seems decent to me.

However, testing on a new dev system, somehow my new user entity got linked but not saved to CouchDB. Needs testing again.

@Abhinegi2 , could you pick this up together with a final review?

src/app/core/setup/system-init-assistant/system-init-assistant.component.html
Abhinegi2: `demo-actions` has no styles anywhere (the component scss is empty), so it computes as `display: block` and `gap` has no effect on it. Used `flex-column gap-regular` instead, verified in the running app: `display: flex; gap: 14px`.

src/app/core/setup/system-init-assistant/system-init-assistant.component.ts
Abhinegi2: Extracted to `system-init-assistant/user-profile-setup/`. The parent now only holds `profileStepDone` / `profileStepNotice` and renders `<app-user-profile-setup (completed)="...">`.

src/app/core/setup/user-entity-link.service.ts
Abhinegi2: Done, the service is gone and its functions are private methods of the new component.

While moving the tests I also found the cause of the entity not reaching CouchDB: the save is local-only, and linking changes what `${user.entityId}` resolves to, so the next login interpolates different permission rules and the enforcer discards local data they no longer cover (on the legacy `idb` adapter by destroying the local database, unsynced documents included). The step now waits for `ensureSynced()` before linking, and skips linking entirely if that push fails.

Abhinegi2:
@coderabbit review

TheSlimvReal (approved):
Tested it locally and it worked well for me. Code looks finde too.

PR #4377, opened 2026-09-10
fix(import): resolve entity reference fields that allow several record types

### LINKED ISSUE #4275: Import wizard crashes with an unhandled TypeError cascade when an entity type is not registered in EntityRegistry
Found by the automated monitoring routine, window 2026-08-14 01:22 CEST → 2026-08-17 01:10 CEST. New this run.

_Client system identifiers are omitted per the standing rule; the affected instance and the exact entity key are visible on the linked Sentry issues._

## What is happening

On one production instance, the `/import` screen is unusable. Four Sentry issues fire together on that single instance and single route, all on `ndb-core@3.88.0`:

| Issue | Message | Events |
|---|---|---|
| [AAM-DIGITAL-76F](https://aam-digital.sentry.io/issues/AAM-DIGITAL-76F) | `RegistryLookupError: Requested item is not registered in EntityRegistry. Key: <custom entity type>` | 5 |
| [AAM-DIGITAL-76H](https://aam-digital.sentry.io/issues/AAM-DIGITAL-76H) | `TypeError: undefined is not an object (evaluating 'e.form.getRawValue')` | 3 |
| [AAM-DIGITAL-76G](https://aam-digital.sentry.io/issues/AAM-DIGITAL-76G) | `TypeError: undefined is not an object (evaluating '…_control.onContainerClick')` | 2 |
| [AAM-DIGITAL-76D](https://aam-digital.sentry.io/issues/AAM-DIGITAL-76D) | `DatabaseException: Document update conflict` on `/import` | 1 |

## Ordering — the registry error is the trigger

Event sequence on 2026-08-16 (UTC), all on `/import`:

```
17:59:44  RegistryLookupError
18:06:11  TypeError  (onContainerClick)
18:09:06  RegistryLookupError
18:11:33  TypeError  (getRawValue)
18:11:45  RegistryLookupError
18:11:50  TypeError  (getRawValue)
18:11:52  RegistryLookupError
18:12:01  RegistryLookupError
18:12:02  TypeError  (getRawValue)
18:15:00  TypeError  (onContainerClick)
```

The registry lookup fails first and then interleaves with the two TypeErrors across a ~15-minute retry session. The reading is that the entity type cannot be resolved, the import wizard throws before its form is constructed, and the two TypeErrors are wizard form/control code subsequently dereferencing a form that was never built.

**Confirmed:** the co-occurrence, the single-instance and single-route scope, and the ordering above.
**Inference:** that the registry failure *causes* the TypeErrors. No shared `trace_id` was found linking them, so this is ordering-based, not trace-confirmed.

## Why the entity type is probably not genuinely absent

Two events earlier the same day on the same instance suggest an admin was editing entity configuration when this started:

- `11:38:41Z` — [AAM-DIGITAL-76C](https://aam-digital.sentry.io/issues/AAM-DIGITAL-76C) `DatabaseException: Document update conflict` on the entity-list route **for that same entity type**
- `13:41:44Z` — [AAM-DIGITAL-71W](https://aam-digital.sentry.io/issues/AAM-DIGITAL-71W) `[ENTITY_SELECT] Error loading selected entity` on `/admin/entity/School`

So the entity type has a working list route, i.e. it exists in the instance's config — it is the *registry* that does not have it. That points at the registry not being (re)populated after a config edit rather than at a genuinely missing entity type. **Hypothesis, not confirmed** — settling it needs that instance's entity config and the ordering of the config write against the session that hit `/import`.

## Two separable defects

1. **The cascade.** Whatever the correct behaviour is when an entity type cannot be resolved, it should not be two unhandled `TypeError`s on `undefined`. A failed `EntityRegistry` lookup in the import flow should surface as a handled, user-visible error ("this import target is not available") and stop, rather than letting the wizard proceed to touch a form that was never created. This is the part that is safe to fix regardless of the root cause.

2. **The registry/config staleness.** If editing entity configuration can leave `EntityRegistry` without the edited type until a reload, that is worth confirming and fixing separately — it would make this reachable on any instance, not just this one.

I have not opened a PR: fix (1) needs a decision about where the guard belongs in the import wizard and what the user should see, and fix (2) cannot be confirmed from telemetry alone.

## What would settle it

- The affected instance's entity configuration — is the entity type present, and was it written shortly before 17:59Z?
- Whether `EntityRegistry` is repopulated on config change, or only on app bootstrap.

### LINKED ISSUE #4360: Import column settings break for a field that can link to several record types
**Describe the bug**

When importing a file, mapping a column to a field that links to other records only works if that field is limited to a single record type. If the field is configured to allow several types (for example a field that may point at either a child or a school), opening the column's settings fails and no matching records are offered.

**To Reproduce**

1. Go to the admin area and edit a record type.
2. Pick a field that links to other records, and configure it to allow **more than one** target record type.
3. Go to **Import** and upload a file for that record type.
4. In the column mapping step, map one of the columns to that field and open the column's settings.
5. The settings do not open / no records are offered to match against.

**Expected behavior**

The column settings should open, and matching should offer records from every record type the field allows.

**Additional context**

Reported from a production instance: [AAM-DIGITAL-78Y](https://aam-digital.sentry.io/issues/AAM-DIGITAL-78Y) and [AAM-DIGITAL-78Z](https://aam-digital.sentry.io/issues/AAM-DIGITAL-78Z) — the same problem twice, because the report varies with the order of the configured types. Found while triaging the Sentry issue list.

### LINKED ISSUE #4364: The import screen crashes while setting up column mappings
**Describe the bug**

While mapping the columns of an uploaded file to fields, the import screen crashes and the user has to start the import over from the beginning.

**To Reproduce**

1. Go to **Import** and upload a file.
2. Work through the column mapping step, opening the settings for individual columns.
3. The screen becomes unresponsive or blank and the import has to be restarted.

The exact column that triggers it is not yet identified — it appears to involve a column whose input control is not fully set up.

**Expected behavior**

Column mapping should not crash, and if one column cannot be configured, the rest of the import should still be usable.

**Additional context**

Reported from production: [AAM-DIGITAL-72P](https://aam-digital.sentry.io/issues/AAM-DIGITAL-72P) (9 occurrences), [AAM-DIGITAL-76G](https://aam-digital.sentry.io/issues/AAM-DIGITAL-76G), [AAM-DIGITAL-76H](https://aam-digital.sentry.io/issues/AAM-DIGITAL-76H) — 14 occurrences in total, all on the import screen. Found while triaging the Sentry issue list.

Abhinegi2:
- closes #4360
- closes #4364
- closes #4275

## Manual testing steps

Setup: Create System → use case "[All Features]" (demo mode) → open **Import**.

1. Upload a small CSV, e.g.: 
[todo-import.csv](https://github.com/user-attachments/files/32091755/todo-import.csv)

2. Import as → **Task**.
3. Map Columns → map the second column to **"Related Records"** (a field that allows Child/School/RecurringActivity).
   - Before fix: settings area next to the dropdown stays blank.
   - After fix: a "Match by Child / School / ..." dropdown appears with the union of matchable properties (ID, Name, ...).
4. Pick a property (e.g. Child Name) → Continue → "Review & Edit Data".
   - Before fix: screen goes fully blank.
   - After fix: a normal preview table renders with the mapped data. The child is correctly linked
   - test the same by mapping "School Title" --> the school should now also be shown correctly mapped
5. Pick a property of a complex field type (e.g. "gender") --> should offer the correct field mapping dialog button
6. Open the "Match by ..." dropdown again and check the type hints next to each property (added from review feedback):
   - each entry names the record type(s) that declare it, in smaller secondary text — e.g. `ID (internal unique ID) — Child, School, Recurring Activity`, `Name — Child, School`, `Project Number — Child`
   - a property that exists on several of the allowed types stays a single entry (matching compares it against candidates of every type), and its hint lists all of them
   - map a column to a single-type reference field instead (e.g. a field allowing only School) → no type hints are shown there, as before
7. Pick a property that only exists on one of the later allowed types (e.g. a School-only field) and check its value-transformation button still appears
   - before this round: the transformation UI silently disappeared for such properties, because the config was built from the first allowed type only

## PR Checklist

---

### DIFF
--- a/src/app/core/basic-datatypes/entity/entity-import-config/entity-import-config.component.ts
+++ b/src/app/core/basic-datatypes/entity/entity-import-config/entity-import-config.component.ts
@@ -23,6 +23,7 @@ import {
 } from "../entity.datatype";
 import { isInheritanceSourceReferenceField } from "../../../import/import-inheritance-warning.util";
 import { EntitySchemaService } from "../../../entity/schema/entity-schema.service";
+import { asArray } from "../../../../utils/asArray";
 
 /**
  * Inline import configuration component for entity reference fields.
@@ -61,30 +62,62 @@ export class EntityImportConfigComponent {
     return current?.refField ?? "";
   });
 
-  referencedEntity = computed<EntityConstructor | null>(() => {
+  /**
+   * All entity types this column's target field allows referencing.
+   * Normally a single type, but a field can allow linking to several record
+   * types, in which case the schema's `additional` is an array rather than a
+   * single type name - each is resolved individually here (an unregistered
+   * type is skipped rather than throwing, since `EntityRegistry.get()` only
+   * accepts one registered key at a time).
+   */
+  referencedEntities = computed<EntityConstructor[]>(() => {
     const col = this.col();
     const entityType = this.entityType();
-    if (!col?.propertyName || !entityType) return null;
+    if (!col?.propertyName || !entityType) return [];
 
     const fieldSchema = entityType.schema.get(col.propertyName);
-    const entityName = fieldSchema?.additional;
-    if (!entityName) return null;
-
-    return this.entityRegistry.get(entityName) ?? null;
+    return asArray(fieldSchema?.additional)
+      .filter((entityName) => this.entityRegistry.has(entityName))
+      .map((entityName) => this.entityRegistry.get(entityName));
   });
 
+  /**
+   * The first of the allowed referenced types, for UI that can only
+   * reasonably work off a single type (e.g. the referenced-entity label, the
+   * nested value-mapping component's target schema).
+   */
+  referencedEntity = computed<EntityConstructor | null>(
+    () => this.referencedEntities()[0] ?? null,
+  );
+
+  /** Display label for the "Match by ... property" dropdown, joining every allowed type's label. */
+  referencedEntityLabel = computed(() =>
+    this.referencedEntities()
+      .map((entity) => entity.label)
+      .join(" / "),
+  );
+
+  /**
+   * The union of matchable properties across every allowed referenced type,
+   * so a multi-type field still offers a full "match by" property list
+   * instead of only the first type's properties.
+   */
   availableProperties = computed(() => {
-    const entity = this.referencedEntity();
-    if (!entity) return [];
-    return [...entity.schema.entries()]
-      .filter(
-        ([prop, schema]) =>
-          (!!schema.label && !schema.isInternalField) || prop === "_id",
-      )
-      .map(([prop, schema]) => ({
-        label: schema.label ?? prop,
-        property: prop,
-      }));
+    const seenProperties = new Set<string>();
+    const properties: { label: string; property: string }[] = [];
+
+    for (const entity of this.referencedEntities()) {
+      for (const [prop, schema] of entity.schema.entries()) {
+        if (seenProperties.has(prop)) continue;
+        if (!((!!schema.label && !schema.isInternalField) || prop === "_id"))
+          continue;
+
+        seenProperties.add(prop);
+        properties.push({ label: schema.label ?? prop, property: prop });
+      }
+    }
+
+    return properties;
   });
 
   showInheritanceHint = computed(() => {
--- a/src/app/core/basic-datatypes/entity/entity-import-config/entity-import-config.component.ts
+++ b/src/app/core/basic-datatypes/entity/entity-import-config/entity-import-config.component.ts
@@ -82,18 +82,27 @@ export class EntityImportConfigComponent {
   });
 
   /**
-   * The first of the allowed referenced types, for UI that can only
-   * reasonably work off a single type (e.g. the referenced-entity label, the
-   * nested value-mapping component's target schema).
+   * The allowed type that declares the currently selected match property.
+   *
+   * With several allowed types the selected property may exist on only one of
+   * them, so the value-transformation config has to be built from that type's
+   * schema rather than from an arbitrary (e.g. the first) one.
    */
-  referencedEntity = computed<EntityConstructor | null>(
-    () => this.referencedEntities()[0] ?? null,
-  );
+  selectedRefFieldEntity = computed<EntityConstructor | null>(() => {
+    const refField = this.selectedRefField();
+    if (!refField) return null;
+
+    return (
+      this.referencedEntities().find((entity) => entity.schema.has(refField)) ??
+      null
+    );
+  });
 
   /** Display label for the "Match by ... property" dropdown, joining every allowed type's label. */
   referencedEntityLabel = computed(() =>
     this.referencedEntities()
-      .map((entity) => entity.label)
+      // a type configured without a label falls back to its id, as elsewhere
+      .map((entity) => entity.label ?? entity.ENTITY_TYPE)
       .join(" / "),
   );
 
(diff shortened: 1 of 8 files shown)


Abhinegi2:
@coderabbit review

george-neha:
Tested. Looks good to me

src/app/core/basic-datatypes/entity/entity.datatype.ts
```diff
@@ -441,13 +468,13 @@ class EntityFieldImportContext {
...
-  get refEntityCtor(): EntityConstructor | undefined {
-    return this.globalContext[`ctor_${this.schemaField.additional}`];
+  get refEntityCtors(): EntityConstructor[] | undefined {
+    return this.globalContext[`ctors_${this.schemaField.additional}`];
   }
 
-  set refEntityCtor(value: EntityConstructor) {
-    this.globalContext[`ctor_${this.schemaField.additional}`] = value;
+  set refEntityCtors(value: EntityConstructor[]) {
+    this.globalContext[`ctors_${this.schemaField.additional}`] = value;
```
sleidig: shouldn't this rather stay as before: a cache with the entities of one type stored under one key?

The `loadImportMapEntities` could then simply run for two types independently filling the caches for these (which will also reused the cache for fields that just need one or the other, probably)

src/app/core/basic-datatypes/entity/entity-import-config/entity-import-config.component.ts
```diff
@@ -61,30 +62,62 @@ export class EntityImportConfigComponent {
...
   });
 
+  /**
+   * The first of the allowed referenced types, for UI that can only
+   * reasonably work off a single type (e.g. the referenced-entity label, the
+   * nested value-mapping component's target schema).
+   */
+  referencedEntity = computed<EntityConstructor | null>(
+    () => this.referencedEntities()[0] ?? null,
+  );
```
sleidig: I am not fully sure what this implies: Will some matching then always assume we match against only the first type? Or where is this used?

Glancing over the code it may still be used in ` /** Config for the sub-field's inline component (for value transformation config) */
  subFieldInlineConfig`. But if the transformation always assumes the first type, then we can't even let users use fields from the second type, I think?

src/app/core/basic-datatypes/entity/entity-import-config/entity-import-config.component.ts
```diff
@@ -61,30 +62,62 @@ export class EntityImportConfigComponent {
...
+    const properties: { label: string; property: string }[] = [];
+
+    for (const entity of this.referencedEntities()) {
+      for (const [prop, schema] of entity.schema.entries()) {
+        if (seenProperties.has(prop)) continue;
+        if (!((!!schema.label && !schema.isInternalField) || prop === "_id"))
+          continue;
+
+        seenProperties.add(prop);
+        properties.push({ label: schema.label ?? prop, property: prop });
```
sleidig: We should show the entity type together with the field label here if it is in the multi-type mode. Otherwise users will struggle a lot finding the correct field they want.

(maybe a bit smaller in the UI, to keep it easier to visually scan)

sleidig: or does this collapse fields that have the exact same ID across multiple entity types? That is a nice feature but then maybe also needs some explanation :thinking:

Abhinegi2: right now the ui looks like this 
image

src/app/core/basic-datatypes/entity/entity.datatype.ts
Abhinegi2: switched to per-type caching

src/app/core/basic-datatypes/entity/entity-import-config/entity-import-config.component.ts
sleidig: works well :+1:

sleidig (approved):
Thanks for the cleanup. Looks good to me now!

PR #4382, opened 2026-09-11
feat: resolved relations for pdf exports

### LINKED ISSUE #4232: generate pdf files for entities with an entity reference array
_As a user I want to export a PDF listing all participants linked within an event in order to have a nicely formatted, printable participant list._

Our "Generate File" feature already allows to manage `ExportTemplate` docs that include placeholders. Placeholders refer to fields of a single entity and are replaced with actual data (e.g. name, address) by the system while generating the PDF. Carbon also supports array properties that are iterated over to build the final report [docs](https://carbone.io/documentation/design/repetitions/with-arrays.html). Especially for the case with an attendance list, this is not sufficient yet. The entity which is currently sent to carbon would only contain an array of IDs, to create a proper report it should be possible to print the actual name of the entity with the ID that is referenced. This means somehow the relationships need to be resolved before providing the data to carbon.

---

## Implementation Approach

As a first approach, the first level of relations should all be resolved on the actual entity before sending the data to carbon. This will allow users to access properties of related entities directly in the report.

TheSlimvReal:
- closes #4232 

## PR Checklist

---

## Summary by CodeRabbit

* **New Features**
  * Template exports now support linked entity data in both single-record and batch rendering.
  * Batch exports provide consistent results for ZIP and combined-output modes.
  * Export filenames now use a fallback when response metadata is unavailable or invalid.

* **Bug Fixes**
  * Improved handling of missing, invalid, or unavailable linked records during template rendering.

* **Refactor**
  * Single and batch template rendering now share consistent processing and response handling.

### DIFF
--- a/src/app/core/entity/entity-relation-resolver/entity-relation-resolver.service.ts
+++ b/src/app/core/entity/entity-relation-resolver/entity-relation-resolver.service.ts
@@ -1,0 +1,215 @@
+import { inject, Injectable } from "@angular/core";
+import { Entity } from "../model/entity";
+import { EntityMapperService } from "../entity-mapper/entity-mapper.service";
+import { EntitySchema } from "../schema/entity-schema";
+import { EntitySchemaField } from "../schema/entity-schema-field";
+import { EntityDatatype } from "../../basic-datatypes/entity/entity.datatype";
+import { Logging } from "../../logging/logging.service";
+
+/**
+ * A related entity that could not be found in the database is represented by
+ * this minimal stub instead, so that templates can still access its id.
+ */
+export interface UnresolvedEntityStub {
+  _id: string;
+}
+
+/**
+ * Resolve entity-reference fields (`dataType: "entity"`) within a record's data,
+ * replacing the plain id string(s) with the actual referenced {@link Entity} objects.
+ *
+ * This is used to give file/PDF templates access to the details of linked records
+ * (e.g. a Note's related Child) instead of only their id.
+ *
+ * Only the relations found directly on the given data (including relations nested
+ * inside embedded objects/arrays that have their own schema, e.g. `AttendanceItem`)
+ * are resolved. The relations *of* a newly resolved entity are deliberately left
+ * untouched, to avoid pulling in an ever-growing tree of linked records.
+ *
+ * The input data is never modified - resolved data is always built as a new,
+ * deep copy so that entity objects still in use elsewhere in the app remain unaffected.
+ *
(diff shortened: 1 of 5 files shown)


TheSlimvReal:
This has been deployed to https://dev.aam-digital.net`
To test:

1. go to a [event details](https://dev.aam-digital.net/c/eventnote/00575188-7bb8-461f-878f-73b5a97def9d)
2. 3 dot menu -> generate file
3. Select template "note pdf"
4. Generate file

The created pdf uses information of multiple relations: a child's name and gender, iterated over (this sometimes shows `[object object]`, which is due to a wrong format in the current data, not a problem for this PR), the author's name, in case of multiple, separated by a comma. The format of the template can be found here https://dev.aam-digital.net/c/admin/template-export/02a0d226-d9fc-47f1-8cb6-9e0b612fd009

TheSlimvReal:
@coderabbitai review

Abhinegi2:
I did functional testing, and everything seems to be working fine

src/app/core/entity/entity-relation-resolver/entity-relation-resolver.service.ts
```diff
@@ -0,0 +1,218 @@
...
+    }
+    return undefined;
+  }
+
+  /**
+   * Load all entities of the given type, indexed by id.
+   * Memoized in `loadedEntitiesByType` so repeated ids/records of the same
+   * type only trigger a single `loadType` database request.
+   */
+  private loadAllOfType(
```
Abhinegi2: this loads all records of the referenced entity type, even if we only need to resolve 2-3 ids. for an app with a large Child/User collection, that means every PDF export pulls the whole table into memory.

`EntityDatatype.loadRelatedEntitiesToString() `(used for csv export) does it differently - it loads only the specific ids needed, one by one. Could we do the same here instead of loading everything?

If loading everything is intentional (e.g. batch exports usually need most records anyway, so one bulk read beats many small ones), a short comment explaining that would help.

Abhinegi2 (commented):
@TheSlimvReal  code already looks good. I just have one thought/question,  could you check if this makes sense?

src/app/core/entity/entity-relation-resolver/entity-relation-resolver.service.ts
TheSlimvReal: We did it here this way because having that many parallel requests often is much slower than a single, bigger one. If it turns out to be an issue with the transferred data we might have to switch to one request per ID but this will be checked in production.

I added a clarification comment.

Abhinegi2 (approved):

PR #4387, opened 2026-09-14
fix: skip an incompletely configured field instead of failing the whole form

### LINKED ISSUE #4365: One incompletely configured field stops a whole form from opening
**Describe the bug**

If a single field in a form's configuration is incomplete, the entire form fails to build instead of skipping that one field. The user cannot open or edit the record at all.

**To Reproduce**

1. In the admin area, edit a form or a record type's details view.
2. Leave one of the fields without a field name selected (or otherwise incompletely configured) and save the configuration.
3. Open a record of that type.
4. The form does not appear.

**Expected behavior**

A field that cannot be built should be left out with a warning for administrators, and the rest of the form should still open so users can keep working.

**Additional context**

Reported from production: [AAM-DIGITAL-6YP](https://aam-digital.sentry.io/issues/AAM-DIGITAL-6YP) — 11 occurrences, last seen a few days ago. Found while triaging the Sentry issue list.

Abhinegi2:
- closes #4365

## Summary

One incompletely configured field in a form's or table's column config (e.g. `null`/`undefined`, or an object with no `id`) threw inside the array-building step and aborted the *whole* form/table — not just that one field. This can't be produced through the normal admin UI (the field editor already bails when an added row has no `id` selected), but is reachable via a direct config JSON edit (Admin > Edit Application Configuration (JSON)) or a raw CouchDB document edit.

- `EntityFormService.createEntityForm` now builds fields one at a time; a field that throws is logged (`Logging.error`, static message + context) and skipped instead of failing the whole form.
- `EntitiesTableComponent._customColumns` gets the same treatment for table columns.
- `EntityFormComponent.filterFieldGroupsByPermissions` (runs on the raw field config *before* `createEntityForm`) also excludes a field with no resolvable id, instead of crashing on `field.id`.
- `EntitiesTableComponent._columnsToDisplay` handles a different config shape — a `columnGroup`'s `columns: string[]` (which column ids to show, e.g. from `EntityListConfig.columnGroups`) — validating each id against the table's known columns instead of trusting it blindly. A `null`/non-string entry or a dangling id that matches no real column is now skipped and logged, rather than throwing or breaking mat-table's `displayedColumns`.
- `EntitiesTableComponent._customColumns` also now catches a malformed column when the table has no `entityType` bound — that path uses `toFormFieldConfig`, which (unlike `extendFormFieldConfig`) never throws, so a `null`/id-less entry was previously slipping through uncaught and breaking `_columnsToDisplay`/`_allColumns` downstream.

## PR Checklist

---

### Manual testing steps

1. Open a record's details view (e.g. any Child or School record) — confirm it loads normally.
2. Go to **Admin > Edit Application Configuration (JSON)**.
3. Find the entity type's `panels`/form field config (or a table's `columns` config) and insert an invalid entry into a field/column array, e.g. `null` or `{}` with no `id`, alongside the existing valid entries. Save.
4. Open a record of that entity type again (or the list view, for a table column config change).
   - **Before this fix:** the form/table fails to render at all (blank / stuck), console shows `Could not create form config for undefined: TypeError: ...`.
   - **After this fix:** the form/table still renders with all the *other* valid fields/columns present; only the broken one is missing. An error is logged (visible in Sentry in production) so admins can find and fix the bad config entry.
5. For the columnGroup case specifically: put `null` or a made-up/nonexistent column id into a list view's `columnGroups.groups[].columns` array and reload the list — the table should still render with its other columns, skipping only the bad entry.
6. Revert the config JSON change from steps 3 and 5.

## Summary by CodeRabbit

* **Bug Fixes**
  * Forms now continue loading when individual field configurations are incomplete or invalid.
  * Invalid fields are skipped while valid fields remain available.
  * Entity tables now ignore malformed, unknown, or reserved column entries instead of failing to render.
  * Custom column errors no longer prevent the remaining columns from displaying.
  * Errors are logged when invalid field or column configurations are encountered.


TheSlimvReal (commented):
Code wise it looks good.
I am not sure what the full extend of this fix should be. I tried a few different things: 

List:
- Set `null` or `"wrong"` in `columns` of a `columngGroup` -> still fails
- Set `null` or `{}` in `columns` -> works now

Details:
- Sett `null` in `fields` of a `Form` -> still breaks
- Set `"wrong"` or `{}` in fields of a `Form` -> worked before and works now

Especially for the details view I am not sure what has actually changed with this fix. For the list view I guess it depends on how many problematic configurations should be covered. I am not sure what the most common issues there are. I guess what should be covered is deleting a property completely, while it is still mentioned in a list. Then it should not break. Which I think is currently the case. This can only be tested on a deployed setup because the config is built only once and it needs a reload to correctly re-built the pages.

Abhinegi2:
@coderabbit review

TheSlimvReal (approved):
Nice, works well now

PR #4388, opened 2026-09-14
feat: support pdf exports with lists of entities on top-level

### LINKED ISSUE #4379: Support pdf exports with lists of entities on top-level
_As a project manager I want to export a list of people in a single pdf in order to create summary reports._

Currently, when selecting multiple entities using the bulk docs it is only possible to have templates that are executed for each record individually. To make more flexible reports possible, our system should also allow to access all records in a single report template using [carbons array syntax](https://carbone.io/documentation/design/repetitions/with-arrays.html). When saving the `TemplateExport` entity it should also store the information whether it is a report that requires a top-level array. If this is the case, the frontend should only show this report in the bulk-action mode and if selected, not show the options whether a `.zip` or a combined `.pdf` should be generated but instead only create that one report. The report should not be available in the entity details view.

## Implementation Approach

- [ ] Extend `TemplateExport` entity with new property for `arrayReport: boolean` 
- [ ] If `arrayReport === true` report is only shown in bulk selection
- [ ] If report with `arrayReport === true` is selected, do not show the options whether a `.zip` or a combined `.pdf` should be generated
- [ ] Make sure the backend/carbon correctly receives the data so that it is a top-level array instead of sending each entity as a single doc

TheSlimvReal:
- closes #4379

## Summary

`TemplateExport` gets a new `arrayReport: boolean` field. Enable this when a template uses [carbone's array/repetition syntax](https://carbone.io/documentation/design/repetitions/with-arrays.html) to list all selected records within a single generated document (e.g. one continuous table of multiple people), instead of generating one file per record.

- Such templates are only offered in the file-generation dialog when generating for **multiple** selected records at once (bulk action) — never for a single record's details/popup view, and never for a bulk selection of exactly one record (treated the same as single-entity mode).
- When an `arrayReport` template is selected, the "zip vs. combined PDF" choice is replaced with a short note explaining the records are combined into one report.
- Generating the file calls the existing single-record render endpoint (`POST /v1/export/render/:templateId`) with the **whole selected array** as `data`, so carbone receives one top-level array in a single render request — rather than the existing zip/combined batch path (`/render-batch`), which renders the template once per record and merges/zips the results afterward.

## Testing

This has been deployed to https://dev.aam-digital.net
To test:

1. go to the [children list](https://dev.aam-digital.net/c/child)
2. 3 dot menu -> bulk action
3. Select at least 2 children
4. Action -> 'Generate File'
5. Select `List Report'
6. Click 'Generate File'

This creates a pdf export containing a list with the names, project numbers and genders for all selected children.

## Summary by CodeRabbit

* **New Features**
  * Added support for array-based report templates that combine multiple selected records into one PDF.
  * Bulk exports using array reports now download as a single PDF, while other export options remain unchanged.
  * Added an “array report” setting to template configuration.
  * Array-report templates are available only when multiple records are selected.


TheSlimvReal:
@coderabbitai review

Abhinegi2 (approved):
Looks good to me , nice 👍

Issue #16, opened 2016-06-14
Open Source License


sleidig:
The project doesn't have any license at the moment. We agreed to make this Open Source but there still are a few choices about the specific license to pick.

Helpful resources:
- Comparison of licenses with main points summarized: [choosealicense.com](http://choosealicense.com/)
- Overview and discussion of different aspects: [Guide to choosing an open-source license](https://opensource.com/law/13/1/which-open-source-software-license-should-i-use)

Other Open Source projects are using:
- Angular 2: _MIT License_
- PouchDB: _Apache License_
- Drupal (CMS): _GPL_
- Wordpress: _GPL_
- LibreOffice: _Mozilla Public License_

The libaries we are using (Angular, PouchDB) are licensed permissively so we don't have limitations on our license here.

sebastian-leidig:
It comes down to the choice between "strict" Open Source (GPL) or also allowing our code to be used in projects with a different license, so it can be picked up more easily (e.g. MIT license).

fschoell:
Another thought that crossed my mind just now: We could use MIT for the ndb-core project and GPL for the actual implementations. This way we would be able to create a proprietary project based on ndb-core if we ever wanted to, but nobody could "steal" our actual projects.

sleidig:
The decision about a license won't get easier after more and more people joined the team again. I only go by simple heuristics here but:
As our software is somewhat like a CMS (Drupal, Wordpress) and I am fond of "real" Open Source, I suggest we just go with GPL.

What do you say? @fschoell @ojauch ?

fschoell:
Agree. The GPL prevents us from running into license incompatibilities with 3rd party libraries, too.

sleidig:
closed through PR #57

PR #4391, opened 2026-09-15
refactor(change-history): render the Change Log with the generic entity list

### LINKED ISSUE #4039: Change Log: UI to view all changes across system

Abhinegi2:
Refactors the Change Log onto the generic entity list, as discussed in review on #4264. Targets that PR's branch, so its diff shows only what changed since the last review.

## Manual Testing Steps

Needs the replication-backend with change logging switched on — it cannot be tested on the PR demo deployment, which has no backend.

**See the log**

1. Sign in as an app-admin user, then make a few changes so the log has content: edit a child's phone number, create a note, delete a record.
2. Go to **Admin → Admin Overview → Export and Backups → Change Log**.
3. The table lists changes newest first with six columns: when (with "2 days ago" underneath), which record, its type, a coloured action tag, who made it, and the changed fields as chips.
4. The record you deleted still appears, marked "Deleted", with "record removed" instead of chips.
5. Record names are real names, not ids. A type that configures no name shows "&lt;Type&gt; Record"; a record that no longer exists shows a greyed-out "not available".
6. Under each record name is a small grey id. Click it: it copies and confirms with "Record ID copied", and must **not** open the row's dialog. Clicking anywhere else in the row does open it.

**Paging**

7. Set "Items per page" to 10 and page forward and back. The same rows appear each time, and the range label reads "1 - 10 of 10+" until the last page.
8. On the last page the "next" arrow is greyed out — you should never reach a page saying "No changes recorded".

**Filters**

9. Pick a record type, e.g. "Child". Only that type remains and the list returns to the first page.
10. Click the small **x** on the filter. All types come back, and the dropdown does not pop open when you click the x.
11. Pick "Deleted" under Action — only deletions remain. Pick a name under "Changed by" — only that person's changes remain.
12. In "Date range", choose "Last month" from the calendar icon. Only changes from last month remain. Typing a start and end date directly works too.
13. Paste a record id (copied in step 6) into "Related record". You see changes involving that record, including changes to *other* records that reference it. Record type, action and changed-by grey out with a tooltip; the date range stays usable.

**Entry points**

14. Open an entity list (e.g. Children), three-dot menu → **Change Log**. It opens with that list's record type already filtered.
15. Open a record's details page, three-dot menu → **View change log**. The dialog shows that record's full history, with its id under the name.

**Off states**

16. Sign in as a non-admin and open the route directly — you should not get in.
17. On a system with change logging switched off, the page shows "This feature is currently not enabled for your system" instead of the table.

---

- closes #4039

## PR Checklist

## Summary by CodeRabbit

* **New Features**
  * Change history now uses a faster, filterable table with date, action, record type, author, and related-record filtering.
  * Audit entries show changed fields, timestamps, authors, and record details—including deleted records.
  * Related-record history and cursor-based pagination are supported.
* **Permissions**
  * Viewing change history now requires explicit audit-record read permission in addition to administrator access.
* **Bug Fixes**
  * Archive controls are no longer shown for records that cannot be archived.
  * Record IDs now provide a clearer indication when clickable.


Abhinegi2:
@coderabbit review

src/assets/base-configs/basic/Config_Permissions.json
```diff
@@ -44,6 +44,10 @@
       {
         "subject": "all",
         "action": "manage"
+      },
+      {
+        "subject": "AuditRecord",
+        "action": "read"
```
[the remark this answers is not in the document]
Abhinegi2: Agreed the exposure is real, but I don't think the inverted rule is safe as written, and I'd rather not land it in this PR.

**Why the deny is risky.** `AbilityService.getRulesForUser` builds the rule list by pushing `_default` first, then each role's section in the order the roles appear on the session:

```ts
sessionInfo.roles.forEach((role) => {
  ...
  rawUserRules.push(...rulesForRole);
});
```

CASL resolves by last-match-wins, so the outcome depends on role order in the token. A user holding both roles gets:

- `["user_app", "admin_app"]` → deny, then admin's allow → **access** ✅
- `["admin_app", "user_app"]` → admin's allow, then deny → **no access** ❌

Role order comes from the JWT and isn't guaranteed, so an admin who also carries `user_app` could silently lose the change log on some deployments and not others. That's a worse failure than the one being fixed, and a hard one to diagnose.

Abhinegi2:
@coderabbit review

src/app/core/common-components/entities-table/entities-table.component.ts
```diff
@@ -105,6 +105,14 @@ export class EntitiesTableComponent<T extends Entity>
...
   editable = input<boolean>(true);
   selectable = input<boolean>(false);
+  /**
+   * Whether records of this type can be archived.
+   *
+   * Turn this off for types that have no `inactive` flag: the archived
+   * condition would be added to every database query as an `$or` that no index
+   * can serve, and the toggle would offer a filter that can never match.
+   */
+  supportsArchiving = input<boolean>(true);
```
TheSlimvReal, line 115: There is an input signal `showInactive`. To make it consistent I would align the wording
```suggestion
  showInactiveToggle = input<boolean>(true);
```

src/app/core/common-components/entities-table/entities-table.component.ts
```diff
@@ -151,6 +159,9 @@ export class EntitiesTableComponent<T extends Entity>
   // --- Filtering (stateless derivation) ---
   readonly effectiveFilter = computed<DataFilter<T>>(() => {
     const nextFilter = { ...this.filter() };
+    if (!this.supportsArchiving()) {
+      return nextFilter;
+    }
```
TheSlimvReal, line 164:
```suggestion
```
If `showInactive` is set to `true` in the config, this code is not needed

src/app/core/database/database-resolver.service.spec.ts
```diff
@@ -153,6 +164,82 @@ describe("DatabaseResolverService", () => {
...
+      // sessionType is already mock. The data exists only on the server, so a
+      // local database could not answer for it - and this is the address the
+      // requests have to go to for a test or a proxy to intercept them.
+      vi.spyOn(factory, "createRemoteDatabase");
+
+      const db = service.getDatabase(REMOTE_ONLY_DB);
+
+      expect(factory.createRemoteDatabase).toHaveBeenCalledWith(REMOTE_ONLY_DB);
+      expect(db.isInitialized()).toBe(true);
+    });
```
TheSlimvReal, line 222: Not really sure what this tests as compared to the previous test `should hand out an already initialized remote database`. The wording is very different for the fact that the previous one just has one more line of setup.

src/app/core/database/database-resolver.service.ts
```diff
@@ -70,14 +99,22 @@ export class DatabaseResolverService {
   }
 
   async resetDatabases() {
-    for (const db of this.databases.values()) {
+    for (const [dbName, db] of this.databases.entries()) {
+      if (this.remoteOnlyDbNames.has(dbName)) {
```
TheSlimvReal, line 103: Is there a reason why you don't rely on `db instanceof RemotePouchDatabase`? This seems more in-line with the pattern used for other functions (e.g. `SyncedPouchDatabase`)

src/app/core/entity-list/entity-list/entity-list.component.ts
```diff
@@ -122,13 +125,30 @@ export class EntityListComponent<T extends Entity> implements OnInit {
...
+   * Recomputed on a rule update as well as on a new session: the rules load
+   * asynchronously, so a session that arrives first would otherwise leave this
+   * evaluated against no rules at all, and never re-evaluated.
    */
   readonly canViewChangeHistory = toSignal(
-    this.sessionSubject.pipe(
-      map((session) => session?.roles?.includes(ADMIN_APP_ROLE) ?? false),
+    combineLatest([
+      this.sessionSubject,
+      this.abilityUpdated.pipe(startWith(undefined)),
```
TheSlimvReal, line 145: I feel like a more general pattern for this issue would be useful. Already the next property (`canImport`) also uses the ability to check for create permissions but there it only waits for the ability to be initialised. A uniform approach would be better.

src/app/core/permissions/permission-types.ts
```diff
@@ -52,6 +52,15 @@ export const RESERVED_ROLE_PREFIX = "_";
...
 export const ADMIN_APP_ROLE = "admin_app";
 
+/**
+ * The CASL subject that audit records are classified under.
+ *
+ * The backend derives it from each audit document's id prefix, so it is
+ * all-or-nothing across record types: whoever may read it sees which records
+ * changed and when, including records they may not open themselves.
+ */
+export const AUDIT_RECORD_SUBJECT = "AuditRecord";
```
TheSlimvReal, line 62: Why do we need this constant and don't just use `AuditRecord.ENTITY_TYPE`?

src/app/features/change-history/change-history-changed-fields/change-history-changed-fields.component.ts
```diff
@@ -0,0 +1,57 @@
...
+})
+export class ChangeHistoryChangedFieldsComponent extends ViewDirective<
+  string[]
+> {
+  private readonly record = computed(() => this.entity() as AuditRecord);
+
+  readonly fields = computed(() => this.value() ?? []);
+  readonly action = computed(() => this.record()?.action);
+  /** the changed record's type, which the field labels are resolved against */
+  readonly recordType = computed(() => this.record()?.recordType);
```
TheSlimvReal, line 56: I think that makes it more clear what is happening
```suggestion
  override entity = input<AuditRecord>();

  readonly fields = computed(() => this.value() ?? []);
  readonly action = computed(() => this.entity()?.action);
  /** the changed record's type, which the field labels are resolved against */
  readonly recordType = computed(() => this.entity()?.recordType);
```

src/app/features/change-history/change-history-list/change-history-list.component.ts
```diff
@@ -197,61 +202,26 @@ export class ChangeHistoryListComponent {
...
-              : this.entityTypeFilter(),
-            changedBy: this.otherFiltersDisabled()
-              ? undefined
-              : this.changedByFilter(),
-            action: this.otherFiltersDisabled()
-              ? undefined
-              : this.actionFilter(),
-            relatedEntityId: this.relatedEntityFilter(),
+  readonly filter = computed(() =>
+    buildAuditFilter(
```
TheSlimvReal, line 211: Is there a reason why this custom filter logic is implemented instead of using the existing ones built with the `FilterGeneratiorService` and maybe creating a new one for the `EntityType`? Is there any super-special logic that does not apply to other lists?
If the `operation: { $ne: BASELINE_OPERATION }` is an issue, I guess the `EntityListConfig` could be extended with a `fixedFilter` property, that is always applied. Otherwise, why not just exclude this already in the index, if it should never be shown anyways?

src/app/features/change-history/change-history-list/change-history-list.component.ts
```diff
@@ -267,145 +237,104 @@ export class ChangeHistoryListComponent {
...
 
   /**
    * Whether this row can be opened as a record history. Requires a known record
    * type, which an old audit record of a type that is no longer registered does
    * not have (the same case the record-type column falls back for).
    */
-  canOpenHistory(entry: ChangeHistoryEntry): boolean {
-    return !!entry.entityId && this.entityRegistry.has(entry.entityType);
+  canOpenHistory(record: AuditRecord): boolean {
+    return !!record.record && this.entityRegistry.has(record.recordType);
```
TheSlimvReal, line 298: Is it really that important that the entity type still exists? For the loading of the records, all that is needed is the ID of the entity which we can know if we have the audit record. The other points where the entity object is used is the `EntityBlockComponent` for displaying it and the `created` and `updated` block which are already both optional. So I don't really see a big reason why we should not offer the popup also for entity types that don't exist anymore.

src/app/features/change-history/change-history-list/change-history-list.component.ts
```diff
@@ -267,145 +237,104 @@ export class ChangeHistoryListComponent {
...
+      return await this.entityMapper.load(record.recordType, record.record);
     } catch (err) {
       Logging.debug(
         "change log: record not available, showing its history only",
-        entry.entityId,
+        record.record,
         err,
       );
-      return new (this.entityRegistry.get(entry.entityType))(entry.entityId);
+      return new (this.entityRegistry.get(record.recordType))(record.record);
```
TheSlimvReal, line 337: This could also be the fallback if the entity type is unknown. Just use the `Entity` constructor instead.

src/app/features/change-history/display-audit-user/display-audit-user.component.ts
```diff
@@ -0,0 +1,51 @@
+import { ChangeDetectionStrategy, Component, computed } from "@angular/core";
+import { ViewDirective } from "../../../core/entity/default-datatype/view.directive";
+import { DynamicComponent } from "../../../core/config/dynamic-components/dynamic-component.decorator";
+import { EntityBlockComponent } from "../../../core/basic-datatypes/entity/entity-block/entity-block.component";
+
+/** the author as the backend records it */
+export interface AuditUser {
```
TheSlimvReal, line 7: Let's move this interface to the `AutidRecord` and already use it there. It might have other uses in the future.

src/app/features/change-history/display-audit-user/display-audit-user.component.ts
```diff
@@ -0,0 +1,51 @@
...
+  id?: string;
+  name?: string;
+  roles?: string[];
+}
+
+/**
+ * The author as an entity id, when one was recorded.
+ * An entity id is type-prefixed, which a plain username is not.
+ */
+export function authorEntityId(author: string): string | undefined {
```
TheSlimvReal, line 17: A bit confusing function name. Maybe `ensureValidEntityId`.

src/app/features/change-history/model/audit-record.ts
```diff
@@ -0,0 +1,97 @@
...
+   * For create/update/delete a jsondiffpatch delta, for baseline the full
+   * previous document. Passed through untyped: its keys are field names of
+   * whichever entity type was changed, so no schema of this type can describe
+   * them.
+   */
+  @DatabaseField() diff?: any;
+
+  /** the changed record's id, e.g. `Child:123` */
+  get record(): string {
+    return this.getId(true).replace(AUDIT_ID_SUFFIX, "");
```
TheSlimvReal, line 69: It's a bit difficult to understand what this code actually does. Either make the static name more descriptive (e.g. `AUDIT_TIMESTAMP_SUFFIX_PATTERN` or work with a split on `:` which is a bit easier to follow (e.g. we want the string between the first and the third `:`).

src/app/features/change-history/model/audit-record.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  /** the changed record's type, e.g. `Child` */
+  get recordType(): string {
+    return Entity.extractTypeFromId(this.record);
+  }
+
+  /** the displayed action, which unlike {@link operation} is past tense */
+  get action(): ChangeAction {
+    return OPERATION_TO_ACTION[this.operation];
```
TheSlimvReal, line 79: Could we use a configurable enum here to also make it translatable? This would also remove a lot of the back-and-forth mapping between the display and the database value

src/app/features/change-history/model/audit-record.ts
```diff
@@ -0,0 +1,97 @@
...
+  /** source database name, e.g. `app` */
+  @DatabaseField() database: string;
+
+  /**
+   * For create/update/delete a jsondiffpatch delta, for baseline the full
+   * previous document. Passed through untyped: its keys are field names of
+   * whichever entity type was changed, so no schema of this type can describe
+   * them.
+   */
+  @DatabaseField() diff?: any;
```
TheSlimvReal, line 65: Can you extend the doc to actually describe all possible states. How does the diff look on `create`, `update`, `delete`, `baseline`. Reading through the code I see a lot of different expected formats and it would be good to have one point of reference here.

src/app/features/change-history/change-history-list/change-history-list.component.ts
```diff
@@ -197,61 +202,26 @@ export class ChangeHistoryListComponent {
...
-      } catch (err) {
-        Logging.error("Failed to load the change log", err);
-        return { entries: [], hasMore: false, failed: true };
-      }
-    },
-  });
+    ),
+  );
 
   private readonly authorsResource = resource({
```
TheSlimvReal, line 226: I think this is a very odd approach. It loads many documents redundantly (it loads 1k docs ignoring pagination) and still doesn't ensure a full list of authors to be available (still only the most recent). Is this really offering more than using a `EntityFilter` on the type `User` and maybe extending it with a option `other` which uses `$not: { $includes: [<list of users>] } }`. This would load much less docs and at least provide deterministic results.

src/app/features/change-history/model/audit-record.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  it("should derive the changed record and its type from the id", () => {
+    const record = load(rawDoc);
+
+    expect(record.record).toBe("Child:123");
+    expect(record.recordType).toBe("Child");
+  });
+
+  it("should derive a changed record whose own id contains colons", () => {
+    // a User's id is its username, which may contain anything
```
TheSlimvReal, line 72: This actually sounds dangerous 😅 are we sure we can support this kind of ID everywhere in the app? Is this a known issue?

src/app/features/change-history/audit-data-source.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+  await fetchPage(first.bookmark);
+
+  expect(first.bookmark).toBe("5");
+  expect(lastQuery().skip).toBe(5);
+  // a repeated range start would return the same rows again
+  expect(lastQuery().startkey).toBeUndefined();
+  expect(lastQuery().endkey).toEqual(["User:1"]);
+});
+
+it("should drop rows the backend's permission filter emptied", async () => {
```
TheSlimvReal, line 86: When would this happen?

src/app/features/change-history/audit-data-source.ts
```diff
@@ -0,0 +1,113 @@
...
+  private readonly service = inject(ChangeHistoryService);
+  private readonly schemaService = inject(EntitySchemaService);
+
+  /** when set, the log answers "what involved this record?" instead */
+  readonly relatedRecord = signal<RelatedRecordQuery | undefined>(undefined);
+
+  /**
+   * Switch the log between its two questions. Everything paged through so far
+   * answered the previous one, so it is discarded.
+   */
```
TheSlimvReal, line 44: I don't understand this comment. What are the two questions? Does this refer to the two modes (load all `AuditRecords` vs load `AuditRecords` related to specific entity?)

src/app/features/change-history/audit-data-source.ts
```diff
@@ -0,0 +1,113 @@
...
+    const skip = page.bookmark ? Number(page.bookmark) : 0;
+    const query: Record<string, unknown> = {
+      // a one-element array sorts before every `[id, ...]`, so an unbounded
+      // range ends before the oldest record of this id
+      endkey: related.from
+        ? [related.recordId, related.from.toISOString()]
+        : [related.recordId],
+      descending: true,
+      // the backend only permission-filters - and only pages - a view response
+      // when the documents are included
```
TheSlimvReal, line 84: I find this comments (I saw it at other places too) a bit confusing. This code is loading the docs - of course  `include_docs: true` is added, otherwise it would not be loading anything at all. So this is not just added because it needs the permissions filtering.

src/app/features/change-history/audit-filter.ts
```diff
@@ -0,0 +1,73 @@
...
+    timestamp.$gte = filters.from.toISOString();
+  }
+  if (filters.to) {
+    timestamp.$lte = endOfDay(filters.to);
+  }
+
+  const selector: Record<string, unknown> = {
+    // the sort field must be constrained for the index to be usable; a date
+    // bound already does that, otherwise match any record that has a timestamp
+    timestamp: Object.keys(timestamp).length > 0 ? timestamp : { $gt: null },
```
TheSlimvReal, line 38: Why is this actually the case?

src/app/features/change-history/change-history.service.ts
```diff
@@ -3,80 +3,53 @@ import { HttpClient } from "@angular/common/http";
...
-  entries: ChangeHistoryEntry[];
-  /** whether at least one further page exists after this one */
-  hasMore: boolean;
-}
+/**
+ * How many of the most recent audit records the author filter samples. The
+ * audit database has no index of its authors, so the options can only come from
+ * a bounded look at recent data.
+ */
+const AUTHOR_SAMPLE_SIZE = 1000;
```
TheSlimvReal, line 32: If it's really important, maybe an index can actually be added that only returns the unique author names. Not fully sure if it is possible with CouchDB. Probably with a map-reduce view.

src/app/utils/findable-memory-pouch-database.ts
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * An in-memory database that also supports {@link PouchDatabase.find}, so that
+ * tests can exercise an entity type end-to-end through
+ * `EntityMapperService.findType` -> `Database.find`.
+ *
+ * PouchDB's local Mango engine has no bookmark cursor, so the opaque bookmark
+ * is the positional offset of the next unseen document. Callers only pass it
+ * back unread, which is all the remote contract promises them.
+ */
+export class FindableMemoryPouchDatabase extends MemoryPouchDatabase {
```
TheSlimvReal, line 12: If you already did the work to implement this, you might as well add it to the core to reduce code duplication. 
I think there was one more problem with the local `_find` queries though. When sort and filter is used at the same tame and filter uses more than just the property on which it also sorts, then I think it also throws an error. Somehow it has a different behaviour there than the remote one.

TheSlimvReal (commented):
I now only looked at the code. The functionality already looked very good the last time. 

I think most of my comments boil down to two goals I would have for this feature:

1. Extend the `PaginatedDataSource` to also support pagination on data which is loaded via a view and not just with the find queries. This is something that might also be useful in the future and I therefore would like to already solve this here on a more abstract level instead of having a separate `AuditDataSource` which brings a lot of duplication. I imagine that the hard rule which we currently have that says "if there is a `loaderMehtod` use the `InMemoryDataSource`" could be removed and instead some `loaderMethod` might also support pagination. In this case the `PaginatedDataSource` still holds the pagination state but uses the loader method and provides it with the necessary information to make the correct request. What still might be an issue is the dynamic switching in the `ChangeHistoryListComponent` whether  it should use the index or a find query. But I think this could also be supported by allowing later overwrites of the `LoadRecordConfig` or at least the `loaderMethod`. I am happy to help with the proper design of the datasource as I was already playing around with it.
2. Use the `EntityListComponent` and instead of the `ChangeHistoryListComponent`. There also is a lot of duplication especially around the filtering which I don't thing is really necessary. Maybe it is not fully possible to get all the required functionality into the `EntityListComponent` because of the `loaderMethod` switching based on the entity ID filter but I still think the component can be made much simpler. Maybe `FilterComponent` + `EntitiesTableComponent` + one field for enabling the "changes for specific record" functionality could be sufficient. Nearly no logic would be required inside the component and it would only properly configure everything downstream.

src/app/core/database/pouchdb/findable-memory-pouch-database.ts
```diff
@@ -8,6 +8,11 @@ import { MemoryPouchDatabase } from "../core/database/pouchdb/memory-pouch-datab
  * PouchDB's local Mango engine has no bookmark cursor, so the opaque bookmark
  * is the positional offset of the next unseen document. Callers only pass it
  * back unread, which is all the remote contract promises them.
+ *
+ * It is stricter than CouchDB in one way worth knowing: it sorts only from an
+ * index that also covers every filtered field, so sorting on one field while
+ * filtering on another is rejected here although a real CouchDB serves it (see
+ * the spec). A test needing both has to filter on the field it sorts by.
  */
 export class FindableMemoryPouchDatabase extends MemoryPouchDatabase {
```
TheSlimvReal, line 17: You could also just move the function implementation to the `MemoryPouchDatabase` and delete this helper class or would there be any issue with that?

src/app/core/entity/entity-special-loader/entity-special-loader.service.ts
```diff
@@ -1,19 +1,41 @@
...
+  /** audit records that involve one given record, served by a CouchDB view */
+  AuditRecordsRelatedToEntity = "AuditRecordsRelatedToEntity",
   HistoricalDataService = "HistoricalDataService",
   ChildrenServiceQueryRelations = "ChildrenServiceQueryRelations",
   NotesRelatedToEntity = "NotesRelatedToEntity",
   TodosRelatedToEntity = "TodosRelatedToEntity",
 }
 
+/** One page of records, continuing from an opaque cursor. */
+export interface LoaderPage<E extends Entity = Entity> {
```
TheSlimvReal: This interface could already be used at other places too (`fetchPage` of `PagiantedDataSource` or `loadType` of `EntityMapperService`) lets declare it at the most general place and use it everywhere.

src/app/core/entity/entity-special-loader/entity-special-loader.service.ts
```diff
@@ -59,6 +87,30 @@ export class EntitySpecialLoaderService {
     return updatedEntity;
   }
 
+  /**
+   * Load one page of the records this loader serves for the given entity.
+   *
+   * Only for loaders {@link supportsPagination} reports: the rest have no notion
+   * of a page and return everything through {@link loadDataFor}.
+   */
+  loadPage<E extends Entity = Entity>(
```
TheSlimvReal: Lets call it `loadPageFor` to be in-line with the other functions (`loadData` and `loadDataFor`)

src/app/features/change-history/audit-reference-loader.service.ts
```diff
@@ -0,0 +1,80 @@
...
+      limit: page.limit,
+    };
+    if (skip > 0) {
+      query.skip = skip;
+    } else {
+      // `{}` sorts after every string, so the range starts past the newest
+      query.startkey = [recordId, bounds?.$lte ?? {}];
+    }
+
+    const response: ViewResponse = await this.service.queryReferenceView(query);
```
TheSlimvReal: Is there a reason why you are not using the `DatabaseIndexingService`? I had a look at the code and for me it seemed, that all functionality the `ChangeHistoryService` is offering here, would already be covered by this service. So no duplication needed and much less dependencies here.

TheSlimvReal (commented):
Nice, I am surprised, how little code changes the datasource actually needed for this. It think this is a good way to go. Looks very clean.

Abhinegi2:
> 2. Use the `EntityListComponent` and instead of the `ChangeHistoryListComponent`. There also is a lot of duplication especially around the filtering which I don't thing is really necessary. Maybe it is not fully possible to get all the required functionality into the `EntityListComponent` because of the `loaderMethod` switching based on the entity ID filter but I still think the component can be made much simpler. Maybe `FilterComponent` + `EntitiesTableComponent` + one field for enabling the "changes for specific record" functionality could be sufficient. Nearly no logic would be required inside the component and it would only properly configure everything downstream.

@TheSlimvReal Opened #4407 for the second goal, based on this branch

src/app/core/database/pouchdb/findable-memory-pouch-database.ts
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * An in-memory database that also supports {@link PouchDatabase.find}, so that
+ * tests can exercise an entity type end-to-end through
+ * `EntityMapperService.findType` -> `Database.find`.
+ *
+ * PouchDB's local Mango engine has no bookmark cursor, so the opaque bookmark
+ * is the positional offset of the next unseen document. Callers only pass it
+ * back unread, which is all the remote contract promises them.
+ */
+export class FindableMemoryPouchDatabase extends MemoryPouchDatabase {
```
Abhinegi2: It is in core — `src/app/core/database/pouchdb/findable-memory-pouch-database.ts`. And you're right about the second problem: sorting on one field while filtering on another is rejected locally though CouchDB serves it. That's documented at the top of the class, I hit exactly that.

src/app/features/change-history/model/audit-record.spec.ts
Abhinegi2: Not new here — user ids are usernames and can already contain colons. This record
only has to parse its own id back, which it does by counting from the end, so the
trailing timestamp and rev are unambiguous regardless. Nothing else about such ids
changes in this PR.

src/app/core/entity-list/entity-list/entity-list.component.ts
Abhinegi2: Agreed, `canImport` only waiting for init is inconsistent. It touches every ability-derived signal in core though, so I'd rather not grow this PR — opening a separate issue one for it.

src/app/core/database/database-resolver.service.ts
Abhinegi2: That would catch too much: in online mode `createDatabase()` returns a `RemotePouchDatabase` for the app database too, so `resetDatabases` and `destroyDatabases` would skip the app db as well. The name set only ever holds the databases that exist server-side, which is the actual condition.

src/app/features/change-history/change-history.service.ts
Abhinegi2: An index would give a complete list rather than a sample, agreed. What it can't fix is what the options are: the author is not one entity type — there's no `User` type, and the backend records either a type-prefixed entity id or a bare username depending on the instance. So this is the same design question as the sampling itself; better solved with #16 than separately here.

Abhinegi2: Correction to the above — I was wrong. I checked a real instance: every audited change carries `user.name` and every one is a type-prefixed entity id, so authors do resolve to records. My "there's no `User` type" came from grepping the source for `@DatabaseEntity`, but `EntityConfigService.createNewEntity` builds types from config at runtime, so it exists on a real deployment without appearing in source. The all-UUID authors I based it on came from a restored backup, not a live setup.

So an `EntityFilter` is viable here and your suggestion stands.

src/app/features/change-history/audit-reference-loader.service.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+  await fetchPage(first.bookmark);
+
+  expect(first.bookmark).toBe("5");
+  expect(lastQuery().skip).toBe(5);
+  // a repeated range start would return the same rows again
+  expect(lastQuery().startkey).toBeUndefined();
+  expect(lastQuery().endkey).toEqual(["User:1"]);
+});
+
+it("should drop rows the backend's permission filter emptied", async () => {
```
Abhinegi2: When the backend permission-filters the view. It checks each document the viewer may read and, where they may not, returns the row with `doc` omitted rather than removing it — so the response still carries a row for a change you cannot see.

That's why the row is dropped from the records but still counted for the cursor: `offset + rows.length` has to include it, or the next page starts before the rows already returned and repeats them.

Reachable on the change log because the audit database spans every record type, so an admin browsing it will hit changes to records their role does not grant read on.

Abhinegi2:
`loadType` won't work here: it's `getAll(ENTITY_TYPE + ":")`, so the prefix is always `"AuditRecord:"` — the whole audit database rather than one record's history. There's no argument to narrow it to `Child:1`.

`findType` does scope it, but it's a paged API — you'd need to follow the bookmark until exhausted to be sure a history is complete, where `getAll(prefix)` is one range request that returns all of it and needs no index. And the actual work in `getHistory` is `buildChangeEvents`, which replays the documents in order to derive per-field before/after; no loader replaces that.

The fair part is that the service reaches past the entity layer to do it. The clean version would be a prefix-scoped load on `EntityMapper` — happy to add that as a follow-up rather than in this PR.

Abhinegi2:
@TheSlimvReal I had to work on some of the pending items, so I’ve now merged this into the parent PR. I also created a separate issue for the remaining work and will create a follow-up PR later

PR #4394, opened 2026-09-16
fix: Note details view stuck loading forever from "Go to details"

TheSlimvReal:
## Steps to reproduce the bug

1. Open any list/dashboard view that shows Notes (e.g. the Notes tab, or a note popup opened from an entity's history).
2. Click a note to open its popup/dialog.
3. In the popup, click **"Go to details"**.
4. The app navigates to `/c/note/<id>`, but the page renders empty with an indefinite loading spinner — it never resolves.

## Root cause

The `/c/note/:id` route renders `NoteDetailsComponent` via the generic `RoutedViewComponent`, which passes the view's `config` object (plus route params) as `@Input`s.

- `NoteDetailsComponent` overrides `entityConstructor` directly to `Note` (`note-details.component.ts`), instead of relying on the base class's `entityType` input.
- `AbstractEntityDetailsComponent`'s entity-loading effect only calls `loadEntity()` when the `entityType()` **input signal** is set (`abstract-entity-details.component.ts`):
  ```ts
  effect((onCleanup) => {
    const id = this.id();
    if (!this.entityType() || !id) {
      return; // never proceeds for Notes
    }
    ...
  });
  ```
- The `view:note/:id` config never set `entityType`, unlike other entity views (e.g. `view:child/:id` sets `"entityType": "Child"`), for two separate reasons (see Fix below):
  1. `addDefaultNoteDetailsConfig` (`add-default-note-views.ts`), which auto-injects this config at runtime when an org's config doesn't define it yet, never set `entityType`.
  2. Several base-configs used by `SetupService` to seed a *new* organization already ship a **persisted** `view:note/:id` entry without `entityType` (`education`, `education/de`, `education/fr`, `all-features`). Because these are baked directly into the org's config at setup time, `addDefaultNoteDetailsConfig`'s "inject only if the key is absent" guard never even runs for these orgs.

Net effect: `entityType()` is always `undefined` for the Note details route, so `loadEntity()` is never called, `isLoading`/`entity` never update, and the template's `@if (isLoading() || !tmpEntity())` guard shows the spinner forever.

This is a regression from the signals migration in #3973, which changed the loading trigger to gate on the `entityType` input instead of the previously-used `id`-only `ngOnChanges` check.

## Fix

- Add `entityType: "Note"` to `getDefaultNoteDetailsConfig()`'s returned config (and add the field to the `NoteDetailsConfig` interface), so the runtime default-injection path supplies `entityType` like every other entity view does.
- Add `"entityType": "Note"` directly to the `view:note/:id` entry in the `education` (+ `de`/`fr` locales) and `all-features` base-config JSON files, since those already have a persisted entry that bypasses the runtime injection above. (Verified: `basic` and `event-management` base-configs don't define `view:note/:id` at all, so they were already correctly covered by the first fix.)
- As a related fix, `AdminNoteDetailsComponent.toConfig()` (the "Configure Note Details" admin editor) was rebuilding the config object from scratch on every save, silently dropping any field it doesn't know about — including `entityType`. This meant that even after the fixes above, an admin editing the Note details form layout and saving would inadvertently strip `entityType` again and reintroduce the bug for that organization. `toConfig()` now spreads the existing config first so unknown fields survive a save.

## Notes for reviewers

- This also has a side benefit: `EntityPermissionGuard` reads `entityType` from route config to enforce CASL `read` permission checks. Since `/c/note/:id` previously had no `entityType`, that route had **no permission check at all** — any authenticated user could open it. With this fix, normal "read Note" permission enforcement now applies to that route too.
- Organizations running an older config that already has a persisted `view:note/:id` entry from a source *other than* the four base-configs patched here (e.g. a custom/hand-edited config) will still need a config update, or an admin re-save of the Note details layout (now that saves preserve `entityType`), to pick up the fix.

## Test plan

- [x] `npx tsc --noEmit` — passes
- [x] `eslint` on changed files — passes
- [x] Existing + new unit tests for `add-default-note-views`, `admin-note-details`, `note-details`, `dialog-buttons`, `setup.service`, `config-migrations` — all pass
- [x] Verified against the actual bug report: the initial fix (runtime-injected default only) did **not** fix the behavior on the `pr-4394.aam-digital.net` preview deployment — traced this to the preview being seeded from a base-config (`education` or `all-features`) that already had a persisted `view:note/:id` without `entityType`, added the base-config fix above to address it
- [ ] Manual re-verification on the redeployed preview once it rebuilds from the latest commit — recommend confirming before merge

## Summary by CodeRabbit

- **Improvements**
  - Note details views now explicitly identify the Note entity type, supporting consistent loading and permission handling.
  - Note detail configuration changes now preserve existing configuration properties while updating form fields.

- **Bug Fixes**
  - Updated default and localized configurations to ensure Note details views use the correct entity type.

- **Tests**
  - Added coverage confirming that existing configuration fields remain intact when note detail form settings change.


TheSlimvReal:
Pushed a follow-up fix (4e33ec1d4): the initial commit only patched the runtime-injected default config, but the `education` (+ de/fr) and `all-features` base-configs already ship a **persisted** `view:note/:id` entry without `entityType` — that bypasses the runtime injection entirely since it only fires when the key is absent.

TheSlimvReal:
@coderabbitai review

src/app/child-dev-project/notes/add-default-note-views.ts
Abhinegi2, line 17: new `entityType: "Note"` only reaches orgs that don't have a `view:note/:id` entry yet, because `getDefaultNoteDetailsConfig()` is called only inside this branch. Orgs that already have that entry saved without `entityType` skip it entirely, and the base-config JSON changes only apply to newly created orgs — so wouldn't those existing orgs still see the endless spinner after this PR?

Abhinegi2 (commented):
code looks good. Just one question:

src/app/child-dev-project/notes/add-default-note-views.ts
TheSlimvReal: Yes, I did not add a config migration yet. Do you think that would be a useful approach?

Abhinegi2: I think we should first check the production system to see how many cases like these we have. If the number is small, we can handle them manually; otherwise, we’ll need to plan a migration.

@sleidig, could you help us get these numbers?

sleidig: most instances seem to have `"view:note"` within their Config doc, so I think a migration will be important (and probably not very hard to implement)

TheSlimvReal:
Added a permanent fix (16eb304fa): a new `migrateNoteDetailsEntityType` entry in the shared `configMigrations` pipeline (`config-migrations.ts`) that backfills `entityType: "Note"` onto any `view:note/:id` config missing it, without touching a custom value if one's already set.

This is the general, durable fix for the class of bug (not just the four base-configs from the previous commit): it's part of the migration pipeline the CLI's `latest-config-formats` migration already runs against production config documents (`applyConfigMigrations()`), so any self-hosted or hand-edited organization's config can be repaired by running that migration, in addition to the automatic in-memory fix on every app load.

src/app/child-dev-project/notes/add-default-note-views.ts
TheSlimvReal: Migration has been added

Abhinegi2 (approved):

PR #4395, opened 2026-09-16
feat: generic schema-embed-array field type for embedded tables (#2197)

### LINKED ISSUE #2197: New Field Type: Embedded table of multiple fields + multiple values
*As a user I want to add multiple values for some complex field combination, like multiple family members with a category (mother/father/..) + name field.*

**Objective**
To give users the option of adding as many entries as they want, we currently create a separate entity type and link it using RelatedEntities table. 
However, both to make config/setup easier and to optimize database performance, in some (possibly most?) cases it would be preferred to embed such an array of multi-value objects in the primary entity. (also see discussion on embedding ChildSchoolRelations into Child #693). 

**Proposed Solutions & Alternatives**
Implement a EditComponent for the "schema-embed-array" datatype, which already allows a complex object with an inside schema defining fields. This should display a table, matching the UX of the current flow with related-entities tables, but not requiring to set up a separate entity type.

>❗ this is a generalization of the existing `attendance` datatype, which does exactly this already.

_use for: individuals related to organisation; identification documents for individual; employment history details for individual/organisation;_

## Configuration

schema config (already exists for `attendance` dataType, make it usable with completely custom `schema-embed` dataType also, by adding the UI components for that):
```
entity.attributes: {
  name: { dataType: string },
  participants: {
          "label": "Participants List",
          "dataType": "attendance",
          "additional": {
            "participant": {
              "label": "Participant",
              "dataType": "entity",
              "additional": ["Child", "User"]
            },
            "role": {
              "label": "Role",
              "dataType": "string"
            }
          },
          "isArray": true
        }
}
```

## Implementation

- [ ] implement an `EditComponent` that displays all schema-embed fields (generalize the `EditAttendance` component)
  - possibly we could also implement this as a table layout? But for a first pass, keep it simple. So maybe generalizing EditAttendance or using a similar design is the best approach
- [ ] implement a `ViewComponent` for display in list views (maybe just a simple comma-separated text for the start?)

TheSlimvReal:
## Summary
- Adds a generic, config-driven "embedded table" field type (`schema-embed-array`), generalizing the attendance datatype's embedded-array pattern so any entity attribute can declare an array of custom sub-objects and get a working edit/view UI - without creating a separate entity type and RelatedEntities table. Closes #2197.
- Editing opens a popup dialog (button showing "Manage entries (N)") with the full table - add/remove rows, one generic `EntityFieldEditComponent` per cell so any inner dataType works automatically. `EntitiesTableComponent` was considered for this but isn't reusable: it's built entirely around persisted `Entity` records (DB-backed loading, entity ids for routing/selection, per-row saves, CASL permissions), none of which apply to an in-memory array nested in one field.
- Both the edit button (while disabled/view-mode) and the view component show a small text/count with a hover preview table, matching the pattern the attendance datatype's display component already uses.
- Also fixes issue described in #4399

## Where to find the demo/testing scaffolding
The commit, `demo: add identificationDocuments field to Child for manual testing`, is **not part of the feature** - it only adds a sample field to the "basic" demo config's `Child` entity so the feature can be tried in the running app. Drop that commit out before merging.

## Test plan
- [x] Unit tests for the datatype (array transform round-trip, config-based and `embeddedType`-based), the edit dialog (add/remove/edit rows, disabled state), the view component, and the shared preview-table component
- [x] Full `ng build` (dev config) to typecheck templates
- [x] Manually verified in the browser: add/edit/remove entries, save, reload, read-only view, hover preview in both edit and view modes, list-column rendering
- [ ] Reviewer: try the demo field described above, then confirm the last commit gets dropped before merge


src/assets/base-configs/basic/Config_CONFIG_ENTITY.json
```diff
@@ -355,6 +355,24 @@
...
+            },
+            "documentNumber": {
+              "dataType": "string",
+              "label": "Document Number"
+            },
+            "issueDate": {
+              "dataType": "date",
+              "label": "Issue Date"
+            }
+          }
```
sleidig, line 375: let's move this to the "all-features" use case and keep the "basic" use case minimal.

src/app/core/basic-datatypes/schema-embed/display-schema-embed-array/display-schema-embed-array.component.html
```diff
@@ -0,0 +1,12 @@
+@if (rows().length) {
+  <span class="entry-count" [appTemplateTooltip]="tooltip">
+    {{ rows().length }} <span i18n>entries</span>
+  </span>
+}
+
+<ng-template #tooltip>
+  <app-schema-embed-array-table
```
sleidig, line 8: the tooltip needs a bit of padding. but otherwise I like the UX of the details available through tooltip.

src/app/core/basic-datatypes/schema-embed/display-schema-embed-array/display-schema-embed-array.component.html
```diff
@@ -0,0 +1,12 @@
+@if (rows().length) {
+  <span class="entry-count" [appTemplateTooltip]="tooltip">
+    {{ rows().length }} <span i18n>entries</span>
```
sleidig, line 3: maybe some kind of icon could indicate to users that more details are available on hover? Not sure what, however ...

src/app/core/basic-datatypes/schema-embed/edit-schema-embed-array/edit-schema-embed-array.component.html
```diff
@@ -0,0 +1,22 @@
+<span
+  class="trigger-wrapper"
+  [appTemplateTooltip]="tooltip"
+  [tooltipDisabled]="rowCount() === 0"
+>
+  <button
```
sleidig, line 6: Did you consider something that makes the embedded rows visible directly here? I think for the details view something like this would make it more convenient to use:

image

src/assets/base-configs/basic/Config_CONFIG_ENTITY.json
TheSlimvReal: Yes, it will be removed before merging. This was just for easily playing around with it.

sleidig: Let's add a field like this permanently to the all-features config, also :-)

src/app/core/basic-datatypes/schema-embed/edit-schema-embed-array/edit-schema-embed-array.component.html
TheSlimvReal: I am a bit worried of the impact a long list or many columns will have on the details view then. We don't really have control on how many properties will be added and this just doesn't feel flexible enough. Our controls are all quite big and I think it would quickly create an odd-looking view. But if you have some good ideas for a more tailored design feel free to mock it.

src/app/core/basic-datatypes/schema-embed/display-schema-embed-array/display-schema-embed-array.component.html
TheSlimvReal: Hmm, maybe we could use the same (disabled) button from the edit component to make it clear there is something to interact with?

src/app/core/basic-datatypes/schema-embed/edit-schema-embed-array/edit-schema-embed-array.component.html
TheSlimvReal: I was thinking however, to also allow the data to be displayed in a table similar to `RelatedEntitiesComponent`. This would probably need a new component and admin UI for it, but both pretty straight forward.

sleidig: That sounds good, initially I had the same idea in mind. Probably the more reliable approach afterall :+1: 

I guess technically this also should be very simple, as our embedded data has the same format as external entities, only stored differently.

TheSlimvReal:
What needs to be done for this PR

- [ ] usability testing
- [ ] (optional) better UI in list view
- [ ] new entity details component that is similar to `RelatedEntitiesComponent` to display related data living on the entity itself. `EntitiesTableComponent` will not be usable for this because it expects rows to be entities
- [ ] how should this mode be communicated to user as compared to having a separate related entity.

PR #4399, opened 2026-09-18
fix: inline-editing a table row crashed instead of rendering fields

TheSlimvReal:
## Summary
- `EntitiesTableComponent` bound `app-entity-field-edit`'s `form` input to `row.formGroup` - a bare `FormGroup` - but that input expects a full `EntityForm` object (`{ formGroup, entity, fieldConfigs, inheritedParentValues, watcher }`).
- `EntityFieldEditComponent`'s `formControl` computed does `form.formGroup.get(...)`, which throws once `form` actually *is* the `FormGroup` itself (a `FormGroup` has no nested `.formGroup` property).
- TypeScript never caught this: `*matCellDef`'s row context is untyped (`CdkCellDef`'s `template` is `TemplateRef<any>`, no `ngTemplateContextGuard`), so `row` - and `row.formGroup` - type-check as `any` regardless of `strictTemplates` being enabled.
- Every column in a row being inline-edited hit this, so the feature never actually worked: clicking the edit-pencil icon just threw repeatedly and no input fields rendered.
- Fix: `TableRow` now also keeps the full `EntityForm` built during `edit()` (`.form`, alongside the existing `.formGroup` still used for the enabled/disabled checks), and the table passes that to `app-entity-field-edit` instead of the bare `FormGroup`.

## Steps to reproduce (on the "basic" demo setup)
1. `npm start`, then open `http://localhost:4200/?useCase=basic` and click "Start Exploring".
2. Open the browser console (to see the errors).
3. Click **Schools** in the sidebar, then open any school from the list.
4. Go to the **Students** tab.
5. Click the pencil (✎) icon on the left of any student row.

**Before this fix:** the console fills with repeated `TypeError: Cannot read properties of undefined (reading 'get')` and the row never becomes editable - no input fields appear for Child/Start date/End date/Class/Result.

**After this fix:** the row switches to editable inputs (an entity-autocomplete for Child, date-pickers for Start/End date, a number input for Class, etc.), and Save/Cancel/Delete icons appear, exactly as inline editing is supposed to work.

This reproduces the same way for any other `EntitiesTableComponent`/`RelatedEntities` list with its default `editable: true` (e.g. `Participants` list, if given editable custom columns) - the School → Students table is just a convenient built-in example already present in the "basic" config.

## Test plan
- [x] Added/extended unit tests in `entity-inline-edit-actions.component.spec.ts` asserting `row().form` is populated as a full `EntityForm` on edit, and cleared on save/reset
- [x] Full `ng build` (dev config) to typecheck templates
- [x] Manually reproduced the crash on `master`, then verified the fix resolves it end-to-end following the steps above


TheSlimvReal:
Will be fixed in #4395

PR #4400, opened 2026-09-18
fix(entities-table): select only free-text filtered records with "select all"

### LINKED ISSUE #4398: Bug: "Select all" ignores the free-text filter and selects all records of the entity type
**Describe the bug**

When a list is narrowed with the free-text filter, the "select all" checkbox in the table header selects every record matching the structured filters — not only the rows shown after free-text filtering.

This is dangerous: a following bulk action (e.g. bulk edit) then executes on all records of the entity type. Observed in a real list: 6 rows visible after free-text filtering, while the bulk action bar reported "Action will execute on 1789 selected records".

Structured filters (dropdown, date range, related entity) are **not** affected — those correctly select only the filtered records.

**To Reproduce**

1. Open a list with many records that has the free-text filter enabled (e.g. Participants)
2. Type a term into the "Filter" field so only a few rows remain
3. Click the "select all" checkbox in the table header
4. Open a bulk action from the bulk action bar
5. See that it reports all records of the entity type as selected, not the filtered rows

**Expected behavior**

"Select all" selects only the records currently matching *all* active filters, including the free-text filter. Changing the free-text filter afterwards also resets an existing selection.

**Additional context**

The free-text filter is set as `MatTableDataSource.filter` and applied lazily through `filterPredicate` when rows are rendered. It never updates `displayedData`, which is the set that "select all" reads. Structured filters instead go through `dataFilter` -> `filteredRecords` -> `displayedData`, which is why they behave correctly.

Abhinegi2:
- closes #4398

"Select all" read the table's unfiltered row set, so on a list narrowed by the free-text filter it selected every record of the entity type instead of the visible rows — a following bulk action would then have run on all of them. The selectable rows now honour the free-text filter as well, and changing the filter term resets an existing selection.

## Steps for manual testing

1. Open a list with many records and an enabled free-text filter (e.g. Participants)
2. Type a term into the "Filter" field so that only a few rows remain
3. Click the "select all" checkbox in the table header -> only the visible rows get checked
4. Open a bulk action -> the bar reports only the filtered rows as selected
5. Change the free-text filter term -> the existing selection is cleared
6. Repeat with a dropdown / date-range filter to confirm those still behave as before

## PR Checklist

---

## Summary by CodeRabbit

- **Bug Fixes**
  - Free-text filtering now keeps the displayed table rows synchronized with the active filter.
  - “Select all” now applies only to records matching the current filter, including filtered records across pages.
  - Updating the free-text filter clears existing row selections to prevent stale selections.

- **Documentation**
  - Clarified that “select all” applies to all filtered rows across every page.


Abhinegi2:
@coderabbit review

Abhinegi2:
Thanks for the alternative approach
Squash auto-merge is enabled , could you approve when you get a chance?

TheSlimvReal (approved):

